PR #8, opened 2021-11-10
fix(clippy): fix some clippy lints

mattsse:
fix some clippy lints, clippy still complains about things like mod inception and unusual default impl though

### DIFF
--- a/bins/revm-ethereum-tests/src/runner.rs
+++ b/bins/revm-ethereum-tests/src/runner.rs
@@ -216,16 +216,18 @@ pub fn execute_test_suit(
                 // do the deed
 
                 let timer = Instant::now();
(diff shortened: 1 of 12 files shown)


rakita:
This is great, thank you! didn't even think about running clippy :)
There is some junk in `main` that I will need to clean (I really need to start using separate branches) will do the cleanup probably today/tomorrow.

mattsse:
my favorite clippy command `cargo +nightly clippy --fix --all-features -Z unstable-options` :)

this looks great and compile time is surprisingly very low!

rakita:
nice command!
compile-time and binary bloat are something that I was afraid of after using generics soo heavily, but it came up great.

PR #42, opened 2022-01-28
fix(revm_precompiles): lock borsh semver tighter

### LINKED ISSUE #41: Borsh 0.9.2 breaks compilation of `revm_precompiles`
We hit the described change here. We should hardcode as 0.9.1 most likely.

https://github.com/near/borsh-rs/blob/master/CHANGELOG.md#092---2022-01-25

brockelmore:
Closes #41

### DIFF
--- a/crates/revm_precompiles/Cargo.toml
+++ b/crates/revm_precompiles/Cargo.toml
@@ -10,7 +10,7 @@ version = "0.4.0"
 
 [dependencies]
 bn = { package = "substrate-bn", version = "0.6", default-features = false }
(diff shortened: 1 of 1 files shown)


gakonst (approved):
Good one

rakita:
It seems borsh is not even used inside precompiles. Removing it, is probably fixed the problem: https://github.com/bluealloy/revm/pull/43

gakonst:
Ha fantastic

rakita:
remove borsh completely here: https://github.com/bluealloy/revm/pull/43

PR #46, opened 2022-02-02
Inspector fixup

### LINKED ISSUE #45: Inspectors
A list of bugs(?) and missing things for inspectors before we can use revm in Foundry:

- It seems that `call_end` is called instead of `create_end`, which is never called
- We should add more parameters to `call_end` and `create_end` (such as gas left after the call, return values etc.)
- It seems like `Inspector::call` is only called for the upper-most call(? not sure on this), not for internal calls

onbjerg:
An attempt at fixing #45 

- Adds parameters to `create_end` and `call_end`
- Calls `create_end` and `call_end` in the correct places

Putting this in draft to get comments - is this the right direction and/or does this break anything?

Closes #45

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -398,9 +398,6 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB, 
                 (exit_reason, ret, machine.gas, machine.return_value())
             }
         };
-        if INSPECT {
-            self.inspector.call_end();
-        }
         ret
     }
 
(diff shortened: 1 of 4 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -586,42 +583,84 @@ impl<'a, GSPEC: Spec, DB: Database + 'a, const INSPECT: bool> Host
...
+                &init_code,
+                gas_limit,
+            );
             if ret != Return::Continue {
-                return (ret, gas, bytes, out);
+                return (ret, address, gas, out);
             }
         }
-        self.create_inner::<SPEC>(caller, scheme, value, init_code, gas)
+        let (ret, address, gas, out) =
```
onbjerg, line 601: A note here, it seems that `out` is always `Bytes::empty`?

rakita:
Nice, this is mostly what is missng. I would like to check few more things over the weekend, but at first look it looks great.

btw you dont need to create issue every time, just creating pull request is enought,

crates/revm/src/evm_impl.rs
rakita: It most cases it is. Only here it returns something: https://github.com/bluealloy/revm/blob/c0bab703000e125a81b2ef35210af55a56d5664f/crates/revm/src/evm_impl.rs#L398

rakita (approved):
will merge after ci finishes

rakita:
I will fix later this monster number of inputs for create/call functions

PR #47, opened 2022-02-03
Remove all `unsafe` code

sbillig:
Replacing all unsafe code with safe equivalents resulted in a minor speed penalty, but that was easy to correct.

revm-test on current main branch:
```
0: 72.280883ms
1: 66.22685ms
2: 63.54688ms
3: 63.925157ms
4: 63.710048ms
5: 63.555763ms
6: 64.017038ms
7: 63.486204ms
8: 63.844067ms
9: 63.534811ms
```

after removing `unsafe`:
```
0: 75.006729ms
1: 70.62909ms
2: 65.881008ms
3: 65.937669ms
4: 65.571569ms
5: 65.658069ms
6: 66.056095ms
7: 65.523081ms
8: 65.74447ms
9: 65.561382ms
```

with some small performance improvements to speed it back up (mostly just avoiding some redundant bounds checks):
```
0: 72.207021ms
1: 66.552397ms
2: 64.265398ms
3: 63.883503ms
4: 64.146832ms
5: 63.8677ms
6: 64.196815ms
7: 64.071832ms
8: 63.978789ms
9: 63.875884ms
```

The `perf` stats are interesting. main branch:
```
             4,214      page-faults:u             #    6.393 K/sec                  
     2,315,735,367      cycles:u                  #    3.513 GHz                      (36.95%)
     5,222,993,976      instructions:u            #    2.26  insn per cycle           (44.84%)
       868,177,930      branches:u                #    1.317 G/sec                    (45.44%)
         9,404,860      branch-misses:u           #    1.08% of all branches          (46.05%)
     1,309,655,465      L1-dcache-loads:u         #    1.987 G/sec                    (46.66%)
         4,189,485      L1-dcache-load-misses:u   #    0.32% of all L1-dcache accesses  (47.26%)
           225,908      LLC-loads:u               #  342.730 K/sec                    (31.55%)
             9,317      LLC-load-misses:u         #    4.12% of all LL-cache accesses  (31.55%)
   <not supported>      L1-icache-loads:u                                           
         2,234,462      L1-icache-load-misses:u                                       (31.50%)
     1,314,240,932      dTLB-loads:u              #    1.994 G/sec                    (30.89%)
             6,336      dTLB-load-misses:u        #    0.00% of all dTLB cache accesses  (30.28%)
            10,869      iTLB-loads:u              #   16.490 K/sec                    (29.68%)
            50,216      iTLB-load-misses:u        #  462.01% of all iTLB cache accesses  (29.13%)

[...]

PR #48, opened 2022-02-05
[WIP] Use Result type for instruction functions

sbillig:
This is sort of an experiment, and isn't entirely finished; if you're not interested in these changes, no hard feelings.

This builds on #47, and removes some macros in favor of using rust's `Result` type and `?` operator. (Note that `Result<(), Return>` is the same size as `Return`; there should be no downside to using Result/? instead of explicit tests and returns).

This is motivated by the fact that some of the macros are doing almost exactly what the more idiomatic, safe equivalent does.

For example,
```
pop!(machine, op1);

macro_rules! pop {
    ( $machine:expr, $x1:ident) => {
        if $machine.stack.len() < 1 {
            return Return::StackUnderflow;
        }
        let $x1 = unsafe { $machine.stack.pop_unsafe() };
    };
    ...
}
pub unsafe fn pop_unsafe(&mut self) -> U256 {
    let mut len = self.data.len();
    len -= 1;
    self.data.set_len(len);
    *self.data.get_unchecked(len)
}
```
should compile down to exactly the same thing as this:
```
let op1 = machine.stack.pop()?;

pub fn pop(&mut self) -> Result<U256, Return> {
    self.data.pop().ok_or(Return::StackUnderflow)
}
```
Of course, it's not *exactly* the same :)
There's a small speed loss again, which could maybe be mitigated.

revm-test on main branch:
`9: 63.577744ms`
this pr:
`9: 66.229494ms`

### DIFF
--- a/bins/revme/src/debugger/ctrl/ctrl.rs
+++ b/bins/revme/src/debugger/ctrl/ctrl.rs
@@ -110,7 +110,7 @@ impl<DB: Database> Inspector<DB> for Controller {
         machine: &mut revm::Machine,
         data: &mut EVMData<'_, DB>,
         _is_static: bool,
-    ) -> Return {
+    ) -> Result<(), Return> {
         loop {
             match Ctrl::next(self.state_machine, &self.history_path) {
                 Ctrl::Help => {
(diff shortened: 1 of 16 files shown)


rakita:
maybe with inline always you can get better perf but i would like to leave this as it is, it is internal thing not exposed to outside of lib and with having macro here i think it reduces compilation and as you said it is faster, and for this project, this matters. Decision making are related to `evm compatibility` > `speed` > `simpicity`.

PR #50, opened 2022-02-14
propagate the back the error parsing k256 ecrecover

ntrippar:
k256 ecrecover currently is unwrapping and if the signature is invalid this will panic. because the error of `recoverable::Signature::try_from` is the same as `k256::ecdsa::Error` I decide to propagate back the error that then in handled in the `ec_recover_run` function.


rakita:
Good catch and thank you for contributing! 

Will merge after tests go green.

rakita (approved):

PR #52, opened 2022-02-20
Refactor

rakita:
Machine -> Interpreter
some name changes and refactoring of structure.


onbjerg:
Any breaking changes I should be aware of for Foundry? :smile:

crates/revm/src/inspector.rs
```diff
@@ -44,24 +48,16 @@ pub trait Inspector<DB: Database> {
...
         &mut self,
         data: &mut EVMData<'_, DB>,
-        call: H160,
-        context: &CallContext,
-        transfer: &Transfer,
-        input: &Bytes,
-        gas_limit: u64,
+        inputs: &CallInputs,
         is_static: bool,
     ) -> (Return, Gas, Bytes);
```
rakita, line 53: @onbjerg this is only change that is going to have a effect outside of revm. all those fields are now inside `CallInputs`/`CreateInputs` struct, so it will be relatively easy to upgrade.

PR #55, opened 2022-02-22
fix: export missing machine structs

onbjerg:
Parts of the `Machine` API is unusable without these exports (which is important for inspectors)


rakita:
fix `fmt` or `clippy`

onbjerg:
Lint is due to missing docs about safety, I'll add it

onbjerg:
Should be gtg now

rakita:
This is new, didnt know that we need unsafe doc section. Will force merge this and fix it later.

PR #57, opened 2022-02-23
refactor: improve inspector ergonomics

onbjerg:
Improves inspector ergonomics by providing a default implementation for all the functions. For example, if you want to create an `OverrideSpecInspector` you currently have to stub all of the other functions.

Also correctly calls `step` and `step_end` for inspectors.

I noticed that `Inspector::initialize` is not used anywhere - is it supposed to?

### DIFF
--- a/crates/revm/src/inspector.rs
+++ b/crates/revm/src/inspector.rs
@@ -43,106 +43,6 @@ pub trait Inspector<DB: Database> {
     #[allow(clippy::too_many_arguments)]
     fn call(
         &mut self,
-        data: &mut EVMData<'_, DB>,
-        call: H160,
-        context: &CallContext,
-        transfer: &Transfer,
-        input: &Bytes,
-        gas_limit: u64,
-        is_static: bool,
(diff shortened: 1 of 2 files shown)


crates/revm/src/inspector.rs
```diff
@@ -197,4 +99,27 @@ impl<DB: Database> Inspector<DB> for NoOpInspector {
     }
 
     fn selfdestruct(&mut self) {}
+
+    /// If needed you can override some of the spec configurations when running with inspector
+    fn override_spec(&self) -> &OverrideSpec {
+        &OVERRIDE_SPEC_DEFAULT
+    }
```
rakita, line 106: Small note: I will probably remove `override_spec` from inspector and just include flag `skip_eip170_contract_code_size_limit` in `env.cfg`. When i first added this i was obsess with making it as fast as possible with macroes and stuff, litlle bit bias decision for this particular case. :)

onbjerg: That will work too :)

PR #71, opened 2022-03-07
feat: cache block hashes

onbjerg:
Caches block hashes, but also computes a block hash based on the block number if no block hash was returned from the underlying database.

Retrieving block hashes can be expensive, but it is also important that block hashes for different blocks differ from each other

@draganrakita I was unsure where to put this - we need it for `cheats.roll` to work correctly in Foundry, but it would also be weird to have a `set_block_hash` function in the database trait.

Two alternatives I could think of:

- Allow inspectors to override block hashes
- Implement our own database in Foundry, although this would pretty much be a 1:1 copy of `CacheDB` here with this patch

Curious to hear what you think


rakita:
Having cached hashes seems okay, having custom hash created on H256::zero() seems like that it can be added to `ExtDB` for CacheDB feels like a small pitfall.

gakonst:
@rakita Agree that doing on `CacheDB` seems weird. What about doing it on `EmptyDB`'s `fn block_hash`? When we're forking, we're gonna have a value that corresponds to the number, but when we're not forking we're using `EmptyDB`, so maybe that's the right place for this logic. @onbjerg https://github.com/bluealloy/revm/blob/main/crates/revm/src/db/in_memory_db.rs#L206

PR #83, opened 2022-03-17
Gas fixes

onbjerg:
The gas returned by the EVM did not account for refunds, and `Interpreter::gas` took `&mut self` without needing to

I also made `finalize` not return a `Result` since there was no path where it would return `Err`

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -147,10 +147,9 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact
         if crate::USE_GAS {
             gas.reimburse_unspend(&exit_reason, ret_gas);
         }
-        match self.finalize::<GSPEC>(caller, &gas) {
-            Err(e) => (e, out, gas.spend(), Map::new(), Vec::new()),
-            Ok((state, logs)) => (exit_reason, out, gas.spend(), state, logs),
-        }
+
+        let (state, logs, gas_used) = self.finalize::<GSPEC>(caller, &gas);
+        (exit_reason, out, gas_used, state, logs)
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -213,11 +212,13 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
             self.data
                 .subroutine
                 .balance_add(coinbase, coinbase_gas_price * (gas.spend() - gas_refunded));
+            gas.spend() - gas_refunded
         } else {
             // touch coinbase
             self.data.subroutine.load_account(coinbase, self.data.db);
             self.data.subroutine.balance_add(coinbase, U256::zero());
-        }
+            0
```
onbjerg, line 220: Unsure if we still want to do the gas refund calculation stuff here? Does `crate::USE_GAS` toggle if we even record gas at all?

crates/revm/src/interpreter.rs
```diff
@@ -53,7 +53,7 @@ impl Interpreter {
         &self.contract
     }
 
-    pub fn gas(&mut self) -> &Gas {
+    pub fn gas(&self) -> &Gas {
```
gakonst, line 56: 👍

crates/revm/src/evm_impl.rs
gakonst: Good q. @rakita wdyt?

crates/revm/src/evm_impl.rs
```diff
@@ -192,9 +191,9 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
         &mut self,
         caller: H160,
         gas: &Gas,
-    ) -> Result<(Map<H160, Account>, Vec<Log>), Return> {
+    ) -> (Map<H160, Account>, Vec<Log>, u64) {
```
gakonst, line 194: The function signature change should be OK given that this function is not exposed

crates/revm/src/evm_impl.rs
```diff
@@ -147,10 +147,9 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact
         if crate::USE_GAS {
             gas.reimburse_unspend(&exit_reason, ret_gas);
         }
-        match self.finalize::<GSPEC>(caller, &gas) {
```
gakonst, line 150: Ack - the error branch would never get hit before

crates/revm/src/evm_impl.rs
rakita: This is fine, USE_GAS in general is to disable gas calculation if evm is needed in some wasm env where pricing is done differently.

rakita (approved):
nice :)

PR #88, opened 2022-03-21
feat: allow inspectors to override block env

onbjerg:
I'm not sure if this is the best way to go about it: we have some cheatcodes in Foundry that alter information about blocks (`cheats.fee` to set `basefee`, `cheats.warp` to set `timestamp` and `cheats.roll` to set `number`). 

When we run tests in fork mode (i.e. using state from mainnet) we also set these things to reflect mainnet (`basefee` etc). Because the basefee is high in this scenario, and because the gas limit is always high for tests, we end up in a scenario where REVM will say we are out of funds, when in reality we do not care if we are out of funds for the main call.

This PR adds the ability to override the block environment, which fixes our problem. We can set all the gas prices etc. to 0 on our side, and in our inspector we can override the block environment with the actual information

Ref https://github.com/gakonst/foundry/pull/997/commits/37c2bbcade00fe32e63bfcb117e44e6a7b6f9764

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -5,7 +5,8 @@ use crate::{
     models::SelfDestructResult,
     return_ok,
     subroutine::{Account, State, SubRoutine},
-    CallContext, CallInputs, CreateInputs, CreateScheme, Env, Gas, Inspector, Log, Return, Spec,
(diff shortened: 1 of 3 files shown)


rakita:
You have access to those things in EVMData -> Env -> BlockEnv

onbjerg:
You're right, I had a brainfart - I'll close this and adjust https://github.com/gakonst/foundry/pull/997 to just override `env.block` when `initialize_interp` is called :smile:

PR #93, opened 2022-03-27
Various fixes

onbjerg:
- Prevents overflows in `SubRoutine::transfer` for the recipient
- Allows inspectors to see what kind of call is being made - some inspectors in Foundry manage state that depend on this info

### DIFF
--- a/crates/revm/src/models.rs
+++ b/crates/revm/src/models.rs
@@ -161,8 +161,6 @@ pub struct CallContext {
     /// Apparent value of the EVM.
     pub apparent_value: U256,
     /// The scheme used for the call.
-    ///
-    /// Defaults to [CallScheme::Call].
     pub scheme: CallScheme,
 }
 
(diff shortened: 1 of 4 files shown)


onbjerg:
There's an issue with transfers now, will take a look at what is going on

crates/revm/src/models.rs
```diff
@@ -142,15 +149,33 @@ pub enum CallScheme {
...
     pub address: H160,
     /// Caller of the EVM.
     pub caller: H160,
+    /// The address the contract code was loaded from, if any.
+    pub code_address: H160,
     /// Apparent value of the EVM.
     pub apparent_value: U256,
+    /// The scheme used for the call.
+    ///
+    /// Defaults to [CallScheme::Call].
```
rakita, line 165: rm comment about defaults

rakita (approved):
one small nit, lgtm

PR #94, opened 2022-03-27
Update AccountInfo#code documentation

alcuadrado:
Hey @rakita 👋

I guess this is a question and PR at the same time. 

If I understand `AccountInfo` correctly, the `code` being `None` doesn't mean that the account doesn't have any code, but that you should load it from a `Database`. Is that right? If so, maybe you want to merge this PR that documents it.

### DIFF
--- a/crates/revm/src/models.rs
+++ b/crates/revm/src/models.rs
@@ -18,7 +18,7 @@ pub struct AccountInfo {
     pub balance: U256,
     /// code hash,
     pub code_hash: H256,
-    /// code: if None, it must be loaded from a `Database`
+    /// code: if None, `code_by_hash` will be used to fetch it if code needs to be loaded from inside of revm.
     #[cfg_attr(feature = "with-serde", serde(with = "serde_hex_bytes_opt"))]
     pub code: Option<Bytes>,
     /// Account nonce.
--- a/crates/revm/src/models.rs
+++ b/crates/revm/src/models.rs
@@ -18,7 +18,7 @@ pub struct AccountInfo {
     pub balance: U256,
     /// code hash,
     pub code_hash: H256,
-    /// code
+    /// code: if None, it must be loaded from a `Database`
     #[cfg_attr(feature = "with-serde", serde(with = "serde_hex_bytes_opt"))]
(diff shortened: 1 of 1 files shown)


onbjerg:
I don't think this is correct - `AccountInfo::code` can be `None` in the DB even if the account has code, e.g. if the code was set in `SubRoutine` but that info has not been commited yet. Conversely, `code` may also be `None` in `SubRoutine` but `Some` in DB if the contract was self-destructed but the results were not committed yet

rakita:
@alcuadrado @onbjerg AccountInfo::code None means it is not loaded from db. This is part of code that handles that:
https://github.com/bluealloy/revm/blob/211c83ba85518d82a1674eb6e5520cf3efbc617a/crates/revm/src/subroutine.rs#L518-L525

crates/revm/src/models.rs
```diff
@@ -18,7 +18,7 @@ pub struct AccountInfo {
     pub balance: U256,
     /// code hash,
     pub code_hash: H256,
-    /// code
+    /// code: if None, it must be loaded from a `Database`
```
rakita, line 21:
```suggestion
    /// code: if None, it means that it is not loaded from a `Database`
```

onbjerg:
> @alcuadrado @onbjerg AccountInfo::code None means it is not loaded from db. This is part of code that handles that:
> 
> https://github.com/bluealloy/revm/blob/211c83ba85518d82a1674eb6e5520cf3efbc617a/crates/revm/src/subroutine.rs#L518-L525

Yeah but this is only the case if you load it from the `SubRoutine`, not the case for `AccountInfo` as a whole is my point. So maybe these docs should be in `SubRoutine`?

rakita:
> Yeah but this is only the case if you load it from the `SubRoutine`, not the case for `AccountInfo` as a whole is my point. So maybe these docs should be in `SubRoutine`?

Yea, you are looking at AccountInfo as something external, in that sense original statement is missleading. I changes meaning of it now.

alcuadrado:
Thanks for your explanations, @rakita @onbjerg !

PR #96, opened 2022-04-02
chore: refactor to exact option combinators

huitseeker:
Plus use [captured identifiers from rust 1.58](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings),

Tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust)


rakita (approved):
ty for sharing, this is amazing

PR #102, opened 2022-04-20
chore: add missing derives

mattsse:
add some derives


rakita:
i need to fix clippy, lgtm

PR #110, opened 2022-05-07
chore: typo fixes

guanqun:


rakita (approved):
I am always a little bit embarrassed when something like this comes up. Thank you for contributing @guanqun!

PR #112, opened 2022-05-16
fix: BLOCKHASH should return 0 if number not in last 256 blocks

wardbradt:


crates/revm/src/instructions/host.rs
```diff
@@ -90,7 +91,19 @@ pub fn blockhash<H: Host>(interp: &mut Interpreter, host: &mut H) -> Return {
...
+    let lower = if upper <= U256::from(256u64) {
+        U256::zero()
+    } else {
+        upper.sub(U256::from(256u64))
+    };
+    if number >= lower && number < upper {
+        push_h256!(interp, host.block_hash(number));
+    } else {
+        push_h256!(interp, H256::zero());
+    }
```
rakita, line 106: Nice, I left this for the DB interface to have the restriction there but that was not okay, it is the proper way to do it here.
```suggestion
    // gas!(interp, gas::BLOCKHASH);

    pop_top!(interp, number);

    if let Some(diff) = host.env().block.number.checked_sub(*number) {
        let diff = as_usize_saturated!(diff);
        // blockhash should push zero if number is same as current block number. Checked with evmone logic.
        if diff <= 256 && diff != 0 {
            *number = U256::from_big_endian(host.block_hash(*number).as_ref());
            return Return::Continue;
        }
    }
    *number = U256::zero();
    Return::Continue
```
This is a bit better code as it does not use U256 that much. I still need to recheck it later but it should be okay :)

rakita: don't forget to remove first `top!` as it is replaced with `pop_top`

PR #119, opened 2022-06-03
feat: add getters for cachedb

mattsse:
adds getters to gain access to the underlying Database


rakita:
need to fix that lint in another pull request

PR #122, opened 2022-06-06
chore: export evm_inner

mattsse:
access to this is useful to replicate `Evm::inspect_ref` with a custom `Database`, other than `RefDBWrapper`

I guess alternatively this could be achieved by making `RefDBWrapper` generic but would require breaking changes.


rakita:
https://crates.io/crates/revm/1.4.1

PR #123, opened 2022-06-08
fix: remove stipend from sstore gas calculations

bernard-wagner:
Not sure what the reasoning was for including the stipend for SSTORE originally. Couldn't find any other EVM implementations doing this check.

Encountered the issue when testing a number of scenarios using [cast](https://github.com/foundry-rs/foundry/tree/master/cast).

Specifically:
```
cast run 0x10c305ada9cb1066b396d6558d1d195d21abfaef6c76e5ae1568354b58daab64
```

The transaction would fail due the storage slot of the `reentrancy` guard being reset.

### DIFF
--- a/crates/revm/src/gas/calc.rs
+++ b/crates/revm/src/gas/calc.rs
@@ -215,9 +215,8 @@ pub fn sstore_cost<SPEC: Spec>(
     original: U256,
     current: U256,
     new: U256,
-    gas: u64,
     is_cold: bool,
-) -> Option<u64> {
+) -> u64 {
     // TODO untangle this mess and make it more elegant
     let (gas_sload, gas_sstore_reset) = if SPEC::enabled(BERLIN) {
         (STORAGE_READ_WARM, SSTORE_RESET - SLOAD_COLD)
@@ -225,10 +224,6 @@ pub fn sstore_cost<SPEC: Spec>(
         (sload_cost::<SPEC>(is_cold), SSTORE_RESET)
     };
     let gas_cost = if SPEC::enabled(CONSTANTINOPLE) {
-        if SPEC::enabled(CONSTANTINOPLE) && gas <= CALL_STIPEND {
-            return None;
-        }
-
         if new == current {
             gas_sload
         } else {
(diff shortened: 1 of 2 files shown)


rakita:
It seems that eth tests do not agree with you. Here is EIP: https://eips.ethereum.org/EIPS/eip-2200
But I need to check when it is enabled, [here](https://github.com/ethereum/execution-specs) it say it is Instanbul

mattsse:
from the yellow paper

> the execution is in an exceptional halting state if the gas left prior to executing an SSTORE instruction is less than or equal to the call stipend

bernard-wagner:
Might have been over simplistic with my approach. 

Looking at GETH, it seems the stipend is added to the gas before call. 

https://github.com/ethereum/go-ethereum/blob/440c9fcf75d9d5383b72646a65d5e21fa7ab6a26/core/vm/instructions.go#L687

bernard-wagner:
Happy to close this and reopen as an issue. Clearly I misidentified the root cause.

rakita:
@bernard-wagner this is part in revm where call stipend is added: https://github.com/bluealloy/revm/blob/538a2e55e55a8ff8c25dee765b6342ae0811daa6/crates/revm/src/instructions/host.rs#L362-L365

bernard-wagner:
> @bernard-wagner this is part in revm where call stipend is added:
> 
> https://github.com/bluealloy/revm/blob/538a2e55e55a8ff8c25dee765b6342ae0811daa6/crates/revm/src/instructions/host.rs#L362-L365

Yeah. Back to the drawing board on why replaying the following transactions fails:

```
cast run 0x10c305ada9cb1066b396d6558d1d195d21abfaef6c76e5ae1568354b58daab64
```

rakita:
> > @bernard-wagner this is part in revm where call stipend is added:
> > https://github.com/bluealloy/revm/blob/538a2e55e55a8ff8c25dee765b6342ae0811daa6/crates/revm/src/instructions/host.rs#L362-L365
> 
> Yeah. Back to the drawing board on why replaying the following transactions fails:
> 
> ```
> cast run 0x10c305ada9cb1066b396d6558d1d195d21abfaef6c76e5ae1568354b58daab64
> ```

There is probably some problem in a gas calculation that takes a little bit more than it should and it just hits gas limit point here

PR #124, opened 2022-06-09
Consensus error with gas block for SSTORE stipend check

rakita:


rakita:
Firstly noticed here: https://github.com/bluealloy/revm/pull/123

PR #126, opened 2022-06-10
don't delete account and storage entries on commit

mattsse:
As discovered here https://github.com/foundry-rs/foundry/issues/1894

deleting storage or account entries does not play nice with fork mode which basically does:

```rust
if !db.contains(acc) {
    // fetch from remote 
}
```

So deleting accounts/slots (when empty) would result in the fork's account/slot being served and not 0 or the empty account.

# Solution

* on `CacheDb::commit` don't delete `AccountInfo` or storage entries in the CacheDb's maps 

* also bumps to v1.6.0

### DIFF
--- a/crates/revm/src/db/in_memory_db.rs
+++ b/crates/revm/src/db/in_memory_db.rs
@@ -1,4 +1,4 @@
-use crate::{subroutine::Filth, Database, KECCAK_EMPTY};
+use crate::{Database, KECCAK_EMPTY};
 
 use alloc::vec::Vec;
 use hashbrown::{hash_map::Entry, HashMap as Map};
@@ -87,26 +87,12 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
 impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
     fn commit(&mut self, changes: Map<H160, Account>) {
         for (add, acc) in changes {
-            if acc.is_empty() || matches!(acc.filth, Filth::Destroyed) {
-                self.cache.remove(&add);
-                self.storage.remove(&add);
-            } else {
-                self.insert_cache(add, acc.info);
-                let storage = self.storage.entry(add).or_default();
-                if acc.filth.abandon_old_storage() {
-                    storage.clear();
-                }
-                for (index, value) in acc.storage {
-                    if value.is_zero() {
-                        storage.remove(&index);
-                    } else {
-                        storage.insert(index, value);
-                    }
-                }
-                if storage.is_empty() {
-                    self.storage.remove(&add);
-                }
+            self.insert_cache(add, acc.info);
+            let storage = self.storage.entry(add).or_default();
+            if acc.filth.abandon_old_storage() {
+                storage.clear();
             }
+            storage.extend(acc.storage);
         }
     }
 }
--- a/crates/revm/src/db/in_memory_db.rs
+++ b/crates/revm/src/db/in_memory_db.rs
@@ -24,6 +24,7 @@ impl InMemoryDB {
 pub struct CacheDB<ExtDB: DatabaseRef> {
     /// Dummy account info where `code` is always `None`.
     /// Code bytes can be found in `contracts`.
+    changes: Map<H160, AccountInfo>,
     cache: Map<H160, AccountInfo>,
     storage: Map<H160, Map<U256, U256>>,
     contracts: Map<H256, Bytes>,
@@ -38,6 +39,7 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
         contracts.insert(KECCAK_EMPTY, Bytes::new());
         contracts.insert(H256::zero(), Bytes::new());
         Self {
+            changes: Map::new(),
             cache: Map::new(),
             storage: Map::new(),
             contracts,
@@ -45,6 +47,9 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
             block_hashes: Map::new(),
             db,
         }
+    }
+    pub fn changes(&self) -> &Map<H160, AccountInfo> {
+        &self.changes
     }
 
     pub fn cache(&self) -> &Map<H160, AccountInfo> {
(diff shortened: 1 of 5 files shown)


mattsse:
hmm, perhaps keeping destroyed accounts has some side-effects?

crates/revm/src/db/in_memory_db.rs
```diff
@@ -87,26 +87,12 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
 impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
     fn commit(&mut self, changes: Map<H160, Account>) {
         for (add, acc) in changes {
-            if acc.is_empty() || matches!(acc.filth, Filth::Destroyed) {
-                self.cache.remove(&add);
-                self.storage.remove(&add);
```
rakita, line 92: Here, you need to set cache to `Account::default` and iterate over `storage.get(add).iter()` and set them all to zero

crates/revm/src/db/in_memory_db.rs
```diff
@@ -87,26 +87,12 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
...
-            } else {
-                self.insert_cache(add, acc.info);
-                let storage = self.storage.entry(add).or_default();
-                if acc.filth.abandon_old_storage() {
-                    storage.clear();
```
rakita, line 97: here iterate over all storage and set all values to zero

crates/revm/src/db/in_memory_db.rs
```diff
@@ -87,26 +87,12 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
...
-                for (index, value) in acc.storage {
-                    if value.is_zero() {
-                        storage.remove(&index);
-                    } else {
-                        storage.insert(index, value);
-                    }
-                }
-                if storage.is_empty() {
-                    self.storage.remove(&add);
-                }
```
rakita, line 108: for this `for` and `if` use extend

rakita (commented):
selfdestruct, what else to say. :)
I am merkelizing everything in the state when running tests. I think i am skipping empty/zero values but I am not sure tbh, will check later.

rakita:
@mattsse in `state_merkle_trie_root` skip `AccountInfo`s that are `is_empty`

mattsse:
I guess we'd also need to skip 0 in storage now when computing `trie_root`, right?

rakita:
> I guess we'd also need to skip 0 in storage now when computing `sec_trie_root`, right?

zero storage slots are already skipped. I am not sure why it is not working. I vaguely remember there are some funky things with merkelization in tests where some empty accounts get merkelized I will need to manually run it and compare it.

rakita:
The problem was with state cleanup that was introduced way back in eth: https://eips.ethereum.org/EIPS/eip-161
The state can contain "empty" accounts, with this EIP, only states that are accessed are going to be removed from state.

For tests that means that there are preloaded empty accounts and not all of them are touched (called) and those empty accounts need to appear in Merkle trie.

I added `changes` field to CacheDB and I am open to modifying this to a more flexible form needed, but at least we know what is going on.

crates/revm/src/db/in_memory_db.rs
```diff
@@ -86,26 +100,25 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
 // TODO It is currently only committing to cached in-memory DB
 impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
     fn commit(&mut self, changes: Map<H160, Account>) {
+        // clear storage by setting all values to `0`
+        fn clear_storage(storage: &mut HashMap<U256, U256>) {
+            storage.values_mut().for_each(|val| {
```
gakonst, line 105: something to consider in the future, maybe we want to convert these to rayon parallel iters? maybe not worth it

crates/revm/src/db/in_memory_db.rs
gakonst: @rakita you said set "cache to Account::default" but we're currently setting the `changes`? is that right, or a mistake?

crates/revm/src/db/in_memory_db.rs
```diff
@@ -86,26 +100,25 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
...
         for (add, acc) in changes {
             if acc.is_empty() || matches!(acc.filth, Filth::Destroyed) {
-                self.cache.remove(&add);
-                self.storage.remove(&add);
+                // clear account data, but don't remove entry
+                self.changes.insert(add, Default::default());
+                self.storage.get_mut(&add).map(clear_storage);
             } else {
-                self.insert_cache(add, acc.info);
+                self.insert_change(add, acc.info);
```
gakonst, line 116: same q here, what's the diff between cache / change?

gakonst:
LGTM - couple comments above, should we try to release this ASAP to fix the downstream bugs in Foundry?

crates/revm/src/db/in_memory_db.rs
rakita: @gakonst `changes` are added after, because of this: https://github.com/bluealloy/revm/pull/126#issuecomment-1152717104

crates/revm/src/db/in_memory_db.rs
rakita: For this particular thing, it would be best to introduce storage incarnation: https://github.com/ledgerwatch/erigon/blob/devel/docs/programmers_guide/db_walkthrough.MD#table-plainstate
In general, for introducing rayon, we need to be careful that we have a good amount of data and processing that we want to iterate/Parallelize on.

crates/revm/src/db/in_memory_db.rs
rakita: It is because of this: https://github.com/bluealloy/revm/pull/126#issuecomment-1152717104
the `cache` is loaded from the external DB.
the `changes` is what is applied from `commit`.

gakonst:
@mattsse @rakita do we intend to merge this or are there any blockers?

mattsse:
I think this is good to merge, don't think this has any downside

gakonst:
Good point - will keep it in mind before we update^

PR #132, opened 2022-06-15
fix: insert in cache map

mattsse:
changing this line made foundry tests pass again

Ref https://github.com/bluealloy/revm/pull/126

original question: https://github.com/bluealloy/revm/pull/126#discussion_r895058605


rakita:
There is probably somewhere that `cache` is used, i need to make this properly with flag that says account status.

gakonst:
bump - any next steps here?

meetmangukiya:
ref https://github.com/bluealloy/revm/pull/136#discussion_r907224087

PR #136, opened 2022-06-27
Return specific `Return` statuses in `CALL`s

meetmangukiya:
- Bubble up the specific `Return` status of a `CALL`.
- Call `call_end` inspector hook in all terminating paths(a couple missed places).
- Interpreter run should ignore non-ok `CALL` statuses

c/f https://github.com/foundry-rs/foundry/issues/1387

### DIFF
--- a/crates/revm/src/db/in_memory_db.rs
+++ b/crates/revm/src/db/in_memory_db.rs
@@ -113,7 +113,7 @@ impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
                 self.changes.insert(add, Default::default());
                 self.storage.get_mut(&add).map(clear_storage);
             } else {
-                self.insert_cache(add, acc.info);
+                self.insert_change(add, acc.info);
                 let storage = self.storage.entry(add).or_default();
                 if acc.filth.abandon_old_storage() {
                     clear_storage(storage);
--- a/crates/revm/src/db/in_memory_db.rs
+++ b/crates/revm/src/db/in_memory_db.rs
@@ -113,7 +113,7 @@ impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
                 self.changes.insert(add, Default::default());
                 self.storage.get_mut(&add).map(clear_storage);
             } else {
-                self.insert_change(add, acc.info);
+                self.insert_cache(add, acc.info);
                 let storage = self.storage.entry(add).or_default();
                 if acc.filth.abandon_old_storage() {
                     clear_storage(storage);
--- a/crates/revm/src/interpreter.rs
+++ b/crates/revm/src/interpreter.rs
@@ -8,7 +8,7 @@ pub use stack::Stack;
 
 use crate::{
     instructions::{eval, Return},
-    Gas, Host, Spec, USE_GAS,
+    return_ok, Gas, Host, Spec, USE_GAS,
 };
 use bytes::Bytes;
 use core::ops::Range;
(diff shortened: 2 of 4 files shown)


crates/revm/src/interpreter.rs
```diff
@@ -123,6 +123,18 @@ impl Interpreter {
...
+            if matches!(
+                opcode,
+                super::instructions::opcode::CALL
+                    | super::instructions::opcode::STATICCALL
+                    | super::instructions::opcode::DELEGATECALL
+            ) {
+                if !matches!(ret, return_ok!() | Return::OutOfGas) {
+                    ret = Return::Continue
+                }
+            }
```
rakita, line 136: This loop is hot part of evm, as it is an execution loop and these checks does not seem right.

crates/revm/src/db/in_memory_db.rs
```diff
@@ -113,7 +113,7 @@ impl<ExtDB: DatabaseRef> DatabaseCommit for CacheDB<ExtDB> {
                 self.changes.insert(add, Default::default());
                 self.storage.get_mut(&add).map(clear_storage);
             } else {
-                self.insert_change(add, acc.info);
+                self.insert_cache(add, acc.info);
```
rakita, line 116: In current code, it is correct to use `insert_change`,
this part is done as bug fix for cleared storage (selfdestruct), I will change this when i get more free time but for now it is what it is.

crates/revm/src/interpreter.rs
meetmangukiya: > This loop is hot part of evm, as it is an execution loop and these checks does not seem right.

Do you mean checking on each step if opcode was `CALL | DELEGATECALL | STATICCALL` might slow things down?

> This seems not correct if ret is Return::OutOfGas we can't just override that to Return::Continue

We do not override `Return::OutOfGas` to `Return::Continue`. We only override when it is neither of `return_ok!() | Return::OutOfGas` to `Return::Continue` to keep the behaviour same as earlier https://github.com/bluealloy/revm/pull/136/files#diff-6e6245d0a2d234f8bc8ea76eefc437a171d3dfc1fe97098603966161ac491244L403 where we only ever returned `Return::Continue` or `Return::OutOfGas`(from `next_gas_block` checks). Execution was not halted on anything other than `OutOfGas`. Here, we are just trying to bubble up the actual reason / status of `CALL` evaluation, but it still shouldn't be halted unless it is out of gas.

rakita: Yes, it will slow it down.
Yeah i missed `!`  in `if !matches!(ret,...` logic is the same.

I am not seeing who is using this? Additional inspector calls for `call_end` are okay and should be added, but this code is called right after `CALL | CALLCODE | DELEGATECALL | STATICCALL` ( you are missing CALLCODE) and it does nothing, or I am missing something?

crates/revm/src/db/in_memory_db.rs
rakita: Foundry ignores `change` field, and that is where new contract is saved, it is a reason why eth tests fail

crates/revm/src/interpreter.rs
meetmangukiya: Whoops, you are right. I was thinking of providing the status to `step_end`, I need to move the `step_end` call above this `if`. However, thinking about it now, I think we'd need to change the interface a bit to give the correct status(the overriden status) vs intermediate status(call's return) since by default an inspector does nothing and returns the same status that it was given. If we replace the current status with intermediate status, we'd trigger termination.

Will think about this more whether we even need to know the intermediate status of `CALL` step because we can use `call_end` to record the same in the inspector.

crates/revm/src/db/in_memory_db.rs
meetmangukiya: I see, I can look into making foundry use the `change` field.

rakita: Revert this and rest can be merged

rakita, line 116:
```suggestion
                self.insert_change(add, acc.info);
```

Issue #121, opened 2022-06-04
feat: Introduce ByteCode format


rakita:
With a restriction that contacts can not start with `0xEF` and with the possibility that in future we will have different `ByteCode` for EVM Object Format, it makes sense to introduce support for it into the revm. https://eips.ethereum.org/EIPS/eip-3541

The biggest benefit atm would be introducing `ByteCode` type that will contain jumptable and gas blocks and with that we would skip the needed analysis of the contract. This is a fairly big speed-u for some use cases noticed in the foundry as fuzzing tests call a lot of contracts with not much execution and analysis, in that case, is taking a lot of time.

The proposal is to introduce enum `ByteCode` that would somehow return: `code`, `code_size` and `jumptable`.

Basically replace this input with `ByteCode`: https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/interpreter/contract.rs#L60
and propagate change wherever needed.
`ByteCode` should set these fields in the contract:
* https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/interpreter/contract.rs#L11-L13
* https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/interpreter/contract.rs#L21

and do the same for database struct:
* Return `ByteCode`: https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/db/traits.rs#L14
* Replace bytes with `BytesCode`: https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/models.rs#L24

To fully circle the story with EVM Object Format, CREATE and CREATE2 will need to return 0xEF type format in bytes that would get transformed to `ByteCode` for evm usage here:  
https://github.com/bluealloy/revm/blob/08528247759bcd113c0c5385ee081e20e11cb80b/crates/revm/src/evm_impl.rs#L395-L399
But this is a consensus rule and out of scope for this feat, we should just continue using legacy `ByteCode` here. If needed we could in Inspector `create_end` transform that into `ByteCode`.

gakonst:
cc @onbjerg @brockjelmore for when we'll want to further speedup fuzz tests

onbjerg:
[...]

PR #156, opened 2022-07-30
feat: Introduce ByteCode format

rakita:
Related to: #121


crates/revm/src/interpreter/contract.rs
```diff
@@ -180,13 +109,16 @@ impl Contract {
...
 /// Mapping of valid jump destination from code.
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct ValidJumpAddress {
-    first_gas_block: u64,
-    analysis: Vec<AnalysisData>,
+    pub first_gas_block: u32,
+    /// Rc is used here so that we dont need to copy vector. We can move it to more suitable more accessable structure
+    /// without copying underlying vec.
+    pub analysis: Rc<Vec<AnalysisData>>,
```
onbjerg, line 117: Can we make this an `Arc`? Forge has some shared backend logic that sends `AccountInfo` across thread bounds and the `Rc` makes it `!Send` :smile:

crates/revm/src/interpreter/bytecode.rs
```diff
@@ -0,0 +1,271 @@
...
+    },
+    Analysed {
+        len: usize,
+        jumptable: ValidJumpAddress,
+    },
+}
+
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Bytecode {
```
onbjerg, line 24: Can we get `Eq` and `PartialEq`?

crates/revm/src/interpreter/contract.rs
gakonst: ^+1 we need this

PR #165, opened 2022-08-09
Bytecode hash, remove override_spec,

rakita:
Save hash of bytecode inside `Bytecode` structure, this is allowing us to have hash ready and not calculate it every time.

Remove override_spec and move configuration to `env.cfg`

and small cleanup.


rakita:
Foundry just got even faster

previous graph:
182139228-35afb27f-caf3-4db7-8f1e-31c4ee849ac7

new graph:
bytecode_hash_flamegraph

`is_success` is not existing in new version as bytecode hash is only done once.

PR #171, opened 2022-08-12
revm: Update account storage methods in CacheDB

ngotchac:
So this PR does 2 things in `revm::CachDB`:
1. It updates the `insert_account_storage` so that it doesn't override the underlying account info
2. It adds a new method `replace_account_storage` to completely replace the account storage

For (1), I don't actually know whether it was intentional to override the account info with the default (empty) value. I don't quite see when this would be useful.

For (2), it would be useful for implementing the `StateOverride` that's in [geth](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set), where there is an option to completely override the storage.

I added tests for both methods.

### DIFF
--- a/crates/revm/src/db/in_memory_db.rs
+++ b/crates/revm/src/db/in_memory_db.rs
@@ -94,6 +94,17 @@ impl<ExtDB: DatabaseRef> CacheDB<ExtDB> {
             .storage
             .insert(slot, value);
     }
+
+    /// replace account storage without overriding account info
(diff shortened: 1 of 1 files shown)


rakita:
I added CacheDb as i needed utility for tests and didnt think about implication if it is used by others, but it has changed to something usable as foundry is relaying on it. Both changes make sense.

@onbjerg this shouldn't change anything major, but just wanted to notify you as it can has different side effect.

PR #175, opened 2022-08-15
JournaledState

rakita:
Reworked how we store changes of state (needed for revert). Journaling is something that other EVM-a uses and it simplifies some logic inside revm.

Some needed renamings from SubRouting to JournaledState
And fixes the problem with selfdestruct removing storage while the transaction is still in progress.


onbjerg:
Is there a small TL;DR of the breaking changes here? It seems like there would be a lot? :smile:

rakita:
> Is there a small TL;DR of the breaking changes here? It seems like there would be a lot? 😄

They are mostly internal, renaming is maybe only one external

PR #177, opened 2022-08-17
refactor(precompiles): Vec -> BTreeMap

shekhirin:


rakita:
I thought that `vec` would be more effective for <50 items, but it seems (from this statistic) that cutoff line is around ~15 items
https://gist.github.com/daboross/976978d8200caf86e02acb6805961195

rakita (approved):

rakita:
@shekhirin can you do `cargo fmt --all`?

PR #182, opened 2022-08-22
fix(revm): Fix balance overflow in `finalize`

ngotchac:
It seems that #175 introduced a `panic` when finalizing a transaction; which occurs at least in some Foundry tests. This PR restores the old behavior as before #175 (although we might want to use a `saturating_add`?)

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -223,13 +223,14 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB, 
             let max_refund_quotient = if SPEC::enabled(LONDON) { 5 } else { 2 }; // EIP-3529: Reduction in refunds
 
             let gas_refunded = min(gas.refunded(), gas.spend() / max_refund_quotient);
-            self.data
-                .journaled_state
-                .state()
-                .get_mut(&caller)
-                .unwrap()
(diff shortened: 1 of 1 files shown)


rakita:
Just, are you sure that balance is panicking, it is U256?

Most correct way would be to error, but because this is unusual situation not possible to happen in mainnet we can relax constraint, saturation add would probably be better but this is good.

ngotchac:
> Just, are you sure that balance is panicking, it is U256?

Yeah it's from a test in `foundry`, which you can see the result of here: https://github.com/foundry-rs/foundry/runs/7955552675?check_suite_focus=true

> Most correct way would be to error, but because this is unusual situation not possible to happen in mainnet we can relax constraint, saturation add would probably be better but this is good.
Yeah I thought I'd mimic the previous behavior. Let me know, I can make another PR if needed :)

rakita:
> > Just, are you sure that balance is panicking, it is U256?
> 
> Yeah it's from a test in `foundry`, which you can see the result of here: https://github.com/foundry-rs/foundry/runs/7955552675?check_suite_focus=true

Ah i see now. 

> > Most correct way would be to error, but because this is unusual situation not possible to happen in mainnet we can relax constraint, saturation add would probably be better but this is good.
> 
> Yeah I thought I'd mimic the previous behavior. Let me know, I can make another PR if needed :)

If pr is made it will be merged 😁

PR #183, opened 2022-08-22
chore: add refunds saturated

mattsse:
in theory, adding refunds should not lead to overflows since it is subtracted beforehand:

https://github.com/bluealloy/revm/blob/24622a4c9e450217f6a723008a75e3faa3b7cc5d/crates/revm/src/evm_impl.rs#L90-L104

however in forge test there are scenarios where we start with an `effective_gas_price` of `0` and an initial balance of `U256::MAX` and modify the env (gas price) via cheat codes so that we could theoretically end up with a refund that will increase the balance in some situations.

using saturated add here would prevent a panic in such situations.


ngotchac:
@mattsse I think we had the same idea :sweat_smile: https://github.com/bluealloy/revm/pull/182

mattsse:
lol :)

closing as dupe #182 then

PR #191, opened 2022-08-26
Add support for old forks.

rakita:
Add support for all ethereum forks. From homestead onwards.

Constantinople is not added as it was superseded with Petersburg:
https://github.com/ethereum/execution-specs

All legacy eth tests are passing and CI is updated to run them.


crates/revm/src/db.rs
```diff
@@ -17,14 +17,14 @@ use auto_impl::auto_impl;
 #[auto_impl(& mut, Box)]
 pub trait Database {
     /// Get basic account information.
-    fn basic(&mut self, address: H160) -> AccountInfo;
+    fn basic(&mut self, address: H160) -> Result<Option<AccountInfo>, &'static str>;
```
mattsse, line 20: would it be possible to make this a `Cow`, this way we could include additional context in the error like address etc

mattsse, line 20: also I think it'd be useful to use an error alias and reexport it

something like:
```rust
pub type DbResult<T> = Result<T, &'static str>;

```

rakita: Adding associate type Error to database trait would maybe make this better

rakita: You can now specify `Error` type inside `Database`.

rakita: `Error` type fixes this

gakonst:
legend

PR #196, opened 2022-09-01
use Infallible for memory db's error type

mattsse:
The `Database(Ref) impl for `EmptyDb` never returns an error.

use `Infallible` as error type to indicate that `<EmptyDb as Database(Ref)::...` will never return an error


crates/revm/src/db/in_memory_db.rs
```diff
@@ -5,6 +5,7 @@ use alloc::vec::Vec;
 use hashbrown::{hash_map::Entry, HashMap as Map};
 use primitive_types::{H160, H256, U256};
 use sha3::{Digest, Keccak256};
+use std::convert::Infallible;
```
rakita, line 8:
```suggestion
use core::convert::Infallible;
```

rakita (approved):

rakita:
@mattsse can you `fmt` this I am not by my PC

mattsse:
done

PR #215, opened 2022-09-12
feat: add Memory::shrink_to_fit

mattsse:
add another delegate function to `Memory` that allows shrinking the allocated 4 * 1024 bytes down to what's actually used.

this is likely the cause for the memory hog in foundry's tracer impl

cc @shekhirin


rakita:
I see, this makes sense

PR #222, opened 2022-09-16
feat(revm, revme): gas inspector

shekhirin:
Gas counting became a bit harder with the introduction of gas blocks composed during analysis. So `GasInspector` was introduced to make it easier to count remaining gas after each executed step.

Approach was adopted from https://github.com/bluealloy/revm/blob/1e25c99b544863d15610e9af8e623dd173397b48/bins/revme/src/statetest/trace.rs#L10-L17

### DIFF
--- a/bins/revme/src/statetest/trace.rs
+++ b/bins/revme/src/statetest/trace.rs
@@ -3,26 +3,18 @@ use primitive_types::H160;
 pub use revm::Inspector;
 use revm::{
     opcode::{self},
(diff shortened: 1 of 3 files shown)


gakonst (commented):
is it possible this hurts performance meaningfully?

shekhirin:
@gakonst you mean new `GasInspector`? I don't think so, it's just very simple math operations and O(1) lookups.

rakita (approved):
Thank you

PR #223, opened 2022-09-17
Add distinct error code for state changes triggered during a staticcall

dyedm1:
The current behavior when a state change is attempted (through sstore, log, selfdestruct, or create) during a static call is to return the NotActive error code, which refers to something not being supported on the current EVM version. 

When encountering this, I found it to be confusing and somewhat inaccurate. I think this specific instance should be given its own error code, StateChangeDuringStaticCall.

### DIFF
--- a/crates/revm/src/instructions.rs
+++ b/crates/revm/src/instructions.rs
@@ -54,6 +54,7 @@ pub enum Return {
     InvalidJump,
     InvalidMemoryRange,
     NotActivated,
+    StateChangeDuringStaticCall,
     StackUnderflow,
     StackOverflow,
     OutOfOffset,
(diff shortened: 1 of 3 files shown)


rakita:
I agree it is both confusing and inaccurate, I missed to change it (and didn't put a lot of thought into it), the proposed change is good and would be included but I am kinda in process of modifying how opcodes are called so this is going to be overwritten soon.

https://github.com/bluealloy/revm/blob/bd8eb6c031509a980b37982addb9a19e1f43017a/crates/revm/src/instructions/opcode/jump_table.rs#L248-L251

I will see to update the proper Return error to the proposed one.

rakita:
Finally introduced here: https://github.com/bluealloy/revm/pull/349

PR #231, opened 2022-10-11
Borrow self and add derive traits for OpCode

iFrostizz:
Making `OpCode` easier to convert into `u8` using `u8()` method by not taking ownership and adding some derive traits to the `OpCode` struct

### DIFF
--- a/crates/revm/src/instructions/opcode.rs
+++ b/crates/revm/src/instructions/opcode.rs
@@ -1,6 +1,7 @@
 use crate::gas;
 use crate::SpecId;
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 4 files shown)


crates/revm/src/instructions/opcode.rs
```diff
@@ -164,7 +165,7 @@ impl OpCode {
     }
 
     #[inline(always)]
-    pub const fn u8(self) -> u8 {
+    pub const fn u8(&self) -> u8 {
```
rakita, line 168:
```suggestion
    pub fn u8(&self) -> u8 {
```
let it build but const probably needs to be removed

rakita (approved):
lgtm

PR #234, opened 2022-10-19
fix: return out of gas code for precompiled contracts

### LINKED ISSUE #233: Precompiled contracts do not return OOG error
When a precompiled contract does not have sufficient gas, a generic `Return::PrecompileError` is returned. In the case of `precompiles::Return::OutOfGas` we can actually return `Return::OutOfGas` to clearly signal why the precompiled contract failed.

Wodann:
Fixes #233


rakita:
need to remove that `TODO` :)

rakita (approved):

PR #237, opened 2022-10-21
feat: allow eip3607 to be toggled off

### LINKED ISSUE #235: Ability to perform call from contract
At hardhat we support the ability to call contracts from another contract when developing your own smart contract. So far we've been using `ethereumjs`, which supports this behaviour.

We're rewriting part of our stack in Rust and using `revm` to do so. However, `revm` strictly adheres to `EIP-3607` to "Reject transactions from senders with deployed code": https://github.com/bluealloy/revm/blob/main/crates/revm/src/evm_impl.rs#L90
```
        // EIP-3607: Reject transactions from senders with deployed code
        // This EIP is introduced after london but there was no colision in past
        // so we can leave it enabled always
        if self.data.journaled_state.account(caller).info.code_hash != KECCAK_EMPTY {
            return exit(Return::RejectCallerWithCode);
        }
```

In development, the security concern does not apply, so it would be nice to still be allowed to call contracts from a contract address. Is this something that you'd be willing to accept a PR for? If so, do you have a preference for how it should be implemented?

One option would be to use a `dev` feature flag that would disable that check.

Wodann:
Fixes #235.

### DIFF
--- a/crates/revm/Cargo.toml
+++ b/crates/revm/Cargo.toml
@@ -28,10 +28,12 @@ web3 = { version = "0.18", optional = true }
 
 [features]
 default = ["std", "secp256k1"]
+dev = ["memory_limit", "optional_eip3607"]
+memory_limit = []
 no_gas_measuring = []
+optional_eip3607 = []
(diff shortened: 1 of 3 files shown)


rakita (approved):
Lgtm

PR #245, opened 2022-10-27
fix sanity check

Owen66:
Currently this check ensures the passed in value is greater than self.block_number. I think is likely a bug as I've seen transactions in the wild which succeed on geth but revert when using this revm code due to the sanity check.

I think ensuring the passed in value is less than self.block_number makes more sense as that excludes block numbers for blocks ahead of the self.block_number which may not exist yet.

### DIFF
--- a/crates/revm/src/db/web3db.rs
+++ b/crates/revm/src/db/web3db.rs
@@ -105,14 +105,8 @@ impl Database for Web3DB {
         if number > U256::from(u64::MAX) {
             return Ok(KECCAK_EMPTY);
         }
-        let number = u64::try_from(number).unwrap();
-        if let Some(block_num) = self.block_number {
-            match block_num {
-                BlockNumber::Number(t) if t.as_u64() > number => return Ok(KECCAK_EMPTY),
-                _ => (),
-            }
-        }
-        let number = wU64::from(number);
+        
+        let number = wU64::from(u64::try_from(number).unwrap());
         let f = async {
             self.web3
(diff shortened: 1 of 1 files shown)


rakita:
Check for "valid" block hash is present inside revm here: https://github.com/bluealloy/revm/blob/9f8cdbd5cbeb831c1f6bb28b913e83b2a467700b/crates/revm/src/instructions/host.rs#L118
let us remove this check fully.

Owen66:
OK cool. I've amended the commit to remove the check.

rakita (approved):
lgtm

Owen66:
Had to push a change to fix a conflict following the merge of another PR. What happens next in terms of getting this merged?

rakita:
It is good! I had a little bit of personal time the last few days so I am a little bit slow to responde, just waiting for tests to pass and that's it, it will get merged.

crates/revm/src/db/web3db.rs
```diff
@@ -105,14 +105,8 @@ impl Database for Web3DB {
...
         }
-        let number = u64::try_from(number).unwrap();
-        if let Some(block_num) = self.block_number {
-            match block_num {
-                BlockNumber::Number(t) if t.as_u64() > number => return Ok(KECCAK_EMPTY),
-                _ => (),
-            }
-        }
-        let number = wU64::from(number);
+
```
rakita, line 108:
```suggestion

```

PR #252, opened 2022-11-03
fix: compiler error when secp256k1 is disabled

Wodann:
Resolves an error that was happening when you disabled default features for revm:

```
failed to resolve: use of undeclared crate or module `secp256k1`
  --> revm_precompiles-1.1.1\src\secp256k1.rs:80:21
   |
80 |     let out = match secp256k1::ecrecover(&sig, &msg) {
   |                     ^^^^^^^^^ use of undeclared crate or module `secp256k1`
```


rakita:
I am fine with build failure, as you would not have erecover precompile if you don't enable `k256` or default `secp256k1`.

Probably need to add comments somewhere, `Cargo.toml` and `readme`.

PR #256, opened 2022-11-08
fix: feature flag compiler errors

Wodann:


crates/revm/Cargo.toml
```diff
@@ -46,14 +46,13 @@ dev = [
...
+k256 = ["revm_precompiles/k256_ecrecover"]
 memory_limit = []
 no_gas_measuring = []
 optional_block_gas_limit = []
 optional_eip3607 = []
 optional_gas_refund = []
-std = ["bytes/std", "num_enum/std", "rlp/std"]
 secp256k1 = ["revm_precompiles/secp256k1"]
-k256 = ["revm_precompiles/k256_ecrecover"]
-web3db = []
```
Wodann, line 57: Removed deprecated feature flag, as it will cause `cargo t --all-features` to fail

crates/revm_precompiles/src/secp256k1.rs
```diff
@@ -59,7 +57,13 @@ mod secp256k1 {
     }
 }
 
+#[cfg(any(feature = "k256_ecrecover", feature = "secp256k1"))]
 fn ec_recover_run(i: &[u8], target_gas: u64) -> PrecompileResult {
+    use alloc::vec::Vec;
```
Wodann, line 62: Moved `use` statements and `const` into local scope to avoid feature `cfg`

crates/revm_precompiles/src/secp256k1.rs
```diff
@@ -1,9 +1,7 @@
+#[cfg(any(feature = "k256_ecrecover", feature = "secp256k1"))]
```
Wodann, line 1: Missing feature `cfg`, as this only works when either the `k256_ecrecover` or `secp256k1` feature flags are on. Ensures that `cargo t --no-default-features` succeeds.

Wodann:
@rakita if there's anything you'd like me to change, please let me know.

crates/revm_precompiles/src/lib.rs
```diff
@@ -112,6 +112,7 @@ impl Precompiles {
         static INSTANCE: OnceCell<Precompiles> = OnceCell::new();
         INSTANCE.get_or_init(|| {
             let fun = vec![
+                #[cfg(any(feature = "k256_ecrecover", feature = "secp256k1"))]
```
rakita, line 115: I dont want to disable ECRECOVER if default is not present, it seems like a footgun. As I said here I am fine with build failure: https://github.com/bluealloy/revm/pull/252#issuecomment-1301929069

Having this would make this more user-friendly: https://doc.rust-lang.org/std/macro.compile_error.html

Something like
```
#[cfg(all(not(feature = "k256_ecrecover"), not(feature = "secp256k1")))]
compile_error!(
    "To support ecrecover precompile please enable one of these two features, `k256_ecrecover` or `secp256k1`
);
```

Wodann: Added the above mentioned `compile_error` to the `secp256k1` module

rakita:
to unblock this I decided to have `k256` as default and have only `secp256k1` feature.

Reasonng behind  this can be found here:
https://github.com/bluealloy/revm/pull/256/files#diff-dddad4f9af653f2ddf5fb5cd993a7873baed17e3ca758f4acf205a529663bd24R29-R32

@Wodann please tell me if this works for you

Wodann:
The problem by reverting those changes (even with a default feature set) is that when people compile with `--no-default-features` they get non-descript compiler errors that they have to chase to realise that they don't have the proper flag. A crate should always succeed compilation (unless there is a `compile_error!`), even if default features are disabled. Those were avoided by the `#[cfg(..)` macros that I added in this PR originally

rakita:
> The problem by reverting those changes (even with a default feature set) is that when people compile with `--no-default-features` they get non-descript compiler errors that they have to chase to realise that they don't have the proper flag. A crate should always succeed compilation (unless there is a `compile_error!`), even if default features are disabled. Those were avoided by the `#[cfg(..)` macros that I added in this PR originally

Totally agreeing with this, I had bad choice of the word `default`, I meant as `k256` is not optional and `seckp256k1` is a feature flag that would use C lib instead of `k256`. It will always compile.

rakita:
@Wodann does this work for you?

Wodann:
Yup, LGTM! Thanks for clarifying 😊

Issue #255, opened 2022-11-07
Tx unexpectedly reverts with "out of gas" reason


Owen66:
Hi,

When processing the below tx with the revm I get an out of gas revert?
https://snowtrace.io/tx/0x5d9999e9e8ccdea2d87736cc39c38f51863f6f23169ac32ef2d9a1842aa94c23

Gas limit on the tx is 348,243 and gas used is 348,243. Seems strange to me that it fails with revm but doesn't fail on avalanche?
Avalanche just use a copy of geth for this stuff afaik.

gd-0:
Just to add some extra info to this.

It seems like every "withdraw" call to this contract: https://snowtrace.io/address/0xf6ccf601bd024612aaf85440153c2df0524e4607 (same fn and contract called by the 0x5d99.. tx linked above) costs 3 GAS more using the REVM compared to Avalanche Coreth.

rakita:
That is suspicious

gd-0:
Some more info:

Here are 4 txs that cost exactly 3 extra gas using the revm vs Avalanche Coreth/ Sputnik EVM:
```
- 0x2a592ab55d9823aa6e25386a361569ae74c274d5352214632505daefdfd885c3
- 0xe6f7f5fe3287e922bf6ecc17104d243bb6c9234574bba2b01c90b90fd883a77f
- 0x5d9999e9e8ccdea2d87736cc39c38f51863f6f23169ac32ef2d9a1842aa94c23
- 0x832e5a4b5f3596eeba4c88aef567a73fd50cebb5d2d46a4710ee1ef1e36428fc
```

I ran each of these txs with REVM and Sputnik logging each opcode that was used and the number of time each was used. Here are the differences:

```
1) 
Different counts! Opcode: 91. REVM count: 4756. SPUTNIK count: 4700
REVM contains an extra opcode: 254. COUNT: 9
Different counts! Opcode: 0. REVM count: 10. SPUTNIK count: 8 

2) 
Different counts! Opcode: 91. REVM count: 1465. SPUTNIK count: 1446
REVM contains an extra opcode: 254. COUNT: 2
Different counts! Opcode: 0. REVM count: 10. SPUTNIK count: 8

3) 
Different counts! Opcode: 0. REVM count: 8. SPUTNIK count: 6
REVM contains an extra opcode: 254. COUNT: 9
Different counts! Opcode: 91. REVM count: 5361. SPUTNIK count: 5294

4)
[...]

PR #259, opened 2022-11-11
Bugfix: if returndatacopy is len 0 return after initial cost

gd-0:
Fixes this issue: https://github.com/bluealloy/revm/issues/255 "Tx unexpectedly reverts with "out of gas" reason"


rakita:
unrelated eth test is failing that is merged yesterday: https://github.com/ethereum/tests/blob/639af20dfbe961376d75ce3596182a777d02c417/src/GeneralStateTestsFiller/stTransactionTest/HighGasPriceFiller.yml#L1

it is edge case where `gas_price` is a big number

PR #264, opened 2022-11-14
fix_gas_charge

Owen66:
While analysing this tx https://snowtrace.io/tx/0xac70b6e2e046be765ac18c1b4e74a342015575eebf778f49eaee5ed1565903c1
I noticed that on chain the this tx reduced the balance on account `0x77420e96f0e5f1dcfa8fcdd6ba6a3a9d3ad0c6c4` by `12510080000000000`, while the revm reduced the balance by only `12435880000000000`.

The seems to a refund being deducted in revm. 2800 * (25000000000 + 1500000000).

This line also seems to indicate that the refund is already included in the spend so shouldn't be added to remaining above?
https://github.com/bluealloy/revm/blob/main/crates/revm/src/evm_impl.rs#L277

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -249,7 +249,7 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB, 
             acc_caller.info.balance = acc_caller
                 .info
                 .balance
-                .saturating_add(effective_gas_price * U256::from(gas.remaining() + gas_refunded));
+                .saturating_add(effective_gas_price * U256::from(gas.remaining()));
 
             // EIP-1559
             let coinbase_gas_price = if SPEC::enabled(LONDON) {


rakita:
how did you get that 2800 and both 250000.. and 1500.. are those println? diff is 12510080000000000-12435880000000000=74200000000000

and effective gas price is:0.000000025 (base fee) + 0.0000000015 (max priority fee)= 0.0000000265

0w3n-d:
> how did you get that 2800 and both 250000.. and 1500.. are those println? diff is 12510080000000000-12435880000000000=74200000000000
> 
> and effective gas price is:0.000000025 (base fee) + 0.0000000015 (max priority fee)= 0.0000000265

Yeah I printed out those values. 
I got `gas.spend()` = `94,720` and `gas_refunded` = `2,800` for this line https://github.com/bluealloy/revm/blob/main/crates/revm/src/evm_impl.rs#L252

So I end us with `(94,720 + 2,800) * effective_gas_price` being added back here.
Seems to be too much being added back.

The the function returns `(91920, 2800)` as `gas used` and `gas refunded` ?
https://github.com/bluealloy/revm/blob/main/crates/revm/src/evm_impl.rs#L277

Owen66:
So it takes away the limit at the start `97,957` then adds back remaining `3,237` plus refund `2,800`. Gas charged is `91,920` times the effective price.

With geth the gas charged is total used `94,720` times the effective price. So the diff is the extra refund amount using the revm.

Owen66:
Turns out this is an avalanche thing. revm is correctly including the refund but avalanche doesn't include refunds.

rakita:
You can add a config flag inside `env` to turn refunds off, that would be acceptable (just add good comment why it is introduced)

PR #265, opened 2022-11-14
Export StorageSlot

ufoscout:
I am evaluating the migration from SputnikVM to REVM for a project of my company. 
Anyway, I am currently unable to migrate some of our unit tests because the Account storage is defined as `Map<U256, StorageSlot>` but the `StorageSlot` is not visible outside of this crate. This PR would fix the issue by exporting it.


rakita (approved):
ty

ufoscout:
Thank you, would it be possible to release it?

rakita:
published

PR #267, opened 2022-11-15
disable gas refunds with env flag

gd-0:
Resolution for this PR: https://github.com/bluealloy/revm/pull/264


crates/revm/src/evm_impl.rs
```diff
@@ -185,6 +185,7 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact
             match exit_reason {
                 return_ok!() => {
                     gas.erase_cost(ret_gas.remaining());
+                    #[cfg(not(feature = "optional_gas_refund"))]
```
rakita, line 188: You introduced a config flag, but you are not using it :D
And I think a better place for this is where the refund calculation is done in the `finalize` function.

gd-0: + actually using cfg flag 😅 
+ moved to finalize

PR #276, opened 2022-11-22
feat: eip 1153

tynes:
[EIP-1153](https://eips.ethereum.org/EIPS/eip-1153) is now CFI for Shanghai. I prototyped an implementation that I'd like to test with in foundry. If the implementation was cleaned up, would you consider this for merge?

### DIFF
--- a/bins/revme/src/statetest/models/spec.rs
+++ b/bins/revme/src/statetest/models/spec.rs
@@ -23,6 +23,7 @@ pub enum SpecName {
     Merge,             //done
     #[serde(alias = "Merge+3540+3670")]
     MergeEOF,
+    Shanghai,
 }
 
 impl SpecName {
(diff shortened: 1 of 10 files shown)


rakita:
I am surprised it is this simple and congrats on moving the point and making EIP-1153 CFI.
There is 4 CFI atm, it still needs to be decided what is going in, I will wait for that decision.

tynes:
It has been implemented and merged into a bunch of clients:
- https://github.com/NethermindEth/nethermind/pull/4126
- https://github.com/ethereumjs/ethereumjs-monorepo/pull/1768
- https://github.com/ethereum/go-ethereum/pull/26003

Still waiting merge:
- https://github.com/ledgerwatch/erigon/pull/6133
- https://github.com/hyperledger/besu/pull/4118

There is also test coverage in `ethereum/tests` https://github.com/ethereum/tests/pull/1091

The plan is to include it in the upcoming shangdong testnet, will need to merge here if reth is planning on participating

rakita:
I will close this as there is no indication when this EIP is going to be included in mainnet

rakita:
> Perhaps it makes sense to revisit this now that this EIP is to be included?

Too early to start working on it now, but I am glad it got included.

tynes:
When would be a good time to start on it then? Its already merged in all other clients. This would help reth be ready for a cancun testnet

rakita:
> When would be a good time to start on it then? Its already merged in all other clients. This would help reth be ready for a cancun testnet

Dont worry about, it will be delivered, but be mindful that I do work on other things

0xJepsen:
I'd like to help work on this if possible when the time is right

snreynolds:
This would be really great to get in so ppl can start to play around with it along with a custom solc binary. Is there any updates. Happy to help where I can also!

PR #278, opened 2022-11-26
native bits

rakita:
wip


gakonst:
got a flamegraph for perf comparison?

rakita:
Times are mostly similar, no diff present. It just allows me for revm to have its own bits (array of bits) and I don't need to include primitive_types as they have more than revm needs and ruint bits are an array of u64 that is not expected to be. This gives me more control over future improvements.

If you want to test some perf, look at this: https://github.com/bluealloy/revm/tree/inst_fn_same_sig :)

PR #283, opened 2022-12-04
feat(interpreter): Unify instruction fn signature

rakita:
All opcode instructions are now functions  with same signature:
`pub fn jumpi(interpreter: &mut Interpreter, _host: &mut dyn Host)` 

This is revm internal change and it does not affect external API.

This allows the compiler to optimize calls in the interpreter hot loop and it gives around ~10% faster `snailtracer` execution on i7. I am interested to see if the same boost can be found on macs with `M` processors.

This is one of steps needed to define multiple dispatch strategies related to interpreter loop. As instructions now have same signature they can be put in an array, this can have potentially new ways to use revm as it is easier to just switch function pointer to something else (custom function to sload/sstore?)

`Spec` was simplified and `is_static` is moved as variable to interpreter (Previously it was inside Spec)

There was one regression with ruint in casting U256 to saturated usize [here](https://github.com/bluealloy/revm/blob/acdbaaca5d110c32bdd1f7ecd7974a306a40a5e2/crates/revm/src/instructions/macros.rs#L236-L253). It was faster to just check limbs.

`microbench` added to `revm-tests`


rakita:
new run (with microbench)
```
Snailtracer benchmark (3.4s) ...  41_297_247.647 ns/iter (0.999 R²)
```

old run:
```
elapsed: 47.478018ms
0: 47.639335ms
1: 47.21674ms
2: 47.112436ms
3: 46.998897ms
4: 46.768525ms
5: 46.722415ms
6: 46.711912ms
7: 46.705151ms
8: 46.658515ms
9: 46.649658ms
10: 46.642895ms
11: 46.631915ms
12: 46.622408ms
13: 46.618776ms
14: 46.617359ms
15: 46.612672ms
16: 46.605036ms
17: 46.58608ms
18: 46.543365ms
19: 46.531848ms
20: 46.524966ms
21: 46.523186ms
22: 46.521023ms
23: 46.514446ms
24: 46.496036ms
end!
```

There is not a lot that can be seen from flamegraph, but there they are:

flamegraph new:
revm_unification


flamegraph old:
revm_main_before_unitication

0xDmtri:
Test info for rakita/unify_instr_fn:
```
Hello, world!
Snailtracer benchmark (3.1s) ...  33_138_990.401 ns/iter (0.999 R²)
end!
```

Test info for main:

```
Hello, world!
elapsed: 35.072378ms
0: 35.08125ms
1: 35.058541ms
2: 34.864458ms
3: 34.665416ms
4: 34.451916ms
5: 34.328375ms
6: 34.279208ms
7: 34.239041ms
8: 34.213ms
9: 34.201791ms
10: 34.199416ms
11: 34.196083ms
12: 34.192875ms
13: 34.191083ms
14: 34.188583ms
15: 34.1665ms
16: 34.153166ms
17: 34.152708ms
18: 34.127666ms
19: 34.097875ms
20: 34.091416ms
21: 34.091083ms
22: 34.085625ms
23: 34.074ms
24: 34.054541ms
end!
```

Cargo.toml
```diff
@@ -8,13 +8,10 @@ members = [
 default-members = ["crates/revm"]
 
 [profile.release]
-# debug = true
+debug = true
```
gakonst, line 11: Remove?

bins/revm-test/src/bin/snailtracer.rs
```diff
@@ -23,21 +23,12 @@ pub fn simple_example() {
...
     evm.env.tx.data = Bytes::from(hex::decode("30627b7c").unwrap());
 
-    let mut elapsed = std::time::Duration::ZERO;
-    let mut times = Vec::new();
-    for _ in 0..30 {
-        let timer = Instant::now();
+    // Microbenchmark
+    let bench_options = microbench::Options::default().time(Duration::from_secs(3));
+
+    microbench::bench(&bench_options, "Snailtracer benchmark", || {
```
gakonst, line 29: Are you sure the speedup you're seeing isn't bench framework related for any reason?

bins/revm-test/src/bin/snailtracer.rs
```diff
@@ -23,21 +23,12 @@ pub fn simple_example() {
...
         let (_, _) = evm.transact();
```
gakonst, line 30: Should this be black boxed?

crates/revm/src/evm_impl.rs
```diff
@@ -167,8 +166,9 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact
                     input: data,
                     gas_limit,
                     context,
+                    is_static: false,
```
gakonst, line 169: Isn't this slower than putting at compile time? Can't we have a simpler sig that get us the advantage of this PR without putting staticcheck at runtime?

crates/revm/src/instructions.rs
```diff
@@ -74,175 +72,165 @@ pub enum Return {
...
+        opcode::BYTE => bitwise::byte(interp, host),
+        opcode::SHL => bitwise::shl::<S>(interp, host),
+        opcode::SHR => bitwise::shr::<S>(interp, host),
+        opcode::SAR => bitwise::sar::<S>(interp, host),
+        opcode::SHA3 => system::sha3(interp, host),
+        opcode::ADDRESS => system::address(interp, host),
+        opcode::BALANCE => host::balance::<S>(interp, host),
+        opcode::SELFBALANCE => host::selfbalance::<S>(interp, host),
+        opcode::CODESIZE => system::codesize(interp, host),
+        opcode::CODECOPY => system::codecopy(interp, host),
```
gakonst, line 120: Do you for sure know that the specific change is what introduced the speed improvement? What about making the PR only change the benchmarks first, and then check the is_static change in one PR then this function sig change?

crates/revm/src/instructions.rs
```diff
@@ -74,175 +72,165 @@ pub enum Return {
...
+        opcode::LOG0 => host::log::<0, S>(interp, host),
+        opcode::LOG1 => host::log::<1, S>(interp, host),
+        opcode::LOG2 => host::log::<2, S>(interp, host),
+        opcode::LOG3 => host::log::<3, S>(interp, host),
+        opcode::LOG4 => host::log::<4, S>(interp, host),
+        opcode::SELFDESTRUCT => host::selfdestruct::<S>(interp, host),
+        opcode::CREATE => host::create::<false, S>(interp, host), //check
+        opcode::CREATE2 => host::create::<true, S>(interp, host), //check
+        opcode::CALL => host::call::<S>(interp, host),            //check
+        opcode::CALLCODE => host::call_code::<S>(interp, host),   //check
```
gakonst, line 230: Makes sense that the fn over enum is faster here

crates/revm/src/instructions.rs
```diff
@@ -74,175 +72,165 @@ pub enum Return {
...
+pub fn return_invalid(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    interpreter.instruction_result = Return::InvalidOpcode;
+}
+
+pub fn return_not_found(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    interpreter.instruction_result = Return::OpcodeNotFound;
+}
+
 #[inline(always)]
-pub fn eval<H: Host, S: Spec>(opcode: u8, interp: &mut Interpreter, host: &mut H) -> Return {
```
gakonst, line 78: Why no return?

crates/revm/src/instructions/arithmetic.rs
```diff
@@ -1,45 +1,59 @@
 use super::i256::{i256_div, i256_mod};
-use crate::{gas, Interpreter, Return, Spec, U256};
+use crate::{gas, Host, Interpreter, Return, Spec, U256};
 
-pub fn div(op1: U256, op2: U256) -> U256 {
-    op1.checked_div(op2).unwrap_or_default()
+pub fn wrapped_add(interpreter: &mut Interpreter, _host: &mut dyn Host) {
```
gakonst, line 4: Are we Ok with dyn dispatch in all opcodes? Looks like we have it inside the it the interp loop and it may be impactful for performance?

crates/revm/src/instructions/arithmetic.rs
```diff
@@ -57,79 +71,13 @@ pub fn eval_exp<SPEC: Spec>(interp: &mut Interpreter) -> Return {
...
-#[cfg(test)]
-mod tests {
-    use alloc::vec;
-
-    use super::{signextend, U256};
-
-    /// Test to ensure new (optimized) `signextend` implementation is equivalent to the previous
-    /// implementation.
-    #[test]
-    fn test_signextend() {
```
gakonst, line 86: Where did this test go?

crates/revm/src/instructions/bitwise.rs
```diff
@@ -1,80 +1,135 @@
...
+    check!(interpreter, SPEC::enabled(CONSTANTINOPLE));
+    pop_top!(interpreter, op1, op2);
+    *op2 <<= as_usize_saturated!(op1);
 }
 
-pub fn shr(shift: U256, value: U256) -> U256 {
-    value >> usize::try_from(shift).unwrap_or(256)
+pub fn shr<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    // EIP-145: Bitwise shifting instructions in EVM
+    check!(interpreter, SPEC::enabled(CONSTANTINOPLE));
```
gakonst, line 103: Can we do a static analysis pass at the bytecode in the beginning for the existence of these opcodes against the spec, instead of doing them in the loop?

gakonst (changes requested):
Love this PR as it has a ton of useful changes. Could we please isolate them in separate PRs, starting with the bench change, followed by the individual opcode changes, followed by the function signature unification? That way we'll know for sure what change caused what performance gain.

bins/revm-test/src/bin/snailtracer.rs
rakita: Yes, I am sure, this is the second iteration of unification that I have done, both with similar results.

bins/revm-test/src/bin/snailtracer.rs
rakita: Should not matter? it is called for 3s and not N times and for some warmup.

crates/revm/src/evm_impl.rs
rakita: It is, it does not matter a lot, it is just a  few `if` and diff is small

crates/revm/src/instructions.rs
rakita: Yes, i tested it while working on it. And this is second time I did this change. Similar results.

crates/revm/src/instructions.rs
rakita: For snailtracer it is not even called. Interpreter loop and maybe even `match` faster when it calls all same functions.

crates/revm/src/instructions.rs
rakita: Not needed, `interpreter_result` is changed only when loop needs to stop, for most instructions it was same `Result::Continue`. i didn't get big perf by removing this but it is cleaner.

crates/revm/src/instructions/arithmetic.rs
rakita: Good point! Will return it back

crates/revm/src/instructions/bitwise.rs
rakita: Will think about it, it is small thing as branch prediction does a great job, but will have this idea in my head, it does make sense.

crates/revm/src/instructions/arithmetic.rs
rakita: That is the point, the compiler seems to handle it better when all of call inside `match` has the same sig.

rakita:
thanks for the review @gakonst!

I experimented with this idea (and a few others) 3 months ago and got pretty decent results, this was one of them. Will try to play with additional dispatch strategies but I am assuming that benefit will not be significant but would potentially allow more creative ways to use revm (as in injecting the instruction fn, I will need to see how it goes).

So this pull request is what I expected to happen it was not surprising.

crates/revm/src/instructions/arithmetic.rs
gakonst: Very interesting

gakonst:
That makes sense - your PR of course, no strong opinions. Love the outcome.

gakonst:
@ziyadedher can you rerun https://github.com/ziyadedher/evm-bench with latest main? ^^

ziyadedher:
@gakonst ooh cool! Yeah, will do tonight.

rakita:
I have one more round of small optimizations, but this should give a good boost in performance.

ziyadedher:
I also see ~10% bumps on snailtracer! But interestingly, I don't see any noticeable differences on other benches, see https://github.com/ziyadedher/evm-bench/commit/aad8de9d7de5a301f834c45a2983a5d8646ea1c5

This might make sense since the optimization is primarily in dispatches and whatnot it seems. Just wanted to float the results!

rakita:
I assumed it would be more, around 15%. Yeah, it is mostly dispatch optimization and interpreter loop, I will play with it a little bit more in future but I don't expect a lot.

revm is a little bit different from other evm's as it has both the Host and Interpreter together. This means that for test  execution, SSLOAD and SSTORE access two HashMaps to find value, one for substate and one for cached value, If I remove this (planned here: https://github.com/bluealloy/revm/issues/307), test for revm will be more accurate.

PR #286, opened 2022-12-07
EIP 3155 tracing

shemnon:
Partial implementation of 3155 tracing.

This covers the critical fields "pc", "op", "gas", "gasCost", "memSize", "stack", "depth", "opName".
gasCost is known to be wonky.


rakita:
Merged here: https://github.com/bluealloy/revm/pull/356

PR #287, opened 2022-12-07
bug: Overflow while calculating the remaining gas in GasInspector

rakita:
`reduced_gas_block` was not reset.


rakita:
tests fails bcs of `--all-features`. tests locally

PR #297, opened 2022-12-16
feat: allow disabling of balance checks

Wodann:
Depends on/incorporates #256

Ethereumjs has the ability to disable balance checks. When this toggle is on and an account does not have sufficient balance to complete a transaction, the transaction will still be completed and the balance set to `0` (zero).

### DIFF
--- a/crates/revm/Cargo.toml
+++ b/crates/revm/Cargo.toml
@@ -42,12 +42,14 @@ sha3 = { version = "0.10", default-features = false, features = [] }
 default = ["std", "secp256k1"]
 dev = [
     "memory_limit",
+    "optional_balance_check",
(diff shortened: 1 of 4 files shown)


0xDmtri:
Oh yeah that would be sweet!

rakita:
I m OK with adding this but remove the dependency on the previous PR (I addressed the reason there).

For the future I feel like we can have someking unified flags for initial check, that would potentially disable them all, but lets leave this for later, as I am not sure how exactly this would look.

Wodann:
> but remove the dependency on the previous PR

Done!

rakita (approved):
lgtm

PR #299, opened 2022-12-17
refactor: split block_hash function into Blockchain trait

### LINKED ISSUE #298: Can we split `block_hash` into a separate `Blockchain` trait?
In our implementation of Project Rethnet for Hardhat, we're running into an issue because we have separate data models for the blockchain (blocks) and state management (accounts, account storage, code) when implementing the `Database[Ref]` traits.

A possible solution would be to separate that trait into two traits: `Database[Ref]` for state queries and a `Blockchain` trait that directly deals with the blockchain for retrieving block hashes by their block number`.

Is this something that you think would make sense for `revm`'s internals?

If so, would you want to have two separate containers for caching internally?

Wodann:
Potentially resolves #298

### DIFF
--- a/bins/revm-test/src/bin/analysis.rs
+++ b/bins/revm-test/src/bin/analysis.rs
@@ -1,7 +1,7 @@
 use std::time::Instant;
 
 use bytes::Bytes;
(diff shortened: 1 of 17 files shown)


rakita:

PR #312, opened 2022-12-28
add `revm-primitives` crate

joshieDo:
Moves some things around so we can have a `revm-primitives` crate.

Adds `proptest::Arbitrary` and `arbitrary::Arbitrary` on `#cfg[test]` or `feature =arbitrary` to `B160`, `B256` and `I256`.


rakita:
I have already started splitting revm into interpreter and host here: https://github.com/bluealloy/revm/tree/revm_interpreter/crates
`revm-primitives` should additionally extract types from interpreter.

gakonst:
Sweet. To be clear, let's still have Joshie finish this refactor for the primitives, so we can unblock reth, and you can continue the interpreter refactor after primitives is done? So we minimize blocking?

rakita:
@gakonst how does this sound to you: https://github.com/paradigmxyz/reth/pull/513#issuecomment-1367237256 ?

The only blocking thing for microbenchmarking is having `arbitrary` added, we can make this change directly to `main`.

Intention for `primitives` is to have additional types not just `U256/B256/I256/etc`, I still need to find the balance of what exactly is going inside, that is why I am taking it step by step and first extracted interpreter to get more sense of what is used where.

joshieDo:
I'm fine with that , the bulk of changes I'm facing are actually in `reth` with changing the `U256` type and `B` types because of their interface differences.

Will open a PR into interpreter just to add `arbitrary` feature to the `B160` and `B256` types, and continue the work on the `reth` side.

PR #318, opened 2023-01-06
feat: return Database::error

### LINKED ISSUE #309: Return more sane value that replaces `Return`
`Return` is a enum and it is internal structure, exposing it outside of revm creates not that great interface. 

There are four states that can be returns:
* `Success`: Transaction executed and it was success
* `Failure`:  Transaction executed but it failed: Check this: https://eips.ethereum.org/EIPS/eip-658
* `Error`: Transaction __not__ executed. Not enough balance in account to cover max gas consumes. 
* `ExternalError`: Transaction __not__ executed, external database error.

Wodann:
First step to resolve #309 

I wanted to get this under your eyes, to see what you think. This only aims to bubble up the actual `FatalExternalError`s that should not result in any gas cost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1583,6 +1583,7 @@ dependencies = [
  "revm-interpreter",
  "revm-precompiles",
  "serde",
+ "thiserror",
  "tokio",
 ]
 
--- a/bins/revm-test/src/bin/analysis.rs
+++ b/bins/revm-test/src/bin/analysis.rs
@@ -31,12 +31,12 @@ fn main() {
 
     // just to spead up processor.
     for _ in 0..10000 {
-        let (_, _) = evm.transact();
+        let (_, _) = evm.transact().expect("Transaction should succeed");
     }
 
     let timer = Instant::now();
     for _ in 0..30000 {
-        let (_, _) = evm.transact();
+        let (_, _) = evm.transact().expect("Transaction should succeed");
     }
     println!("Raw elapsed time: {:?}", timer.elapsed());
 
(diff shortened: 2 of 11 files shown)


rakita:
This is not optimal as there is a lot of boilerplate code and if you are in 1000 call depth you would copy/move the DB error 1000 times. That is why I have only the simple stupid enum `Return` so I am moving only one byte and checking only one byte in the interpreter hot loop.

and is `Debug` no_std, I want revm to be able to be compiled for wasm?

Wodann:
To make better usage of the memory, we can include the `Reason` in the `Result::Ok` to share the memory. That would allow us to remove it from the interpreter.

Also, the compiler should be smart enough to move the memory in one go for the `Reason` instead of moving it a 1000x to. If I'm not mistaken, me flagging it as `Infallible` for some paths should even result in complete optimisation of the code path. So for a `Database::Error = Infallible`, it should have no performance overhead. Benchmarks would be required to confirm that, though

Wodann:
> and is Debug no_std, I want revm to be able to be compiled for wasm?

I can include `core::fmt::Debug` instead

rakita:
Either way, `Reason` change should be localized, and this PR has a lot of unnecessary changes to the code.

Wodann:
Which changes would you say are unnecessary?

rakita:
I never intended to remove `Return` but just to cast it to something more reasonable to external call (as in from `transact`).

In the interpreter, `Return` is not made idiomatic and the reason why I didn't use `?` is there were reports of it being slower and LLVM was not optimizing it: https://agourlay.github.io/rust-performance-retrospective-part3/
 
Other than that you are adding more code to every instruction to fit `Result` pattern, it is unnecessary.

Wodann:
I made a minimal implementation, as (I think) you envisioned. This should be a lot less intrusive

rakita:

PR #319, opened 2023-01-07
fix: broken feature flags

Wodann:


crates/revm/src/inspector.rs
```diff
@@ -6,7 +6,6 @@ use crate::{
 use auto_impl::auto_impl;
 use bytes::Bytes;
 
-#[cfg(feature = "std")]
```
rakita, line 9: customprinter uses `println!` that is why it is behind `std`.

rakita (approved):
lgtm. just revert std for customprinter

Wodann:
> lgtm. just revert std for customprinter

Gotcha. Then I'll add the `std` feature flag in the `pub use` underneath as well

Wodann:
Re-inserted the `std` feature flag

PR #323, opened 2023-01-09
chore(interpreter): enable `proptest` with `arbitrary` feature

joshieDo:
Also bumps `proptest-derive` since it was wrongly set to a previous version


rakita (approved):
will merge after tests

PR #324, opened 2023-01-09
refactor: split Database trait into sub-traits

### LINKED ISSUE #298: Can we split `block_hash` into a separate `Blockchain` trait?

Wodann:
Resolves #298 based on the design discussed

Alternative to #299

### DIFF
--- a/crates/revm/src/db.rs
+++ b/crates/revm/src/db.rs
@@ -40,8 +40,8 @@ where
 }
 
 pub struct DatabaseComponents<BH: BlockHash, S: State> {
-    block_hash: BH,
-    state: S,
+    pub block_hash: BH,
+    pub state: S,
 }
 
 pub enum ComponentError<BHE, SE> {
--- a/crates/revm/src/db.rs
+++ b/crates/revm/src/db.rs
@@ -2,84 +2,91 @@ mod in_memory_db;
 
 #[cfg(feature = "ethersdb")]
 pub mod ethersdb;
-#[cfg(feature = "ethersdb")]
-pub use ethersdb::EthersDB;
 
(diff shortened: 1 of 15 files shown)


crates/revm/src/db.rs
```diff
@@ -2,84 +2,91 @@ mod in_memory_db;
...
+
+use crate::{
+    blockchain::{BlockHash, BlockHashRef},
+    state::{State, StateRef},
+    StateCommit,
+};
+
+pub use self::in_memory_db::{AccountState, BenchmarkDB, CacheDB, DbAccount, EmptyDB, InMemoryDB};
+
+#[impl_tools::autoimpl(for<T: trait> &mut T, Box<T>)]
```
Wodann, line 24: `auto_impl` didn't work when the trait's associated types has trait bounds, so I had to use `impl_tools` instead

crates/revm/src/db.rs
```diff
@@ -2,84 +2,91 @@ mod in_memory_db;
...
+pub trait Database: BlockHash + State {
+    type DatabaseError: From<<Self as BlockHash>::Error> + From<<Self as State>::Error>;
```
Wodann, line 26: A limitation of this approach it that these constraints will pose problems for `DatabaseComponents` when the two error types are the same

rakita:

PR #328, opened 2023-01-12
`revm`: mark `with-serde` feature as deprecated

### LINKED ISSUE #327: `compile_error!` on deprected `with-serde` feature flag
Put it in revm lib file as `with-serde` was renamed to `serde`

Evalir:
Closes #327.

Marks the `with-serde` feature as deprecated, pointing out that it has been renamed to `serde.`

### DIFF
--- a/crates/revm/Cargo.toml
+++ b/crates/revm/Cargo.toml
@@ -54,4 +54,4 @@ ethersdb = ["tokio", "futures", "ethers-providers", "ethers-core"]
 serde = ["dep:serde", "hex/serde", "hashbrown/serde", "revm-interpreter/serde"]
 # deprecated feature
 web3db = []
-with-serde = ["dep:serde", "hex/serde", "hashbrown/serde", "revm-interpreter/serde"]
+with-serde = []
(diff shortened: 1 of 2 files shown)


crates/revm/Cargo.toml
```diff
@@ -54,3 +54,4 @@ ethersdb = ["tokio", "futures", "ethers-providers", "ethers-core"]
 serde = ["dep:serde", "hex/serde", "hashbrown/serde", "revm-interpreter/serde"]
 # deprecated feature
 web3db = []
+with-serde = ["dep:serde", "hex/serde", "hashbrown/serde", "revm-interpreter/serde"]
```
rakita, line 57:
```suggestion
with-serde = []
```

No need for other

crates/revm/src/lib.rs
```diff
@@ -6,6 +6,9 @@ mod evm_impl;
 mod inspector;
 mod journaled_state;
 
+#[cfg(feature = "with-serde")]
```
rakita, line 9:
```suggestion
#[cfg(all(feature = "with-serde", not("serde" )]
```

Added `all (not serde), (with serde)` as tests turn all features on and this compile error would trigger.

rakita (approved):
Merge suggestions and it is looking good!

crates/revm/src/lib.rs
Evalir: ahh mb! overlooked this

PR #330, opened 2023-01-13
fix: feature flags

Wodann:
Propagates features to `revm-interpreter`


Wodann:
@rakita, FYI you'll need to do something similar for the `primitives` crate in your branch: https://github.com/bluealloy/revm/tree/primitives

rakita:
Yeah! Thank you for noting it!

PR #339, opened 2023-01-23
refactor: simplify DatabaseComponentError

Wodann:


rakita:
ref: https://github.com/bluealloy/revm/issues/309#issuecomment-1399269000

PR #341, opened 2023-01-25
refactor: simplify DatabaseComponentError

Wodann:
Also makes fields public, such that an instance of the struct can be constructed


Wodann:
Created a duplicate by accident

PR #342, opened 2023-01-25
fix: disable balance check

Wodann:
In the refactor of `evm_impl.rs`, the behaviour of `disable_balance_check` changed. It was no longer setting the balance to zero when there is not enough balance in the caller's account.


rakita (approved):
Yeah, i was tinking, if we dont have enough balance for gas, then value if not zero would fail to transfer ether.

This is probably fine for usecase as gas limit can be u64::max, but I am wondering if we need to left the `value` amount of balance in account, just a thought.

Wodann:
We can add a test to validate that edge case

rakita:
> We can add a test to validate that edge case

It is just a side note, nothing more. If `disable_balance_check` is enabled and gas_limit*gas_price is high, any send value would make transaction revert as we made the balance zero.

PR #343, opened 2023-01-25
improvement: add logs & return value to revert

Wodann:
The EVM allows you to call the `revert` opcode with an error message, which needs to be returned as the return value.

Additionally, I added the logs that are collected up to the point `revert` was called. This allows debugging tools built on top of `revm` without adding additional cost - as we're just bubbling up the data.

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -24,7 +24,11 @@ pub enum ExecutionResult {
         output: Output,
     },
     /// Reverted by `REVERT` opcode that doesn't spend all gas.
-    Revert { gas_used: u64 },
+    Revert {
+        gas_used: u64,
+        logs: Vec<Log>,
+        return_value: Bytes,
+    },
     /// Reverted for various reasons and spend all gas.
     Halt {
         reason: Halt,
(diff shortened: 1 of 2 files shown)


crates/primitives/src/result.rs
```diff
@@ -24,7 +24,11 @@ pub enum ExecutionResult {
         output: Output,
     },
     /// Reverted by `REVERT` opcode that doesn't spend all gas.
-    Revert { gas_used: u64 },
+    Revert {
+        gas_used: u64,
+        logs: Vec<Log>,
```
rakita, line 29: Logs are getting reverted, they will always be empty, it does not make sense to have them here: https://github.com/bluealloy/revm/blob/ba393d75cf12ec0b1446fd5ce147b338de6bb662/crates/revm/src/evm_impl.rs#L738-L742

crates/primitives/src/result.rs
```diff
@@ -24,7 +24,11 @@ pub enum ExecutionResult {
...
+        return_value: Bytes,
```
rakita, line 30: And Maybe rename it to `output` as to have similar name to `Success::output`

Wodann: Fixed in latest commit

crates/primitives/src/result.rs
Wodann: Removed in latest commit

rakita (approved):

Issue #344, opened 2023-01-25
Help with exports


andy-thomason:
Excellent work folks.

Currently the only exports from `journaled_state` are JournalEntry and JournaledState.

https://github.com/bluealloy/revm/blob/main/crates/revm/src/lib.rs#L18

It would be useful to have Map and State and Storage as these are used as outputs to transact()
and will avoid a duplicate dependency on hashbrown.

rakita:
Hi Andy,

This deserves a little bit of cleanup. The best case would be to have hashbrown hashmap in one place inside `revm-primitives` and export them everywhere. And move away from `Map` alias to `HashMp`, just map is not intuitive what it is.

If you want to unblock yourself just expose the Map and it will get merged.

andy-thomason:
Ok. Good to start somewhere :wink:

andy-thomason:
It does look like `primitives` is the best place to expose this.

We already have `Account` and `StorageSlot` so only the `HashMap` is needed.

rakita:
some imports cleanup done here; https://github.com/bluealloy/revm/pull/348

PR #345, opened 2023-01-25
expose hashbrown::HashMap in primitives

andy-thomason:
Issue: #344 

`EVM::transact` returns `ResultAndState` which contains `type State = HashMap<B160, Account>;`

State is internal to `revm` and requires `hashbrown` to form the type in user-called code.

Likewise, `Account` contains:

```
    pub storage: HashMap<U256, StorageSlot>,
```

By exposing `HashMap` in primitives it provides user the ability to use this type independently and insulates them from changes to the type.

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -22,6 +22,7 @@ pub type Hash = B256;
 
 pub use bytecode::*;
 pub use env::*;
+pub use hashbrown::{hash_map, HashMap};
 pub use log::Log;
 pub use precompile::*;
 pub use result::*;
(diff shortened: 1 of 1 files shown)


crates/primitives/src/lib.rs
```diff
@@ -30,3 +30,4 @@ pub use ruint::aliases::U256;
 pub use specification::*;
 pub use state::*;
 pub use utilities::*;
+pub use hashbrown::HashMap;
```
rakita, line 33:
```suggestion
pub use hashbrown::{HashMap, hash_map};
```

rakita (approved):

PR #346, opened 2023-01-27
fix: underflow panic when using "no_gas_measuring" feature

ckoopmann:
When using the "no_gas_measuring" feature, I encountered various underflow panics which seem to stem from calls to the "erase_cost" method which should not be called in this case. 

This PR should fix that by wrapping those calls in an `if crate::USE_GAS` block analogous to other parts of the code.


rakita (approved):
Good catch!

PR #354, opened 2023-01-31
feat: Different OutOfGas Error types

### LINKED ISSUE #336: Different OOG error types
It would be nice for foundry (and likely reth) if there were a few different kinds of `OutOfGas` error types. For example, in https://github.com/foundry-rs/foundry/issues/3971 I had to patch foundry and add debugging lines to `revm` to figure out this was a memory limit OOG error.

Some variants may be:
```rust
pub enum OutOfGasError {
    // Basic OOG error
    OOG,
    // Tried to expand past REVM limit
    MemoryLimit,
    // Basic OOG error from memory expansion
    Memory,
    // Precompile threw OOG error
    Precompile,
    // When performing something that takes a U256 and casts down to a u64, if its too large this would fire
    // i.e. in `as_usize_or_fail`
    InvalidOperand
}
```

then the `Return::OutOfGas` would become `Return::OutOfGas(OutOfGasError::OOG)` or whatever variant best suits the error

chirag-bgh:
Fixes #336

### DIFF
--- a/crates/interpreter/src/instruction_result.rs
+++ b/crates/interpreter/src/instruction_result.rs
@@ -17,6 +17,10 @@ pub enum InstructionResult {
 
     // error codes
     OutOfGas = 0x50,
+    MemoryOOG = 0x51,
+    MemoryLimitOOG = 0x52,
+    PrecompileOOG = 0x53,
+    InvalidOperandOOG = 0x54,
     OpcodeNotFound,
     CallNotAllowedInsideStatic,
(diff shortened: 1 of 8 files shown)


rakita:
Nice, but we still need to use them, replacing old OutOfGas with new codes.

chirag-bgh:
@rakita replaced the old OOG errors.

rakita (approved):
nice!

PR #361, opened 2023-02-08
improvement: implement BlockHash for Arc<BlockHashRef>

Wodann:


crates/primitives/src/db/components/block_hash.rs
```diff
@@ -30,3 +30,16 @@ where
         BlockHashRef::block_hash(*self, number)
     }
 }
+
+#[cfg(feature = "std")]
```
rakita, line 34: There is `alloc::sync::Arc` but to be honest I don't know if it would make a problem and if they can be interchangeably used

rakita (approved):

PR #362, opened 2023-02-08
fix: call create_end for all code paths

Wodann:


crates/revm/src/evm_impl.rs
```diff
@@ -554,12 +554,23 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+                            bytes,
+                        )
+                    } else {
+                        (
+                            InstructionResult::CreateContractSizeLimit,
+                            ret,
+                            interpreter.gas,
+                            bytes,
+                        )
+                    };
```
rakita, line 573: This is not exactly the pattern used elsewhere

https://github.com/bluealloy/revm/blob/5d6ecd03ca8cf749b5f0d38a95fa628c32fc8387/crates/revm/src/evm_impl.rs#L405-L412

by default `create_end` would return `::Continue`

Wodann: Isn't Inspector::create_end supposed to be called for every possible reason the create function might end?

Wodann: > by default `create_end` would return `::Continue`

And doesn't the default implementation of `create_end` return `ret`?

https://github.com/bluealloy/revm/blob/5d6ecd03ca8cf749b5f0d38a95fa628c32fc8387/crates/revm/src/inspector.rs#L123

Wodann:
I made sure `create_end` is now called for all return code paths

rakita (approved):
nice pattern!

PR #363, opened 2023-02-08
feat: implement Debug for DatabaseComponentError if supported

Wodann:


crates/primitives/src/db/components.rs
```diff
@@ -2,6 +2,8 @@
 pub mod block_hash;
 pub mod state;
 
+use std::fmt::Debug;
```
rakita, line 5:
```suggestion
use core::fmt::Debug;
```

SuperFluffy:
When you `#[derive(Debug)]` for a type `Foo<T>` with generic type parameters, then the proc macro will implicitly add `impl<T: Debug> Debug for Foo<T>` for you. This means that the manual implementation of Debug is the same as the derived one. Proof:

```rust
use std::fmt::Debug;

#[derive(Debug)]
enum Foo<T, U> {
    State(T),
    BlockHash(U),
}

enum Bar<T, U> {
    State(T),
    BlockHash(U),
}

impl<T: Debug, U: Debug> Debug for Bar<T, U> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::State(arg0) => f.debug_tuple("State").field(arg0).finish(),
            Self::BlockHash(arg0) => f.debug_tuple("BlockHash").field(arg0).finish(),
        }
    }
}

fn main() {
    let foo_state: Foo<&str, &str> = Foo::State("Hello");
    let foo_blockhash: Foo<&str, &str> = Foo::BlockHash("World");
    let bar_state: Bar<&str, &str> = Bar::State("Hello");
    let bar_blockhash: Bar<&str, &str> = Bar::BlockHash("World");

    println!("{foo_state:?}");
    println!("{foo_blockhash:?}");
    println!("{bar_state:?}");
    println!("{bar_blockhash:?}");
}
```

Output:
```sh
❯ cargo run
   Compiling check_debug v0.1.0 (/home/janis/dev/scratchpad/check_debug)
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/check_debug`
State("Hello")
BlockHash("World")
State("Hello")
BlockHash("World")
```

rakita (approved):
few nits

SuperFluffy:
> @SuperFluffy does that still allow instances of the type that don't implement Debug for T?

Of course! The `#[derive(Debug)]` is only relevant for the actual `impl <Trait> for Type` block, it does not affect the type definition itself. The derive macro literally boils down to what you wrote. Here is the output of `cargo expand` for my example above:

```rust
enum Foo<T, U> {
    State(T),
    BlockHash(U),
}
#[automatically_derived]
impl<T: ::core::fmt::Debug, U: ::core::fmt::Debug> ::core::fmt::Debug for Foo<T, U> {
    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
        match self {
            Foo::State(__self_0) => {
                ::core::fmt::Formatter::debug_tuple_field1_finish(f, "State", &__self_0)
            }
            Foo::BlockHash(__self_0) => {
                ::core::fmt::Formatter::debug_tuple_field1_finish(f, "BlockHash", &__self_0)
            }
        }
    }
}
```

You are probably thinking of something like `struct Foo<T: Debug>`, which would indeed restrict `Foo` to only contain `T: Debug`. I wrote some code like that a few years ago and it was awful. :D

Wodann:
Good to know, thanks!

Wodann:
Applied all suggestions

rakita:
@SuperFluffy nice!

PR #380, opened 2023-02-15
Add B160/B256 From primitive_types traits

ufoscout:
REVM 3 does not use `primitive_types` anymore. Numerical types compatibility is already provided by `ruint` `primitive_types` feature. This PR adds type conversion between the hashes types (i.e. `B160` and `B256`) and their counterparts in `primitive_types`.


rakita (approved):
ty

PR #383, opened 2023-02-21
impl NonceTooHigh/ NonceTooLow checks

gd-0:
Implements missing nonce checks. 

Using a feature flag to disable the nonce checks was discussed in telegram but I noticed that the `nonce` parameter in the `TxEnv` struct was optional so I've put the check behind that. If this isn't what the Optional nonce was for then I can put it behind a feature flag, just let me know :D.

Matched to Geth: https://github.com/ethereum/go-ethereum/blob/fe01a2f63b8591d8226742726d8d6aaad4cd981e/core/state_transition.go#L240

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -102,8 +102,16 @@ pub enum InvalidTransaction {
     },
     /// Overflow payment in transaction.
     OverflowPaymentInTransaction,
-    /// Nonce overflows in transaction,
+    /// Nonce overflows in transaction.
     NonceOverflowInTransaction,
+    NonceTooHigh {
+        tx: u64,
+        state: u64,
+    },
+    NonceTooLow {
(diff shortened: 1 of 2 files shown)


rakita:
> Implements missing nonce checks.
> 
> Using a feature flag to disable the nonce checks was discussed in telegram but I noticed that the `nonce` parameter in the `TxEnv` struct was optional so I've put the check behind that. If this isn't what the Optional nonce was for then I can put it behind a feature flag, just let me know :D.
> 
> Matched to Geth: https://github.com/ethereum/go-ethereum/blob/fe01a2f63b8591d8226742726d8d6aaad4cd981e/core/state_transition.go#L240

I forgot that Nonce was optional, this is great!
Just fix the clippy and it will be merged

rakita (approved):

PR #393, opened 2023-02-28
feat: Add check for chainID

### LINKED ISSUE #386: Add check for `env.cfg.chain_id != env.tx.chain_id`
This check should be done in transact function and would return `EVMError` to mark an invalid transaction.

chirag-bgh:
Closes #386

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -99,6 +99,11 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
             return Err(InvalidTransaction::RejectCallerWithCode.into());
         }
 
+        // Check that the transaction's chain id is correct
+        if self.data.env.cfg.chain_id != U256::from(self.data.env.tx.chain_id.unwrap_or_default()) {
+            return Err(InvalidTransaction::InvalidChainId.into());
+        }
+
         // Check that the transaction's nonce is correct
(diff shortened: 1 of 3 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -99,6 +99,11 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
             return Err(InvalidTransaction::RejectCallerWithCode.into());
         }
 
+        // Check that the transaction's chain id is correct
+        if self.data.env.cfg.chain_id != U256::from(self.data.env.tx.chain_id.unwrap_or_default()) {
```
rakita, line 103: There are legacy transactions that does not have a chain_id. It is optional field and if  tx does not have it we don't need to check it. do `if let Some(tx_chain_id) = self.data.env.tx.chain_id {`

crates/revm/src/inspector/gas.rs
```diff
@@ -282,6 +282,7 @@ mod tests {
         evm.env.tx.transact_to =
             TransactTo::Call(B160(hex!("0000000000000000000000000000000000000000")));
         evm.env.tx.gas_limit = 21100;
+        evm.env.tx.chain_id = Some(1);
```
rakita, line 285:
```suggestion
```

rm it, it is good to test if chain_id is not set.

PR #398, opened 2023-03-04
feat: add b256<->u256 from impls

mattsse:
converting hash types to u256 is _incredibly_ annoying.

ethereum-types solved this via a BigendianHash trait

https://github.com/paritytech/parity-common/blob/27b17c0d5d32ebf384dd4b535d9045598904519d/ethereum-types/src/hash.rs#L18

however, I really would like to have From impls instead, so you don't have to do.

`U256::from_be_bytes(b256.0)` or `U256::from_be_bytes(b256.to_fixed_bytes())`

### DIFF
--- a/crates/primitives/src/bits.rs
+++ b/crates/primitives/src/bits.rs
@@ -54,6 +54,32 @@ impl From<B256> for primitive_types::H256 {
     }
 }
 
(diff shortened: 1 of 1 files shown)


rakita:
`From` impl looks better

PR #402, opened 2023-03-07
feat: use singular bytes for the jumpmap

onbjerg:
Switches from `usize` to `u8` for easier encoding, and adds some methods to build a jump map from a `u8` slice


rakita:
Is it better to have u8 or u64?

onbjerg:
There is probably not much of a difference performance wise, but encoding/decoding (which almost always is represented as some operation from X -> &[u8] and &[u8] -> X) will be a lot more annoying with u64, since you would need to turn slices of u64 to slices of u8, and slices of u8 to slices of u64.

rakita:
Yeah, will check in the future if there is some degradation (I assume there is none), and having u8 is easier to handle.

PR #412, opened 2023-03-09
fix(interpreter): halt on `CreateInitcodeSizeLimit`

rkrasiuk:
From [EIP 3860](https://eips.ethereum.org/EIPS/eip-3860#rules):

> If length of initcode to CREATE or CREATE2 instructions exceeds MAX_INITCODE_SIZE, instruction execution exceptionally aborts (as if it runs out of gas).


rkrasiuk:
@rakita this fixes the failing state tests

rakita:
Yes it does, initcode size check is done before the call is made, so the `InstructionResult` can be exposed to the first call, missed that.

PR #414, opened 2023-03-09
fix(db): preserve existing account state

rkrasiuk:
## Bug

If the account already existed, its state would get reset to `Touched` after any change.
This is not ideal for accounts which had their storage cleared.

## Fix
Preserve existing account state if any.


rakita (approved):
Amazing find!

PR #418, opened 2023-03-14
feat: change memory resize gas cost recording to deltas

ArtificialPB:
While debugging the following [issue](https://github.com/foundry-rs/foundry/issues/4523), I found out that memory gas cost is replaced with new total cost on each memory resize. This does not play nicely with gas metering functionality of `foundry` which expects gas cost deltas to be recorded instead.

I've prepared a minimal change which converts memory gas recording to use deltas as well, and produces expected result in the issue linked above. Added one additional `subtraction` and `addition` on `u64` compared to previous implementation so performance shouldn't be impacted too much.

If this gets merged I'll prepare a PR for `main` branch as well.

### DIFF
--- a/crates/revm/src/gas.rs
+++ b/crates/revm/src/gas.rs
@@ -65,15 +65,14 @@ impl Gas {
     }
 
     /// used in memory_resize! macro
-    pub fn record_memory(&mut self, gas_memory: u64) -> bool {
-        if gas_memory > self.memory {
-            let (all_used_gas, overflow) = self.used.overflowing_add(gas_memory);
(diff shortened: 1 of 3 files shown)


rakita:
@ArtificialPB this can be done inside `Inspector` and not `Interpreter`.

Somewhere here: https://github.com/bluealloy/revm/blob/main/crates/revm/src/inspector/gas.rs

ArtificialPB:
@rakita thank you for your feedback. The `GasInspector` might already contain everything we need to properly track gas usage. Will open a new PR if any additional requirement for it pops up.

PR #430, opened 2023-03-21
Comment Fix

Kuly14:
Isn't this supposed to be the last 20 bytes of the hash of the public key?

I might be wrong tho.


Kuly14:
Great, glad I could help!

PR #436, opened 2023-03-28
add feature for ignoring base fee check

Rjected:
This is analogous to the geth EVM `NoBaseFee` option:
https://github.com/ethereum/go-ethereum/blob/ee8e83fa5f6cb261dad2ed0a7bbcde4930c41e6c/core/vm/interpreter.go#L30

Which is used in `eth_call` and other RPCs that make calls, since they may have a gas price of zero:
https://github.com/ethereum/go-ethereum/blob/ee8e83fa5f6cb261dad2ed0a7bbcde4930c41e6c/internal/ethapi/api.go#L985

### DIFF
--- a/crates/interpreter/Cargo.toml
+++ b/crates/interpreter/Cargo.toml
@@ -38,6 +38,7 @@ dev = [
     "optional_block_gas_limit",
     "optional_eip3607",
     "optional_gas_refund",
+    "optional_no_base_fee",
 ]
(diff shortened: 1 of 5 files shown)


gakonst:
Seems reasonable

rakita (approved):
lgtm,
Will refactor this in future to separate tx validity check from other things, it would make this a lot easier to do.

will merge after CI goes green

PR #450, opened 2023-04-06
feat(revm): return journal entries

rakita:
expose and return internal journal entries as they can potentially help with the optimization.


kliao:
This seems useful for inspecting changes to state for a transaction. Just curious why this was closed instead of merged?

rakita:
> This seems useful for inspecting changes to state for a transaction. Just curious why this was closed instead of merged?

There can be multiple changes in journal in regards to same storage/account, this can get confusing and can be a footgun imo. And any account returned from revm can be checked if it got changed

Issue #449, opened 2023-04-06
`k256` version of `mod secp256k1` causes compile error


Wodann:
Upon compilation with `cargo t --all --no-default-features --features std`, the following compile error results:

```
error[E0432]: unresolved import `k256::ecdsa::recoverable`
  --> crates\precompile\src\secp256k1.rs:13:17
   |
13 |         ecdsa::{recoverable, Error},
   |                 ^^^^^^^^^^^ no `recoverable` in `ecdsa`
```

I tried fixing the issue in [this branch](https://github.com/Wodann/revm/commit/f4b9addc2082fdba1fb91581715f6af1e5f8f63f), but it's causing tests to fail due to the signature not matching the request. I'm not familiar with the crate though; my attempt was merely trying to mimic the `secp256k1` implementation.

rakita:
good catch! I really need to add CL build without features.

The function that you needed to use was `VerifyingKey::recover_from_prehash(` as the msg in precompile is considered hashed.

Wodann:
> On what versions of k256 are you on? 

k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }

This was the latest release of `revm`, if I'm not mistaken

Wodann:
Even if I try with `VerifyingKey::recover_from_prehash` I get an error in OpenZeppelin's contracts: `signature does not match request`

Wodann:
I'll give your PR a try to see whether it works

Wodann:
Update: Your PR fixes the problem correctly. I was missing the hash 😁

rakita:
> > On what versions of k256 are you on?
> 
> k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
> 
> This was the latest release of `revm`, if I'm not mistaken

Yep it was latest, i had trouble to trigger it, that is why i asked for version.

Added ci for it, it should not happen in future.

PR #451, opened 2023-04-06
fix: k256 compile error

rakita:
Make k256 compile.

closes: #449


crates/precompile/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![no_std]
+//#![no_std]
```
rakita, line 1:
```suggestion
#![no_std]
```

Wodann (approved):

PR #485, opened 2023-05-09
feat: add transfer bench

rakita:
Add simple ethere transfer bench


rakita:
Added with this PR: #486

PR #486, opened 2023-05-09
feat: separate initial checks

rakita:
There is a lot of refactor of initial EVM call

Added a bench for the simple transfer of value transaction, it got slightly faster from 700ns to 500ns but that is like ~5-10min saved for 2trilion of eth transactions, so it does not have a big impact.

Most of changes are internal, only external change is that previously only touched accounts (touch means something is done on them) would be returned but now all account from the journal state is returned.

### DIFF
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -1,3 +1,5 @@
+use revm_primitives::{Bytes, B160};
+
 use super::constants::*;
 use crate::{
     inner_models::SelfDestructResult,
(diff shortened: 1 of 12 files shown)


crates/interpreter/src/gas/calc.rs
```diff
@@ -325,3 +327,50 @@ pub fn memory_gas(a: usize) -> u64 {
...
+
+    // get number of access list account and storages.
+    if SPEC::enabled(BERLIN) {
+        let accessed_slots = access_list
+            .iter()
+            .fold(0, |slot_count, (_, slots)| slot_count + slots.len() as u64);
+        initial_gas += access_list.len() as u64 * ACCESS_LIST_ADDRESS;
+        initial_gas += accessed_slots * ACCESS_LIST_STORAGE_KEY;
+    }
+    // access list stipend
```
rakita, line 355:
```suggestion
```

PR #491, opened 2023-05-24
feat: Add initcode limit to be double of bytecode limit

rakita:
Set initcode limit from EIP-3860 to be double of ma bytecode size if custom max bytecode size is set inside `env.cfg.limit_contract_code_size`.


mattsse (approved):
I guess that makes sense,
this would also prevent a breaking issue

rakita:
> I guess that makes sense, this would also prevent a breaking issue

This looks a little bit nicer to me, but if needed we can add an additional limit.

PR #497, opened 2023-06-01
fix: typo in eip-3155 output

perama-v:
The output from the eip-3155 tracer returns `gasUser` rather than `gasUsed` ([spec](https://eips.ethereum.org/EIPS/eip-3155))

current:
```
"{\"output\":\"b\\\"\\\"\",\"gasUser\":\"0xd772\"}"
```
spec:
```
"{\"output\":\"b\\\"\\\"\",\"gasUsed\":\"0xd772\"}"
```


rakita:
ty!

PR #498, opened 2023-06-01
feat(revm): add post-tx state root to 3155 tracer

perama-v:
The eip-3155 ([spec](https://eips.ethereum.org/EIPS/eip-3155)) tracer does not return the post-transaction state root.

The `revme` binary crate had an implementation that I plugged in with a few modifications.

This involved adding a few dependencies, which seemed appropriate to place under the `ethersdb` feature flag.

- [ ] Test for correctness

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,13 +1668,17 @@ dependencies = [
  "ethers-core",
  "ethers-providers",
  "futures",
+ "hash-db",
  "hex",
  "hex-literal",
+ "plain_hasher",
  "revm-interpreter",
  "revm-precompile",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


perama-v:
This PR was misplaced, closing

The context I was using was using `revm` to execute state provided in a collection of merkle proofs of state accessed (read-written) in a block. All trie nodes in that context are known, so the new state root can be calculated. With the result of the EVM execution, the new modified state values can be updated in the proof and nodes recalculated (leaf-to-root).

However, these nodes (in the state proof) would need to be passed to `revm` for the state root calculation to be done here. Instead it likely makes more sense to do this in the calling crate/context rather than here.

PR #500, opened 2023-06-04
consume all gas on invalid opcode

### LINKED ISSUE #335: Inspector does not return `gas.limit()` as cost when an invalid opcode is detected
In Rethnet we implemented stack traces by using the `Inspector` trait. When comparing results against ethereumjs, we noticed that `revm` does not report the `Gas::limit()` in `Gas::spend()` when an invalid opcode is detected, whereas ethereumjs does (see [discussion](https://github.com/NomicFoundation/hardhat/pull/3593#discussion_r1080791076) or [`Inspector` trait implementation](https://github.com/NomicFoundation/hardhat/pull/3593/files#diff-4c5e3f43ac374461ee4d7d668e16bc3d6f5dfda3d4924b10439a56ba76af104eR491)).

I don't recall off the top of my head whether we return the gas limit for other exceptional halts, but it raises the question whether `revm` should be going this too?

teddav:
I believe this should close #335 as an invalid opcode should consume all the remaining gas

### DIFF
--- a/crates/interpreter/src/instructions.rs
+++ b/crates/interpreter/src/instructions.rs
@@ -19,12 +19,10 @@ pub fn return_stop(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     interpreter.instruction_result = InstructionResult::Stop;
 }
 pub fn return_invalid(interpreter: &mut Interpreter, _host: &mut dyn Host) {
-    gas!(interpreter, interpreter.gas.remaining());
     interpreter.instruction_result = InstructionResult::InvalidFEOpcode;
 }
 
 pub fn return_not_found(interpreter: &mut Interpreter, _host: &mut dyn Host) {
-    gas!(interpreter, interpreter.gas.remaining());
     interpreter.instruction_result = InstructionResult::OpcodeNotFound;
 }
 
--- a/crates/interpreter/src/instructions.rs
+++ b/crates/interpreter/src/instructions.rs
@@ -19,10 +19,12 @@ pub fn return_stop(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     interpreter.instruction_result = InstructionResult::Stop;
 }
 pub fn return_invalid(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    gas!(interpreter, interpreter.gas.remaining());
     interpreter.instruction_result = InstructionResult::InvalidFEOpcode;
 }
 
 pub fn return_not_found(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    gas!(interpreter, interpreter.gas.remaining());
     interpreter.instruction_result = InstructionResult::OpcodeNotFound;
 }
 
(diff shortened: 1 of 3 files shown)


rakita:
I am unsure if this does it, as `gas!` check if the limit is hit. and would change to different instruction result.

I think this should be more located inside GasInspector if it is possible to set gas.

teddav:
I saw in the issue discussion that you talked about adding it to an Inspector, but that means you'd have to use this specific Inspector to have the correct behavior. So I don't think that's something that should be implemented in an Inspector.  
In the EVM the correct behavior is to consume the entire gas for the call if you hit an Invalid opcode, so if you just "transact" without passing an inspector and hit an invalid opcode you want the gas consumption to be correct

rakita:
It is cause the evm internals desn't care about this, when this is triggered it will just ignore gas on return, as gas is calculated differently inside evm.

But this `GasInspector` is utility and so the assumption is that other inspectors will use this one so it is good place to add it.

crates/interpreter/src/instructions.rs
```diff
@@ -19,10 +19,12 @@ pub fn return_stop(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     interpreter.instruction_result = InstructionResult::Stop;
 }
 pub fn return_invalid(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    gas!(interpreter, interpreter.gas.remaining());
     interpreter.instruction_result = InstructionResult::InvalidFEOpcode;
 }
 
 pub fn return_not_found(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    gas!(interpreter, interpreter.gas.remaining());
```
rakita, line 27:
```suggestion
```
This can be removed

crates/interpreter/src/instructions.rs
```diff
@@ -19,10 +19,12 @@ pub fn return_stop(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     interpreter.instruction_result = InstructionResult::Stop;
 }
 pub fn return_invalid(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    gas!(interpreter, interpreter.gas.remaining());
```
rakita, line 22:
```suggestion
```
This can be removed

crates/revm/src/inspector/gas.rs
```diff
@@ -69,7 +69,15 @@ impl<DB: Database> Inspector<DB> for GasInspector {
         ret: InstructionResult,
         out: Bytes,
     ) -> (InstructionResult, Gas, Bytes) {
-        (ret, remaining_gas, out)
+        match ret {
+            InstructionResult::InvalidFEOpcode | InstructionResult::OpcodeNotFound => {
```
rakita, line 73: About this. This should emulate all `InstructionResult` that will spend all gas.

Check this part:
https://github.com/bluealloy/revm/blob/c54f079a4517c0c58e2808605208278338c24211/crates/interpreter/src/instructions/host.rs#L495-L522

It would be good to have function in `InstructionResult` `fn is_error(&self) -> bool` to check all cases of error kind cases.
All error kind cases are here:
https://github.com/bluealloy/revm/blob/c54f079a4517c0c58e2808605208278338c24211/crates/interpreter/src/instruction_result.rs#L19-L45

rakita (approved):
This look nice! Can you just fix clippy, it requires use of `matches!` macro

rakita (approved):
nice!

PR #504, opened 2023-06-05
chore: Move Precompiles to EVMData so Inspector can access it

rakita:
Move Precompiles so they can be accessed from Inspector.


mattsse (approved):
nice!
that works

PR #522, opened 2023-06-19
optimize stack usage for recursive `call` and `create` programs

### LINKED ISSUE #305: Measure Interpreter call stack footprint
For some tests that call multiple contracts in higher depths, revm has a overflow. It will be great to measure and quantitate where is the place that takes most data and how much does the stack rises with every call.

For info pthread_crate default is around 2-4MB. Check `Notes` here https://man7.org/linux/man-pages/man3/pthread_create.3.html

valo:
Fixes #305 

With these changes the EVM state tests does not require increased stack in release mode. In debug mode, the stack need to be setup to 3MB.

The changes are in the `create` and `call` EVM implementations and optimize the stack usage of these code paths. The code is broken down into smaller functions, so that the stack memory footprint is kept low. Some of the larger parameters are moved to the heap using boxing.


crates/interpreter/src/instructions/host.rs
```diff
@@ -321,6 +327,24 @@ pub fn create<const IS_CREATE2: bool, SPEC: Spec>(
...
+
+    if create_input.is_none() {
+        return;
+    }
+
+    let mut create_input = create_input.unwrap();
+
+    let ret = host.create(&mut create_input);
+
+    handle_create_result(interpreter, ret);
```
rakita, line 345: This is not needed as after `host.create` we are not in recursive call so it will not increment stack that much on recursion.

crates/interpreter/src/instructions/host.rs
```diff
@@ -521,3 +547,47 @@ pub fn call_inner<SPEC: Spec>(
...
+        &mut out_offset,
+        &mut call_input,
+    );
+
+    if call_input.is_none() {
+        return;
+    }
+
+    // The Option is not needed form this point
+    let mut call_input = call_input.unwrap();
```
rakita, line 580: Can use `let else` here.
```suggestion
    if call_input.is_none() {
        return;
    }

    // The Option is not needed form this point
    let Some(mut call_input) = call_input else { return };
```

crates/revm/src/evm_impl.rs
```diff
@@ -525,18 +570,33 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+    }
+
+    /// Main contract call of the EVM.
+    fn call_inner(&mut self, inputs: &mut CallInputs) -> (InstructionResult, Gas, Bytes) {
+        let res = self.prepare_call(inputs);
+        if let Err(ret) = res {
+            return ret;
+        }
+
+        let (gas, bytecode, checkpoint, contract) = res.unwrap();
```
rakita, line 592: Same here for `let else`

crates/interpreter/src/interpreter/contract.rs
```diff
@@ -19,8 +19,16 @@ pub struct Contract {
...
-    pub fn new(input: Bytes, bytecode: Bytecode, address: B160, caller: B160, value: U256) -> Self {
-        let bytecode = to_analysed(bytecode).try_into().expect("it is analyzed");
+    pub fn new(
+        input: Bytes,
+        bytecode: &Bytecode,
+        address: B160,
+        caller: B160,
+        value: U256,
+    ) -> Self {
+        let bytecode = to_analysed(bytecode.clone())
```
rakita, line 29: I would remove the clone here, and instantiate the Bytecode without `Box`, the clone seems redundant.

crates/revm/src/evm_impl.rs
```diff
@@ -525,18 +570,33 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+            return Err((e, gas, Bytes::new()));
         }
 
+        let contract = Box::new(Contract::new_with_context(
+            inputs.input.clone(),
+            &bytecode,
+            &inputs.context,
+        ));
+
+        Ok((gas, Box::new(bytecode), checkpoint, contract))
```
rakita, line 582: There is no need to return bytecode here.

crates/revm/src/evm_impl.rs
```diff
@@ -525,18 +570,33 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+        if let Err(ret) = res {
+            return ret;
+        }
+
+        let (gas, bytecode, checkpoint, contract) = res.unwrap();
+
         let ret = if is_precompile(inputs.contract, self.precompiles.len()) {
-            self.call_precompile(gas, inputs.contract, inputs.input.clone())
+            self.call_precompile(inputs, gas)
         } else if !bytecode.is_empty() {
```
rakita, line 596: This line can be replaced with 
```suggestion
        } else if !contract.is_empty() {
```

if fn `is_empty` is not present you can make it and inside of it call `bytecode.is_empty()`

crates/revm/src/evm_impl.rs
```diff
@@ -275,34 +275,36 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
 
-    /// EVM create opcode for both initial crate and CREATE and CREATE2 opcodes.
-    fn create_inner(
+    fn prepare_create(
         &mut self,
         inputs: &CreateInputs,
-    ) -> (InstructionResult, Option<B160>, Gas, Bytes) {
+    ) -> Result<
+        (Gas, B160, JournalCheckpoint, Box<Contract>),
+        (InstructionResult, Option<B160>, Gas, Bytes),
```
rakita, line 283: This looks ugly, but I dont have any better way atm.

Jus define `pub type .. = ..` at the beginning of the file to satisfy clippy.

rakita (approved):
This looks great! Left a few comments but there are mostly related to moving some code, nothing serious, very nice PR!

valo:
Ran a benchmark comparing `main` and this PR using https://github.com/ziyadedher/evm-bench. The performance seems to be pretty much the same.

```
|                         | evmone | revm-main | revm-local |
|-------------------------|--------|-----------|------------|
| **sum**                 | 56.2ms | 72.4ms    | 72.8ms     |
| **relative**            | 1.000x | 1.288x    | 1.295x     |
| erc20.approval-transfer | 6.2ms  | 7ms       | 8ms        |
| erc20.mint              | 4ms    | 4.4ms     | 4ms        |
| erc20.transfer          | 7.4ms  | 8.2ms     | 8.6ms      |
| snailtracer             | 36.6ms | 49ms      | 48.4ms     |
| ten-thousand-hashes     | 2ms    | 3.8ms     | 3.8ms      |
```

rakita (approved):

PR #528, opened 2023-06-20
feat(cancun): EIP-5656: MCOPY - Memory copying instruction

### LINKED ISSUE #523: EIP-5656: MCOPY - Memory copying instruction
All info can be found here: https://eips.ethereum.org/EIPS/eip-5656

0xJepsen:
Started working on this. Still needs work. 

Will Close #523

### DIFF
--- a/crates/interpreter/src/instructions/memory.rs
+++ b/crates/interpreter/src/instructions/memory.rs
@@ -60,10 +60,8 @@ pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
     let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
     // temporary value for memory resize
-    let mut resize = dest;
-    if src + len > dest {
-        resize = src + len;
-    }
+    // memory resize
+    let resize = max(dest, len).saturating_add(len);
     // resize memory
     memory_resize!(interpreter, src, resize);
     // copy memory in place
--- a/crates/interpreter/src/instructions/memory.rs
+++ b/crates/interpreter/src/instructions/memory.rs
@@ -43,21 +43,34 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     push!(interpreter, U256::from(interpreter.memory.effective_len()));
 }
 // From EIP-5656 MCOPY
-pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
     check!(interpreter, SPEC::enabled(CANCUN));
     pop!(interpreter, dest, src, len);
     if len == U256::ZERO {
         return;
     }
+
     let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
     gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
 
+    // memory_resize!(interpreter, dest, len);
+    memory_copy_resize(interpreter, host);
     let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
-    memory_resize!(interpreter, dest, len);
+
 
     let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
     // Read data with length len from src
-    let data = interpreter.memory.copy(src, len);
+    let binding = interpreter.memory.copy_to_vec(src, len);
+    let data = binding.as_slice()  ;
     // Write data to dest
-    interpreter.memory.set_data(src, dest, len, &data);
+    interpreter.memory.set_data(src, dest, len, data);
+}
+
+pub fn memory_copy_resize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    let mut mstart = as_usize_or_fail!(interpreter, interpreter.stack.peek(0).unwrap()); // stack[0]: dst
+
+    if as_usize_or_fail!(interpreter, interpreter.stack.peek(1).unwrap()) > mstart {
+        mstart = as_usize_or_fail!(interpreter, interpreter.stack.peek(1).unwrap()); // stack[1]: source
+    }
+    memory_resize!(interpreter, mstart, as_usize_or_fail!(interpreter, interpreter.stack.peek(2).unwrap()));
 }
--- a/crates/interpreter/src/instructions/memory.rs
+++ b/crates/interpreter/src/instructions/memory.rs
@@ -41,22 +41,28 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
 pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     check!(interpreter, SPEC::enabled(CANCUN));           
     pop!(interpreter, dest, src, len);
+    if len == U256::ZERO {
+        return;
+    }
     let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
     gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
+
+
     let dest = as_usize_or_fail!(
         interpreter,
         dest,
         InstructionResult::InvalidOperandOOG
     );
+    memory_resize!(interpreter, dest, len);
+
+
     let src = as_usize_or_fail!(
         interpreter,
         src,
         InstructionResult::InvalidOperandOOG
     );
-    let (data_end, overflow) = dest.overflowing_add(len);
-    if overflow || data_end > interpreter.return_data_buffer.len() {
-        interpreter.instruction_result = InstructionResult::OutOfOffset;
-        return;
-    }
-    interpreter.memory.set_data(src, dest, len, &interpreter.contract.input);
+    // Read data with length len from src
+    let data = interpreter.memory.clone();
+    // Write data to dest
+    interpreter.memory.set_data(src, dest, len, data.get_slice(src, len));
 }
--- a/crates/interpreter/src/instructions/memory.rs
+++ b/crates/interpreter/src/instructions/memory.rs
@@ -63,7 +63,6 @@ pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
 
     let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
     let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
-    // temporary value for memory resize
     // memory resize
     let resize = max(dest, len).saturating_add(len);
     // resize memory
--- a/crates/interpreter/src/instructions/memory.rs
+++ b/crates/interpreter/src/instructions/memory.rs
@@ -1,6 +1,11 @@
 use revm_primitives::SpecId::CANCUN;
 
-use crate::{gas, interpreter::Interpreter, primitives::{U256, Spec}, Host, InstructionResult};
+use crate::{
+    gas,
+    interpreter::Interpreter,
+    primitives::{Spec, U256},
+    Host, InstructionResult,
+};
 
 pub fn mload(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::VERYLOW);
@@ -39,7 +44,7 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
 }
 // From EIP-5656 MCOPY
 pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
-    check!(interpreter, SPEC::enabled(CANCUN));           
+    check!(interpreter, SPEC::enabled(CANCUN));
     pop!(interpreter, dest, src, len);
     if len == U256::ZERO {
         return;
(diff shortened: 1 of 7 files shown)


crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +37,26 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::BASE);
     push!(interpreter, U256::from(interpreter.memory.effective_len()));
 }
+// From EIP-5656 MCOPY
+pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    check!(interpreter, SPEC::enabled(CANCUN));           
+    pop!(interpreter, dest, src, len);
+    let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
```
rakita, line 44: What should happen if `len==0`?
* should we just continue and skip rest of opcode.
* Should we cast dest to size.
* Should we expand memory until `dest` this would mean we could get OOG on that operation.

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +37,26 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+        dest,
+        InstructionResult::InvalidOperandOOG
+    );
+    let src = as_usize_or_fail!(
+        interpreter,
+        src,
+        InstructionResult::InvalidOperandOOG
+    );
+    let (data_end, overflow) = dest.overflowing_add(len);
+    if overflow || data_end > interpreter.return_data_buffer.len() {
```
rakita, line 57: wait silly question but isn't mcopy just copying data from memory to memory and not from return data buffer.

You should have `memory_resize` somewhere here

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +37,26 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+        interpreter.instruction_result = InstructionResult::OutOfOffset;
+        return;
+    }
+    interpreter.memory.set_data(src, dest, len, &interpreter.contract.input);
```
rakita, line 61: And here we are coping the `contract.input` that is different from the thing we check above

0xJepsen:
> left few comments

Awesome thanks for the comments. Will work some more on this today

crates/interpreter/src/instructions/memory.rs
0xJepsen: I think returning early would be correct based on the geth implementation.

crates/interpreter/src/instructions/memory.rs
0xJepsen: Yeah this is correct.

crates/interpreter/src/instructions/memory.rs
0xJepsen: Yep this is right, i read the memory from slice to pass the right data here

crates/interpreter/src/instructions/memory.rs
rakita: Can you share the code line in geth just to compare it.

And if we just return, there is probably some base gas amount that we need to add.

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +37,32 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    memory_resize!(interpreter, dest, len);
+
+
+    let src = as_usize_or_fail!(
+        interpreter,
+        src,
+        InstructionResult::InvalidOperandOOG
+    );
+    // Read data with length len from src
+    let data = interpreter.memory.clone();
```
rakita, line 65: Cloning memory is a no go, it is expensive.

You could probably add `copy` function to memory that would done copying. Just figure out how to do in in place with vec

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +37,32 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
+    gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
+
+
+    let dest = as_usize_or_fail!(
+        interpreter,
+        dest,
+        InstructionResult::InvalidOperandOOG
+    );
+    memory_resize!(interpreter, dest, len);
```
rakita, line 56: I would assume that both dest and src can increment memory. So you need to find what is bigger and increment memory by a bigger value

crates/interpreter/src/interpreter/memory.rs
```diff
@@ -67,6 +67,14 @@ impl Memory {
         &self.data[offset..offset + size]
     }
 
+    /// Memory copy
+    /// Copies memory inplace to a returns a [u8]
+    #[inline(always)]
+    pub fn copy(&self, offset: usize, size: usize) -> [u8; 32] {
+        let mut copy = [0u8; 32];
```
tynes, line 74: This looks like `copy_word` because `copy` is only `32`, it should be `size`

crates/interpreter/src/instructions/memory.rs
0xJepsen: Yeah of course, I was referencing this [PR](https://github.com/ethereum/go-ethereum/pull/26181/files#diff-ab2e5d690957bc365d96e997b236ce9333d343b50dbb674d8a018aa578584cdb)

crates/interpreter/src/instructions/memory.rs
0xJepsen: I made a function to check and handle this `memory_copy_resize` that was inspired by the geth pr's implementation in the `memory_table.go`

crates/interpreter/src/interpreter/memory.rs
0xJepsen: addressed this with returning a vec of bytes and then passing as a slice to the data

0xJepsen:
So the thing i want to try and fix now is that the `memory_copy_resize` has some nasty unwraps because the peek returns a result. I thought to use the peek because its how i was reading the `Back` method in the go implementation. But im not sure we have a similar implementation. I could try to write a something with similar functionality. What do you think?

0xJepsen:
> So the thing i want to try and fix now is that the `memory_copy_resize` has some nasty unwraps because the peek returns a result. I thought to use the peek because its how i was reading the `Back` method in the go implementation. But im not sure we have a similar implementation. I could try to write a something with similar functionality. What do you think?

In the geth implementation they validate the stack length in the interpreter here
https://github.com/ethereum/go-ethereum/blob/942ba4ddaa0e2591b77a42f40059babbaaa44154/core/vm/interpreter.go#L183C1-L187C4

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,35 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+
+    let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
+    // Read data with length len from src
+    let binding = interpreter.memory.copy_to_vec(src, len);
+    let data = binding.as_slice()  ;
+    // Write data to dest
+    interpreter.memory.set_data(src, dest, len, data);
+}
+
+pub fn memory_copy_resize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
```
rakita, line 69: This is faulty, and you already have `dest,src, len` in `fn mcopy` so this fn is not needed

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,35 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
+
+    // memory_resize!(interpreter, dest, len);
+    memory_copy_resize(interpreter, host);
+    let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
+
+
+    let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
+    // Read data with length len from src
+    let binding = interpreter.memory.copy_to_vec(src, len);
```
rakita, line 63: It is not optimal to create temporary vec. There is a memory operation from `C` that does copying of overlapping data so I assume there is something similar in rust. Make `memory.copy` that does internal copying of data.

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,35 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::BASE);
     push!(interpreter, U256::from(interpreter.memory.effective_len()));
 }
+// From EIP-5656 MCOPY
+pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
+    check!(interpreter, SPEC::enabled(CANCUN));
+    pop!(interpreter, dest, src, len);
+    if len == U256::ZERO {
```
rakita, line 49: This is not correct, even if len is zero, some gas needs to be deduced.
maybe it is best to check if `len == 0` after `[gas_or_fail](gas::verylowcopy_cost)` is added.

crates/interpreter/src/instructions/memory.rs
0xJepsen: I found and used `copy_within` 
https://doc.rust-lang.org/std/ptr/fn.copy.html

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,31 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
+    let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
+    let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
+    // deduce gas
+    gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
+
+    // check if len is 0
+    if len == 0 {
+        return;
+    }
```
rakita, line 62: There is a bug here. in order of checks that are done.

If `len == 0` we should just take BASE gas and return, but if `dest` or `src` are over `usize` we would have a consensus error and report `InvalidOperatorOOG`.

The order should be:

```suggestion
    let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
    // deduce gas
    gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
    if len == 0 {
        return;
    }
    
    let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
    let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
```

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,31 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    // temporary value for memory resize
+    let mut resize = dest;
+    if src + len > dest {
+        resize = src + len;
+    }
```
rakita, line 67: There are two problems here. The first consensus is that `dest+len` should be compared (not just `dest`) and the second is that `+` can overflow.

```suggestion
    // memory resize
    let resize = max(dest, len).saturating_add(len);
```

crates/interpreter/src/interpreter/memory.rs
```diff
@@ -66,7 +66,6 @@ impl Memory {
     pub fn get_slice(&self, offset: usize, size: usize) -> &[u8] {
         &self.data[offset..offset + size]
     }
-
```
rakita, line 69: nit: add new line here.

crates/interpreter/src/interpreter/memory.rs
```diff
@@ -110,6 +109,15 @@ impl Memory {
...
     }
+    /// In memory copy given a src, dst, and offset
+    /// Assumes that we check that src + offset > self.data.len()
+    #[inline(always)]
+    pub fn copy(&mut self, src: usize, dst: usize, offset: usize) {
+        if src + offset > self.data.len() {
+            return;
+        }
+        self.data.copy_within(src..src + offset, dst);
+    }
```
rakita, line 120: Few nits. wouth call it `lenght` and would remove this `if` as we already resized memory before hand.

```suggestion
    /// In memory copy given a src, dst, and length
    ///
    /// # Safety
    /// The caller is responsible to check that we resized memory properly.
    #[inline(always)]
    pub fn copy(&mut self, dst: usize, src: usize, length: usize) {
        self.data.copy_within(src..src + length, dst);
    }
```

crates/interpreter/src/instructions/memory.rs
```diff
@@ -35,3 +42,31 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::BASE);
     push!(interpreter, U256::from(interpreter.memory.effective_len()));
 }
+// From EIP-5656 MCOPY
+pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
+    // check that Cancun spec is enabled
+    check!(interpreter, SPEC::enabled(CANCUN));
```
rakita, line 48:
```suggestion
}

// From EIP-5656 MCOPY
pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
    // Opcode enabled in Cancun.
    // EIP-5656: MCOPY - Memory copying instruction
    check!(interpreter, SPEC::enabled(CANCUN));
```

rakita:
I think we are close!

crates/interpreter/src/instructions/memory.rs
rakita: This works perfectly

0xJepsen:
> I think we are close!

Awesome, thanks for all the feedback it's been really helpful in learning all the details here. Let me know if there are any more problems you see here.

rakita:
> > I think we are close!
> 
> Awesome, thanks for all the feedback! it's been really helpful in learning all the details here. Let me know if there are any more problems you see here.

Can you check why eth tests are failing?

0xJepsen:
Got the spec to pass, wrote a test for the copy too

0xJepsen:
For some reason the lint is failing with 
```error[E0635]: unknown feature proc_macro_span_shrink
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.56/src/lib.rs:92:30
      |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
      |                              ^^^^^

I looked into it and seems like this error usually occurs when using an incompatible version of the proc-macro2 crate with the current Rust toolchain. The lint passes locally by fails here. Interestingly it only started failing with the edition of my test

**Comment by 0xJepsen** [comment]
Wrote some tests and got CI passing!

File: crates/interpreter/src/instructions/memory.rs [hunk 15]
```diff
@@ -35,3 +44,30 @@ pub fn msize(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    let len = as_usize_or_fail!(interpreter, len, InstructionResult::InvalidOperandOOG);
+    // deduce gas
+    gas_or_fail!(interpreter, gas::verylowcopy_cost(len as u64));
+    if len == 0 {
+        return;
+    }
+
+    let dest = as_usize_or_fail!(interpreter, dest, InstructionResult::InvalidOperandOOG);
+    let src = as_usize_or_fail!(interpreter, src, InstructionResult::InvalidOperandOOG);
+    // temporary value for memory resize
```
rakita, line 66:
```suggestion
```

rakita (approved):
gj!

PR #536, opened 2023-07-01
Waylon/eip4844

### LINKED ISSUE #526: EIP-4844: Shard Blob Transactions
There are three things related to evm.

* `BLOBHASH` (with opcode 0x49) which reads index from the top of the stack as big-endian uint256, and replaces it on the stack with `env.tx.blob_versioned_hashes[index]`
* And precompile that needs to be added at `POINT_EVALUATION_PRECOMPILE_ADDRESS` that verifies a KZG proof which claims that a blob (represented by a commitment) evaluates to a given value at a given point.
* Data gas accounting.

More info here: https://eips.ethereum.org/EIPS/eip-4844

0xJepsen:
Closes #526 
Should be good for review!


Autoparallel:
Some tasks for us:
- [ ] Write `verify_kzg_proof()`
- [ ] Fix the type for `BLS_MODULUS` (no more `BigInt`)
- [ ] Write tests and hunt for bugs.
- [ ] Get CI/CD to pass.

Autoparallel:
We will test the `point_evaluation_run` now

crates/precompile/src/lib.rs
```diff
@@ -177,7 +178,7 @@ impl Precompiles {
     pub fn berlin() -> &'static Self {
         static INSTANCE: OnceCell<Precompiles> = OnceCell::new();
         INSTANCE.get_or_init(|| {
-            let mut precompiles = Self::istanbul().clone();
+            let mut precompiles = Self::berlin().clone();
```
rakita, line 181: This is an infinitive loop, it should be `instanbul`

crates/interpreter/src/instructions/memory.rs
```diff
@@ -68,3 +64,25 @@ pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+
+/// see https://eips.ethereum.org/EIPS/eip-4844
+pub fn blob_hash<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
+    check!(interpreter, SPEC::enabled(CANCUN));
+    // We add an instruction BLOBHASH (with opcode HASH_OPCODE_BYTE) which reads index from the top of the stack as big-endian uint256,
+    // and replaces it on the stack with tx.blob_versioned_hashes[index] if index < len(tx.blob_versioned_hashes),
+    // and otherwise with a zeroed bytes32 value. The opcode has a gas cost of HASH_OPCODE_GAS.
+    gas!(interpreter, gas::HASH_OPCODE_GAS);
+    pop!(interpreter, index);
+    let index = as_usize_or_fail!(interpreter, index, InstructionResult::InvalidOperandOOG);
```
rakita, line 76: Should it fail or not?

crates/interpreter/src/instructions/memory.rs
```diff
@@ -68,3 +64,25 @@ pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
 }
+
+/// see https://eips.ethereum.org/EIPS/eip-4844
+pub fn blob_hash<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
+    check!(interpreter, SPEC::enabled(CANCUN));
+    // We add an instruction BLOBHASH (with opcode HASH_OPCODE_BYTE) which reads index from the top of the stack as big-endian uint256,
+    // and replaces it on the stack with tx.blob_versioned_hashes[index] if index < len(tx.blob_versioned_hashes),
+    // and otherwise with a zeroed bytes32 value. The opcode has a gas cost of HASH_OPCODE_GAS.
+    gas!(interpreter, gas::HASH_OPCODE_GAS);
+    pop!(interpreter, index);
```
rakita, line 75: see `pop_top!` in other instructions

crates/interpreter/src/instructions/memory.rs
```diff
@@ -68,3 +64,25 @@ pub fn mcopy<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
+    if index < host.env().tx.blob_versioned_hashes.len() {
+        // Replace the top of the stack with the versioned hash here
+        push!(
+            interpreter,
+            *host.env().tx.blob_versioned_hashes.index(index)
+        );
+    } else {
+        // else write out a zerod out 32 bytes
+        let bytes: [u8; 32] = [0; 32];
+        push!(interpreter, U256::from_be_bytes(bytes));
```
rakita, line 86: is this just U256::zero?

crates/precompile/src/point_evaluation.rs
```diff
@@ -0,0 +1,55 @@
...
+/// in big endian format
+const BLS_MODULUS: [u8; 32] = [
+    115, 237, 167, 83, 41, 157, 125, 72, 51, 57, 216, 8, 9, 161, 216, 5, 83, 189, 164, 2, 255, 254,
+    91, 254, 255, 255, 255, 255, 0, 0, 0, 1,
+];
+
+/// See: https://eips.ethereum.org/EIPS/eip-4844
+pub fn point_evaluation_run(input: &[u8], gas_limit: u64) -> PrecompileResult {
+    // The data is encoded as follows: versioned_hash | z | y | commitment | proof | with z and y being padded 32 byte big endian values
+    assert!(input.len() == 192);
```
rakita, line 20: Is this good assumption?

crates/precompile/src/point_evaluation.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    // We can always be sure that these will be 48 bytes so this unwrap should be okay
+    let z = Bytes32::from_bytes(&input[32..64]).unwrap();
+    let y = Bytes32::from_bytes(&input[64..96]).unwrap();
+    let commitment = Bytes48::from_bytes(&input[96..144]).unwrap();
+    let versioned_hash = Bytes48::from_bytes(&input[0..32]).unwrap();
+    let proof = Bytes48::from_bytes(&input[144..192]).unwrap();
+    let kzg_settings = c_kzg::KzgSettings::load_trusted_setup_file(
+        "crates/precompile/src/trusted_setup4.txt".into(),
```
rakita, line 33: Do you need this file in runtime? Where should be placed

crates/revm/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tokio = { version = "1.28", features = ["rt-multi-thread", "macros"], optional =
 ethers-providers = { version = "2.0", optional = true }
 ethers-core = { version = "2.0", optional = true }
 futures = { version = "0.3.27", optional = true }
+rust-kzg = "0.0.0"
```
rakita, line 27: Is this used?

crates/interpreter/src/instructions/memory.rs
0xJepsen: No i don't think it should, can i use `as_usize_saturated!()` here instead?

crates/interpreter/src/instructions/memory.rs
0xJepsen: yes, thats cleaner.

crates/precompile/src/point_evaluation.rs
0xJepsen: Perhaps not, Can we through an error if they aren't equal. Something like 
```rust
    if input.len() != 192 {
        return Err(Error::InvalidInputLength);
    }

```

crates/precompile/src/point_evaluation.rs
0xJepsen: This is a mock trusted setup from that was used for testing with the EF audited bindings from c-kzg, afaik we do need it at runtime and should replace it with the official trusted set up once the ceremony is finished. I am not sure where it should live so i put it here. Do you think it should go in a different directory.

crates/revm/Cargo.toml
0xJepsen: No i was using it at one point when writing the point verification by hand prior to using the EF bindings.

rakita:
Closing in support for https://github.com/bluealloy/revm/pull/668

PR #537, opened 2023-07-03
book workflow

0xJepsen:
Workflow to lint, test, build, and deploy the mdbook. Can change the deploy parts once pages are set up. This is the same workflow from reth book [here](https://github.com/paradigmxyz/reth/blob/main/.github/workflows/book.yml). We will need to edit the example https://github.com/bluealloy/revm/blob/10f81ba1b3cd6171c25d3346f1800bbb04f5ead1/examples/fork_ref_transact.rs#L1-L109 to get the test passing or pull it out of the book

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -1,0 +1,126 @@
+name: book
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  merge_group:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: test
+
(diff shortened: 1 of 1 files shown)


0xJepsen:
Hmm, when writing the docs the code annotations were mainly just references to the implementation, they weren't intended to compile at all (big reason tests are failing). Maybe that was a bad idea and I should go back and change them all. The alternative would be to remove the `book/test` check. What do you think @rakita?

rakita (approved):
lgtm!

Issue #154, opened 2022-07-24
EIP-1153: Transient storage opcodes


moodysalem:
useful for testing and developing to have these opcodes optionally enabled
https://eips.ethereum.org/EIPS/eip-1153

veteransmod:
https://direct.lc.chat/14345826/

rakita:
Hm what is the usecase for using this opcodes for testing?

moodysalem:
developing protocols that will make use of it. you can just use regular storage, but then your gas measurements are off and you're testing against a pre-transient storage version of the code

rakita:
Yeah but transient storage is still going to spend the gas.
I am okay to spend time including this but only when it gets certain it is going to be included in the Ethereum mainnet.

moodysalem:
> Yeah but transient storage is still going to spend the gas. 

what do you mean? it spends much less gas. hard to compare gas snapshots of contracts if you can't test with the proposed opcodes

> I am okay to spend time including this but only when it gets certain it is going to be included in the Ethereum mainnet.

what's the downside of including it before it is certain, behind a flag? ethereumjs did this, which allows for testing in hardhat

rakita:
> > Yeah but transient storage is still going to spend the gas.
> 
> what do you mean? it spends much less gas. hard to compare gas snapshots of contracts if you can't test with the proposed opcodes
My bad, I though you meant testing smart contract, i was confused bcs of that, but you meant testing this EIP?

> 
> > I am okay to spend time including this but only when it gets certain it is going to be included in the Ethereum mainnet.
> 
> what's the downside of including it before it is certain, behind a flag? ethereumjs did this, which allows for testing in hardhat

I would reader spend doing something else than investing my time in this particular EIP atm, and even if the work is done by somebody else there is still a maintenance cost of this (small but still there). As it is uncertain when or even if it is going to be included and design as with every EIP is going to be changed, so for me, as I am doing revm on my own time, the best way is to wait and see where this goes.

moodysalem:
The EIP is now CFI: https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md#eips-considered-for-inclusion

Would love to see this in revm and subsequently forge so users can start playing with it

rakita:
Hello, I am following what is going on. There were a few major changes happening inside revm as usage of ruint so I am focusing on that.

[...]

PR #546, opened 2023-07-07
feat: EIP-1153 Transient storage opcodes

tynes:
Implement EIP-1153 for the Cancun hardfork

The changes to the specs in this PR I'd like some feedback on, happy to revert things that aren't directly related to the EIP implementation

Need to update the opcode numbers before this PR is ready, they are still on the older values

https://eips.ethereum.org/EIPS/eip-1153

Closes https://github.com/bluealloy/revm/issues/154


crates/interpreter/src/instructions/host.rs
```diff
@@ -179,6 +179,24 @@ pub fn sstore<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
     refund!(interpreter, gas::sstore_refund::<SPEC>(original, old, new));
 }
 
+pub fn tstore<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
+    check_staticcall!(interpreter);
```
rakita, line 183: Need additional check for hardfork. Both here and for tload

crates/interpreter/src/instructions/host.rs
```diff
@@ -179,6 +179,24 @@ pub fn sstore<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
...
+
+    pop!(interpreter, index, value);
+
+    host.tstore(interpreter.contract.address, index, value);
+}
+
+pub fn tload<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
+    gas!(interpreter, gas::WARM_STORAGE_READ_COST);
+
+    pop!(interpreter, index);
```
rakita, line 194: you can use `pop_top!` here and you set tload value to it.

crates/revm/src/evm.rs
```diff
@@ -196,7 +196,6 @@ pub fn evm_inner<'a, DB: Database, const INSPECT: bool>(
         }
         SpecId::MERGE => create_evm!(MergeSpec, db, env, insp),
         SpecId::SHANGHAI => create_evm!(ShanghaiSpec, db, env, insp),
-        SpecId::CANCUN => create_evm!(LatestSpec, db, env, insp),
-        SpecId::LATEST => create_evm!(LatestSpec, db, env, insp),
+        SpecId::CANCUN | SpecId::LATEST => create_evm!(LatestSpec, db, env, insp),
```
rakita, line 199: LATEST should be LatestSpec, Cancun should have its own spec

crates/revm/src/journaled_state.rs
```diff
@@ -657,6 +690,23 @@ impl JournaledState {
         Ok((slot.original_value, present, new, is_cold))
     }
 
+    pub fn tstore(&mut self, address: B160, key: U256, new: U256) {
+        let present = self.tload(address, key);
```
rakita, line 694: We are not calling `tstore` anywhere

crates/revm/src/journaled_state.rs
```diff
@@ -393,6 +407,13 @@ impl JournaledState {
                         storage.remove(&key);
                     }
                 }
+                JournalEntry::TransientStorageChange {
+                    address,
+                    key,
+                    had_value,
+                } => {
+                    transient_storage.set(address, key, had_value.unwrap());
```
rakita, line 415: If `had_value` is `None` we should remove the entry from transient storage.

crates/interpreter/src/interpreter.rs
```diff
@@ -2,6 +2,7 @@ pub mod analysis;
 mod contract;
 pub(crate) mod memory;
 mod stack;
+pub mod transient;
```
rakita, line 5: Would be good to have full name `transient_storage`

tynes:
Thank you for picking this up @rakita. For test coverage, do you think using the Ethereum state tests are sufficient?

The implementation looks good to me, looks much more idiomatic than before

rakita:
> Thank you for picking this up @rakita. For test coverage, do you think using the Ethereum state tests are sufficient?
> 
> The implementation looks good to me, looks much more idiomatic than before

using eth tests should be sufficient to cover edge cases

PR #559, opened 2023-07-24
feat: Implement `Error` for EVMError

oblique:


rakita (commented):
Intentionally left it without thiserror

oblique:
Just curious, what was the intention?

oblique:
Is it because of `no_std`? In that case I can manually impl the `Error` and feature gate it with `#[cfg(feature = "std")]`

oblique:
What about now? `Error` is implemented only in case of `std` feature.

rakita (approved):
This looks great, lets add it as is!

PR #570, opened 2023-07-29
fix(interpreter): mcopy call order

DaniPopes:
`Memory::copy` argument order is `dst, src, len`, while it is being called with `src, dst, len`.


rakita:
@0xJepsen fyi

rakita:
I added additonal commit as there is one more bug related to mcopy: https://github.com/DaniPopes/revm/tree/fix-mcopy
But for some reason, github does not include it inside this PR 🤷🏻‍♂️

rakita:
Just for reference, initial impls was here: https://github.com/bluealloy/revm/pulls?q=is%3Apr+mcopy

rakita (approved):

Issue #432, opened 2023-03-27
Refactor: Split transaction pre verification to separate function


rakita:
Preverification that could make transactions fail should be split to separate function so that transactions can execute without those checks.

lorenzofero:
Hi! Can I take a look at this issue? It seems a first good one to get familiar with the codebase. Thanks!

rakita:
Done

PR #499, opened 2023-06-01
feat: State with account status

rakita:
Model account states to cover all possible state that transaction can change it

Model can be found here
https://mermaid.live/edit#pako:eNqNU7FOwzAQ_RXLI2qQysDgAQnRDEhQITFiBss-Uku1XTkXIKr677gxIa5jlWZInHvv3t3L5fZUOgWU0RYFwkqLxgtTfd5wS8K1gha960GRqrqb3tbwRRix4S49hLRIftgI2-TUOfR7jsBRqcAfw-GZ089UydA8-aT93M-FzPtGaJuRhthMcK5U7Dwv_eSEAlWbHfb148vydjn_Zv9qJk2W5aaxMjLMncgtCH99LufPUoTXDutv3aK2TQmelVGno0pYowVGZOry7ep9gCOzFEtaKMGJAbqgBrwRWoUffX8kc4obMMApC0cFH6LbIqfcHgJVdOheeyspQ9_BgnY7Na3GGASl0fnnuDvDCh1-AFmUEJU

And create cached/transition/bundle states that can easily create reverts on those states.

States interconnection can be seen here:
```
EVM State
(It has both original/present storage and new account and storage acts as a access list.)
|
|
V
[Cache State] Fetches data from database and get updated from EVM state. Allows us fewer layer to fetch data from database.
|        \
|         \
|          [Transitions State] contains changes related to transition (Mostly block). It has original account/storage/status (Needed for Loaded account). It is used for creating Reverts. 
|          |
|          V
|          [Bundle state] (It has Reverts and plain state, Original storage is not needed) One of reasons why this is the case, is because when reverting of canonical chain we can't get previous storage value. And it is not needed.
|       /
|    /
v
database
```

this is still WIP!

PR #573, opened 2023-07-31
Refactor: Split transaction pre verification to separate function

lorenzofero:
Hi! I tried to give a shot to https://github.com/bluealloy/revm/issues/432 in this PR. I hope to have understood and addressed the problem indicated by the issue.

### DIFF
--- a/bins/revme/src/statetest/runner.rs
+++ b/bins/revme/src/statetest/runner.rs
@@ -268,8 +268,7 @@ pub fn execute_test_suit(
                 let exec_result = if trace {
                     evm.inspect_commit(TracerEip3155::new(Box::new(stdout()), false, false))
                 } else {
-                    evm.preverify_transaction()
-                        .and_then(|_| evm.transact_commit())
+                    evm.transact_commit()
                 };
                 let timer = timer.elapsed();
 
--- a/bins/revme/src/statetest/runner.rs
+++ b/bins/revme/src/statetest/runner.rs
@@ -268,7 +268,8 @@ pub fn execute_test_suit(
                 let exec_result = if trace {
                     evm.inspect_commit(TracerEip3155::new(Box::new(stdout()), false, false))
                 } else {
-                    evm.transact_commit()
+                    evm.preverify_transaction()
+                        .and_then(|_| evm.transact_commit())
                 };
                 let timer = timer.elapsed();
 
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -60,6 +60,7 @@ struct CallResult {
 }
 
 pub trait Transact<DBError> {
+    /// Do checks that could make transaction fail before call/create
     fn preverify_transaction(&mut self) -> Result<(), EVMError<DBError>>;
 
     /// Do transaction.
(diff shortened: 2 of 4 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -85,10 +88,35 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+        let (caller_account, _) = journal
+            .load_account(tx_caller, self.data.db)
+            .map_err(EVMError::Database)?;
+
+        self.data.env.validate_tx_against_state(caller_account)?;
+
+        Ok(())
+    }
+
+    fn transact(&mut self) -> EVMResult<DB::Error> {
```
rakita, line 119: It will be a foot gun if we silently remove some checks. 
We should probably add an additional function `transact_preverified` to make it explicit.

rakita (commented):
Left comments.

And I apologise for a long wait I wanted to think about some additional things to make this more flexible but it missed me, we can push this over the line.

bins/revme/src/statetest/runner.rs
```diff
@@ -268,7 +268,8 @@ pub fn execute_test_suit(
                 let exec_result = if trace {
                     evm.inspect_commit(TracerEip3155::new(Box::new(stdout()), false, false))
                 } else {
-                    evm.transact_commit()
+                    evm.preverify_transaction()
+                        .and_then(|_| evm.transact_commit())
```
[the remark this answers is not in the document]
lorenzofero: Thanks! Now I moved it to the `transact` function which does `preverify_transaction` and then `transact_preverified`, according to the other comment you made

crates/revm/src/evm_impl.rs
lorenzofero: You're right! Sorry for the dumb take. Now I added the `transact_preverified` function in the `Transact` trait

lorenzofero:
> Left comments.
> 
> And I apologise for a long wait I wanted to think about some additional things to make this more flexible but it missed me, we can push this over the line.

Hi @rakita! I made some small changes according to the comments. Thanks for the feedback. I hope it is ok now.

crates/revm/src/evm_impl.rs
```diff
@@ -85,10 +91,35 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
 impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
     for EVMImpl<'a, GSPEC, DB, INSPECT>
 {
-    fn transact(&mut self) -> EVMResult<DB::Error> {
-        self.env().validate_block_env::<GSPEC, DB::Error>()?;
-        self.env().validate_tx::<GSPEC>()?;
+    fn preverify_transaction(&mut self) -> Result<(), EVMError<DB::Error>> {
```
gakonst, line 94: Interesting - I wonder how much time we spend doing these checks, and whether we could potentially run all these checks in parallel at the beginning of each block in Reth?

I have seen similar separations to be valuable in the context of account abstraction where you'd want to verify e.g. all signatures in parallel (we do that alrd) and then execute.

rakita: It is negligible, check these measurments https://github.com/bluealloy/revm/pull/499#issuecomment-1589610777
It is 250ms for 100k blocks that took 56s (on sepolia). Although tx num varies it is under 0.5% of the time spent there.

PR #558, opened 2023-07-24
perf: refactor interpreter internals

DaniPopes:
This PR focuses on refactoring some parts of the interpreter which lead to, although minor, performance gains.

Closes #310, not fixes, as the current `match` performs better on debug builds, and performs the same in release builds. This is not shown in the results.

# Results

## Cachegrind

[cachegrind.tar.gz](https://github.com/bluealloy/revm/files/12210142/cachegrind.tar.gz)

|                      Commit                     | Snailtracer (I) |
|:-----------------------------------------------:|:---------------:|
| 968fe98fcb590e5d67d690ce3c6a370834dc0541 (main) |   515,151,261   |
|     2a06916ef30e93b9b23907624b8e04ee63994693    |   489,886,182   |
| Improvement                                     | -4.9%           |

## Benches

~~Not entirely sure why the non-generic version performs better, since I would expect the checks to be optimized out at compile time on the generic version.~~

~~I have included the patch which can be used to change between the two versions.~~

Time benches shouldn't be relied upon for these type of small changes, and these ones are also outdated.

### Main

36de35b25d5941ac257e1877696e894223445c30

```julia
snailtracer/transact/analysed                                                                           
                        time:   [46.154 ms 46.666 ms 47.183 ms]
snailtracer/eval        time:   [40.100 ms 40.293 ms 40.487 ms]                            


snailtracer/transact/analysed                                                                           
                        time:   [45.875 ms 46.356 ms 47.026 ms]
snailtracer/eval        time:   [40.874 ms 41.289 ms 41.946 ms]                            


snailtracer/transact/analysed                                                                           
                        time:   [46.696 ms 46.993 ms 47.451 ms]
snailtracer/eval        time:   [40.366 ms 40.707 ms 41.054 ms]                            
```

### Generic (patch)

3658fa7a16e2dc3ea7b125bd836380f4a3ed678f

<details><summary><strong>Patch</strong></summary>
<p>

```diff
diff --git a/crates/interpreter/src/gas/calc.rs b/crates/interpreter/src/gas/calc.rs
index fa93b6a..dbd05be 100644
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -1,13 +1,13 @@
 use super::constants::*;
 use crate::inner_models::SelfDestructResult;
 use alloc::vec::Vec;
-use primitives::{Bytes, SpecId, SpecId::*, B160, U256};
[...]

PR #579, opened 2023-08-03
refactor: rewrite revm-test as a criterion bench

DaniPopes:
Split off from #558

[`microbench`](https://github.com/KyleMayes/microbench) is unmaintained, and the other benches are just raw for loops.


rakita (approved):
lgtm

PR #581, opened 2023-08-03
chore: avoid unnecessary allocations

DaniPopes:
Split off from #558

And some other minor clean-ups

### DIFF
--- a/crates/primitives/src/bytecode.rs
+++ b/crates/primitives/src/bytecode.rs
@@ -1,5 +1,5 @@
 use crate::{keccak256, B256, KECCAK_EMPTY};
-use alloc::{sync::Arc, vec, vec::Vec};
+use alloc::{sync::Arc, vec::Vec};
 use bitvec::prelude::{bitvec, Lsb0};
 use bitvec::vec::BitVec;
 use bytes::Bytes;
@@ -21,16 +21,19 @@ impl Debug for JumpMap {
 
 impl JumpMap {
     /// Get the raw bytes of the jump map
+    #[inline]
     pub fn as_slice(&self) -> &[u8] {
         self.0.as_raw_slice()
     }
 
     /// Construct a jump map from raw bytes
+    #[inline]
     pub fn from_slice(slice: &[u8]) -> Self {
         Self(Arc::new(BitVec::from_slice(slice)))
     }
 
     /// Check if `pc` is a valid jump destination.
+    #[inline]
     pub fn is_valid(&self, pc: usize) -> bool {
         pc < self.0.len() && self.0[pc]
     }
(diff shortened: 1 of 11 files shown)


crates/primitives/src/specification.rs
```diff
@@ -1,9 +1,10 @@
+#![allow(non_camel_case_types)]
```
DaniPopes, line 1: `rust-analyzer` gives a warning for me on the struct without this attribute at the module level

crates/primitives/src/bytecode.rs
```diff
@@ -44,16 +47,18 @@ pub struct Bytecode {
...
+    #[inline]
     fn default() -> Self {
         Bytecode::new()
     }
 }
 
 impl Bytecode {
-    /// Create [`Bytecode`] with one STOP opcode.
+    /// Creates a new [`Bytecode`] with exactly one STOP opcode.
+    #[inline]
```
rakita, line 58: The compiler is smart and in most cases will do a good job on the inline front, so adding `#[inline]` everywhere is just noisy.

crates/primitives/src/bytecode.rs
```diff
@@ -75,11 +82,15 @@ impl Bytecode {
...
 
-    /// Create new raw Bytecode with hash
+    /// Creates a new raw Bytecode with the given hash.
     ///
     /// # Safety
-    /// Hash need to be appropriate keccak256 over bytecode.
+    ///
+    /// The given `hash` has to be equal to the [`keccak256`] hash of the given `bytecode`.
+    #[inline]
+    #[cfg_attr(debug_assertions, track_caller)]
```
rakita, line 91: Why is this added, in essence, `unsafe` is here for a reason user of this should be careful what he does

DaniPopes: Right, it's just an extra check in debug mode. I can remove it if you want

crates/primitives/src/bytecode.rs
DaniPopes: This is not the case for public functions, which will not get inlined across crate boundaries without LTO: <https://nnethercote.github.io/perf-book/inlining.html>

Also `#[inline]` is just a hint, and given that these functions are very trivial I don't see why not to hint to the compiler that they should be inline (like `Vec::len`).

I've had this discussion before, I don't want to argue about this again so if you're really keen on this I'll just remove all of the hints.

rakita: I was thinking just about in crate use, I tried to play with it with the opcode function and would have better results when I just leave it to the compiler. Let's leave it with `#[inline]` as it makes sense for `Bytecode`.

crates/primitives/src/bytecode.rs
rakita: Let us remove it, I would assume there is a possibility for one to intentionally switch the bytecode of a different hash.

PR #582, opened 2023-08-03
perf: refactor interpreter internals (take 2)

DaniPopes:
Split off from (and supersedes) #558, see that PR for benches and list of changes.

Best reviewed on a commit-by-commit basis.

# Results

## Cachegrind

- download [cachegrind.tar.gz](https://github.com/bluealloy/revm/files/12284693/cachegrind.tar.gz), extracts into crates/cachegrind
- run:
```sh
cargo b -r -p revm-cachegrind --bin snailtracer
valgrind --tool=cachegrind target/release/snailtracer
```

|                      Commit                     | Snailtracer (I) | Improvement |
|:-----------------------------------------------:|:---------------:|:----:|
| 449d6b99ce19c51900782db1c4ea2b9c1c8a5199 (main) |   505,445,058   | N/A |
| b3440b1f83fd08c3c067be57bf0f0c58dfc79b02 |   441,772,332   |-12.6%|


crates/interpreter/src/instructions/host_env.rs
```diff
@@ -33,24 +31,24 @@ pub fn difficulty<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut
...
     push!(interpreter, host.env().effective_gas_price());
 }
 
-pub fn basefee<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
+// EIP-3198: BASEFEE opcode
+pub(super) fn basefee<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
+    check!(interpreter, LONDON);
     gas!(interpreter, gas::BASE);
-    // EIP-3198: BASEFEE opcode
-    check!(interpreter, SPEC::enabled(LONDON));
```
DaniPopes, line 49: I feel like the spec check should be before any gas checks, is this right @rakita?

crates/interpreter/src/instructions/mod.rs
```diff
@@ -0,0 +1,23 @@
+#[macro_use]
```
rakita, line 1: Inside revm i don't use `mod.rs` style of naming.

crates/interpreter/src/instructions/host_env.rs
rakita: It is not a bug as both actions would result in the same behaviour (REVERT), but it is a lot nicer and more consistent to  do `check` first then `gas`

DaniPopes:
Marking as draft until the other PRs are merged, so I can then rebase on top of main.

crates/interpreter/src/interpreter/memory.rs
DaniPopes, line 1: - changed runtime panics to be `unreachable_unchecked` in release mode since they should never happen

DaniPopes (commented):
Added descriptions of changes. Other minor (non functional) changes include:
- pub -> pub(super) to denote we are not exporting these instructions individually
- remove last argument of `as_usize_or_fail!` macro calls when it's `InvalidOperandOOG`, since it's the default

crates/interpreter/src/instructions/control.rs
DaniPopes, line 1: - factored out the set up in common for return and revert

crates/interpreter/src/instructions/bitwise.rs
DaniPopes, line 1: - cmp opcodes are now branchless
- `byte` indexes into the slice instead of performing shifts
- `sar` uses `U256::MAX` instead of computing it at runtime

crates/interpreter/src/instructions/arithmetic.rs
DaniPopes, line 1: - avoid reassigning to 0 when already 0 in div and rem

crates/interpreter/src/instructions.rs
DaniPopes, line 1: moved to `eval` to `opcode.rs`, and the return instructions to `control`

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -1,453 +1,313 @@
...
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+use super::*;
+use crate::{
+    gas,
+    primitives::{Spec, SpecId},
+    Host, Interpreter,
+};
+use core::fmt;
+
+macro_rules! opcodes {
```
DaniPopes, line 11: This macro declares all of:
- `const NAME: u8 = $value;`
- `OPCODE_JUMPMAP[$value] = Some(stringify!($name));`
- `eval` dispatch

Left out gas, which is improved below.

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -1,453 +1,313 @@
...
+            f(interpreter, host);
+        }
+    };
+}
+
+// When adding new opcodes:
+// 1. add the opcode to the list below; make sure it's sorted by opcode value
+// 2. add its gas info in the `opcode_gas_info` function below
+// 3. implement the opcode in the corresponding module;
+//    the function signature must be the exact same as the others
```
DaniPopes, line 76: Added instructions on how to add new opcodes with the new macro.

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -1,453 +1,313 @@
...
+        let n = self.get();
+        if let Some(val) = OPCODE_JUMPMAP[n as usize] {
+            f.write_str(val)
+        } else {
+            write!(f, "UNKNOWN(0x{n:02X})")
+        }
+    }
+}
 
 impl OpCode {
```
DaniPopes, line 258: Deprecated:
- `try_from_u8` -> `new`
- `u8` -> `get`

This is the same naming scheme as `NonZero*` std numbers.

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -460,469 +320,336 @@ const IS_PUSH_MASK: u32 = 0x20000000;
...
-            /* 0xf9 */ OpInfo::none(),
-            /* 0xfa  STATICCALL */ OpInfo::gas_block_end(0),
-            /* 0xfb */ OpInfo::none(),
-            /* 0xfc */ OpInfo::none(),
-            /* 0xfd  REVERT */ OpInfo::gas_block_end(0),
-            /* 0xfe  INVALID */ OpInfo::gas_block_end(0),
-            /* 0xff  SELFDESTRUCT */ OpInfo::gas_block_end(0),
-        ];
-    };
+const fn opcode_gas_info(opcode: u8, spec: SpecId) -> OpInfo {
```
DaniPopes, line 384: moved gas declarations from macros to const fns.

gakonst: Is this something you'd want to move on the main OPcode macro? i.e make it point to a statement that returns `OpInfo`?

crates/interpreter/src/instructions/opcode.rs
gakonst: was wondering if `gas` should be part of the opcode, but maybe not? it'd be nice if they were all declared in one place, but ig fine as-is?

gakonst (commented):
generally supportive - this seems hard to break down further (except the visibility changes I guess). i like the new opcode declaration method. seems like const fns are paying off. 

the changes outside of opcode.rs seem non-controversial/pure improvements, but defer to Dragan for all that

crates/interpreter/src/instructions/opcode.rs
DaniPopes: I considered doing this but each match arm is pretty substantial for each opcode and would clutter the main invocation

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -1,454 +1,314 @@
...
+        */
+
+        /// Evaluates the opcode in the given context.
+        #[inline(always)]
+        pub(crate) fn eval<SPEC: Spec>(opcode: u8, interpreter: &mut Interpreter, host: &mut dyn Host) {
+            // See https://github.com/bluealloy/revm/issues/310#issuecomment-1664381513
+            // for previous efforts on optimizing this function.
+            let f: Instruction = match opcode {
+                $($name => $f as Instruction,)*
+                _ => control::not_found as Instruction,
```
rakita, line 65: I like this a lot! casting to `Instruction` allows us to be sure that every `fn` has the same signature. @DaniPopes do you think this is the reason for the performance boost?

DaniPopes: IIRC it's a bit more than half of the total perf improvement of this pr

rakita:
@DaniPopes added mostly nits, moved all instructions to be public as that can be useful, and added some small changes.

very nice PR!

rakita (approved):
the test is failing for an unrelated reason.

PR #585, opened 2023-08-06
fix(transient_storage): set previous value in journal

rakita:
Journaled values (Value that we revert to) should have been the previous value and not the new value

Introduced in initial EIP here: https://github.com/bluealloy/revm/pull/546
found when running eth tests. Tests will be added in ci in next commit.

@tynes fyi


tynes:
Thank you for fixing this

PR #586, opened 2023-08-06
fix(mcopy): memory expansion and add eth tests to ci

rakita:
`memory_resize` resize macro was not correctly used, inputs of it are `offset` and `length` but in code we had `src` and `resized` value of `max(src,dst)+len` that was not correct.

Added eth test for MCOPY and TransientStorage to ci. Found this problem while running those tests.
Bug was introduced in initial impl: https://github.com/bluealloy/revm/pull/528

@0xJepsen @DaniPopes fyi


0xJepsen:
Nice catch

PR #587, opened 2023-08-08
feat: introduce initcode size limit check taking config into account

Evalir:
Essentially re-adding https://github.com/bluealloy/revm/commit/dd4a13c418e5dd0ab89dfb91d0c89a17806b2f47 , but updated to use in relevant places.

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -318,25 +318,6 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB, 
             });
         }
 
-        // Check initcode size limit on Shanghai spec.
-        if GSPEC::enabled(SHANGHAI) {
-            let max_initcode_size = self
-                .data
-                .env
-                .cfg
-                .limit_contract_code_size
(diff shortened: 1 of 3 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -318,6 +318,25 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+            if inputs.init_code.len() > max_initcode_size {
+                return Err(CreateResult {
+                    result: InstructionResult::CreateInitcodeSizeLimit,
+                    created_address: None,
+                    gas,
+                    return_value: Bytes::new(),
+                });
+            }
+        }
+
```
rakita, line 339: This seems not needed, this check was moved inside `primitives/src/env`

Evalir: right, was wondering if this check was redundant as we were checking this before

Evalir: done in https://github.com/bluealloy/revm/pull/587/commits/345d69c8786f357cd88b68dab736d62b21ab03ed

rakita (approved):

PR #592, opened 2023-08-09
remove unneccesary var and if branch

bemevolent:
Super minor, but given the amount of times this gets called during execution it makes sense to optimize cpu cycles.

This removes the unnecessary `overflow` var and the branch in `if overflow || ` by using `saturating_add`. This seems fine since `all_used_gas` is not updated unless under the limit


rakita (approved):
lgtm!

PR #598, opened 2023-08-15
chore: export some `unreachable_pub` items

DaniPopes:
Some items, like `to_precompile_id`, aren't exported at the crate level, even if their visibility is `pub`.


rakita (approved):
good cleanup!

PR #609, opened 2023-08-17
fix(inspector): call call_end/create_end when inspector shortcircuits calls

Evalir:
If the Inspector shortcircuits the call of `call/create`, functions for call end `call_end/create_end` were not called, these calls were expected by the foundry and it broke a few things.

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -417,9 +417,8 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB, 
 
     /// EVM create opcode for both initial crate and CREATE and CREATE2 opcodes.
     fn create_inner(&mut self, inputs: &CreateInputs) -> CreateResult {
-        let res = self.prepare_create(inputs);
-
-        let prepared_create = match res {
+        // Prepare crate.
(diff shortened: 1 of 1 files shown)


crates/revm/src/evm_impl.rs
```diff
@@ -415,16 +415,55 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
     }
 
+    fn create_end(
+        &mut self,
+        inputs: &CreateInputs,
+        ret: InstructionResult,
+        address: Option<Address>,
+        gas: Gas,
+        out: Bytes,
+    ) -> (InstructionResult, Option<Address>, Gas, Bytes) {
```
rakita, line 425: Can you revert back to how it was, as this function (and others `create_precheck `,`call_..`) just wrap around inspector calls and hides just very small functionality and is called only from one place, it is not worth imo.

rakita: Not exactly what I asked, will make a change to push it over the line.

rakita (approved):
@Evalir can you double check the PR?

Evalir:
@rakita yep double checking rn

Evalir (commented):
Alright @rakita — triple checked, confirmed it works as expected. Thank you!

PR #614, opened 2023-08-18
test(state): account & storage revert value preservation

rkrasiuk:
## Description

Add a test that applies multiple transitions to the state and asserts that the old values of the existing and new accounts and their respective storages are preserved.


rakita (approved):
lgtm

PR #625, opened 2023-08-22
feat: Optional coinbase tip

clabby:
## Overview

Adds a new ~~feature flag, `optional_coinbase_tip`~~ `CfgEnv` option, `disable_coinbase_tip`, that allows for disabling the tip to coinbase during transaction finalization. This option will be used in https://github.com/anton-rs/op-reth to support the Regolith hardfork.


rakita:
Can you do it without a feature flag, this option can be generally useful.

And (maybe not related) but is this line okay to stay, loading coinbase address as warm if coinbase does not receive a fee:
https://github.com/bluealloy/revm/blob/64b0874c107bf5fe999a3258dc8345a88ef39e68/crates/revm/src/evm_impl.rs#L108-L115

rakita (approved):
lgtm

PR #627, opened 2023-08-22
test(state): bundle selfdestructs

rkrasiuk:
## Description

Add a test checking that the behavior of state handling destroyed account within a block is correct.

_The test is currently failing, anything I'm doing wrong @rakita or is it a bug?_


crates/revm/src/db/states/state.rs
```diff
@@ -547,4 +547,126 @@ mod tests {
...
+                previous_status: AccountStatus::DestroyedChanged,
+                previous_info: Some(existing_account_info.clone()),
+                // storage change should be ignored
+                storage: HashMap::from([(
+                    slot1,
+                    StorageSlot {
+                        previous_or_original_value: U256::from(1),
+                        present_value: U256::from(2),
+                    },
+                )]),
```
rakita, line 612: Destroyed account (`DestroyedAgain` here) will not have storage

Selfdestruct is called here: https://github.com/bluealloy/revm/blob/main/crates/revm/src/db/states/cache.rs#L102
And transition created here:
https://github.com/bluealloy/revm/blob/main/crates/revm/src/db/states/cache_account.rs#L241

We should maybe be more explicit about this set storage to empty hashmap here: https://github.com/bluealloy/revm/blob/main/crates/revm/src/db/states/transition_account.rs#L68

rkrasiuk:
@rakita should be ready for merge now

PR #628, opened 2023-08-22
feat(`interpreter`): add hash to bytecode

Evalir:
Caches the bytecode on `Contract` to only call `hash_slow` once during `Contract`'s creation.

Should speed up the coverage interpreter in foundry. cc @rakita

### DIFF
--- a/crates/interpreter/src/interpreter/contract.rs
+++ b/crates/interpreter/src/interpreter/contract.rs
@@ -1,7 +1,7 @@
 use super::analysis::{to_analysed, BytecodeLocked};
 use crate::primitives::{Bytecode, Bytes, B160, U256};
 use crate::CallContext;
-use revm_primitives::{Env, TransactTo};
+use revm_primitives::{Env, TransactTo, B256};
 
 #[derive(Clone, Default)]
 pub struct Contract {
@@ -10,6 +10,8 @@ pub struct Contract {
     /// Bytecode contains contract code, size of original code, analysis with gas block and jump table.
     /// Note that current code is extended with push padding and STOP at end.
     pub bytecode: BytecodeLocked,
+    /// Bytecode hash.
+    pub hash: B256,
     /// Contract address
     pub address: B160,
     /// Caller of the EVM.
@@ -20,11 +22,13 @@ pub struct Contract {
 
 impl Contract {
     pub fn new(input: Bytes, bytecode: Bytecode, address: B160, caller: B160, value: U256) -> Self {
+        let hash = bytecode.hash_slow();
         let bytecode = to_analysed(bytecode).try_into().expect("it is analyzed");
 
         Self {
             input,
             bytecode,
+            hash,
             address,
             caller,
             value,
--- a/crates/interpreter/src/interpreter/contract.rs
+++ b/crates/interpreter/src/interpreter/contract.rs
@@ -21,8 +21,7 @@ pub struct Contract {
 }
 
 impl Contract {
-    pub fn new(input: Bytes, bytecode: Bytecode, address: B160, caller: B160, value: U256) -> Self {
-        let hash = bytecode.hash_slow();
+    pub fn new(input: Bytes, bytecode: Bytecode, hash: B256,address: B160, caller: B160, value: U256) -> Self {
         let bytecode = to_analysed(bytecode).try_into().expect("it is analyzed");
 
         Self {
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/interpreter/contract.rs
```diff
@@ -20,11 +22,13 @@ pub struct Contract {
 
 impl Contract {
     pub fn new(input: Bytes, bytecode: Bytecode, address: B160, caller: B160, value: U256) -> Self {
+        let hash = bytecode.hash_slow();
```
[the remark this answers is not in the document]
Evalir: right, you mean like this? https://github.com/bluealloy/revm/pull/628/commits/ab9d4b4c67fb143b00efafeb1496096a99d3b1ed

crates/revm/src/evm_impl.rs
```diff
@@ -396,9 +396,13 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
         };
 
+        let bytecode = Bytecode::new_raw(inputs.init_code.clone());
+        let hash = bytecode.hash_slow();
+
         let contract = Box::new(Contract::new(
             Bytes::new(),
-            Bytecode::new_raw(inputs.init_code.clone()),
+            bytecode,
+            hash,
```
rakita, line 405: There is `code_hash` variable above that can be used here. Doing keccak two times is not needed.

crates/revm/src/evm_impl.rs
```diff
@@ -640,9 +644,12 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
             });
         }
 
+        let hash = bytecode.hash_slow();
```
rakita, line 647: On the line 601
You can replace `self.code` with something like
```
         self
            .data
            .journaled_state
            .load_code(address, &mut self.data.db)
            .map_err(|e| ...;
```
This would return an Account that contains both the hash and bytecode.

Evalir: Done in https://github.com/bluealloy/revm/pull/628/commits/f895c57874acd6db2b5877375734d3102a08b4ea, albeit involves a bytecode clone to not use `take_bytecode`.

crates/revm/src/evm_impl.rs
```diff
@@ -598,13 +601,30 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+            Err(_) => {
+                return Err(CallResult {
+                    result: InstructionResult::FatalExternalError,
+                    gas,
+                    return_value: Bytes::new(),
+                })
+            }
+        };
+        let code_hash = account.info.code_hash();
+        let bytecode = match account.info.code {
```
rakita, line 619: you can `unwrap_or_default()` here

crates/revm/src/evm_impl.rs
```diff
@@ -598,13 +601,30 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
+            Some(ref code) => code.clone(),
+            None => {
+                return Err(CallResult {
+                    result: InstructionResult::FatalExternalError,
+                    gas,
+                    return_value: Bytes::new(),
+                })
+            }
```
rakita, line 627:
```suggestion
        let bytecode = match account.info.code.unwrap_or_default();
```

code should be present, but to not use `expect` we can use `unwrap_or_default`

crates/revm/src/evm_impl.rs
```diff
@@ -598,13 +601,30 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> EVMImpl<'a, GSPEC, DB,
...
-                return_value: Bytes::new(),
-            });
+        let account = match self
+            .data
+            .journaled_state
+            .load_code(inputs.contract, self.data.db)
+        {
+            Ok((account, _)) => account,
+            Err(_) => {
+                return Err(CallResult {
```
rakita, line 611:
```suggestion
            Err(e) => {
                self.data.error = e;
                return Err(CallResult {
```

Error should be set here.

crates/revm/src/evm_impl.rs
Evalir: Done in https://github.com/bluealloy/revm/pull/628/commits/6ba9372e6f3255155181175b7440401b8327adfc

crates/revm/src/evm_impl.rs
Evalir: Done in https://github.com/bluealloy/revm/pull/628/commits/6ba9372e6f3255155181175b7440401b8327adfc

rakita (approved):

PR #629, opened 2023-08-22
chore(interpreter): use `let else`

DaniPopes:
Split off from #582

And macros style changes.

Improves readability. MSRV for this feature is 1.65 but I see it's already present in `revm`.

### DIFF
--- a/crates/interpreter/src/instructions/macros.rs
+++ b/crates/interpreter/src/instructions/macros.rs
@@ -39,8 +39,8 @@ macro_rules! gas_or_fail {
     ($interp:expr, $gas:expr) => {
         if crate::USE_GAS {
             match $gas {
-                Some(gas_used) if $interp.gas.record_cost(gas_used) => {}
-                _ => {
+                Some(gas_used) => gas!($interp, gas_used),
+                None => {
                     $interp.instruction_result = InstructionResult::OutOfGas;
                     return;
                 }
--- a/crates/interpreter/src/instructions/macros.rs
+++ b/crates/interpreter/src/instructions/macros.rs
@@ -1,5 +1,3 @@
-pub use crate::InstructionResult;
-
 macro_rules! check_staticcall {
     ($interp:expr) => {
         if $interp.is_static {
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/instructions/macros.rs
```diff
@@ -30,19 +28,19 @@ macro_rules! gas {
...
 }
 
 macro_rules! gas_or_fail {
     ($interp:expr, $gas:expr) => {
         if crate::USE_GAS {
             match $gas {
-                Some(gas_used) => gas!($interp, gas_used),
-                None => {
+                Some(gas_used) if $interp.gas.record_cost(gas_used) => {}
+                _ => {
```
gakonst, line 43: nit any reason to not use the macro

crates/interpreter/src/instructions/macros.rs
```diff
@@ -178,59 +174,57 @@ macro_rules! pop_top {
...
-			match $interp.stack.push_b256($x) {
-				Ok(()) => (),
-				Err(e) => {
-                    $interp.instruction_result = e;
-                    return
-                },
-			}
-		)*
-	)
+	($interp:expr, $($x:expr),* $(,)?) => ($(
```
gakonst, line 177: this now lets you push multiple values? or what is the new `$(,)?` for?

gakonst (approved):
seems fine but defer to dragan for the stylistical choice , know seitz doesn't like let else

crates/interpreter/src/instructions/macros.rs
DaniPopes: Just allows trailing comma.

DaniPopes:
It also avoids unwrap, `let else` was "invented" for this exact use case.

crates/interpreter/src/instructions/macros.rs
DaniPopes: Would duplicate the assign/return on both branches, doesn't really matter since compiler probably optimizes that anyway

rakita: Can you revert this one, I slightly vary on the fn call side-effect inside if, and second reason is that `gas!` is used everywhere so it converges to one place

rakita (commented):
one nit for `gas!` macro. Other things lgtm!

crates/interpreter/src/instructions/macros.rs
DaniPopes: Understandable

DaniPopes: fac35ad7331a39d91eb353a9e9d5366127730cda

PR #630, opened 2023-08-22
perf(interpreter): improve i256 instructions

DaniPopes:
Split off from #582

Summary:
- `i256_sign`:
  - is split into the two possible variants as before (with the const bool), so we can use it with an immutable reference, avoiding copies in `i256_cmp` (see `slt`, `sgt` impls). `i256_sign_compl` gets optimized away like `i256_sign::<true>` before
  - by sorting  `Sign`  variants we can leverage derived `Eq, Ord` impls, which are can be further optimized by the compiler. Also, by using `-1, 0, 1` it will be the exact same `cmp::Ordering` and the `signum` function result, which helps in the sign function (where we transmute `num != 0` directly to `Sign`)
- `i256_div`: if condition generates better code than a fully exhaustive match (branchless vs multiple branches). I don't know why, I would assume the two to be identical ... godbolt: <https://godbolt.org/z/939obPrsj>
- other small changes are for readability: `overflowing().0` == `wrapping()`, `>> == wrapping_shr`

### DIFF
--- a/crates/interpreter/src/instructions/bitwise.rs
+++ b/crates/interpreter/src/instructions/bitwise.rs
@@ -1,4 +1,4 @@
-use super::i256::{i256_cmp, i256_sign, two_compl, Sign};
+use super::i256::{i256_cmp, i256_sign_compl, two_compl, Sign};
 use crate::{
     gas,
     primitives::SpecId::CONSTANTINOPLE,
@@ -31,21 +31,13 @@ pub fn gt(interpreter: &mut Interpreter, _host: &mut dyn Host) {
 pub fn slt(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 = if i256_cmp(op1, *op2) == Ordering::Less {
-        U256::from(1)
-    } else {
-        U256::ZERO
-    }
+    *op2 = U256::from(i256_cmp(&op1, op2) == Ordering::Less);
 }
 
 pub fn sgt(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 = if i256_cmp(op1, *op2) == Ordering::Greater {
-        U256::from(1)
-    } else {
-        U256::ZERO
-    };
+    *op2 = U256::from(i256_cmp(&op1, op2) == Ordering::Greater);
 }
 
 pub fn eq(interpreter: &mut Interpreter, _host: &mut dyn Host) {
@@ -125,26 +117,21 @@ pub fn sar<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
 
-    let value_sign = i256_sign::<true>(op2);
+    let value_sign = i256_sign_compl(op2);
 
     *op2 = if *op2 == U256::ZERO || op1 >= U256::from(256) {
         match value_sign {
             // value is 0 or >=1, pushing 0
             Sign::Plus | Sign::Zero => U256::ZERO,
             // value is <0, pushing -1
-            Sign::Minus => two_compl(U256::from(1)),
+            Sign::Minus => U256::MAX,
         }
     } else {
+        const ONE: U256 = U256::from_limbs([1, 0, 0, 0]);
         let shift = usize::try_from(op1).unwrap();
-
         match value_sign {
-            Sign::Plus | Sign::Zero => *op2 >> shift,
-            Sign::Minus => {
-                let shifted = ((op2.overflowing_sub(U256::from(1)).0) >> shift)
-                    .overflowing_add(U256::from(1))
-                    .0;
-                two_compl(shifted)
-            }
+            Sign::Plus | Sign::Zero => op2.wrapping_shr(shift),
+            Sign::Minus => two_compl(op2.wrapping_sub(ONE).wrapping_shr(shift).wrapping_add(ONE)),
         }
     };
 }
--- a/crates/interpreter/src/instructions/i256.rs
+++ b/crates/interpreter/src/instructions/i256.rs
@@ -1,27 +1,17 @@
 use crate::primitives::U256;
 use core::cmp::Ordering;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-#[cfg_attr(
-    any(test, feature = "arbitrary"),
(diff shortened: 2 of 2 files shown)


crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -125,26 +117,21 @@ pub fn sar<SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
         match value_sign {
-            Sign::Plus | Sign::Zero => *op2 >> shift,
-            Sign::Minus => {
-                let shifted = ((op2.overflowing_sub(U256::from(1)).0) >> shift)
-                    .overflowing_add(U256::from(1))
-                    .0;
-                two_compl(shifted)
-            }
+            Sign::Plus | Sign::Zero => op2.wrapping_shr(shift),
+            Sign::Minus => two_compl(op2.wrapping_sub(ONE).wrapping_shr(shift).wrapping_add(ONE)),
```
gakonst, line 134: Seems fine if we are OK with going from overflowing -> wrapping

crates/interpreter/src/instructions/i256.rs
```diff
@@ -31,102 +30,101 @@ pub const MIN_NEGATIVE_VALUE: U256 = U256::from_limbs([
...
-    } else {
-        if DO_TWO_COMPL {
-            two_compl_mut(val);
-        }
+pub(super) fn i256_sign(val: &U256) -> Sign {
+    if val.bit(U256::BITS - 1) {
         Sign::Minus
+    } else {
+        // SAFETY: false == 0 == Zero, true == 1 == Plus
+        unsafe { core::mem::transmute(*val != U256::ZERO) }
```
gakonst, line 46: i guess ok

crates/interpreter/src/instructions/i256.rs
```diff
@@ -31,102 +30,101 @@ pub const MIN_NEGATIVE_VALUE: U256 = U256::from_limbs([
...
-        (Sign::Plus, Sign::Plus) => first.cmp(&second),
+pub(super) fn i256_cmp(first: &U256, second: &U256) -> Ordering {
+    let first_sign = i256_sign(first);
+    let second_sign = i256_sign(second);
+    match first_sign.cmp(&second_sign) {
+        // note: adding `if first_sign != Sign::Zero` to short circuit zero comparisons performs
+        // slower on average, as of #582
+        Ordering::Equal => first.cmp(second),
+        o => o,
     }
```
gakonst, line 86: seems fine

gakonst (approved):
overall supportive, defer to dragan for more

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -31,21 +31,13 @@ pub fn gt(interpreter: &mut Interpreter, _host: &mut dyn Host) {
...
 
 pub fn sgt(interpreter: &mut Interpreter, _host: &mut dyn Host) {
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 = if i256_cmp(op1, *op2) == Ordering::Greater {
-        U256::from(1)
-    } else {
-        U256::ZERO
-    };
+    *op2 = U256::from(i256_cmp(&op1, op2) == Ordering::Greater);
```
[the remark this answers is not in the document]
rakita: Bool gets casted to u64, so `true == 1` is fine

crates/interpreter/src/instructions/bitwise.rs
rakita: We can simplify this to `((op2-ONE)>>shift)+1)` as checks for `op2 == zero` and `op1 >= 256` are done and `shr` in background calls `overflowing_shr` inside ruint.

crates/interpreter/src/instructions/i256.rs
```diff
@@ -1,28 +1,27 @@
...
+#[cfg_attr(
+    any(test, feature = "arbitrary"),
+    derive(arbitrary::Arbitrary, proptest_derive::Arbitrary)
+)]
+#[repr(i8)]
+pub(super) enum Sign {
+    // same as `cmp::Ordering`
+    Minus = -1,
+    Zero = 0,
+    #[allow(dead_code)] // "constructed" with `mem::transmute`
```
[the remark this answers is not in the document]
rakita: btw, very smart!

crates/interpreter/src/instructions/i256.rs
```diff
@@ -31,102 +30,101 @@ pub const MIN_NEGATIVE_VALUE: U256 = U256::from_limbs([
...
     }
-    let first_sign = i256_sign::<true>(&mut first);
+    let first_sign = i256_sign_compl(&mut first);
     if first_sign == Sign::Minus && first == MIN_NEGATIVE_VALUE && second == U256::from(1) {
         return two_compl(MIN_NEGATIVE_VALUE);
     }
 
-    //let mut d = first / second;
-    let mut d = first.div_rem(second).0;
+    let mut d = first.wrapping_div(second);
```
rakita, line 100: `wrapping_div` calls `div_rem` in background, we can even do `first / second` as it calls same things

rakita (approved):
lgtm! Had a few small nits, really like what you did with `Sign`

crates/interpreter/src/instructions/bitwise.rs
DaniPopes: Makes sense in the context that ruint operations are overloaded to be wrapping, but it might not be obvious without this knowledge. I wrote them with explicit `wrapping`, normally I'd use operators but this is one of the very few instances where I'd rather use explicit methods

crates/interpreter/src/instructions/i256.rs
```diff
@@ -1,28 +1,27 @@
...
+pub(super) enum Sign {
+    // same as `cmp::Ordering`
+    Minus = -1,
+    Zero = 0,
+    #[allow(dead_code)] // "constructed" with `mem::transmute`
+    Plus = 1,
 }
 
-pub const _SIGN_BIT_MASK: U256 = U256::from_limbs([
+pub(super) const _SIGN_BIT_MASK: U256 = U256::from_limbs([
```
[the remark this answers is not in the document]
DaniPopes: I've also removed I256, isn't used anywhere

DaniPopes:
Noticed one more thing (a774b294e50c0c970822105449d8552275b61cb3), we can drop the zero check in div and mod after doing the operation, since it allows for more optimizations at function return and results in better codegen overall.

crates/interpreter/src/instructions/bitwise.rs
rakita: `wrapper_*` are nicer here then `overflowing_*` so I will just merge PR.
But I am curious why you think this is one of those instances, bearing in mind that we already check for `op2 == zero and op1 >= 256`

PR #633, opened 2023-08-22
chore: misc improvements

DaniPopes:
Split off from #582

Note that ruint >1.10 is needed for main #582 changes.


gakonst (approved):
LGTM

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -460,8 +456,7 @@ mod tests {
                 vec![(slot(), U256::from(10))],
             )]],
             vec![],
-        );
-        bundle
+        )
```
DaniPopes, line 459: I think this is a clippy lint

rakita (approved):
ty!

PR #635, opened 2023-08-23
bug(state): remove redundant info revert on destruct

rakita:
On revert check if account info was changed or not.

Notices here: https://github.com/bluealloy/revm/pull/627/files/94c98a615fd35189c4b20a7ca82b44c2bf885dc9#diff-549cde144deee3e0083097545a0eff627800d0ea38bcd720f69477d3136a4a2bR664

Where info transition from Loaded to DestroyedChanged would result the same info, but we would save redundant `RevertTo` of previous account.


rkrasiuk (approved):
lgtm!

PR #637, opened 2023-08-23
style: bundle state

rkrasiuk:
## Description

Minor stylistic changes to `BundleState`. 

@rakita I'm not opinionated, you can pick and choose which ones you like (if any).


rakita (approved):
lgtm, good one

PR #646, opened 2023-08-24
chore(state): bundle state split

rkrasiuk:
## Description

* Consume bundle state entirely upon taking the plain state
* Create additional function for splitting plain state and reverts

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -174,18 +174,46 @@ impl BundleState {
         self.reverts.push(reverts);
     }
 
-    /// Nuke the bundle state and return sorted plain state.
+    /// Return and clear all reverts from BundleState, sort them before returning.
+    pub fn take_reverts(&mut self) -> StateReverts {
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -234,32 +262,14 @@ impl BundleState {
...
-                }
-            }
-            accounts.par_sort_unstable_by(|a, b| a.0.cmp(&b.0));
-            state_reverts.accounts.push(accounts);
-            state_reverts.storage.push(storage);
-        }
-
-        state_reverts
+    /// Consume the bundle state and split it into reverts and plain state.
+    pub fn split_into_plain_and_reverts(
```
rakita, line 266: maybe a name change to 
```suggestion
    pub fn into_sorted_plain_state_and_reverts(
```

Would like to have `into_` first to show that this fn consumes self

rkrasiuk: @rakita fixed

PR #648, opened 2023-08-25
fix(state): Revert from DestroyedChanged to DestroyedAgain

rakita:
TODO was right, it was invalid.

The storage that we should have used was the present storage of the bundle account. As this is transition to `DestroyedAgain` updated_storage would always be empty.


rkrasiuk (approved):
nice!

Issue #644, opened 2023-08-24
Make BundleBuilder


rakita:
The current initialization of the bundle is with `new` function that takes two big iterators. This should be done a lot better with a builder that is going to output the bundle state.

https://github.com/bluealloy/revm/blob/0b9c12e6e9d876a93deaf718d0988e3f6c3c7031/crates/revm/src/db/states/bundle_state.rs#L58-L77

darknight:
I'm reading the codebase and come across this issue, wondering if it's good to be first issue. If so, I'd like to take it.
And If I understand correctly, what is expected here is apply builder pattern to initialize `BundleState`, correct?

rakita:
> @rakita Hi, I'm reading the codebase and come across this issue, wondering if it's good to be first issue. If so, I'd like to take it. And If I understand correctly, what is expected here is apply builder pattern to initialize `BundleState`, correct?

It is a good first issue, go for it!

rakita:
closed https://github.com/bluealloy/revm/pull/649

PR #649, opened 2023-08-26
BundleBuilder for BundleState initialization

darknight:
This PR is for closing issue https://github.com/bluealloy/revm/issues/644

Add `BundleBuilder` and a set of member methods to collect initialization information for `BundleState` struct.

The builder enables chaining of method calls.

Add 2 more test cases.

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -7,6 +7,196 @@ use revm_interpreter::primitives::{
     hash_map::{self, Entry},
     AccountInfo, Bytecode, HashMap, StorageSlot, B160, B256, KECCAK_EMPTY, U256,
 };
+use std::collections::{BTreeMap, BTreeSet, HashSet};
+use std::ops::RangeInclusive;
+
+/// This builder is used to help to facilitate the initialization of `BundleState` struct
+#[derive(Debug)]
+pub struct BundleBuilder {
+    states: HashSet<B160>,
+    state_original: HashMap<B160, AccountInfo>,
+    state_present: HashMap<B160, AccountInfo>,
+    state_storage: HashMap<B160, HashMap<U256, (U256, U256)>>,
+
+    reverts: BTreeSet<(u64, B160)>,
+    revert_range: RangeInclusive<u64>,
+    revert_account: HashMap<(u64, B160), Option<Option<AccountInfo>>>,
+    revert_storage: HashMap<(u64, B160), Vec<(U256, U256)>>,
+
+    contracts: HashMap<B256, Bytecode>,
+}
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -7,6 +7,148 @@ use revm_interpreter::primitives::{
...
+use std::collections::HashSet;
+
+#[derive(Debug, Default)]
+pub struct BundleBuilder {
+    states: HashSet<B160>,
+    state_original: HashMap<B160, AccountInfo>,
+    state_present: HashMap<B160, AccountInfo>,
+    state_storage: HashMap<B160, HashMap<U256, (U256, U256)>>,
+
+    reverts: HashSet<(usize, B160)>,
```
rakita, line 19: `HashSet` would not preserve the order of reverts. So it is better to use `BTreeMap`.

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -7,6 +7,148 @@ use revm_interpreter::primitives::{
...
+                    self.state_present.remove(&address),
+                    storage,
+                    AccountStatus::Changed,
+                );
+                (address, bundle_account)
+            })
+            .collect();
+
+        let block_reverts = self
+            .reverts
```
rakita, line 111: You need to be careful to fill the gaps and insert empty vec if there is no reverts for that block.

For example:
`[[Block1Revert,[],[Block3Revert]]` if block1 and block3 reverts are inserted, block2 empty revert should be present.

darknight: Ah yes, thanks for the catch, I missed it.

darknight:
@rakita Need your review here, thanks :)

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -7,6 +7,185 @@ use revm_interpreter::primitives::{
...
+    revert_storage: HashMap<(usize, B160), Vec<(U256, U256)>>,
+
+    contracts: HashMap<B256, Bytecode>,
+}
+
+impl BundleBuilder {
+    /// Create builder instance
+    ///
+    /// `revert_capacity` indicates the size of BundleState `reverts` field
+    pub fn new(revert_capacity: usize) -> Self {
```
rakita, line 32: Would do it a little bit differently and use `RangeInclusive<u64>` here. Where u64 would represent the block number.

The idea is to have API with block number as this is the number that is expected, `block_index` seems unnatural.

And you have can iterate over it with when creating reverts.

Other things seem good

rakita (approved):
lgtm!

Issue #445, opened 2023-04-04
Share memory between calls


rakita:
more info here: https://github.com/ethereum/evmone/issues/481#issuecomment-1179088645

It seems like a good idea for both memory and stack and would remove unnecessary allocation inside evm.

rakita:
Closed in: https://github.com/bluealloy/revm/pull/673

PR #650, opened 2023-08-26
Share memory between calls

lorenzofero:
I tried to give a shot to https://github.com/bluealloy/revm/issues/445 just for fun and as a challenge with this draft PR, which is still WIP. 
This introduces a new struct called `SharedMemory` which is indeed shared between calls. I'd like a feedback on my implementation to see what it can done better. 

Here are some implementation choices I made:
- its API matches almost completely the 'old' interpreter `Memory` implementation
- memory is allocated using this estimate https://2π.com/22/eth-max-mem/ (more on this below)
- it has a pointer `current_slice` which is internally used to refer to the portion of data reserved for the current context. This requires two `unsafe` methods `get_current_slice` and `get_current_slice_mut` which deferences the raw pointer
- `current_slice` pointer is updated when entering a new context (which is when a new `Interpreter` instance is created) using the `use_new_memory` method (every name is still WIP too) and when exiting it, which happens when the `return_value` method of `Interpreter` is called
- in order to make `SharedMemory` work properly, it is wrapped in a `Rc` for dealing with recursive calls and with an `RefCell` to allow mutability

Little problem: now test passes because I always allocate `u32::MAX` bytes. If I use the estimation however they fail because some tests have an impressive gas limit (at least 4_503_599_627_349_304), and my system fails to allocate so much memory (34_359_713_280 bytes). I don't really know how to handle this at the moment

Thanks in advance for any feedback!

### DIFF
--- a/crates/interpreter/src/host.rs
+++ b/crates/interpreter/src/host.rs
@@ -1,3 +1,5 @@
+use core::cell::RefCell;
+
 use crate::primitives::Bytecode;
 use crate::{
     primitives::{Bytes, Env, B160, B256, U256},
@@ -7,6 +9,8 @@ use alloc::vec::Vec;
 pub use dummy::DummyHost;
 
 mod dummy;
+use alloc::rc::Rc;
+use revm_primitives::shared_memory::SharedMemory;
 
 /// EVM context host.
 pub trait Host {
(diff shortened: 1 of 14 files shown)


lorenzofero:
Closing this as I've found some stuff that needs more work than I thought

PR #651, opened 2023-08-26
fix(state): drop storage only for DestroyedChanged

rakita:
Dropping the storage was a bug as it should have been saved inside the `Revert`.

Storage was drained for `Destroyed`, `DestroyedAgain` and for `DestroyedChanged` it depends on the Transition wipe flag if we are going to move present storage to the `Revert` or not.


rkrasiuk (approved):
lgtm

PR #657, opened 2023-08-29
feat(state): add a flag allowing transition merge without reverts

rkrasiuk:
## Description

Add a flag to `State::merge_transitions` that allows merging transitions to `BundleState` without reverts.

### DIFF
--- a/crates/revm/src/db/states/state.rs
+++ b/crates/revm/src/db/states/state.rs
@@ -125,13 +125,11 @@ impl<'a, DBError> State<'a, DBError> {
     /// This action will create final post state and all reverts so that
     /// we at any time revert state of bundle to the state before transition
     /// is applied.
-    pub fn merge_transitions(&mut self) {
-        if let Some(transition_state) = self.transition_state.as_mut() {
-            let transition_state = transition_state.take();
-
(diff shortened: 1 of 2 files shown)


crates/revm/src/db/states/state.rs
```diff
@@ -125,13 +125,11 @@ impl<'a, DBError> State<'a, DBError> {
...
-            let transition_state = transition_state.take();
-
-            self.bundle_state
-                .get_or_insert(BundleState::default())
-                .apply_block_substate_and_create_reverts(transition_state);
+    pub fn merge_transitions(&mut self, with_reverts: bool) {
+        if let Some(transition_state) = self.transition_state.take() {
+            let bundle_state = self.bundle_state.get_or_insert(BundleState::default());
+            bundle_state.set_should_collect_reverts(with_reverts);
+            bundle_state.apply_block_substate_and_create_reverts(transition_state);
```
rakita, line 132: Shouldn't we just add `with_reverts` to  `apply_block_substate_and_create_reverts`, rn we are saving it inside BundleState and changing it every time?

rakita (approved):
lgtm

Issue #658, opened 2023-08-29
Bug: gas cost calculation is wrong


kunxian-xia:
The gas cost to run pairing precompile is wrong. The right formula is 

```
let gas_used = 
     input.len() as u64 / PAIR_ELEMENT_LEN as u64 * pair_per_point_cost + pair_base_cost; 
```

For example, if the input.len() = 191, then the gas_cost should be `pair_base_cost` instead of `33822 + pair_base_cost` (assuming pair_per_point_cost = 34000).

https://github.com/bluealloy/revm/blob/13e471eeed703e7c4e58c6b0443104711b2b5a40/crates/precompile/src/bn128.rs#L168-L169

PR #659, opened 2023-08-29
fix: pairing cost formula 

kunxian-xia:
This pr fixes the https://github.com/bluealloy/revm/issues/658.

### DIFF
--- a/crates/precompile/src/bn128.rs
+++ b/crates/precompile/src/bn128.rs
@@ -166,7 +166,7 @@ fn run_pair(
     gas_limit: u64,
 ) -> PrecompileResult {
     let gas_used =
-        pair_per_point_cost * input.len() as u64 / PAIR_ELEMENT_LEN as u64 + pair_base_cost;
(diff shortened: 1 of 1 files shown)


rakita:
Posting this here for awareness https://github.com/bluealloy/revm/issues/658#issuecomment-1697206507.

> the gas used is indeed wrong but the precompile expects that len be multiple of PAIR_ELEMENT_LEN so the outcome would be the same:
> 
> https://github.com/bluealloy/revm/blob/main/crates/precompile/src/bn128.rs#L176-L178

roynalnaruto:
@rakita our test case was an out-of-gas case, but with `len(input)` not a multiple of `192`. The trace from go-ethereum indeed catches the out-of-gas error. However we used `revm` in our [`bus-mapping`](https://github.com/scroll-tech/zkevm-circuits/tree/develop/bus-mapping) implementation to get the precompile result, where although the call fails (due to the `len(input)` not being a multiple of `192`), it fails for a different reason (not out-of-gas). And this resulted in a mismatch between traces from `go-ethereum` and the witness we construct in `zkEvm`.

PR #660, opened 2023-08-29
chore: remove all `dyn` for opcode that reduce dyn trait cost

yjhmelody:


rakita:
> yjhmelody

wdym by dyn trait cost? In the end this shoudn't this be dynamically dispatched as it can be called in recursion

rakita:
Resolved here: https://github.com/bluealloy/revm/pull/659

rakita:
> Could not see it will be a problem for recursion. And you could still use dyn outside. I just remove the inner dyn.

Sure, but the present code works as well.

Could you run `cachegrind` from here: https://github.com/bluealloy/revm/pull/582
Just to check performance.

yjhmelody:
You could try. I am using macOS but `valgrind: Linux is required for this software.`

DaniPopes:
Can we try after the interpreter refactor is merged? I wanted to explore this as well after that PR.

rakita:
> You could try. I am using macOS but `valgrind: Linux is required for this software.`
> 
> I don't think this will cause performance degradation at least. After all, the `eval` function mixed the static dispatch and dynamic dispatch of Host before. I just remove the dynamic part.

Same here :) will try it with docker as it is something i will need in future. tbh i don't expect a difference but it is best just to check.

> Can we try after the interpreter refactor is merged? I wanted to explore this as well after that PR (#582).

That makes sense, will come back to this after #582 gets merged.

yjhmelody:
Fine

rakita:
Tested this on macbook inside docker for snailtracer. 
Main is currently at `400,122,558` Instructions
And this branch gives `447,359,474` instructions

Just for reference main before `perf2` branch was showing `464,501,741` instructions.

Will close this for now.

rakita:
@yjhmelody as dani wrote in #739 there is a 2% performance, so we are changing it to `H: Host`

PR #661, opened 2023-08-29
feat: add BundleState::revert_latest

mattsse:
adds `BundleState::revert_latest` for convenience if you want to revert 1 single transition

moves revert logic from `revert` to `revert_latest` and replace the while loop


rakita (approved):
nice!

PR #662, opened 2023-08-29
fix(state): state transition regression

rkrasiuk:
# Description

Revert the change from #657 that introduced a regression.


rakita (approved):
lgtm

PR #663, opened 2023-08-29
Matt/review 1

mattsse:
overall this abstraction is very easy to follow.

most of my comments concern docs, it would be useful to give an overview of the used terms

* what exactly is a Bundle?
* what is an Transition
* and what is a Revert

especially the revert term could potentially be confusing, from what I understand is that this is the inverse of an applied change, so an applied change can be "reverted", which makes sense. But since "revert" is also used by "reverted" transactions, this could be confusing at first.

One suggestion I have regarding the various type aliases is to replace tuples of 3 or more fields with new types, just to make it clearer what exactly these types represent. especially since a lot of them are public types. 

my personal preference for map aliases like
`pub type StorageWithOriginalValues = HashMap<U256, StorageSlot>;` is to avoid them or wrap in a new type because I find them very hard to work with, because I always end up jumping to the alias definition.

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -8,6 +8,8 @@ use revm_interpreter::primitives::{
     AccountInfo, Bytecode, HashMap, StorageSlot, B160, B256, KECCAK_EMPTY, U256,
 };
 
+// mattsse: It's not clear to me what exactly a bundle is, is this a collection of multiple executed transactions that are bundled together?
+
 /// Bundle state contains only values that got changed
 ///
 /// For every account it contains both original and present state.
@@ -19,8 +21,12 @@ use revm_interpreter::primitives::{
 pub struct BundleState {
     /// Account state.
     pub state: HashMap<B160, BundleAccount>,
+    // mattsse: This mentions block, so this is only supposed to represent all changes in a single block? or is this independent of block and just multiple transactions?
+
     /// All created contracts in this block.
     pub contracts: HashMap<B256, Bytecode>,
+    // mattsse: Are these reverted transactions, or Account changes that can be reverted? The `revert` naming makes this a bit ambiguous.
+
     /// Changes to revert.
     ///
     /// If `should_collect_reverts` flag was set to `false`, the revert for any given block will be just an empty array.
@@ -142,9 +148,13 @@ impl BundleState {
         self.contracts.get(hash).cloned()
     }
 
+    // mattsse: what does `substate` mean here exactly?
+
     /// Consume `TransitionState` by applying the changes and creating the reverts
     ///
     /// `with_reverts` flag indicates whether we should collect the reverts.
+
+    // mattsse: what does collecting the reverts mean? keeping the inverse of the transition changes, so they can be reverted/rolled back?
     pub fn apply_block_substate_and_create_reverts(
         &mut self,
         transitions: TransitionState,
@@ -219,9 +229,9 @@ impl BundleState {
 
     /// Consume the bundle state and return sorted plain state.
     ///
-    /// `omit_changed_check` does not check If account is same as
+    /// `omit_changed_check` does not check if account is same as
     /// original state, this assumption can't be made in cases when
-    /// we split the bundle state and commit part of it.
+    /// we split the bundle state and commit parts of it.
     pub fn into_plain_state_sorted(self, omit_changed_check: bool) -> StateChangeset {
         // pessimistically pre-allocate assuming _all_ accounts changed.
         let state_len = self.state.len();
@@ -296,6 +306,7 @@ impl BundleState {
 
     /// Extend the state with state that is build on top of it.
     ///
+    // mattsse: this paragraph is very hard to parse:
     /// For other state, if there a wipe storage flag set inside Revert copy the state
     /// from `this` to `other` revert (if there is no duplicates of course).
     ///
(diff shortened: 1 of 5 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -8,7 +8,9 @@ use revm_interpreter::primitives::{
     AccountInfo, Bytecode, HashMap, StorageSlot, B160, B256, KECCAK_EMPTY, U256,
 };
 
-/// Bundle state contain only values that got changed
+// mattsse: It's not clear to me what exactly a bundle is, is this a collection of multiple executed transactions that are bundled together?
```
rakita, line 11: Created issue for overview docs this here: https://github.com/bluealloy/revm/issues/683

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -142,9 +148,13 @@ impl BundleState {
         self.contracts.get(hash).cloned()
     }
 
+    // mattsse: what does `substate` mean here exactly?
```
rakita, line 151: Should rename this, to `apply_transitions_and_create_reverts`

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -219,9 +229,9 @@ impl BundleState {
 
     /// Consume the bundle state and return sorted plain state.
     ///
-    /// `omit_changed_check` does not check If account is same as
+    /// `omit_changed_check` does not check if account is same as
     /// original state, this assumption can't be made in cases when
-    /// we split the bundle state and commit part of it.
+    /// we split the bundle state and commit parts of it.
```
rakita, line 234: done

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -296,6 +306,7 @@ impl BundleState {
 
     /// Extend the state with state that is build on top of it.
     ///
+    // mattsse: this paragraph is very hard to parse:
```
rakita, line 309: reworded it here: https://github.com/bluealloy/revm/commit/28976553c1110e943c6adfb65a22fd878c957ef4

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -363,9 +374,12 @@ impl BundleState {
         self.reverts.extend(other.reverts);
     }
 
+    // mattsse: It is weird to me that this returns type Self, but only includes the reverts
+    // should this return only the reverts?
```
rakita, line 378: Yeah, it was wierd, made a change here: https://github.com/bluealloy/revm/pull/681
and now return the `Reverts`.

crates/revm/src/db/states/cache_account.rs
```diff
@@ -5,6 +5,8 @@ use super::{
 use revm_interpreter::primitives::{AccountInfo, KECCAK_EMPTY, U256};
 use revm_precompile::HashMap;
 
+// mattsse: can we rephrase this a bit? basically this is an account read from the database that represents the current state of the account inside the database?
```
rakita, line 8: Done

crates/revm/src/db/states/changes.rs
```diff
@@ -15,9 +15,11 @@ pub struct StateChangeset {
     pub contracts: Vec<(B256, Bytecode)>,
 }
 
+// mattsse: a lot of the type aliases are very hard to read, this looks like Address with some bool marker and a list of storage slots with values, but it is not obvious, Personally I'd prefer structs for these tuples or at the very least docs what these types represent.
```
rakita, line 18: Done in this commit: https://github.com/bluealloy/revm/pull/681/files#diff-4695ed30467f83f567379595f36e895b87a25ad850f296c8c40b3528d72e3af1R21

And renamed it for consitency

crates/revm/src/db/states/changes.rs
```diff
@@ -15,9 +15,11 @@ pub struct StateChangeset {
...
 /// Storage changeset
 pub type StorageChangeset = Vec<(B160, (bool, Vec<(U256, U256)>))>;
 
+// mattsse: what is a StateRevert exactly?
 #[derive(Clone, Debug, Default)]
 pub struct StateReverts {
```
rakita, line 24: it is helper structure that helps to structure database aligned output. For example, it splits accounts and storage.

crates/revm/src/db/states/reverts.rs
```diff
@@ -93,6 +93,7 @@ impl AccountRevert {
     }
 }
 
+// mattsse: what does this exactly represent? the action that needs to be taken on revert? maybe rename?
```
rakita, line 96: Added comment here. And will address it more in the docs: https://github.com/bluealloy/revm/issues/683

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -142,9 +148,13 @@ impl BundleState {
         self.contracts.get(hash).cloned()
     }
 
+    // mattsse: what does `substate` mean here exactly?
+
     /// Consume `TransitionState` by applying the changes and creating the reverts
     ///
     /// `with_reverts` flag indicates whether we should collect the reverts.
+
+    // mattsse: what does collecting the reverts mean? keeping the inverse of the transition changes, so they can be reverted/rolled back?
```
rakita, line 157: Yeah in the essence revert is used to rollback the plain state and create the changesets in reth (previous values of change at least).

rakita:
Addressed comments in multiple commits:
https://github.com/bluealloy/revm/pull/675/files, https://github.com/bluealloy/revm/pull/681 and https://github.com/bluealloy/revm/pull/684

And created issue for overview docs that we talked about: https://github.com/bluealloy/revm/issues/683

mattsse:
nice, ty!

closing this now

PR #667, opened 2023-08-30
fix(state): retain destroyed account status on bundle extend

rakita:
Retail destroyed account status when extending bundle state.

This handles extended account status in a more precise way.

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -366,27 +366,7 @@ impl BundleState {
                         }
                     }
                     this.info = other_account.info;
-                    // if this status was destroyed this means we know all the storages
-                    // and new status need to contains that "Destroyed" flag.
-                    this.status = match (
-                        this.status.was_destroyed(),
-                        other_account.status.was_destroyed(),
-                    ) {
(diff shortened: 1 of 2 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -365,8 +365,28 @@ impl BundleState {
                                 .present_value = storage_slot.present_value;
                         }
                     }
-                    this.status = other_account.status;
                     this.info = other_account.info;
+                    // if this status was destroyed this means we know all the storages
+                    // and new status need to contains that "Destroyed" flag.
+                    this.status = match (
```
rkrasiuk, line 371: does it make sense to declare `extend` fn on account status and put this logic there?

rkrasiuk (commented):
fix lgtm, one comment re style/readability

rkrasiuk (approved):

crates/revm/src/db/states/bundle_state.rs
rakita: It does make sense to do that. Moved this logic inside `AccountStatus::transition(..)`

rkrasiuk (approved):
love the comments!

PR #668, opened 2023-08-30
feat: implement EIP-4844

### LINKED ISSUE #526: EIP-4844: Shard Blob Transactions

DaniPopes:
Implements [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844).

cc @rakita @Rjected

Closes #526.
Supersedes #536.


crates/precompile/build.rs
```diff
@@ -0,0 +1,115 @@
...
+fn generate_kzg_settings() {
+    let in_path = "src/kzg/trusted_setup.txt";
+    let out_path = "src/kzg/generated_settings.rs";
+
+    let in_path = Path::new(in_path);
+    println!("cargo:rerun-if-changed={}", in_path.display());
+    assert!(in_path.exists());
+    let contents = format_kzg_settings(in_path);
+    fs::write(out_path, contents).unwrap();
+}
```
Rjected: makes sense - generating consts here makes creating the default kzgsettings easy, and does not require reading the file on startup etc

crates/precompile/src/kzg/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+fn get_kzg_settings() -> &'static KzgSettings {
+    static SETTINGS: OnceCell<KzgSettings> = OnceCell::new();
+    SETTINGS.get_or_init(|| {
+        c_kzg::KzgSettings::load_trusted_setup(
+            generated_settings::G1_POINTS,
+            generated_settings::G2_POINTS,
+        )
+        .expect("failed to load trusted setup")
+    })
+}
```
Rjected: This makes sense given how the precompiles have to access the kzgsettings, but we need to be able to configure this from reth as well somehow.

Maybe we could have a `static mut` and use a pattern similar to this:
https://github.com/tokio-rs/tracing/blob/bf05c61b7234eb1a196ef8d30f773d3ea6b60ce9/tracing-core/src/dispatch.rs#L329-L366

To provide a way to set the `KzgSettings` in reth?

Rjected (commented):
just adding some comments on how we might be able to configure the trusted setup from reth

mattsse:
just flagging that it probably should be Arc<KzgSettings> because KzgSettings is not clone

ref 

https://github.com/paradigmxyz/reth/blob/893f4cf2a27034f4550d3f4443a0fa7c2cd2444b/bin/reth/src/node/mod.rs#L575-L585

crates/precompile/src/lib.rs
```diff
@@ -191,6 +192,22 @@ impl Precompiles {
         })
     }
 
+    pub fn cancun() -> &'static Self {
+        static INSTANCE: OnceCell<Precompiles> = OnceCell::new();
+        INSTANCE.get_or_init(|| {
+            let mut precompiles = Self::berlin().clone();
```
mattsse, line 198: for precompiles, we should consider either an enum

```rust
enum Precompile {
  Fun(fn(&[u8], u64) -> PrecompileResult),
  Fn(Arc<dyn Send + Sync + Fn(&[u8], u64) -> PrecompileResult)
}
```

or always use the Arc variant, 

and a static mut Precompile list that supports configuring global defaults

DaniPopes: This is in my TODO list :)

DaniPopes:
What's clone needed for? AFAIK you just need to pass the instance pointer into C, so I'm thinking just `OnceCell<KzgSettings>`

crates/precompile/src/kzg/mod.rs
DaniPopes: I believe they use `static mut` and manual atomic operations because of MSRV restrictions, `once_cell` is probably better

mattsse:
> What's clone needed for?

we need the KzgSettings in multiple locations (for example tx pool validation) and ideally, we just want to load it once.

mattsse:
hmm, the way precompiles are currently used

`let mut precompiles = Self::berlin().clone();`

it's only really useful to allow custom default precompiles starting with cancun hardfork, pre-cancun this doesn't really make sense.

so actually not sure yet what the ideal static mut should look like.
perhaps just a single `static mut` specifically for the kzg precompile?

DaniPopes:
> we need the KzgSettings in multiple locations (for example tx pool validation) and ideally, we just want to load it once.

You can store a local `&'static KzgSettings` or get it when needed with a global getter function. I don't think `Arc`s are needed here.

See https://github.com/bluealloy/revm/pull/668#discussion_r1310980065 wrt `static mut`s.

mattsse:
okay @Rjected I figured out how the hardfork precompiles are currently instantiated, and since we only need to access the KzgSetting once, a `static Lazy<Mutex<Arc<KzgSettings>>` should be enough to support custom KzgSettings if set at the start of the program

crates/precompile/src/blob/kzg_settings/mod.rs
```diff
@@ -0,0 +1,33 @@
+use c_kzg::KzgSettings;
```
DaniPopes, line 1: This module is exported, lmk wyt @Rjected @mattsse

crates/interpreter/src/instructions/host_env.rs
```diff
@@ -54,3 +57,15 @@ pub fn origin(interpreter: &mut Interpreter, host: &mut dyn Host) {
     gas!(interpreter, gas::BASE);
     push_b256!(interpreter, host.env().tx.caller.into());
 }
+
+// EIP-4844: Shard Blob Transactions
+pub fn blob_hash<SPEC: Spec>(interpreter: &mut Interpreter, host: &mut dyn Host) {
```
DaniPopes, line 62: Having trouble understanding why `EIPTests/StateTests/stEIP4844-blobtransactions` are all failing, the implementation of the opcode is pretty simple. Might've missed setting up the env?

crates/precompile/src/lib.rs
rakita: hm we can simplify a lot of things about initialization by adding function signature that has `CfgEnv` input
No global initialization at all, and we set `KzgSettings` inside `CfgEnv`.

@DaniPopes @mattsse @Rjected wdyt?

crates/precompile/src/blob/kzg_settings/mod.rs
rakita: I tend not to use`mod` inside `revm`

crates/interpreter/src/instructions/host_env.rs
rakita: If `max_fee_per_blob_gas` is set, I assume you need to do some gas calculation and determent the fee that is going to be deduced from the sender.
https://eips.ethereum.org/EIPS/eip-4844

crates/precompile/build.rs
```diff
@@ -0,0 +1,183 @@
...
+    const BYTES: &[u8] = include_bytes!("./g1_points.bin");
+    assert!(BYTES.len() == core::mem::size_of::<G1Points>());
+    unsafe {{ &*BYTES.as_ptr().cast::<G1Points>() }}
+}};
+pub const G2_POINTS: &G2Points = {{
+    const BYTES: &[u8] = include_bytes!("./g2_points.bin");
+    assert!(BYTES.len() == core::mem::size_of::<G2Points>());
+    unsafe {{ &*BYTES.as_ptr().cast::<G2Points>() }}
+}};
+"#
```
gakonst, line 96: This seems fine to me! How would we allow somebody to override the KZG settings with custom params?

Should these be instead `static`, so that somebody can unsafely mutate them via CLI?

crates/precompile/build.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    // load g2 points
+    let mut g2_points = Vec::with_capacity(n_g2);
+    for _ in 0..n_g2 {
+        let line = lines.next().unwrap();
+        let mut bytes = [0; BYTES_PER_G2_POINT];
+        hex::decode_to_slice(line, &mut bytes).unwrap();
+        g2_points.push(bytes);
+    }
```
gakonst, line 64: Nit but as discussed you don't need to read all the points right?

crates/precompile/build.rs
```diff
@@ -0,0 +1,183 @@
...
+/*
+/// Pros:
+/// - no need to call `load_trusted_setup` at runtime
+///
+/// Cons:
+/// - larger runtime static size (722K = `4096*32 + 4096*144 + 65*288`)
+/// - larger generated file size (2M)
+/// - possibly unsafe? `as_ptr() as *mut _`
+/// - depends on `c_kzg`; this might not matter much
+fn format_kzg_settings_large(in_path: &Path) -> String {
```
gakonst, line 109: We should check about this w/ the c-kzg folks.

crates/precompile/src/blob/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+    // Verify KZG proof
+    let commitment = as_bytes48(commitment);
+    let z = as_bytes32(&input[32..64]);
+    let y = as_bytes32(&input[64..96]);
+    let proof = as_bytes48(&input[144..192]);
+    if !verify_kzg_proof(commitment, z, y, proof) {
+        return Err(Error::BlobVerifyKzgProofFailed);
+    }
+
+    Ok((GAS_COST, RETURN_VALUE.to_vec()))
```
gakonst, line 53: Is the return value always the same?

crates/precompile/src/lib.rs
gakonst: > No global initialization at all, and we set KzgSettings inside CfgEnv.

I am open to that, as long as we can set a default that's hard-coded to Cancun/Sepolia/.. mainnet?

crates/precompile/src/blob/mod.rs
DaniPopes: Yes
[Geth impl](https://github.com/ethereum/go-ethereum/blob/5b159498bb73eca1c2864c97811ce0375532762c/core/vm/contracts.go#L1128C46-L1128C46)

crates/precompile/build.rs
DaniPopes: It's done at compile-time. There's only 2 methods exposed to init a `KzgSettings`, with the points bytes, or with a file path.

crates/precompile/build.rs
DaniPopes: These are the defaults set at compile time, see `src/blob/kzg_settings/mod.rs` for the global instance

crates/precompile/build.rs
DaniPopes: I think I'll delete the second option, it's way too big to have at runtime

crates/precompile/src/lib.rs
Rjected: > > No global initialization at all, and we set KzgSettings inside CfgEnv.
> 
> I am open to that, as long as we can set a default that's hard-coded to Cancun/Sepolia/.. mainnet?

yeah this seems much nicer than messing around with globals

bins/revme/src/statetest/runner.rs
```diff
@@ -350,58 +363,82 @@ pub fn run(
...
-                    }
-                    //println!("Test:{:?}\n",test_path);
-                    if let Err(err) = execute_test_suit(&test_path, &elapsed, trace) {
-                        endjob.store(true, Ordering::SeqCst);
-                        println!("Test[{index}] named:\n{test_path:?} failed: {err}\n");
-                        return Err(err);
-                    }
+        if cfg!(debug_assertions) {
+            thread = thread.stack_size(4 * 1024 * 1024);
+        }
```
rakita, line 390: Stack size got increased and default release stack size was not enough. Lets see if tests are going to pass if they overflow revert tihs.

crates/precompile/Cargo.toml
```diff
@@ -21,10 +24,15 @@ secp256k1 = { version = "0.27.0", default-features = false, features = [
 ], optional = true }
 sha2 = { version = "0.10.5", default-features = false }
 sha3 = { version = "0.10.7", default-features = false }
+c-kzg = { workspace = true }
```
rakita, line 27: If it is used just here, no need to make version set on workspace.

crates/precompile/Cargo.toml
```diff
@@ -8,6 +8,9 @@ name = "revm-precompile"
 repository = "https://github.com/bluealloy/revm"
 version = "2.0.3"
 
+# Don't need to run build script outside of this repo
+exclude = ["build.rs", "src/blob/kzg_settings/*.txt"]
```
rakita, line 12: I see the idea and I like the script as we get a very clean and compact variable that we can load and use, but this feels better to be moved to `revme` as a helper command.

We can make an issue for this and do it latter if that is okay with you?

crates/precompile/Cargo.toml
```diff
@@ -21,10 +24,15 @@ secp256k1 = { version = "0.27.0", default-features = false, features = [
...
 sha2 = { version = "0.10.5", default-features = false }
 sha3 = { version = "0.10.7", default-features = false }
+c-kzg = { workspace = true }
 
 [dev-dependencies]
 hex = "0.4"
 
+[build-dependencies]
+hex = "0.4"
+# c-kzg = { workspace = true }
```
rakita, line 34: Should be removed?

crates/precompile/src/blob/mod.rs
```diff
@@ -0,0 +1,139 @@
+use crate::{Error, Precompile, PrecompileAddress, PrecompileResult, B160};
```
rakita, line 1: `mod` is not style used inside revm

crates/precompile/src/blob/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+const GAS_COST: u64 = 50_000;
+const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
+
+#[allow(dead_code)]
+const FIELD_ELEMENTS_PER_BLOB: u64 = 4096;
+
+/// The big-endian representation of the modulus.
+#[allow(dead_code)]
+const BLS_MODULUS: &[u8; 32] =
+    &hex!("73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001");
```
rakita, line 21: Should be removed as it is dead code? Move add comment to `RETURN_VALUE`

crates/primitives/src/env.rs
```diff
@@ -12,54 +12,165 @@ pub struct Env {
...
+    }
+
+    /// Returns `true` if the transaction is `Create` or `Create2`.
+    #[inline]
     pub fn is_create(&self) -> bool {
-        matches!(self, Self::Create(_))
+        matches!(
+            self,
+            Self::Create(CreateScheme::Create | CreateScheme::Create2 { .. })
+        )
```
rakita, line 173: Why this change?

crates/primitives/src/env.rs
```diff
@@ -318,6 +446,11 @@ impl Env {
             return Err(InvalidTransaction::AccessListNotSupported);
         }
 
+        // Check that the max fee per blob gas is not set for transactions before CANCUN
+        if !SPEC::enabled(SpecId::CANCUN) && self.tx.max_fee_per_blob_gas.is_some() {
+            return Err(InvalidTransaction::MaxFeePerBlobGasNotSupported);
+        }
+
```
rakita, line 453: Add additional check for `blob_hashes.is_empty()` before CANCUN

crates/primitives/src/result.rs
```diff
@@ -122,25 +122,26 @@ impl Output {
...
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[non_exhaustive]
 pub enum EVMError<DBError> {
     Transaction(InvalidTransaction),
-    /// REVM specific and related to environment.
+    /// `prevrandao` is not set for Merge and above.
     PrevrandaoNotSet,
+    /// `excess_blob_gas` is not set for Cancun and above.
+    ExcessBlobGasNotSet,
```
rakita, line 131: Would be good to have another type of errors `EnvError` so we would have `Transaction`, `Database` and `EnvError` types, and move `PrevrandaoNotSet` and `ExcessBlobGasNotSet` there.

Can be done after this PR, will make a issue.

crates/primitives/src/specification.rs
```diff
@@ -21,10 +23,10 @@ pub enum SpecId {
     LONDON = 12,          // London	                12965000
     ARROW_GLACIER = 13,   // Arrow Glacier	        13773000
     GRAY_GLACIER = 14,    // Gray Glacier	        15050000
-    MERGE = 15,           // Paris/Merge	        TBD (Depends on difficulty)
-    SHANGHAI = 16,
-    CANCUN = 17,
-    LATEST = 18,
+    MERGE = 15,           // Paris/Merge	        15537394 (TTD: 58750000000000000000000)
+    SHANGHAI = 16,        // Shanghai	            17034870 (TS: 1681338455)
+    CANCUN = 17,          // Cancun	                TBD
```
rakita, line 28: nice

crates/precompile/Cargo.toml
DaniPopes: I don't see why not keep the build script? It's ran when the files change so you don't have to remember/need a CI job to remind you etc.

crates/primitives/src/env.rs
DaniPopes: I prefer exhaustive patterns so when something changes you are reminded to also update everywhere it's used

crates/primitives/src/env.rs
DaniPopes: I can also remove `max_fee_per_blob_gas` since I don't think we'll be using it in REVM?

crates/primitives/src/result.rs
DaniPopes: That's fine

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,56 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
     B160(hasher.finalize().as_slice()[12..].try_into().unwrap())
 }
 
+/// Calculates the `excess_blob_gas` from the parent header's `blob_gas_used` and `excess_blob_gas`.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers).
+#[inline]
+pub fn calc_excess_blob_gas(parent_excess_blob_gas: u64, parent_blob_gas_used: u64) -> u64 {
+    (parent_excess_blob_gas + parent_blob_gas_used).saturating_sub(TARGET_BLOB_GAS_PER_BLOCK)
```
rakita, line 41: TODO for me, why `saturating_sub` this can go in negative. If we return zero does that mean that `base_blob_fee` is going down? Check `def calc_excess_blob_gas(parent: Header) -> int:` later in https://eips.ethereum.org/EIPS/eip-4844

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,56 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
...
+}
+
+/// Approximates `factor * e ** (numerator / denominator)` using Taylor expansion.
+///
+/// This is used to calculate the blob price.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers).
+#[inline]
+pub fn fake_exponential(factor: u64, numerator: u64, denominator: u64) -> u64 {
+    assert!(denominator > 0, "attempt to divide by zero");
```
rakita, line 63: would panic here if dominator is == 0

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,56 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
...
+    let mut numerator_accum = factor * denominator;
+    while numerator_accum > 0 {
+        output += numerator_accum;
+
+        // SAFETY: asserted > 0 above.
+        // This is needed to eliminate the compiler's divide by 0 check.
+        numerator_accum = unsafe {
+            (numerator_accum * numerator)
+                .checked_div(denominator * i)
+                .unwrap_unchecked()
```
rakita, line 79: Is it better to remove unsafe and use `unwrap_or_default()`?

crates/primitives/src/utilities.rs
```diff
@@ -59,3 +111,100 @@ pub mod serde_hex_bytes {
...
+                2 * GAS_PER_BLOB + 1,
+            ),
+            // The excess blob gas should decrease by however much the target was
+            // under-shot, capped at zero.
+            (
+                TARGET_BLOB_GAS_PER_BLOCK,
+                TARGET_BLOB_GAS_PER_BLOCK / GAS_PER_BLOB,
+                TARGET_BLOB_GAS_PER_BLOCK,
+            ),
+            // TODO ?
```
rakita, line 153: todo

crates/revm/Cargo.toml
```diff
@@ -21,10 +21,7 @@ serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 serde_json = { version = "1.0", features = ["preserve_order"], optional = true }
 
 # ethersdb
-tokio = { version = "1.32", features = [
-    "rt-multi-thread",
-    "macros",
-], optional = true }
+tokio = { version = "1.32", features = ["rt-multi-thread", "macros"], optional = true }
```
rakita, line 24: I think we have two different formatter for toml. I am using https://taplo.tamasfe.dev/ but not sure if I was the one formatting this

crates/primitives/src/utilities.rs
DaniPopes: it can only divide by zero if the `denominator` arg is 0 (`i` starts at 1), which is why I added the assertion above

crates/primitives/src/utilities.rs
DaniPopes: Should we return 0/None if `denominator == 0`? It would be divide-by-zero

crates/revm/src/evm_impl.rs
```diff
@@ -139,21 +129,32 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
...
+
+        // Load access list for berlin hardfork.
+        //
+        // Loading of accounts/storages is needed to make them hot.
+        for (address, slots) in self.data.env.tx.access_list.iter() {
+            self.data
+                .journaled_state
+                .initial_account_load(*address, slots, self.data.db)
+                .map_err(EVMError::Database)?;
+        }
```
rakita, line 141: Can you revert this, having it inside function feels better

DaniPopes: The problem here was that the borrow checker was complaining that `load_access_list` borrows `self` mutably

crates/revm/Cargo.toml
DaniPopes: I am too, I have it configured to 100 column width, so that might be it

crates/primitives/src/utilities.rs
DaniPopes: some test cases taken from Geth are outdated, I'll look out for updates in Geth PRs

crates/primitives/src/utilities.rs
Rjected: The `calc_excess_blob_gas` method defined there:
```python
def calc_excess_blob_gas(parent: Header) -> int:
    if parent.excess_blob_gas + parent.blob_gas_used < TARGET_BLOB_GAS_PER_BLOCK:
        return 0
    else:
        return parent.excess_blob_gas + parent.blob_gas_used - TARGET_BLOB_GAS_PER_BLOCK
```

Is the same thing as `saturating_sub`, because the first condition ensures that the function saturates at numeric bounds instead of returning any negative values.

rakita: I see, thank you!

crates/precompile/Cargo.toml
rakita: `.bin` file would stay, would rename it to `.hex` or even `.txt` as it is still readable, and would be used only to store mainnet setup in compact form, this is not something that is expected to be changed at all in future so having build script for it is a overkill. This procedure should be just a helper function inside `revm` and it would be exposed in `revme` if needed to be called in cli.

crates/primitives/src/env.rs
rakita: It seems not needed, can you revert this?

crates/primitives/src/utilities.rs
rakita: with `unwrap_or_default` would remove need of unsafe, or maybe we can just use `.expect("Denominator is checked above")`

crates/primitives/src/utilities.rs
rakita: I would just `panic!` as it is undefined what to expect as result, maybe it is `u64::MAX` maybe `0`

crates/revm/Cargo.toml
rakita: Can you use default value and revert it back to be consistent? Maybe we should add local vscode setup, and change it there.

crates/precompile/src/blob/kzg_settings/mod.rs
rakita: With `EnvCfg` path of setting kzg setting, this is unnecessary.

crates/primitives/src/utilities.rs
mattsse: a panic is fine here imo, for blob fee, the denominator is a constant != 0

crates/precompile/Cargo.toml
rakita: issue: https://github.com/bluealloy/revm/issues/686

rakita (changes requested):
Two things left:
* Enable setting kzg setup/points from `EnvCfg`. Remove global/static setting of setup. And have default setup that is loaded from `.bin` files (Would rename to `.hex` or `.txt`)
* Make eip4844 precompile only available if `std` is enabled.

Additional things that can be done later: https://github.com/bluealloy/revm/issues/685 https://github.com/bluealloy/revm/issues/686

crates/precompile/Cargo.toml
DaniPopes: Alright I guess, but the bin files are not human readable: they're the raw `hex::decode`d bytes of the G1 and G2 points from the trusted setup file

mattsse:
> Enable setting kzg setup/points from EnvCfg. Remove global/static setting of setup. And have default setup that is loaded from .bin files (Would rename to .hex or .txt)

isn't this the same as a lazy static?

crates/precompile/src/blob/kzg_settings.rs
```diff
@@ -0,0 +1,34 @@
+use alloc::boxed::Box;
+use c_kzg::KzgSettings;
+use once_cell::race::OnceBox;
+
+#[rustfmt::skip]
+mod generated;
+
+/// Note: the type of this is an implementation detail, so we don't expose it in the public API.
+static GLOBAL_KZG_SETTINGS: OnceBox<KzgSettings> = OnceBox::new();
```
mattsse, line 9: I'd prefer if this would be `Arc<KzgSettings>` otherwise it would be much harder to install custom settings because KzgSettings is not Clone

crates/precompile/src/lib.rs
```diff
@@ -191,6 +195,22 @@ impl Precompiles {
         })
     }
 
+    pub fn cancun() -> &'static Self {
```
mattsse, line 198: this should mention that this uses the globally installed KzgSettings

mattsse (changes requested):
some comments about global KzgSettings

crates/primitives/src/utilities.rs
DaniPopes: The problem is that the compiler does not optimize the check out, when we know for a fact that this cannot overflow.

DaniPopes:
> * Make eip4844 precompile only available if `std` is enabled.

It works on `no_std` AFAICT. If you mean for the need to compile C, then it can be an optional feature unrelated to `std`

crates/primitives/src/utilities.rs
DaniPopes: Updated: 1aea2a4b2bae4736ad76c333862c916ae8bc97f1

crates/primitives/src/utilities.rs
rakita: You are thinking too deeply about optimization. As this function has a specific purpose that we know we can make an assumption that this is not going to be a hot loop that we expect to be called a few billion times in the loop, so it is okay to have it be slightly inefficient for more readable code.

rakita:
> > * Make eip4844 precompile only available if `std` is enabled.
> 
> It works on `no_std` AFAICT. If you mean for the need to compile C, then it can be an optional feature unrelated to `std`

Had outdated info, and missed the branch: https://github.com/ethereum/c-kzg-4844/pull/347 This is a nice addon and there is no need to flag the lib behind `std`.

crates/precompile/Cargo.toml
gakonst: How do we want to move forward with this? Seeing we still have the build.rs file

crates/precompile/build.rs
```diff
@@ -0,0 +1,129 @@
...
+    const BYTES: &[u8] = include_bytes!("./g2_points.bin");
+    assert!(BYTES.len() == core::mem::size_of::<G2Points>());
+    unsafe {{ &*BYTES.as_ptr().cast::<G2Points>() }}
+}};
+"#
+    )
+}
+
+/// [`Vec::into_flattened`].
+#[inline]
```
gakonst, line 97: dumb Q is this because of the const?

crates/precompile/src/blob/kzg_settings.rs
gakonst: Doesn't the current OnceBox allow that by overriding it once?

crates/precompile/src/blob/kzg_settings.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Returns a reference to the global [`KzgSettings`] instance, initializing it with the default
+/// value if it was not previously set.
+pub fn get_global_or_default() -> &'static KzgSettings {
+    GLOBAL_KZG_SETTINGS.get_or_init(|| Box::new(default()))
```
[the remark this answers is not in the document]
gakonst: > which should be fine, but it would mean that it is impossible to set the global kzg after a transact call.

Hmm does this mean that you cannot test w/ various parameters in a loop? I guess it might be a fine limitation agree about doc

crates/primitives/src/utilities.rs
gakonst: defer to Dragan here - agree we don't need to be "too smart" about it given it's only called once per block?

cc @mattsse,

crates/primitives/src/env.rs
```diff
@@ -210,57 +325,73 @@ impl Default for CfgEnv {
...
     #[inline]
     pub fn validate_block_env<SPEC: Spec, T>(&self) -> Result<(), EVMError<T>> {
-        // Prevrandao is required for merge
+        // `prevrandao` is required for the merge
         if SPEC::enabled(SpecId::MERGE) && self.block.prevrandao.is_none() {
             return Err(EVMError::PrevrandaoNotSet);
         }
+        // `excess_blob_gas` is required for Cancun
+        if SPEC::enabled(SpecId::CANCUN) && self.block.excess_blob_gas.is_none() {
+            return Err(EVMError::ExcessBlobGasNotSet);
```
gakonst, line 393: 👍

rakita:
> > Enable setting kzg setup/points from EnvCfg. Remove global/static setting of setup. And have default setup that is loaded from .bin files (Would rename to .hex or .txt)
> 
> isn't this the same as a lazy static?

it would remove the need for static variables that we have, a singleton pattern is usually not good. Idea mentioned here: https://github.com/bluealloy/revm/pull/668#discussion_r1311150805

crates/precompile/build.rs
DaniPopes: `Vec::into_flattened` is unstable

mattsse:
so all precompiles need to be configured for every CfgEnv?

that would be super inconvenient and would most likely result in lazy statics anyway,

but the CfgEnv could take an Option of Precompiles if None use the existing defaults?

Rjected:
> so all precompiles need to be configured for every CfgEnv?
> 
> that would be super inconvenient and would most likely result in lazy statics anyway,
> 
> but the CfgEnv could take an Option of Precompiles if None use the existing defaults?

wouldn't it be something like:
```rust
enum StandardPrecompile {
  Fun(fn(&[u8], u64) -> PrecompileResult),
  ConfiguredFun(fn(&CfgEnv, &[u8], u64) -> PrecompileResult)
}
```

and then in `call_precompile`:
```rust
let out = match precompile {
    Precompile::Standard(standard) => match standard {
        StandardPrecompile::Fun(fun) => fun(input_data, gas.limit()),
        StandardPrecompile::ConfiguredFun(fun) => fun(self.env().cfg, input_data, gas.limit())
    },
    Precompile::Custom(fun) => fun(input_data, gas.limit()),
};
```

The blobhash precompile would be:
```rust

pub const POINT_EVALUATION: PrecompileAddress =
    PrecompileAddress(ADDRESS, Precompile::Standard(StandardPrecompile::Configured(run)));

fn run(cfg: &CfgEnv, input: &[u8], gas_limit: u64) -> PrecompileResult {
    // ... setup

    // Verify KZG proof
    let commitment = as_bytes48(commitment);
    let z = as_bytes32(&input[32..64]);
    let y = as_bytes32(&input[64..96]);
    let proof = as_bytes48(&input[144..192]);
    // ommitting match for simplicity
    if !KzgProof::verify_kzg_proof(commitment, z, y, proof, cfg.kzg_settings) {
        return Err(Error::BlobVerifyKzgProofFailed);
    }

    Ok((GAS_COST, RETURN_VALUE.to_vec()))
}
```

The current solution seems to work, if we wanted to go static-less for the kzg settings then I think the above would also work. IMO this should be for a follow up

rakita:
Unresolved things:
* https://github.com/bluealloy/revm/pull/668/files#r1313344511
* https://github.com/bluealloy/revm/pull/668#discussion_r1311150805 and another explanation here: https://github.com/bluealloy/revm/pull/668#issuecomment-1714478733

README.md
```diff
@@ -11,17 +11,16 @@ Here is a list of things that I would like to use as guide in this project:
...
     * revm-primitives -> Primitive data types.
     * revm-interpreter -> Execution loop with instructions
     * revm-precompile -> EVM precompiles
 * bins:
-    * revme: cli binary, used for running state test json
-    * revm-test: test binaries with contracts, used mostly to check performance
+    * revme: cli binary, used for running state test jsons
 
-Last checked revm requires rust v1.65 or higher for `core::error::Error`
+Project tend to use the newest rust, so if there is a build error try first to run `rustup update`.
```
mattsse, line 23:
```suggestion
This project tends to use the newest rust version, so if you're encountering a build error try running `rustup update` first.
```

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -0,0 +1,108 @@
...
+const GAS_COST: u64 = 50_000;
+const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
+
+/// `U256(FIELD_ELEMENTS_PER_BLOB).to_be_bytes() ++ BLS_MODULUS.to_bytes32()`
+const RETURN_VALUE: &[u8; 64] = &hex!(
+    "0000000000000000000000000000000000000000000000000000000000001000"
+    "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001"
+);
+
+fn run(input: &[u8], gas_limit: u64, env: &Env) -> PrecompileResult {
```
mattsse, line 18: one line doc that this is the point evaluation precompile

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -0,0 +1,108 @@
...
+    if gas_limit < GAS_COST {
+        return Err(Error::OutOfGas);
+    }
+    if input.len() != 192 {
+        return Err(Error::BlobInvalidInputLength);
+    }
+
+    // Verify commitment matches versioned_hash
+    let commitment = &input[96..144];
+    let versioned_hash = &input[0..32];
```
mattsse, line 28: can we swap the order of these two vars, so we start slicing from the beginning?

mattsse, line 28:
```suggestion
    let versioned_hash = &input[..32];
```

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -0,0 +1,108 @@
...
+const RETURN_VALUE: &[u8; 64] = &hex!(
+    "0000000000000000000000000000000000000000000000000000000000001000"
+    "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001"
+);
+
+fn run(input: &[u8], gas_limit: u64, env: &Env) -> PrecompileResult {
+    if gas_limit < GAS_COST {
+        return Err(Error::OutOfGas);
+    }
+    if input.len() != 192 {
```
mattsse, line 22: can we add a brief note here about the layout of the input slice? so it's obvious where these consts are coming from?

crates/precompile/src/lib.rs
```diff
@@ -191,6 +192,30 @@ impl Precompiles {
...
```
mattsse, line 195: This needs docs that this does not include point eval precompile if no-std

crates/primitives/src/env.rs
```diff
@@ -76,12 +186,17 @@ pub enum CreateScheme {
...
 }
 
+/// EVM configuration.
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[non_exhaustive]
 pub struct CfgEnv {
     pub chain_id: u64,
     pub spec_id: SpecId,
+    ///// KZG Settings
```
mattsse, line 196:
```suggestion
    /// KZG Settings for point evaluation precompile. By default, this is loaded from the ethereum mainnet trusted setup.
```

crates/primitives/src/kzg/env_settings.rs
```diff
@@ -0,0 +1,30 @@
...
+use alloc::{boxed::Box, sync::Arc};
+use once_cell::race::OnceBox;
+
+/// KZG Settings.
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
+pub enum EnvKzgSettings {
+    #[default]
+    Default,
+    Custom(Arc<c_kzg::KzgSettings>),
+}
```
mattsse, line 14: this needs proper docs

* used for point eval precompile
* Default: loaded from eth mainnet trusted setup
* Custom: use the given kzgsettings in the precompile

crates/primitives/src/kzg/env_settings.rs
```diff
@@ -0,0 +1,30 @@
...
+
+impl EnvKzgSettings {
+    pub fn get(&self) -> &KzgSettings {
```
mattsse, line 17: one line doc, that repeats the info above

crates/primitives/src/kzg/generated.rs
```diff
@@ -0,0 +1,24 @@
+// @generated by build.rs from "src/kzg/trusted_setup.txt", do not modify manually.
+
+pub use c_kzg::{BYTES_PER_G1_POINT, BYTES_PER_G2_POINT};
+
+// Ensure that the build script constants are synced with the C bindings ones.
+const _: [(); BYTES_PER_G1_POINT] = [(); 48];
+const _: [(); BYTES_PER_G2_POINT] = [(); 96];
+
+pub const NUM_G1_POINTS: usize = 4096;
+pub const NUM_G2_POINTS: usize = 65;
```
mattsse, line 10: this is unfortunately not pub yet in kzg, so we need to define it, I'll open a PR to make the const public

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,53 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
...
+    let numerator = numerator as u128;
+    let denominator = denominator as u128;
+
+    let mut i = 1;
+    let mut output = 0;
+    let mut numerator_accum = factor * denominator;
+    while numerator_accum > 0 {
+        output += numerator_accum;
+
+        // Check for to see if denominator is zero is done at the beginning of the call.
```
mattsse, line 76: can we rephrase that?

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,53 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
...
+        excess_blob_gas,
+        BLOB_GASPRICE_UPDATE_FRACTION,
+    )
+}
+
+/// Approximates `factor * e ** (numerator / denominator)` using Taylor expansion.
+///
+/// This is used to calculate the blob price.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers).
```
mattsse, line 60: needs `# Panics` section

mattsse (changes requested):
only doc nits

crates/primitives/src/kzg/generated.rs
```diff
@@ -0,0 +1,24 @@
+// @generated by build.rs from "src/kzg/trusted_setup.txt", do not modify manually.
+
+pub use c_kzg::{BYTES_PER_G1_POINT, BYTES_PER_G2_POINT};
+
+// Ensure that the build script constants are synced with the C bindings ones.
+const _: [(); BYTES_PER_G1_POINT] = [(); 48];
+const _: [(); BYTES_PER_G2_POINT] = [(); 96];
+
+pub const NUM_G1_POINTS: usize = 4096;
```
mattsse, line 9: can use `FIELD_ELEMENTS_PER_BLOB` const

crates/primitives/src/utilities.rs
```diff
@@ -31,6 +33,53 @@ pub fn create2_address(caller: B160, code_hash: B256, salt: U256) -> B160 {
...
+#[inline]
+pub fn fake_exponential(factor: u64, numerator: u64, denominator: u64) -> u64 {
+    if denominator == 0 {
+        panic!("attempt to divide by zero");
+    }
```
onbjerg, line 65: nit: can use `assert_ne!` instead

crates/primitives/src/precompile.rs
```diff
@@ -1,10 +1,13 @@
...
 use alloc::vec::Vec;
 
 /// A precompile operation result.
+///
+/// Returns either `Ok((gas_used, return_bytes))` or `Err(error)`.
 pub type PrecompileResult = Result<(u64, Vec<u8>), PrecompileError>;
 
 pub type StandardPrecompileFn = fn(&[u8], u64) -> PrecompileResult;
-pub type CustomPrecompileFn = fn(&[u8], u64) -> PrecompileResult;
+pub type EnvPrecompileFn = fn(&[u8], u64, env: &Env) -> PrecompileResult;
```
onbjerg, line 10: any reason for this rename? i assume it would be a breaking change, i'm not sure if this pr is supposed to be breaking in general?

onbjerg (commented):
small qs/nits, overall looks good to me i think

crates/primitives/src/precompile.rs
rakita: Mostly as it was not used, it was a idea to add "Custom" precompiles and allows an external user to add their precompiles, but usage was lacking, should be fine to break it as the signature is the same as a `StandardPrecompileFn`.

crates/precompile/src/kzg_point_evaluation.rs
rakita: Added format spec at the top of the function,

crates/primitives/src/kzg/generated.rs
rakita: Ref to this issue: https://github.com/bluealloy/revm/issues/686 we can make change there

crates/primitives/src/kzg/generated.rs
rakita: Can probably move this later in this task: https://github.com/bluealloy/revm/issues/686
But those fields are reintroduced to have preserve strict size at the time when points are moved to binary format (in build script).

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -0,0 +1,119 @@
...
+    "73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001"
+);
+
+/// Run kzg point evaluation precompile.
+///
+/// The Env has the KZGSettings that is needed for evaluation.
+///
+/// The input is encoded as follows:
+/// | versioned_hash |  z  |  y  | commitment | proof |
+/// |     32         | 32  | 32  |     48     |   48  |
```
mattsse, line 24: nice!

mattsse (approved):
lgtm

DaniPopes (commented):
lgtm

rakita:
Just to mention here, we have disabled some of `ethereum/tests` related to eip4844 that have `expectException` field set. As the state root is different from test expected value that is generated. 

Those tests are:
"emptyBlobhashList.json" // '>=Cancun': TR_EMPTYBLOB
"wrongBlobhashVersion.json"  // '>=Cancun': TR_BLOBVERSION_INVALID
"createBlobhashTx.json" // '>=Cancun': TR_BLOBCREATE
"blobhashListBounds7.json" // ">=Cancun": "TR_BLOBLIST_OVERSIZE"
 
`blobhashListBounds7` and `blobhashListBounds6` are very similar but I am not sure why `ListBound7` is marked as  `TR_BLOBLIST_OVERSIZE` and it has different behaviour from `ListBounds6`. 

Will treat this as tests not being aligned on what evm should do, and will check later.

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -45,7 +45,7 @@ fn run(input: &[u8], gas_limit: u64, env: &Env) -> PrecompileResult {
     let z = as_bytes32(&input[32..64]);
     let y = as_bytes32(&input[64..96]);
     let proof = as_bytes48(&input[144..192]);
-    if !verify_kzg_proof(commitment, z, y, proof, env.cfg.kzg_settings.get()) {
+    if KzgProof::verify_kzg_proof(commitment, z, y, proof, env.cfg.kzg_settings.get()).is_err() {
```
DaniPopes, line 48: What about when it returns false? That's why I had a separate function for it

rakita: Will revert back

rakita: Good and fast catch. commit: https://github.com/bluealloy/revm/pull/668/commits/eb8b2233713eda89949bb7ebbbfd2dae003d57ec

rakita (approved):

PR #669, opened 2023-08-30
Removed the last dependencies breaking no-std build.

lvella:
Removed to_binary, which was used only in a debug display.

Replaced once_cell with lazy_static. The no-std build of once_cell did not support the Sync version of OnceCell, that was needed when creating static variables. On the other hand, lazy_static is the complete solution for lazyly initialized (Sync) static variables, and is no-std.

With this changes, revm builds on a no-std platform.

### DIFF
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -13,6 +13,7 @@ revm-primitives = { path = "../primitives", version = "1.1.2", default-features 
 bn = { package = "substrate-bn", version = "0.6", default-features = false }
 k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
 num = { version = "0.4.0", default-features = false, features = ["alloc"] }
+once_cell = { version = "1.17", default-features = false }
 ripemd = { version = "0.1", default-features = false }
 secp256k1 = { version = "0.27.0", default-features = false, features = [
     "alloc",
@@ -20,13 +21,14 @@ secp256k1 = { version = "0.27.0", default-features = false, features = [
 ], optional = true }
 sha2 = { version = "0.10.5", default-features = false }
 sha3 = { version = "0.10.7", default-features = false }
-lazy_static = "1.4"
 
 [dev-dependencies]
 hex = "0.4"
 
 [features]
-default = ["secp256k1"]
+default = ["std", "secp256k1"]
+std = ["once_cell/std"]
+critical-section = ["once_cell/critical-section"]
 # secp256k1 is used as faster alternative to k256 lib. And in most cases should be default.
 # Only problem that it has, it fails to build for wasm target on windows and mac as it is c lib.
 # If you dont require wasm on win/mac, i would recommend its usage.
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -13,7 +13,7 @@ revm-primitives = { path = "../primitives", version = "1.1.2", default-features 
 bn = { package = "substrate-bn", version = "0.6", default-features = false }
 k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
 num = { version = "0.4.0", default-features = false, features = ["alloc"] }
-once_cell = { version = "1.17", default-features = false }
+once_cell = { version = "1.17", default-features = false, features = ["alloc"] }
 ripemd = { version = "0.1", default-features = false }
 secp256k1 = { version = "0.27.0", default-features = false, features = [
     "alloc",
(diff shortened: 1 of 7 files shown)


crates/precompile/Cargo.toml
```diff
@@ -9,15 +9,18 @@ repository = "https://github.com/bluealloy/revm"
...
-once_cell = "1.17"
 ripemd = { version = "0.1", default-features = false }
-secp256k1 = { version = "0.27.0", default-features = false, features = ["alloc", "recovery"], optional = true }
+secp256k1 = { version = "0.27.0", default-features = false, features = [
+    "alloc",
+    "recovery",
+], optional = true }
 sha2 = { version = "0.10.5", default-features = false }
 sha3 = { version = "0.10.7", default-features = false }
+lazy_static = "1.4"
```
DaniPopes, line 23: `lazy_static` is unmaintained. Please just use `once_cell` with `default-features = false`.

DaniPopes (changes requested):

crates/precompile/Cargo.toml
lvella: That was my first try, doesn't work.

Without `std`, only `once_cell::unsync::OnceCell` is available, and it can't be used to initialize a static variable, because it is not `Sync`. It seems `lazy_static` manages to do it by using a spin-lock, what `once_cell` doesn't have.

Anyway, the other crate I removed, `to-binary`, is also unmaintained and is older, so I don't see the issue of using `lazy_static`, specially since neither `once_cell` or the unstable alternative `std::sync::LazyLock` is a suitable `Sync` no-std alternative.

DaniPopes: See the [`once_cell` FAQ](https://docs.rs/once_cell/latest/once_cell/#faq):

- Does this crate support `no_std`
- Should I use std::cell::OnceCell, once_cell, or lazy_static?

I say we enable the `critical-section` feature. cc @rakita

lvella: Just found `once_cell::race`. I'll use it instead of `lazy_static`.

rakita: OnceCell was stabilised inside rust, I would check if it works
https://doc.rust-lang.org/nightly/core/cell/struct.OnceCell.html#:~:text=Struct%20core%3A%3Acell%3A%3AOnceCell&text=A%20cell%20which%20can%20be,borrow%20checks%20(unlike%20RefCell%20).

to-binary is not relevant at all as it had very simpler usecase,

crates/primitives/src/bytecode.rs
```diff
@@ -14,7 +13,7 @@ pub struct JumpMap(pub Arc<BitVec<u8>>);
 impl Debug for JumpMap {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.debug_struct("JumpMap")
-            .field("map", &BinaryString::from(self.0.as_raw_slice()))
+            .field("map", &self.0.as_raw_slice())
```
rakita, line 16: Would probably use `hex::encode` here. Forgot the reason why I didn't use that in first place

crates/precompile/Cargo.toml
lvella: It doesn't for static initialization, it is `!Sync`.

lvella: > `core::sync::OnceLock` is the equivalent of `once_cell::sync::OnceCell`. Let's just use this and resolve this discussion :)

There is no `core::sync::OnceLock`. There is either `std::sync::OnceLock` or `core::cell::OnceCell`, the former requires `std`, the later is `!Sync`, thus it can't be a static variable.

Besides `lazy_static`, the valid alternatives are `once_cell::race::OnceCell` or `once_cell`'s `critical-section` feature. Given the revm usage, I think `once_cell::race::OnceCell` is a fine solution.

rakita: > Sorry, it seems `once_cell::race` doesn't provide a generic `OnceCell`, instead it is specialized for some types that can be handled by atomic instructions. I'll change this PR to use critical-section.

Just came this to write this. We can use `once_cell::race::OnlyBox`, it does not require std and it fits nicely with the code

DaniPopes (approved):

crates/primitives/src/bytecode.rs
lvella: done

crates/precompile/Cargo.toml
lvella: Well, I just pushed a version using `critical-section`, had to add a new feature and all.

rakita: > Well, I just pushed a version using `critical-section`, had to add a new feature and all. Should I change?

It would be better to change it, fewer features to maintain and it would work out of the box ( pun intended :))

lvella: done

rakita:
@lvella what is the best build command to add to CL for `no_std` check?

PR #670, opened 2023-08-31
feat(state): bundle size hint

rkrasiuk:
## Description

Add support for estimating the size of the bundle state.


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -130,9 +134,17 @@ impl BundleState {
...
             reverts,
+            reverts_size,
         }
     }
 
+    /// Returns the approximate size of changes in the bundle state.
+    /// The estimation is not precise, because the information about the number of
+    /// destroyed entries that need to be removed is not accessible to the bundle state.
+    pub fn size_hint(&self) -> usize {
+        self.state.len() + self.contracts.len() + self.reverts_size
```
rakita, line 145: state can have both the account and the storage. Maybe we should take that into consideration. But it gets harder to update

rakita:
To be honest, I would like to introduce a full estimate of the memory size that `BundleState` takes. And do the similar things for `CacheState`.

This is a very rough take in that direction, but I see that it can be beneficial so I am okay to merge it.

crates/revm/src/db/states/bundle_state.rs
rakita: Let us be okay with not ideal situation, as adding storage counting of state can get complex and can result in a fault code

PR #673, opened 2023-08-31
Shared memory

lorenzofero:
I tried to give a shot to #445  just for fun and as a challenge with this PR.
This introduces a new struct called `SharedMemory` which is indeed shared between calls. 

### About mplementation
- its API matches almost completely the 'old' interpreter `Memory` implementation
- memory space is allocated using this estimate https://2π.com/22/eth-max-mem/,
- it has a pointer `current_slice` which is internally used to refer to the portion of data reserved for the current context. This requires two `unsafe` methods `get_current_slice` and `get_current_slice_mut` which deferences the raw pointer
- `current_slice` pointer is updated when entering a new context (which is when a new `Interpreter` instance is created) using the `new_context_memory` method and when exiting it, which happens when the `return_value` method of `Interpreter` is called
- in order to make `SharedMemory` work properly, it is wrapped in a `Rc` for dealing with recursive calls and with an `RefCell` to allow mutability

### Performance
I'd like a feedback on this because:
- probably I'm doing something not super optimized: some results are very good, other are awful
- benches code is not ideal for a shared memory: for example, if a tx uses a lot of gas but _no_ memory allocation, performance is slower since I'm allocating for no reason. If I don't go wrong, we always use the same bytecode for benches, which may result is not entirely accurate performance indications on real usage of this shared memory
Anyway, this is the result on running `cargo bench --all` on `main` and then on my branch:
```text
analysis/transact/raw   time:   [9.1437 µs 9.1751 µs 9.2336 µs]
                        change: [+26.333% +27.735% +28.956%] (p = 0.00 < 0.05)
                        Performance has regressed.
analysis/transact/checked
                        time:   [8.9833 µs 9.0089 µs 9.0323 µs]
                        change: [+24.117% +24.585% +24.995%] (p = 0.00 < 0.05)
                        Performance has regressed.
analysis/transact/analysed
                        time:   [6.1140 µs 6.1477 µs 6.1955 µs]
                        change: [+14.154% +14.958% +15.765%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

snailtracer/transact/analysed
                        time:   [5.8516 µs 5.9056 µs 6.0202 µs]
                        change: [-99.991% -99.991% -99.991%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
snailtracer/eval        time:   [5.7537 ms 5.7599 ms 5.7693 ms]
                        change: [-89.993% -89.975% -89.957%] (p = 0.00 < 0.05)
                        Performance has improved.

transfer/transact/analysed
                        time:   [4.7373 µs 4.7475 µs 4.7573 µs]
                        change: [+498.50% +500.64% +502.62%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```
I also tried to run cachegrind as explained here https://github.com/bluealloy/revm/pull/582 and it is indeed slower, of about a 10% or less. Gas limit required is between 2^15 and 2^16. Maybe also here I'm allocating memory which is not used very much. It depends on the bytecode of the snailtracer.

Thanks in advance for any feedback!


crates/interpreter/src/instructions/macros.rs
```diff
@@ -49,32 +49,34 @@ macro_rules! gas_or_fail {
     };
 }
 
-macro_rules! memory_resize {
-    ($interp:expr, $offset:expr, $len:expr) => {
+macro_rules! shared_memory_resize {
+    ($interp:expr, $offset:expr, $len:expr) => {{
+        let len: usize = $len;
+        let offset: usize = $offset;
```
lorenzofero, line 55: Here I'm creating another inner scope to allow a `borrow_mut`. I'd be better to use the `mem` variable also elsewhere in the opcode which calls this macro but I get some compilation problems.. I'd like to get the same result of the `pop!` macro but I can't... My Rust chops are not very good yet. How can I achieve this?
I wonder if this has a particular performance penalty

gakonst:
cc @DaniPopes who may also have thoughts

rakita:
In general very supportive of this, tbh I would have expected the same or better performance. Will see to play with it a little bit to check it out.

Would like to see if we can remove Rc<RefCell<>>, maybe the overhead that we see is related to the dynamic borrow checks inside refcell. Snailtracer should be a good benchmark to check as init of evm is small in comparison of the work that interpreter is doing.

lorenzofero:
> In general very supportive of this, tbh I would have expected the same or better performance. Will see to play with it a little bit to check it out.

I'm very happy you like it! After this PR I can try to do a `shared_stack` as well in the same fashion, by allocating 32MB.
> Would like to see if we can remove Rc<RefCell<>>, maybe the overhead that we see is related to the dynamic borrow checks inside refcell. Snailtracer should be a good benchmark to check as init of evm is small in comparison of the work that interpreter is doing.

lorenzofero:
> #582 was merged, can you rebase this PR? @lorenzofero

I can't really perform a rebase because I have conflicts in multiple commits and it's really cumbersome to get out. I'm working on a merge with conflicts resolution :+1:

lorenzofero:
Hey @rakita @DaniPopes @gakonst, now tests are passing if you want to check it out again. I managed to keep all Dani's updates of memory related functions inside `shared_memory.rs`. Let me know what you think!

DaniPopes:
You can run `cargo criterion` or the Cachegrind test as explained in #582's description

crates/interpreter/src/instructions/macros.rs
```diff
@@ -50,26 +50,27 @@ macro_rules! gas_or_fail {
...
 }
 
-macro_rules! memory_resize {
+macro_rules! shared_memory_resize {
     ($interp:expr, $offset:expr, $len:expr) => {
+        let len: usize = $len;
+        let offset: usize = $offset;
         if let Some(new_size) =
-            crate::interpreter::memory::next_multiple_of_32($offset.saturating_add($len))
+            crate::interpreter::shared_memory::next_multiple_of_32(offset.saturating_add(len))
```
DaniPopes, line 58: can revert, I assume this is a merge fallout

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+            .finish()
+    }
+}
+
+impl SharedMemory {
+    /// Allocate memory to be shared between calls.
+    /// Memory size is estimated using https://2π.com/22/eth-max-mem
+    /// which depends on transaction [gas_limit].
+    /// Maximum allocation size is 2^32 - 1 bytes;
+    pub fn new(gas_limit: u64) -> Self {
```
DaniPopes, line 45: This can call `Self::new_with_memory_limit(u32::MAX as usize)` (also `pow` doesn't get optimized well I think)

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+
+    /// Allocate memory to be shared between calls.
+    /// Memory size is estimated using https://2π.com/22/eth-max-mem
+    /// which depends on transaction [gas_limit].
+    /// Uses [memory_limit] as maximum allocation size
+    pub fn new_with_memory_limit(gas_limit: u64, memory_limit: u64) -> Self {
+        let size = min(
+            SharedMemory::calculate_upper_bound(gas_limit) as usize,
```
DaniPopes, line 68:
```suggestion
            Self::calculate_upper_bound(gas_limit) as usize,
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct SharedMemory {
+    /// Shared buffer
+    data: Vec<u8>,
+    /// Memory checkpoints for each depth
+    checkpoints: Vec<usize>,
+    /// Raw pointer used for the portion of memory used
+    /// by the current context
+    current_slice: *mut [u8],
+    /// How much memory has been used in the current context
+    current_len: usize,
```
DaniPopes, line 24: `*mut [u8]` (just like `&mut [u8]`) holds (ptr, len), so either use `*mut u8, usize` or just `*mut [u8]`

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[inline(always)]
+    fn update_limit(&mut self) {
+        self.limit =
+            (self.data.len() - *self.checkpoints.last().unwrap_or(&0) - self.current_len) as u64;
+    }
+}
+
+/// Rounds up `x` to the closest multiple of 32. If `x % 32 == 0` then `x` is returned.
+#[inline]
+pub(crate) fn next_multiple_of_32(x: usize) -> Option<usize> {
```
DaniPopes, line 257: This function is in primitives utils now

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+    let mut y = (x + n / x) / 2;
+    while y < x {
+        x = y;
+        y = (x + n / x) / 2;
+    }
+    x
+}
+
+#[cfg(test)]
+mod tests {
```
DaniPopes, line 277: These are in primitives utils alongside the function

crates/revm/src/evm_impl.rs
```diff
@@ -125,6 +125,8 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
 
     fn transact_preverified(&mut self) -> EVMResult<DB::Error> {
         let env = &self.data.env;
+        #[cfg(feature = "memory_limit")]
```
DaniPopes, line 128: I don't think we need this feature anymore if we're going to have a limit always? cc @rakita

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+        // Safety: if it is a valid pointer to a slice of `self.data`
+        unsafe { &*self.current_slice }
+    }
+
+    /// Get a mutable refernce to the memory of the current context
+    #[inline(always)]
+    fn get_current_slice_mut(&mut self) -> &mut [u8] {
+        // Safety: it is a valid pointer to a slice of `self.data`
+        unsafe { &mut *self.current_slice }
+    }
```
DaniPopes, line 238: remove the `get_` prefix, Rust getters don't usually use prefixes
<https://rust-lang.github.io/api-guidelines/naming.html>

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Calculates memory allocation upper bound using
+    /// https://2π.com/22/eth-max-mem
+    #[inline(always)]
+    fn calculate_upper_bound(gas_limit: u64) -> u64 {
```
DaniPopes, line 243: And move to the top of the impl block, this could probably be useful elsewhere or downstream

```suggestion
    #[inline]
    pub fn calculate_upper_bound(gas_limit: u64) -> u64 {
```

crates/revm/benches/bench.rs
```diff
@@ -111,7 +124,14 @@ fn bench_eval(g: &mut BenchmarkGroup<'_, WallTime>, evm: &Evm) {
         };
         let mut host = DummyHost::new(evm.env.clone());
         b.iter(|| {
-            let mut interpreter = Interpreter::new(Box::new(contract.clone()), u64::MAX, false);
+            // reset gas limit to the right amount before tx
+            evm.env.tx.gas_limit = gas_limit;
```
DaniPopes, line 128: Why does this need to happen?

DaniPopes (changes requested):
- cachegrind
  - main (8206193e65ad199a1dde1e79e38bf9ffe38118c9): 431,577,511
  - #673 (c7945ebcf375bf629d76fa64d04ff979b7cd15d3): 445,967,393

crates/revm/src/evm_impl.rs
DaniPopes: This enables the env field in the first place, but if we're going to store a limit in the memory I don't see a reason to cfg this

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// - we have enought gas to resize this vector
+    /// - we made new_size as multiply of 32
+    /// - [new_size] is greater than `self.len()`
+    pub fn resize(&mut self, new_size: usize) {
+        // extend with zeros
+        let range = self.current_len..new_size;
+
+        self.get_current_slice_mut()[range]
+            .iter_mut()
+            .for_each(|byte| *byte = 0);
```
DaniPopes, line 129: This can be `ptr::write_bytes(self.ptr.add(self.len), 0, new_size)`

DaniPopes (changes requested):

crates/interpreter/src/interpreter.rs
```diff
@@ -169,8 +146,10 @@ impl Interpreter {
 
     /// Returns a copy of the interpreter's return value, if any.
     #[inline]
-    pub fn return_value(&self) -> Bytes {
-        self.return_value_slice().to_vec().into()
+    pub fn return_value(&mut self) -> Bytes {
+        let return_data = self.return_value_slice().to_vec().into();
+        self.shared_memory.free_context_memory();
```
DaniPopes, line 151: I don't think this is great, can we keep this as is (to return a copy), and move the free to a finish/reset function (maybe it should also consume `self` instead of taking `&mut self`?)

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// Return the memory of the current context
+    pub fn data(&self) -> &[u8] {
+        self.get_current_slice()
+    }
+
+    /// Resize the memory. assume that we already checked if:
+    /// - we have enought gas to resize this vector
+    /// - we made new_size as multiply of 32
+    /// - [new_size] is greater than `self.len()`
+    pub fn resize(&mut self, new_size: usize) {
```
DaniPopes, line 123: Marking this and all the functions above `#[inline(always)]` should help with perf regression. On cachegrind bench I'm seeing 214812 calls to memset (from inlined `Vec::resize`) become calls to `SharedMemory::resize`

DaniPopes (changes requested):

crates/revm/src/evm_impl.rs
lorenzofero: My first idea was to keep the feature memory limit as a cap of how much memory you'd like to allocate, since the upper bound calculation can return a potentially high number of bytes. An example was here:
```rust
pub fn new_with_memory_limit(gas_limit: u64, memory_limit: u64) -> Self {
        let size = min(
            SharedMemory::calculate_upper_bound(gas_limit) as usize,
            memory_limit as usize,
        );
```

 If you want I can remove it, no problem!

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: My bad I write too much TypeScript in my day job :smiling_face_with_tear: :

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,297 @@
...
+        let size = min(
+            SharedMemory::calculate_upper_bound(gas_limit) as usize,
+            2usize.pow(32) - 1,
+        );
+
+        let mut data = vec![0; size];
+        let current_slice: *mut [u8] = &mut data[..];
+        SharedMemory {
+            data,
+            checkpoints: Vec::with_capacity(1024),
```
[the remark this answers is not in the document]
lorenzofero: I need a memory checkpoint for each level of depth. Now that I noticed that was call_stack_limit before merging, which was clearer

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Ok I get your point regarding `new_size` and `len`. However, this function is always used in the `shared_memory_resize` macro where this check is already done. If you want, I can move that check inside here and remove it from the macro

Regarding the code snippet you wrote, I'll try to implement it! thanks

lorenzofero:
> This is great btw, I think we can get perf regression down to neutral or positive

Thanks and thank you for the very detailed feedback you provided; I'll try to resolve all the comments as soon as possible

crates/interpreter/src/interpreter/shared_memory.rs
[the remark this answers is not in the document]
DaniPopes: Sure it can be reduced to u32, but I mostly mean how often do we expect this to increase in length, just so that we don't pre-allocate too much

crates/interpreter/src/interpreter/shared_memory.rs
DaniPopes: Ah I see that, we should move the check to this function since it's public (make it return `Result<(), _>` like the stack functions)

lorenzofero: Sorry for all the questions but
- I tried to move the check inside the resize function, however the check still needs to live inside the macro because of the gas calculation that needs to be skipped, otherwise I get a root mismatch. Is this still convenient to do?
- I'm trying to understand the `data_mut` as suggested from `slice::from_raw_parts_mut(self.ptr, self.data.as_ptr().add(self.data.len()).sub(self.ptr))`. What is `self.ptr` here?

DaniPopes: - that's fine, you can keep it in the macro then
- this https://github.com/bluealloy/revm/pull/673#discussion_r1337555841

crates/interpreter/src/interpreter.rs
lorenzofero: I don't think that consuming `self` is really that useful: this function is called inside `run_interpreter` but there we want to (probably in this order):
- save and return the return_data
- save return the gas used
- free the context memory
I cannot free the context memory after calling `return_value` is it consumes `self`.
However, I moved the free call just before the end of the `run_interpreter` call and I think it makes sense: in the same scope you create and free the context memory.
Let me know if it ok now

crates/interpreter/src/interpreter/shared_memory.rs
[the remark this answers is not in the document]
lorenzofero: You are right. I've reduced the initial capacity to 32 which should be enough for most uses

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Take a look at the `resize` function now, I should have matched what you wanted. The for loop should be faster than calling the closure with its overhead

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Sorry that I misread/didn't understand it the first time, now it should be good!

DaniPopes (changes requested):

crates/interpreter/src/interpreter.rs
```diff
@@ -104,14 +88,7 @@ impl Interpreter {
...
 
-    /// Returns a reference to the interpreter's memory.
-    #[inline]
-    pub fn memory(&self) -> &Memory {
-        &self.memory
-    }
-
-    /// Returns a reference to the interpreter's stack.
-    #[inline]
+    /// Reference of interpreter stack.
```
DaniPopes, line 91: Revert

crates/interpreter/src/interpreter.rs
```diff
@@ -169,7 +146,7 @@ impl Interpreter {
 
     /// Returns a copy of the interpreter's return value, if any.
     #[inline]
-    pub fn return_value(&self) -> Bytes {
+    pub fn return_value(&mut self) -> Bytes {
```
DaniPopes, line 149: Revert

DaniPopes (changes requested):

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,286 @@
...
+            .finish()
+    }
+}
+
+impl SharedMemory {
+    /// Calculates memory allocation upper bound using
+    /// https://2π.com/22/eth-max-mem
+    #[inline]
+    pub fn calculate_upper_bound(gas_limit: u64) -> u64 {
+        4096 * sqrt(2u64.checked_mul(gas_limit).unwrap_or(u64::MAX))
```
DaniPopes, line 45:
```suggestion
        4096 * sqrt(2u64.saturating_mul(gas_limit))
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /// Allocate memory to be shared between calls.
+    /// Memory size is estimated using https://2π.com/22/eth-max-mem
+    /// which depends on transaction [gas_limit].
+    /// Maximum allocation size is 2^32 - 1 bytes;
+    pub fn new(gas_limit: u64) -> Self {
+        Self::new_with_memory_limit(gas_limit, (u32::MAX - 1) as u64)
```
DaniPopes, line 53: `u32::MAX` is `2^32 - 1`

```suggestion
        Self::new_with_memory_limit(gas_limit, u32::MAX as u64)
```

DaniPopes (changes requested):

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,286 @@
...
+    #[cfg_attr(debug_assertions, track_caller)]
+    pub fn copy(&mut self, dst: usize, src: usize, len: usize) {
+        self.current_slice_mut().copy_within(src..src + len, dst);
+    }
+
+    /// Get a reference to the memory of the current context
+    #[inline(always)]
+    fn current_slice(&self) -> &[u8] {
+        // Safety: if it is a valid pointer to a slice of `self.data`
+        unsafe { slice::from_raw_parts(self.current_ptr, self.limit) }
```
DaniPopes, line 221: this should be current length
```suggestion
        unsafe { slice::from_raw_parts(self.current_ptr, self.current_len) }
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /// Get a mutable reference to the memory of the current context
+    #[inline(always)]
+    fn current_slice_mut(&mut self) -> &mut [u8] {
+        // Safety: it is a valid pointer to a slice of `self.data`
+        unsafe { slice::from_raw_parts_mut(self.current_ptr, self.limit) }
```
DaniPopes, line 228:
```suggestion
        unsafe { slice::from_raw_parts_mut(self.current_ptr, self.current_len) }
```

DaniPopes (changes requested):
I think this is what's causing tests to fail

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Changed the `for` loop with the `fill` method for consistency

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        self.current_ptr = self.data[new_checkpoint..].as_mut_ptr();
+        self.current_len = 0;
+    }
+
+    /// Prepares the shared memory for returning to the previous context
+    pub fn free_context_memory(&mut self) {
+        if let Some(old_checkpoint) = self.checkpoints.pop() {
+            let last_checkpoint = self.last_checkpoint();
+            self.current_ptr = self.data[last_checkpoint..].as_mut_ptr();
```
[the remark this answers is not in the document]
lorenzofero: Good idea! Done

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: My bad that I don't know Rust very well yet

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: This is what happens when you commit on a friday after 6pm :smiling_face_with_tear: I'm sorry

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: This doesn't work, or there is a misunderstanding about what this function should do. 
While also `self.limit` was not good but made it pass the tests ("wasting" memory), this should be as it is now, which is `self.data.len() - self.last_checkpoint()` (which is actually all the memory available for the current context).
With `self.current_len` tests fail for the following reason:
- when you resize, you take the range `self.current_len...new_size`, with `self.current_len` < `new_size`
- however, when you take the slice since you take it up to `self.current_len`, you get index out of bounds because it is too short

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Same as above

lorenzofero:
Today if I have time I'll take a look at tests failing with `ethtests` profile

crates/revm/benches/bench.rs
lorenzofero: I remember I needed that a while ago, but now it seems to be useless. Thanks for noticing it!

rakita:
Hi @lorenzofero, how is the performance on this?
What I am thinking rn is this becoming a more complex solution than the current one, additionally the peak memory usage is lower but average memory usage is a few times more than the previous where your memory per call was dynamically incremented while this approach takes its maximum right away.

In essence, I expected this to be more impactful but if this is not the case unfortunately there is no good reason to include it.

lorenzofero:
> * cachegrind
>   
>   * main ([8206193](https://github.com/bluealloy/revm/commit/8206193e65ad199a1dde1e79e38bf9ffe38118c9)): 431,577,511
>   * [Shared memory between calls #673](https://github.com/bluealloy/revm/pull/673) ([c7945eb](https://github.com/bluealloy/revm/commit/c7945ebcf375bf629d76fa64d04ff979b7cd15d3)): 445,967,393

Which is around -3.5% in performance. After current changes, this is what I get:
* main https://github.com/bluealloy/revm/commit/4e78fbe86ef4d030968d00f461926abbcb813943: 469,208,070
* [Shared memory between calls #673](https://github.com/bluealloy/revm/pull/673/commits/6b6119e648013673a66641ebb315f6593da607fd): 482,056,264

which is around -2.7%.

@DaniPopes maybe I can try with this now to see if we can squeeze a little bit more
> I think we can go a step further here and use uninit memory for extra perf. This may not be that big so maybe wait until the end to try this.

lorenzofero:
Hey, I hope you don't find all of this comments pedantic. I tried to use a different model for the shared memory similar to https://github.com/ethereum/evmone/issues/481#issuecomment-1179088645 that you can see here on my fork: https://github.com/lorenzofero/revm/pull/2. This model has some benefits imo:
- very simple -- a little more than a wrapper of original memory for the shared setup
- no big pre-allocation based on gas limit -- it allocates 4KiB as the original memory, and _only when you need more_, it allocates another slot of 4KiB, which will be kept for the next calls (see `resize` method)
- overall better performance in most situation -- if you don't use the memory, no performance penalties

Here is the result of running `cargo bench --all` against `main`:
```
analysis/transact/raw   time:   [4.0090 µs 4.1340 µs 4.2474 µs]
                        change: [-57.683% -52.266% -45.567%] (p = 0.00 < 0.05)
                        Performance has improved.
analysis/transact/checked
                        time:   [3.9221 µs 3.9936 µs 4.0423 µs]
                        change: [-45.316% -44.017% -42.611%] (p = 0.00 < 0.05)
                        Performance has improved.
analysis/transact/analysed
                        time:   [2.3999 µs 2.4642 µs 2.5303 µs]
                        change: [-54.329% -53.175% -52.059%] (p = 0.00 < 0.05)
                        Performance has improved.

snailtracer/transact/analysed
                        time:   [60.762 ms 61.650 ms 62.692 ms]
                        change: [-3.0766% +0.7508% +4.7091%] (p = 0.73 > 0.05)
                        No change in performance detected.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high mild
snailtracer/eval        time:   [58.104 ms 59.000 ms 59.465 ms]
                        change: [-10.689% -7.1282% -3.1639%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

transfer/transact/analysed
                        time:   [960.41 ns 967.11 ns 975.02 ns]
                        change: [+13.760% +16.191% +18.220%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```
If you like it, I can bring the changes to this branch

rakita:
> Hey, I hope you don't find all of this comments pedantic. 

It is fine, this is a tough decision as you invested a lot of time into this, and the benefits are unfortunately small, I have a few examples of this where the idea didn't pan out as expected (gas block i am looking at you, we had performance boost of 5-7% and was merged but it was very hard to use, so after few months it was reverted) so in the end it was more like a research effort to get data. I am judging this only on current and new code.

> I tried to use a different model for the shared memory similar to [ethereum/evmone#481 (comment)](https://github.com/ethereum/evmone/issues/481#issuecomment-1179088645) that you can see here on my fork: [lorenzofero#2](https://github.com/lorenzofero/revm/pull/2). This model has some benefits imo:
> 
> * very simple -- a little more than a wrapper of original memory for the shared setup
> * no big pre-allocation based on gas limit -- it allocates 4KiB as the original memory, and _only when you need more_, it allocates other slots pf 4KiB, which will be kept for the next calls (see `resize` method)
> * overall better performance in most situation -- and if you don't use the memory, no performance penalties

This seems better, not all transactions are going to be 1024 calls deep or use 30M gas, so this is more reasonable. But with this, 
* you would always copy all data to the newly allocated vec when resizing happens. That is why I that comment i showed memory in chunks to mitigate this somehow.
* but it allows more compact usage of allocated memory.
* memory peak stays allocated.

And spilling context (in this case memory) from one Interpreter to another would be fine if we would gain something significant on the other hand having just one place for memory opens things for new ideas. This is probably going to look better if we switch from recursive calls to loop calls (if the loop approach turns out okay).

I am on the fence here but let us include it, will review it in detail in the next few days (@DaniPopes already did an amazing job there).

lorenzofero:
Ok I brought the changes here for reviewing. However, I was wondering if it was worth it to open a new PR which supersedes this one with this new model, since both commit history and conversation here are becoming a little messy imo. Let me know what you think about it!

rakita:
> Ok I brought the changes here for reviewing. However, I was wondering if it was worth it to open a new PR which supersedes this one with this new model, since both commit history and github conversation here are becoming a little messy imo. Let me know what you think about it!

It is messy but it I fine for it to be in one place, so people can follow what is happening.

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -1,88 +1,99 @@
...
+/// a `Vec` for internal representation.
+/// A [SharedMemory] instance should always be obtained using
+/// the `new` static method to ensure memory safety.
+pub struct SharedMemory {
+    /// Shared buffer
     data: Vec<u8>,
+    /// Memory checkpoints for each depth
+    checkpoints: Vec<usize>,
+    /// How much memory has been used in the current context
+    current_len: usize,
```
rakita, line 20: We can probably put memory limit here, it feels like a better place

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -180,7 +197,31 @@ impl Memory {
...
+        let last_checkpoint = self.last_checkpoint();
+        let current_len = self.current_len;
+        // Safety: it is a valid pointer to a slice of `self.data`
+        &mut self.data[last_checkpoint..last_checkpoint + current_len]
+    }
+
+    /// Get the last memory checkpoint
+    #[inline(always)]
+    fn last_checkpoint(&self) -> usize {
+        *self.checkpoints.last().unwrap_or(&0)
```
rakita, line 224:
```suggestion
        self.checkpoints.last().cloned().unwrap_or_default()
```
`cloned` would remove the reference from the option.

rakita (changes requested):
We should reintroduce the memory limit.

Other parts look good!

lorenzofero:
> We should reintroduce the memory limit.

It should be good now!

rakita (approved):
lgtm!

lorenzofero:
I'm very happy we've found the right approach for this and got some good performance improvements. Thanks a lot @rakita and @DaniPopes for all the support in the last month; it has been a pleasure working with you.

PR #682, opened 2023-09-04
feat: Optimism execution changes

clabby:
## Overview

Implements Optimism's execution changes into `revm` behind the `optimism` feature flag. These changes enable native Optimism execution changes in `op-reth`, `foundry`, `anvil`, and other projects building on top of `revm`.

### Goal

With the `optimism` feature flag enabled, a new configuration option is exposed that allows for enabling Optimism's execution changes at runtime. This is so that tools such as Foundry can swap between regular Ethereum execution and Optimism execution as needed (i.e. for multichain fork testing). `revm`'s regular logic should be unaltered if either the `optimism` cfg option is `false` or the `optimism` feature flag is not enabled.

### New Hardforks
- `Bedrock` (Ethereum Base: `Merge`)
- `Regolith` (Base: `Bedrock`)

### TODO
- [x] Tests
    - [x] Unit tests
- [x] Implement the spec types + extra `TxEnv` fields within `revm-primitives`.
- [x] Ignore transaction field validation and nonce / balance checks on deposit transactions.
- [x] Alter gas accounting for deposit transactions in both the Bedrock and Regolith hardforks.
- [x] Mint ETH at the start of the state transition if the transaction is a deposit with a `Some` mint field.
- [x] Pay out fees to the fee vaults rather than coinbase in non-deposit transactions when the `optimism` configuration flag is enabled.
- [x] Implement the L1 fee calculation by reading the storage of the `L1Block` contract.

**Future**
- Benchmarks
    - How much overhead does the runtime branching on the `optimism` cfg option add when the `optimism` feature flag is enabled but the `optimism` cfg option is not? For example, in order for `foundry` to swap between `optimism` and `ethereum` execution modes, the regular release build will have to enable the `optimism` feature flag on `revm`. Does the runtime branching significantly slow down execution for the majority of tests which are not on OP?


crates/revm/src/evm_impl.rs
```diff
@@ -207,11 +272,61 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
...
+                        // Regolith, gas is reported as normal.
+                        gas.erase_cost(ret_gas.remaining());
+                        gas.record_refund(ret_gas.refunded());
+                    } else if is_deposit && tx_system.unwrap_or(false) {
+                        // System transactions were a special type of deposit transaction in
+                        // the Bedrock hardfork that did not incur any gas costs.
+                        gas.erase_cost(tx_gas_limit);
+                    }
                 }
                 return_revert!() => {
```
clabby, line 308: Note: Deposit transactions post-regolith will always bump the nonce of the account if the call was a contract creation and it resulted in an EVM halt. Don't think this property is upheld right now.

crates/primitives/src/env.rs
```diff
@@ -45,6 +45,14 @@ pub struct TxEnv {
...
     pub nonce: Option<u64>,
     pub access_list: Vec<(B160, Vec<U256>)>,
+    #[cfg(feature = "optimism")]
+    pub source_hash: Option<B256>,
+    #[cfg(feature = "optimism")]
+    pub mint: Option<u128>,
+    #[cfg(feature = "optimism")]
+    pub is_system_transaction: Option<bool>,
+    #[cfg(feature = "optimism")]
+    pub l1_cost: Option<U256>,
```
clabby, line 55: Calling this out - having `l1_cost` in the env is a bit of a workaround. In order to compute the L1 cost of a transaction on OP, we need to have access to the full 2718 encoded tx within `revm`, which felt leaky. Instead of computing the L1 cost within revm, I've offloaded this responsibility to the processes that fill the `TxEnv`.

It may make more sense to just pass the enveloped encoding here rather than the L1 cost - we need to load the L1 block information within `revm` anyways in order to determine the fee that heads to the base fee vault, so doing this could save the parent process an extra database load of the `L1Block` account. Open to opinions here 😄

crates/primitives/src/env.rs
```diff
@@ -45,6 +45,14 @@ pub struct TxEnv {
     pub chain_id: Option<u64>,
     pub nonce: Option<u64>,
     pub access_list: Vec<(B160, Vec<U256>)>,
+    #[cfg(feature = "optimism")]
```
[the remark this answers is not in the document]
refcell: Potential fix here: https://github.com/anton-rs/op-revm/pull/2

crates/revm/src/evm_impl.rs
```diff
@@ -143,6 +178,36 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
...
+            // If the transaction is not a deposit transaction, subtract the L1 data fee from the
+            // caller's balance directly after minting the requested amount of ETH.
+            if !is_deposit {
+                if let Some(l1_cost) = tx_l1_cost {
+                    journal
+                        .load_account(tx_caller, self.data.db)
+                        .map_err(EVMError::Database)?
+                        .0
+                        .info
+                        .balance -= l1_cost;
```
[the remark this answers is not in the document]
refcell: FYI @clabby, attempted to refactor and use `saturating_sub` here: https://github.com/anton-rs/op-revm/pull/4

crates/revm/src/evm_impl.rs
```diff
@@ -207,11 +272,61 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
         if crate::USE_GAS {
             match exit_reason {
                 return_ok!() => {
-                    gas.erase_cost(ret_gas.remaining());
-                    gas.record_refund(ret_gas.refunded());
+                    #[cfg(not(feature = "optimism"))]
```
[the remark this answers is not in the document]
refcell: Agreed. FYI @clabby, attempted a refactor to fix this in https://github.com/anton-rs/op-revm/pull/5

crates/revm/src/evm_impl.rs
```diff
@@ -207,11 +272,61 @@ impl<'a, GSPEC: Spec, DB: Database, const INSPECT: bool> Transact<DB::Error>
...
+                    {
+                        // For regular transactions prior to Regolith and all transactions after
+                        // Regolith, gas is reported as normal.
+                        gas.erase_cost(ret_gas.remaining());
+                        gas.record_refund(ret_gas.refunded());
+                    } else if is_deposit && tx_system.unwrap_or(false) {
+                        // System transactions were a special type of deposit transaction in
+                        // the Bedrock hardfork that did not incur any gas costs.
+                        gas.erase_cost(tx_gas_limit);
+                    }
```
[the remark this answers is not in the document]
refcell: This is the case when the transaction is a deposit transaction pre-regolith. In this case (as clabby noted in the comments), the deposit transaction reports the gas limit as its usage since it's pre-paid for on L1. As such, we don't need to erase any gas costs or perform any refunds :)

crates/revm/src/evm_impl.rs
refcell: This should be fixed in https://github.com/anton-rs/op-revm/pull/11

crates/primitives/src/env.rs
refcell: We decided to switch to accept the enveloped tx bytes here instead (implemented in https://github.com/anton-rs/op-revm/pull/16).

The reason to go this route is that specifying the `l1_cost` here means downstream applications/users will need to compute the l1 tx cost themselves using the l1 block info. Since this can, and is done internally, exposing this computation is leaky.

Instead, we can accept the 2718 enveloped tx bytes and then compute the l1 cost internally. This a) reduces the amount of work that users have to perform and b) reduces "dead code" in the form of the L1 block info l1 cost computation already available inside the `optimism` module in the `revm` crate.

Ultimately, either method works but this should be easier to with 😄

clabby:
Think this is ready for a re-review. What are your normal standards for tests @rakita? @refcell and I looked into integrating with t8n and t9n in order to generate state tests for OP, but that may be a bit further out. Glad to attempt to hook this version into `op-reth` and test syncing on top of the unit tests we've added!

Feel free to let us know if there is anything we can change as well 😄

refcell:
Would really really appreciate another round of review here

bins/revme/src/statetest/runner.rs
```diff
@@ -65,6 +65,12 @@ fn skip_test(path: &Path) -> bool {
         // custom json parser. https://github.com/ethereum/tests/issues/971
         | "ValueOverflow.json"
 
+        // TODO remove this after we merge branch to main.
+        | "NoSrcAccountCreate.json"
+        | "NoSrcAccount1559.json"
+        | "NoSrcAccountCreate1559.json"
+        | "NoSrcAccount.json"
+
```
rakita, line 73:
```suggestion
```

rakita (approved):
lgtm! One nit comment.

refcell:
Just FYI @rakita neither @clabby nor I have write access so I think it's up to you to hit merge 😄

PR #687, opened 2023-09-05
chore(state): Make Database more generic.

rakita:
Make the State database more generic and introduce `StateDBBox` as an alternative to the current type.


crates/revm/src/db/states/state_builder.rs
```diff
@@ -28,7 +27,7 @@ pub struct StateBuilder<'a, DBError> {
     pub with_block_hashes: BTreeMap<u64, B256>,
 }
 
-impl Default for StateBuilder<'_, Infallible> {
+impl Default for StateBuilder<Box<EmptyDB>> {
```
DaniPopes, line 30: This can be made generic

```rust
impl<DB: Default> Default for StateBuilder<DB> {
```

PR #688, opened 2023-09-05
feat(StateBuilder): switch builder option from without_bundle to with_bundle

rakita:
In general output state and creating of revert is better to be something to be enabled, and not disabled.

### DIFF
--- a/crates/revm/src/db/states/state_builder.rs
+++ b/crates/revm/src/db/states/state_builder.rs
@@ -6,26 +6,26 @@ use revm_interpreter::primitives::{db::Database, B256};
 
 /// Allows building of State and initializing it with different options.
 pub struct StateBuilder<'a, DBError> {
-    pub with_state_clear: bool,
+    with_state_clear: bool,
     /// Optional database that we use to fetch data from. If database is not present, we will
     /// return not existing account and storage.
     ///
     /// Note: It is marked as Send so database can be shared between threads.
-    pub database: Box<dyn Database<Error = DBError> + Send + 'a>,
+    database: Box<dyn Database<Error = DBError> + Send + 'a>,
     /// if there is prestate that we want to use.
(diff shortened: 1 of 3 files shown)


crates/revm/src/db/states/state_builder.rs
```diff
@@ -18,8 +18,8 @@ pub struct StateBuilder<'a, DBError> {
     /// This will initialize cache to this state.
     pub with_cache_prestate: Option<CacheState>,
     /// Do we want to create reverts and update bundle state.
-    /// Default is true.
-    pub without_bundle_update: bool,
+    /// Default is false.
+    pub with_bundle_update: bool,
     /// Do we want to merge transitions in background.
     /// This will allows evm to continue executing.
     /// Default is false.
```
mattsse, line 25: Should all of these be private(pub crate) instead?
I don't think they should be mut accessible from the outside

crates/revm/src/db/states/state_builder.rs
```diff
@@ -86,13 +86,13 @@ impl<'a, DBError> StateBuilder<'a, DBError> {
...
 
-    /// Don't make transitions and don't update bundle state.
+    /// Make transitions and update bundle state.
     ///
-    /// This is good option if we don't care about creating reverts
-    /// or getting output of changed states.
-    pub fn without_bundle_update(self) -> Self {
+    /// This is needed option if we want to create reverts
+    /// and getting output of changed states.
+    pub fn witho_bundle_update(self) -> Self {
```
mattsse, line 93:
```suggestion
    pub fn with_bundle_update(self) -> Self {
```

mattsse (changes requested):

crates/revm/src/db/states/state_builder.rs
rakita: Makes sense

PR #690, opened 2023-09-06
chore: make `impl Default for StateBuilder` generic

DaniPopes:
Follow-up to #687, see https://github.com/bluealloy/revm/pull/687#discussion_r1316523492

### DIFF
--- a/crates/revm/src/db/states/state_builder.rs
+++ b/crates/revm/src/db/states/state_builder.rs
@@ -8,14 +8,11 @@ use revm_interpreter::primitives::{
 
 /// Allows building of State and initializing it with different options.
 pub struct StateBuilder<DB> {
+    /// Database that we use to fetch data from.
+    database: DB,
     /// Enabled state clear flag that is introduced in Spurious Dragon hardfork.
     /// Default is true as spurious dragon happened long time ago.
     with_state_clear: bool,
-    /// Optional database that we use to fetch data from. If database is not present, we will
-    /// return not existing account and storage.
-    ///
-    /// Note: It is marked as Send so database can be shared between threads.
-    database: DB,
     /// if there is prestate that we want to use.
     /// This would mean that we have additional state layer between evm and disk/database.
     with_bundle_prestate: Option<BundleState>,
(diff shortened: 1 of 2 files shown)


rakita (changes requested):
This is more verbose than it was. Using `EmptyDB` removes the need to specify the type for `default()`

DaniPopes:
This is only true with `StateBuilder`, which can be changed by making the type generic default to `EmptyDB`. But I don't see why `Default` (or other traits in the future) are only implemented for the default type.

crates/revm/src/db/states/state_builder.rs
```diff
@@ -29,26 +26,38 @@ pub struct StateBuilder<DB> {
...
-}
-
-impl<DB: Database> StateBuilder<DB> {
-    /// Create default instance of builder.
-    pub fn new() -> StateBuilder<Box<EmptyDB>> {
-        StateBuilder::<Box<EmptyDB>>::default()
-    }
 
+    #[doc(hidden)]
+    #[deprecated = "instantiate with `new_with_database` instead"]
```
rakita, line 60: I use this fn inside reth, maybe it is not needed as there is `fn with_database` currently in the main, but can you remove `deprecated` for now?

rakita (approved):
Pending conflicts, but lgtm!

I added  `State::builder()` inside State so `StateBuilder` doesn't need to be specified in usually flows. So we can do `State::builder().build()` pattern.

DaniPopes:
Re Based

crates/revm/src/db/states/state_builder.rs
```diff
@@ -32,27 +29,37 @@ pub struct StateBuilder<DB> {
...
-}
-
-impl<DB: Database> StateBuilder<DB> {
-    /// Create default instance of builder.
-    pub fn new() -> StateBuilder<EmptyDB> {
-        StateBuilder::default()
-    }
 
-    /// With generic database.
+    /// Set the database. Prefer using [`new_with_database`](Self::new_with_database) when possible.
```
rakita, line 62:
```suggestion
    /// Set the database.
```

PR #691, opened 2023-09-06
chore: implement `Default` for other databases

DaniPopes:
And some code cleanup and docs

### DIFF
--- a/crates/primitives/src/db.rs
+++ b/crates/primitives/src/db.rs
@@ -1,5 +1,3 @@
-pub mod components;
-
 use crate::AccountInfo;
 use crate::U256;
 use crate::{Account, Bytecode};
@@ -7,21 +5,26 @@ use crate::{B160, B256};
 use auto_impl::auto_impl;
 use hashbrown::HashMap as Map;
 
+pub mod components;
 pub use components::{
     BlockHash, BlockHashRef, DatabaseComponentError, DatabaseComponents, State, StateRef,
 };
 
+/// EVM database interface.
 #[auto_impl(&mut, Box)]
 pub trait Database {
     type Error;
+
     /// Get basic account information.
     fn basic(&mut self, address: B160) -> Result<Option<AccountInfo>, Self::Error>;
-    /// Get account code by its hash
+
+    /// Get account code by its hash.
     fn code_by_hash(&mut self, code_hash: B256) -> Result<Bytecode, Self::Error>;
+
     /// Get storage value of address at index.
     fn storage(&mut self, address: B160, index: U256) -> Result<U256, Self::Error>;
 
-    // History related
+    /// Get block hash by block number.
     fn block_hash(&mut self, number: U256) -> Result<B256, Self::Error>;
 }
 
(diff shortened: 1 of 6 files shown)


crates/revm/src/db/emptydb.rs
```diff
@@ -1,80 +1,87 @@
...
     fn storage(&self, _address: B160, _index: U256) -> Result<U256, Self::Error> {
         Ok(U256::default())
     }
 
-    // History related
+    #[inline(always)]
     fn block_hash(&self, number: U256) -> Result<B256, Self::Error> {
-        Ok(keccak256(&number.to_be_bytes::<{ U256::BYTES }>()))
+        Ok(number.to_be_bytes().into())
     }
```
DaniPopes, line 86: @rakita I assume this was the intended behavior for `keccak_block_hash = false`? I've removed the flag since it was unused

crates/revm/src/db/emptydb.rs
```diff
@@ -1,80 +1,87 @@
...
     fn basic(&self, _address: B160) -> Result<Option<AccountInfo>, Self::Error> {
         Ok(None)
     }
-    /// Get account code by its hash
+
+    #[inline(always)]
     fn code_by_hash(&self, _code_hash: B256) -> Result<Bytecode, Self::Error> {
         Ok(Bytecode::new())
     }
-    /// Get storage value of address at index.
```
DaniPopes, line 71: Docs on `impl Trait` functions are unnecessary if they are the same as the trait's.

crates/primitives/src/db.rs
```diff
@@ -1,27 +1,30 @@
-pub mod components;
-
 use crate::AccountInfo;
 use crate::U256;
 use crate::{Account, Bytecode};
 use crate::{B160, B256};
 use auto_impl::auto_impl;
 use hashbrown::HashMap as Map;
 
+pub mod components;
```
rakita, line 8: Why this style and not moving `pub used componnets` behind `pub mod`?

crates/revm/src/db/emptydb.rs
rakita: `keccak256` is needed for foundry in some way, fuzz tests iirc

crates/revm/src/db/in_memory_db.rs
```diff
@@ -41,6 +33,12 @@ pub struct CacheDB<ExtDB: DatabaseRef> {
     pub db: ExtDB,
 }
 
+impl<ExtDB: DatabaseRef + Default> Default for CacheDB<ExtDB> {
```
rakita, line 36: Not really liking this, having default is okay when you don't have any inputs. but this would require specifying `<DB>` generic item, so it is better to go simple and just have `fn new`

crates/primitives/src/db.rs
DaniPopes: I don't think there's a style guide for this, but I like when it's `mod x; use x::*;` groups, like: <https://github.com/alloy-rs/core/blob/d9d0ded3ca4e2dde992dad1afe6410015158d6b1/crates/syn-solidity/src/lib.rs>

crates/revm/src/db/in_memory_db.rs
DaniPopes: The generic has to be specified when doing `CacheDB::new(Default::default())` too, which now is the same as `CacheDB::<DB>::default()`

rakita: My point exactly, but with usage I would usually type `CacheDB::new(` and add check what rust-analyzer is giving me as use `CacheDB::new(DB::default())`, having generic there complicates this workflow and to the point, it is the same outcome but more complex. Simple and stupid is in most times a better way.

DaniPopes: Right, which is why `fn new` is unchanged for if you want the basic empty DB. Implementing Default allows having `CacheDB<MyDB>: Default` where `MyDB: Default`, which also allows deriving `Default` on structs containg it etc. I don't see anything wrong with this

rakita: Hm, not sure why i thought this was exclusive as one would disable the other. It is good to add it.

Just check `ExtDB: DatabaseRef + Default` it should be `ExtDB: Database + Default` as `Database` is the main used trait.

crates/revm/src/db/emptydb.rs
rakita: A long time ago: https://github.com/bluealloy/revm/pull/71
Supportive, returning number feels better. In unlikely case something breaks related to this, you would know where to lokk

crates/primitives/src/db.rs
rakita: I would prefer a little bit of a different order.

```
// use if needed. or maybe even put reexports.
use std::..

// list all modules in one place
mod components;
mod tests;

// reexport module items here.
pub use components::*
```

Something like this is easier to my eyes: https://github.com/bluealloy/revm/blob/main/crates/revm/src/db/states.rs
having `mod *` grouped together.

Just bikesheding, this change in general is nice

rakita (approved):
Check that `ExtDB: DatabaseRef + Default ` if it can be replaced with `Database` as it is main trait

Other things lgtm!

crates/revm/src/db/in_memory_db.rs
DaniPopes: `CacheDB` is defined with `ExtDB: DatabaseRef`, maybe in another PR?

rakita: I see, strange tbh. Yeah, another time, it will break things if we change it.

rakita (approved):

PR #693, opened 2023-09-06
`CfgEnv.chain_id` should be `u64`

### LINKED ISSUE #671: `CfgEnv.chain_id` should be `u64`
The only check done is by converting `TxEnv.chain_id` to `U256` which is redundant: https://github.com/bluealloy/revm/blob/42725351f4691c99addb1658e5deef7fb66335d0/crates/primitives/src/env.rs#L309-L314

It can be converted to `U256` directly in the `CHAINID` opcode with `U256::from`, since it's very cheap.

lorenzofero:
Closes #671. Let me know if I am missing something from this issue!

### DIFF
--- a/bins/revme/src/statetest/runner.rs
+++ b/bins/revme/src/statetest/runner.rs
@@ -172,7 +172,7 @@ pub fn execute_test_suit(
         }
         let mut env = Env::default();
         // cfg env. SpecId is set down the road
-        env.cfg.chain_id = U256::from(1); // for mainnet
+        env.cfg.chain_id = 1; // for mainnet
 
         // block env
         env.block.number = unit.env.current_number;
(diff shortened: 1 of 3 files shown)


DaniPopes:
hm actually i realized this is wrong, realistically chain id is a u64 but users should be able to set any arbitrary u256 in the EVM

rakita:
> hm actually i realized this is wrong, realistically chain id is a u64 but users should be able to set any arbitrary u256 in the EVM

This is one of those ambiguous things where the upper limit was not specified but all the "chains" use u64, so it is a good call to use u64.

Here is a site that lists chain ids: https://chainlist.org/
from github list is here: https://github.com/DefiLlama/chainlist/blob/main/constants/chainIds.json
And biggest value is: 836542336838601

rakita (approved):
lgtm

DaniPopes:
This was also enforced in Foundry, makes sense then

image

PR #701, opened 2023-09-07
feat: OriginalValueKnown, and expose db box

rakita:
Introduce `OriginalValueKnown` enum that replaces a boolean used as a flag.

Add `State::builder()` that allows better usage of StateBuilder.


rakita:
Was already merged: https://github.com/bluealloy/revm/pull/699

PR #702, opened 2023-09-07
Do not insert empty reverts in state

### LINKED ISSUE #634: Dont insert empty reverts in state
Noticed here: https://github.com/bluealloy/revm/pull/626#discussion_r1301347468

We should check if revert is empty and not insert it.

lorenzofero:
Closes #634. I hope I've got it right! Is there something else to check other than the tests in `states/state.rs`?


crates/revm/src/db/states/bundle_account.rs
```diff
@@ -154,7 +154,7 @@ impl BundleAccount {
             AccountInfoRevert::DoNothing
         };
 
-        match updated_status {
+        let account_revert = match updated_status {
```
rakita, line 157: There is one branch that uses `return`  (like 251) we can just use `else` there to cover all of them

crates/revm/src/db/states/reverts.rs
```diff
@@ -171,6 +171,12 @@ impl AccountRevert {
             wipe_storage: true,
         }
     }
+
+    /// Returns `true` if there is nothing to revert,
+    /// by checking that account info and storage have been left untouched
+    pub fn is_empty(&self) -> bool {
+        self.account == AccountInfoRevert::DoNothing && self.storage.is_empty()
```
rakita, line 178: Just remembered one more thing. check if `wipe_storage` is `false`, if it is `true` it is an action that needs to happen.

crates/revm/src/db/states/bundle_account.rs
rakita: > There is one branch that uses `return` (like 251) we can just use `else` there to cover all of them

crates/revm/src/db/states/reverts.rs
lorenzofero: Yeah I had that doubt but I forgot to mention it. Tomorrow I’ll fix all comments. Thanks!

crates/revm/src/db/states/bundle_account.rs
lorenzofero: Ok I've removed every early return inside that function (there were two). It should be easier to reason about it now

rakita (approved):
lgtm

PR #708, opened 2023-09-09
Run the Ethereum tests both in debug and release mode

valo:
In debug mode the tests are checking for additional things like unsigned int arithmetic overflows. In release mode they are checking for things like proper stack utilization (ex. in release mode the tests are running with the default thread stack limit).

### DIFF
--- a/.github/workflows/ethereum-tests.yml
+++ b/.github/workflows/ethereum-tests.yml
@@ -29,9 +29,18 @@ jobs:
         with:
           cache-on-failure: true
 
-      - name: Run Ethereum tests
+      - name: Run Ethereum tests (Debug mode)
         run: |
           cargo run --profile ethtests -p revme -- \
+          statetest \
+          ethtests/GeneralStateTests/ \
+          ethtests/LegacyTests/Constantinople/GeneralStateTests/ \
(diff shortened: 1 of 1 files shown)


.github/workflows/ethereum-tests.yml
```diff
@@ -29,11 +29,20 @@ jobs:
...
           tests/EIPTests/StateTests/stEIP1153-transientStorage/
+
+      - name: Run Ethereum tests (Release mode)
+        run: |
+          cargo run --release -p revme -- \
+          statetest \
+          ethtests/GeneralStateTests/ \
+          ethtests/LegacyTests/Constantinople/GeneralStateTests/ \
+          tests/EIPTests/StateTests/stEIP5656-MCOPY/ \
+          tests/EIPTests/StateTests/stEIP1153-transientStorage/
```
DaniPopes, line 48: These tests already take a while to compile and run, can you make it a matrix so the two profiles run in parallel?

DaniPopes (changes requested):

DaniPopes (approved):

.github/workflows/ethereum-tests.yml
valo: Great suggestion! Now it's 2x faster 💪

PR #712, opened 2023-09-11
Never inline the prepare functions

### LINKED ISSUE #589: Test overflowed the stack
It seem something got changed and default stack size is not enough for running eth tests, bumped it to 3MB does the job but should check to be sure

valo:
These functions are used for keeping the stack memory clean on the host recursive code paths. EVM supports up to 1024 levels of recursion, so if one is not careful with the stack memory allocations, the stack memory can blow up.

Benchmarks show dramatic stack memory improvements when ont inlining these functions. A recursive bomb uses around 1.1MB of stack when these functions are not inlined. If inlined the recursive bombs blow up the stack.

Fixes #589

### DIFF
--- a/bins/revme/src/statetest/runner.rs
+++ b/bins/revme/src/statetest/runner.rs
@@ -365,12 +365,7 @@ pub fn run(
         let console_bar = console_bar.clone();
         let elapsed = elapsed.clone();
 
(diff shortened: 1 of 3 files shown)


rakita:
nice!

rakita (approved):

PR #713, opened 2023-09-11
Make evm.db not optional

### LINKED ISSUE #697: Make `EVM.db` not optional
Given that basically every `EVM` method requires that the `db` field is `Some`, we should remove the option entirely.

It would look something like this:

```diff
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -41,16 +41,16 @@ use revm_precompile::Precompiles;
 #[derive(Clone)]
 pub struct EVM<DB> {
     pub env: Env,
-    pub db: Option<DB>,
+    pub db: DB,
 }
 
-pub fn new<DB>() -> EVM<DB> {
-    EVM::new()
+pub fn new<DB>(db: DB) -> EVM<DB> {
+    EVM::new(db)
 }
 
-impl<DB> Default for EVM<DB> {
+impl<DB: Default> Default for EVM<DB> {
     fn default() -> Self {
-        Self::new()
+        Self::new(Default::default())
     }
 }


```

alessandromazza98:
Closes #697 

I removed the `Option<DB>` inside the `EVM` struct to just leave `DB`.

Let me know if it's ok.
Thanks

### DIFF
--- a/examples/fork_ref_transact.rs
+++ b/examples/fork_ref_transact.rs
@@ -4,7 +4,7 @@ use ethers_contract::BaseContract;
 use ethers_core::abi::parse_abi;
 use ethers_providers::{Http, Provider};
 use revm::{
-    db::{CacheDB, EmptyDB, EthersDB, states::cache},
+    db::{CacheDB, EmptyDB, EthersDB},
     primitives::{ExecutionResult, Output, TransactTo, B160, U256 as rU256},
     Database, EVM,
 };


crates/revm/src/evm.rs
```diff
@@ -68,133 +73,103 @@ impl<DB: Database + DatabaseCommit> EVM<DB> {
...
-    pub fn db(&mut self) -> Option<&mut DB> {
-        self.db.as_mut()
+    pub fn db(&mut self) -> &mut DB {
+        &mut self.db
     }
 
     pub fn take_db(&mut self) -> DB {
-        core::mem::take(&mut self.db).unwrap()
+        core::mem::take(&mut self.db)
     }
```
DaniPopes, line 173: Same as `fn database`

DaniPopes (changes requested):
This is great, just a few things

bins/revme/src/statetest/runner.rs
```diff
@@ -255,12 +255,12 @@ pub fn execute_test_suit(
...
                 ));
-                let mut state = revm::db::State::builder()
+                let state = revm::db::State::builder()
                     .with_cached_prestate(cache)
                     .with_bundle_update()
                     .build();
-                let mut evm = revm::new();
-                evm.database(&mut state);
+                let mut evm = revm::new(Default::default());
+                *evm.db() = state;
```
DaniPopes, line 263: can move this into the new call above

crates/revm/benches/bench.rs
```diff
@@ -73,8 +73,8 @@ fn snailtracer(c: &mut Criterion) {
 }
 
 fn transfer(c: &mut Criterion) {
-    let mut evm = revm::new();
-    evm.database(BenchmarkDB::new_bytecode(Bytecode::new()));
+    let mut evm = revm::new(Default::default());
+    *evm.db() = BenchmarkDB::new_bytecode(Bytecode::new());
```
DaniPopes, line 77: same here and in this file

DaniPopes (approved):
lgtm, nits you can ignore

crates/revm/src/evm.rs
```diff
@@ -34,31 +34,30 @@ use revm_precompile::Precompiles;
...
 #[derive(Clone)]
 pub struct EVM<DB> {
     pub env: Env,
-    pub db: Option<DB>,
+    pub db: DB,
 }
 
-pub fn new<DB>() -> EVM<DB> {
-    EVM::new()
+pub fn new<DB>(db: DB) -> EVM<DB> {
```
rakita, line 46: Would love to refresh this in the future, so to not break API two times can you revert this?

crates/revm/src/evm.rs
```diff
@@ -68,133 +67,95 @@ impl<DB: Database + DatabaseCommit> EVM<DB> {
...
+    pub fn with_env(env: Env, db: DB) -> Self {
+        Self { env, db }
     }
 
-    pub fn take_db(&mut self) -> DB {
-        core::mem::take(&mut self.db).unwrap()
+    pub fn db(&mut self) -> &mut DB {
+        &mut self.db
     }
 }
```
rakita, line 160: Can you revert this so as to not break functions

`pub fn new() -> Self {` should init with `EmptyDB` this would remove a need to have `Option` and would make code nicer

rakita (changes requested):
Code looks a lot cleaner without `Option`, but I would restrain from breaking current functions as this is something I want to rework in near future so it does not make sense to break it twice

alessandromazza98:
> Code looks a lot cleaner without `Option`, but I would restrain from breaking current functions as this is something I want to rework in near future so it does not make sense to break it twice

Ok I can revert those changes, but I am now struggling because without breaking those functions is not easy for me.

In particular I have the problem that, if I leave the `Option`, then I have problems in `with_env` and `take_db` functions.
```rust
impl<DB> EVM<DB> {
    /// Creates a new [EVM] instance with the default environment,
    pub fn new() -> Self {
        Self::with_env(Default::default())
    }

    /// Creates a new [EVM] instance with the given environment.
    pub fn with_env(env: Env) -> Self {
        Self { env, db: None }
    }                              ^---> here 

    pub fn database(&mut self, db: DB) {
        self.db = db;
    }

    pub fn db(&mut self) -> &mut DB {
        &mut self.db
    }

    pub fn take_db(&mut self) -> DB {
        core::mem::take(&mut self.db)
    }                               ^---> here
}
```
One way to solve those errors is to add `DB: Default` trait in the impl and do like this:
```rust
impl<DB: Default> EVM<DB> {
    /// Creates a new [EVM] instance with the default environment,
    pub fn new() -> Self {
        Self::with_env(Default::default())
    }

    /// Creates a new [EVM] instance with the given environment.
    pub fn with_env(env: Env) -> Self {
        Self { env, db: Default::default() }
    }                              

    pub fn database(&mut self, db: DB) {
        self.db = db;
    }

    pub fn db(&mut self) -> &mut DB {
        &mut self.db
    }

    pub fn take_db(&mut self) -> DB {
        core::mem::take(&mut self.db)
    }                               
}
```
But then I have to put the Default constrain also to the `new` function and also in the impl of the Default trait for `EVM`.

So I tried do do as you suggested: init with an `EmptyDB` db, doing something like this:
```rust
impl<DB> EVM<DB> {
    /// Creates a new [EVM] instance with the default environment,
    pub fn new() -> Self {
        Self::with_env(Default::default())
    }

    /// Creates a new [EVM] instance with the given environment.
    pub fn with_env(env: Env) -> Self {
        Self { env, db: EmptyDB::default() }
    }                                     ^--> here I put EmptyDB

    pub fn database(&mut self, db: DB) {
        self.db = db;
    }

    pub fn db(&mut self) -> &mut DB {
        &mut self.db
    }

    pub fn take_db(&mut self) -> DB {
        core::mem::take(&mut self.db)
    }
}
```
But of course this is not good since the expected type is `DB` and not `EmptyDB`.

If you could give me some pointers about how to start tackling this problem I can try and solve it.

Thank you very much

rakita:
@alessandromazza98 you are right, let us move this to later after we create new revm version

rakita:
This is getting changed a lot inside EvmBuilder https://github.com/bluealloy/revm/pull/888

PR #714, opened 2023-09-11
chore(docs): minor improvements, create separate page for `Host` trait

Evalir:
Was reading the docs seeing what can be improved (and what's out of date) and spotted a bunch of smol things to improve.

The idea here is to slowly start improving the `revm` docs bit by bit until it becomes an amazing resource not only to learn about how to use `revm`, but to learn about the EVM and related EIP themselves and level up in the meanwhile. I'll probably also be submitting code docs as well over time.

As always, feel free to nit on this 😄


rakita (approved):
nice one!

PR #715, opened 2023-09-12
docs: add warning on panic conditions in take_bundle

Rjected:
These docs were added after encountering a panic in reth: https://github.com/paradigmxyz/reth/pull/4561


rakita:
This can be a footgun to unwrap it like that. `Option` or `unwrap_or_default` are good alternatives

rakita (approved):

PR #716, opened 2023-09-12
chore(`perf`): only recalc code hash if its the default

Evalir:
Stems from https://github.com/foundry-rs/foundry/issues/5730#issuecomment-1700458447

Only recalculate the code hash when inserting if it's the default


rakita (approved):
lgtm

PR #719, opened 2023-09-14
  EIP-6780: SELFDESTRUCT only in same transaction

### LINKED ISSUE #524: EIP-6780: SELFDESTRUCT only in same transaction
More info here: https://eips.ethereum.org/EIPS/eip-6780

lorenzofero:
Closes #524. Reading the [EIP](https://eips.ethereum.org/EIPS/eip-6780), when the contract is not created in the same tx, `SELFDESTRUCT` simply becomes a transfer, while keeping the gas cost of the opcode. As such, I tried to modify only the `JournaledState` to reflect that.

### DIFF
--- a/crates/revm/src/journaled_state.rs
+++ b/crates/revm/src/journaled_state.rs
@@ -509,6 +509,10 @@ impl JournaledState {
                 balance,
             })
         } else {
+            // State is not changed:
+            // * if we are after Cancun upgrade and
+            // * Selfdestruct account that is created in the same transaction and
+            // * Specify the target is same as selfdestructed account. The balance stays unchanged.
             None
         };
 
--- a/crates/revm/src/journaled_state.rs
+++ b/crates/revm/src/journaled_state.rs
@@ -21,7 +21,8 @@ pub struct JournaledState {
     /// journal with changes that happened between calls.
     pub journal: Vec<Vec<JournalEntry>>,
     /// Ethereum before EIP-161 differently defined empty and not-existing account
-    /// so we need to take care of that difference.
+    /// Spec is needed for two things SpuriousDragon's `EIP-161 State clear`,
+    /// and for Cancun's `EIP-6780: SELFDESTRUCT in same transaction`
     pub spec: SpecId,
     /// It is assumed that precompiles start from 0x1 address and span next N addresses.
     /// we are using that assumption here
@@ -460,9 +461,11 @@ impl JournaledState {
     }
 
     /// Transfer balance from address to target. Check if target exist/is_cold
-    /// NOTE: balance will be lost if [address] and [target] are the same BUT when
+    ///
+    /// Note: balance will be lost if [address] and [target] are the same BUT when
     /// current spec enables Cancun, this happens only when the account associated to [address]
     /// is created in the same tx
+    ///
     /// references:
     ///  * https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/vm/instructions.go#L832-L833
     ///  * https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/state/statedb.go#L449
@@ -488,10 +491,11 @@ impl JournaledState {
         let previously_destroyed = acc.is_selfdestructed();
         let is_cancun_enabled = SpecId::enabled(self.spec, CANCUN);
 
+        acc.info.balance = U256::ZERO;
+
         // EIP-6780 (Cancun hard-fork): selfdestruct only if contract is created in the same tx
         let journal_entry = if acc.is_created() || !is_cancun_enabled {
             acc.mark_selfdestruct();
-            acc.info.balance = U256::ZERO;
             JournalEntry::AccountDestroyed {
                 address,
                 target,
@@ -499,7 +503,6 @@ impl JournaledState {
                 had_balance: balance,
             }
         } else {
-            Self::touch_account(self.journal.last_mut().unwrap(), &address, acc);
             JournalEntry::BalanceTransfer {
                 from: address,
                 to: target,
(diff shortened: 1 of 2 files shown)


rakita:
Didn't forget about this, checking things related to Verkle tree.

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +459,55 @@ impl JournaledState {
...
+            acc.info.balance = U256::ZERO;
+            JournalEntry::AccountDestroyed {
                 address,
                 target,
                 was_destroyed: previously_destroyed,
                 had_balance: balance,
-            });
+            }
+        } else {
+            Self::touch_account(self.journal.last_mut().unwrap(), &address, acc);
```
rakita, line 502: This line is not needed, `address` is considered `touched` as this is the account that we currently executing.

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +459,55 @@ impl JournaledState {
...
+            JournalEntry::BalanceTransfer {
+                from: address,
+                to: target,
+                balance,
+            }
```
rakita, line 507: We are not decrementing the balance from `address`

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +459,55 @@ impl JournaledState {
...
-
-        // NOTE: In case that target and destroyed addresses are same, balance will be lost.
-        // ref: https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/vm/instructions.go#L832-L833
-        // https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/state/statedb.go#L449
-        if address != target {
-            let target_account = self.state.get_mut(&target).unwrap();
-            // touch target account
+
+        let acc = if address != target {
+            let [acc, target_account] = self.state.get_many_mut([&address, &target]).unwrap();
```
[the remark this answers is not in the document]
rakita: Oh forget this, `Hashbrown` has it

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +459,55 @@ impl JournaledState {
...
-            .push(JournalEntry::AccountDestroyed {
+        let balance = acc.info.balance;
+        let previously_destroyed = acc.is_selfdestructed();
+        let is_cancun_enabled = SpecId::enabled(self.spec, CANCUN);
+
+        // EIP-6780 (Cancun hard-fork): selfdestruct only if contract is created in the same tx
+        let journal_entry = if acc.is_created() || !is_cancun_enabled {
+            acc.mark_selfdestruct();
+            acc.info.balance = U256::ZERO;
+            JournalEntry::AccountDestroyed {
```
rakita, line 495: For JournalEntry, there is some work needed to support cancun balance change

rakita (approved):
good direction, left comments

rakita (changes requested):
Meant to request changes

crates/revm/src/journaled_state.rs
lorenzofero: Oof what a dumb error, thanks for noticing

crates/revm/src/journaled_state.rs
lorenzofero: I'm sorry for the questions but I don't understand what you mean: do we need another enum member for `JournaledState` to track the "cancun balance change"? Why is it needed?

rakita: Nothing to be done there, I thought that we needed to add the modifications for `address != target` as it has a different behaviour but that path is converted with the `JournalEntry::BalanceTransfer`.

lorenzofero: Ok! Thank you for the explanation!

lorenzofero:
I should have resolved all your comments, thanks for the support. I'm happy to see that tests are passing, even if in local they were failing for some reason..
Note that now if cancun is enabled and `address === target`, the `BalanceTransfer` entry in the JournledState is not pushed in since it is useless. As the EIP says, in such case there is no net change of balances. 
Let me know what you think!

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +460,61 @@ impl JournaledState {
...
-
-        // NOTE: In case that target and destroyed addresses are same, balance will be lost.
-        // ref: https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/vm/instructions.go#L832-L833
-        // https://github.com/ethereum/go-ethereum/blob/141cd425310b503c5678e674a8c3872cf46b7086/core/state/statedb.go#L449
-        if address != target {
-            let target_account = self.state.get_mut(&target).unwrap();
-            // touch target account
+
+        let acc = if address != target {
+            let [acc, target_account] = self.state.get_many_mut([&address, &target]).unwrap();
```
rakita, line 482:
```suggestion
            // Both accounts are loaded before this point, `address` as we execute its contract.
            // and `target` at the beginning of the function.
            let [acc, target_account] = self.state.get_many_mut([&address, &target]).unwrap();
```

crates/revm/src/journaled_state.rs
```diff
@@ -468,39 +460,61 @@ impl JournaledState {
...
+            })
+        } else if address != target {
+            acc.info.balance = U256::ZERO;
+            Some(JournalEntry::BalanceTransfer {
+                from: address,
+                to: target,
+                balance,
+            })
+        } else {
+            None
```
rakita, line 512:
```suggestion
            // State is not changed:
            // * if we are after Cancun upgrade
            // * Selfdestruct account that is created in the same transaction
            // * Specify the target to be the same as selfdestructed account. The balance stays unchanged.
            None
```

rakita (approved):
lgtm!
Left nit comments to add doc

rakita:
> I should have resolved all your comments, thanks for the support. I'm happy to see that tests are passing, even if in local they were failing for some reason.. Note that now if cancun is enabled and `address === target`, the `BalanceTransfer` entry in the JournledState is not pushed in since it is useless. As the EIP says, in such case there is no net change of balances. Let me know what you think!

Tests were failing for some other reason, found the problem and merged in on `main`, so we are good

Rjected (approved):
other than the current comments, this looks good to me, will test this with hive once merged!

PR #722, opened 2023-09-14
Document when InvalidTransaction errors are thrown

### LINKED ISSUE #594: Document when InvalidTransaction errors are thrown
error variants are undocumented which makes it very hard to reason about certain errors that could be introduced via env settings

https://github.com/bluealloy/revm/blob/bc4d203bf0b5f5c01868477c8e98d3abd6bb92ab/crates/primitives/src/result.rs#L156

it would be helpful to have some docs when these are thrown

alessandromazza98:
Closes #594 

I added some docs to the `InvalidTransaction` enum.

I just excluded 3 variants because their meaning are obvious:
- `NonceTooHigh`
- `NonceTooLow`
- `InvalidChainId`

I also renamed the first variant: `PriorityFeeGreaterThanMaxFee` that previously was `GasMaxFeeGreaterThanPriorityFee `.
I think it's better named like this as that error arises when the `priority_fee` is actually greater than the `max_fee` and not the contrary.

Let me know WDYT!

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -278,7 +278,7 @@ impl Env {
             if let Some(priority_fee) = self.tx.gas_priority_fee {
                 if priority_fee > self.tx.gas_price {
                     // or gas_max_fee for eip1559
-                    return Err(InvalidTransaction::GasMaxFeeGreaterThanPriorityFee);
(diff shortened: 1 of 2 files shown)


rakita:
Yeah this is correct it should be `PriorityFeeGreaterThanMaxFee`

Issue #685, opened 2023-09-05
Introduce error type for EnvError 


rakita:
https://github.com/bluealloy/revm/blob/c0767fa3e3b49497fa67cef37a9519fa6e55dd12/crates/primitives/src/result.rs#L127-L128C22

After eip4844 we would have two filed that should have been set if spec is Cancun. We can group them as a cleanup.

Do it after eip4844 PR is merged: https://github.com/bluealloy/revm/pull/668

hack3r-0m:
I would like to work on this, am i correct in understanding that EnvError will include `PrevrandaoNotSet` & `ExcessBlobGasNotSet` and would be would be variant of `EVMError` ?

PR #726, opened 2023-09-17
chore: rm nonexhaustive for error

mattsse:
having non-exhaustive on an error forces this clippy error:

```
126 | pub enum EVMError<DBError> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: the matched value is of type `EVMError<T>`
    = note: `EVMError<T>` is marked as non-exhaustive, so a wildcard `_` is necessary to match exhaustively
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
192 ~             EVMError::Database(err) => err.into(),
193 ~             _ => todo!(),
    |

```

this forces a wildcard handling which can introduce bugs if a new evmerror variant is added

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -122,7 +122,6 @@ impl Output {
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[non_exhaustive]
 pub enum EVMError<DBError> {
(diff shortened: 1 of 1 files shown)


rakita:
This error marks the transaction as Invalid, so I would not assume that it will break things if the subvariant is different.


Would move those two errors behind `EnvError`:  https://github.com/bluealloy/revm/issues/685 so `#[non_exhaustive]` is less impactful and would be okay to remove it.

PR #724, opened 2023-09-14
Feat: Reintroduce Alloy

Evalir:
Rebases #617 to be up to date with the newest changes, mainly concerning bundlestate.

PR #732, opened 2023-09-19
style: change type and impl order

mattsse:
changes the order of type def and impls. makes it easier to navigate the file

type
impl type
impl for type

env
cfgenv
blockenv
txenv

I can hold off on this until we have #724 to not introduce more conflicts than necessary?

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -5,363 +5,16 @@ use crate::{
 use bytes::Bytes;
 use core::cmp::{min, Ordering};
 
+/// EVM environment configuration.
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


rakita:
> changes the order of type def and impls. makes it easier to navigate the file
> 
> type impl type impl for type
> 
> env cfgenv blockenv txenv
> 
> I can hold off on this until we have #724 to not introduce more conflicts than necessary?

Yeah, this kind of style related changes are long overdue. Had the intention to do it in one swoop. Would wait for release to merge it, so that we don't break alloy transition a lot.

gakonst:
(failing CI fixed by https://github.com/bluealloy/revm/pull/735/)

mattsse:
closing since diverged too much

PR #733, opened 2023-09-19
fix: use CANCUN precompile id for CANCUN SpecId

Rjected:
ref https://github.com/paradigmxyz/reth/issues/4671

Hive tests discovered that the cancun precompile was not being set even though the input `SpecId` was `SpecId::CANCUN`. It was instead being mapped to `revm_precompile::SpecId::BERLIN`.

This maps `CANCUN` to `CANCUN`, causing the hive tests to pass.

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -215,9 +215,9 @@ pub fn to_precompile_id(spec_id: SpecId) -> revm_precompile::SpecId {
         | SpecId::ARROW_GLACIER
         | SpecId::GRAY_GLACIER
         | SpecId::MERGE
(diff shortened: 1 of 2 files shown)


Rjected:
Tests seem to be failing on master as well 🤔

rakita:
That is a footgun, can you remove that function from code and use:
https://github.com/bluealloy/revm/blob/70cf969a25a45e3bb4e503926297d61a90c7eec5/crates/precompile/src/lib.rs#L102-L114

Rjected:
> That is a footgun, can you remove that function from code and use:

just removed!

PR #734, opened 2023-09-20
fix(test): Check expect exception and revm error

rakita:
Explanation can be found in the code but tldr:
 If there is a tx exception state root is calculated differently for pre state clear hardforks. The solution, don't check the hash if revm returns an error and an exception is set in the test.


DaniPopes:
I guess we can remove the 4844 tests we ignored before with this?

rakita:
@DaniPopes make PR for eip4844: https://github.com/bluealloy/revm/pull/735

PR #735, opened 2023-09-20
fix(eip4844): Pass eth tests, additional conditions added.

rakita:
Added failing condition to `Env`, now we pass all tests related to blobs

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -485,7 +485,7 @@ impl Env {
                 // all versioned blob hashes must start with VERSIONED_HASH_VERSION_KZG
                 for blob in self.tx.blob_hashes.iter() {
                     if blob[0] != VERSIONED_HASH_VERSION_KZG {
-                        return Err(InvalidTransaction::BlobVersionedHashesNotSupported);
+                        return Err(InvalidTransaction::BlobVersionNotSupported);
                     }
                 }
 
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -1,6 +1,7 @@
 use crate::{
     alloc::vec::Vec, calc_blob_fee, Account, EVMError, InvalidTransaction, Spec, SpecId, B160,
-    B256, GAS_PER_BLOB, KECCAK_EMPTY, MAX_INITCODE_SIZE, U256,
+    B256, GAS_PER_BLOB, KECCAK_EMPTY, MAX_BLOB_NUMBER_PER_BLOCK, MAX_INITCODE_SIZE, U256,
+    VERSIONED_HASH_VERSION_KZG,
 };
 use bytes::Bytes;
 use core::cmp::{min, Ordering};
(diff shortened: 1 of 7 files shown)


crates/primitives/src/result.rs
```diff
@@ -188,6 +189,13 @@ pub enum InvalidTransaction {
...
     /// Block `blob_gas_price` is greater than tx-specified `max_fee_per_blob_gas` after Cancun.
     BlobGasPriceGreaterThanMax,
+    /// There should be at least one blob in Blob transaction.
+    EmptyBlobs,
+    /// Blob transaction can't be a create transaction.
+    /// `to` must be present
+    BlobCreateTransaction,
+    /// Transaction has more then [`crate::MAX_BLOB_NUMBER_PER_BLOCK`] blobs
+    TooManyBlobs,
 }
```
Rjected, line 199:
```suggestion
    /// Blob transaction contains a versioned hash with an incorrect version 
    BlobMismatchedVersion,
}
```

DaniPopes (approved):

crates/primitives/src/env.rs
```diff
@@ -458,11 +460,40 @@ impl Env {
...
+
+                // all versioned blob hashes must start with VERSIONED_HASH_VERSION_KZG
+                for blob in self.tx.blob_hashes.iter() {
+                    if blob[0] != VERSIONED_HASH_VERSION_KZG {
+                        return Err(InvalidTransaction::BlobVersionedHashesNotSupported);
+                    }
+                }
+
+                // ensure the total blob gas spent is at most equal to the limit
+                // assert blob_gas_used <= MAX_BLOB_GAS_PER_BLOCK
```
Rjected, line 493: ah, I guess this is the right part of the spec to put as a comment here

Rjected (approved):
lgtm other than the suggestions

crates/primitives/src/env.rs
```diff
@@ -458,11 +460,40 @@ impl Env {
...
+                // a 20-byte address. This means that blob transactions cannot
+                // have the form of a create transaction.
+                if self.tx.transact_to.is_create() {
+                    return Err(InvalidTransaction::BlobCreateTransaction);
+                }
+
+                // all versioned blob hashes must start with VERSIONED_HASH_VERSION_KZG
+                for blob in self.tx.blob_hashes.iter() {
+                    if blob[0] != VERSIONED_HASH_VERSION_KZG {
+                        return Err(InvalidTransaction::BlobVersionedHashesNotSupported);
```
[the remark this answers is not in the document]
rakita: it is different error, one is related to precompile input and second one is related to transaction blobs.

crates/primitives/src/env.rs
rakita: Yep, added a few constants but max blob gas is calculated from blob number, and blob number is cleaner way to specify constraint

crates/primitives/src/result.rs
rakita: BlobMismatchedVersion is related to precompile. I sow that `BlobVersionedHashesNotSupported` was not used, probably we missed it so I used it to report not supported blob hash version

crates/primitives/src/env.rs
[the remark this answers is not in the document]
Rjected: yeah, I'm mainly just suggesting a separate error variant similar to the precompile one. In reth, especially in rpc responses, it would be nice to distinguish between `BlobVersionedHashesNotSupported` and another error type which indicates that a versioned hash has an invalid version

crates/primitives/src/result.rs
Rjected: Ah I see, if unused elsewhere then maybe it could just be renamed. my other comment assumes it was used

rakita: I apologize this was not correct, will add additional field

rakita: Renamed it as `BlobVersionNotSupported`

PR #738, opened 2023-09-21
Move eip4844 points build script to helper fn

### LINKED ISSUE #686: Move eip4844 points build script to helper fn
Discussion: https://github.com/bluealloy/revm/pull/668#discussion_r1313310505

make build script as a helper function inside revm that would load `trusted_setup.txt` and output it to file and load file and create the point.

The function for creating the file can be run from revme as a command to help with the creation of the file/points.

alessandromazza98:
Closes #686 

I moved the build script to a helper CLI function named: `generate-kzg-points` where you pass the `path` of the trusted setup and it generates the `.bin` files related to `g1` and `g2` points in the folder you are when launching that command (I used `env::current_dir()` for that).

 I only changed all the `unwraps()` or `assert!(...)` in order to not make the program panic but instead return errors.


rakita:
Superseded it with this PR: https://github.com/bluealloy/revm/pull/818

PR #739, opened 2023-09-21
perf(interpreter): remove dynamic dispatch from all instructions

DaniPopes:
Snailtracer (I): 442,867,225 -> 431,363,903 (-2.6%)

Mostly generated with the following regex search and replace:

```
fn (\w+)(?:<([\s\S]*?)>)?\(([\n\s\S]*?)&mut dyn Host
fn $1<H: Host, $2>($3&mut H
```

Supersedes #660, all functions must drop the `dyn Host` argument to make a difference


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -52,17 +53,17 @@ macro_rules! opcodes {
...
 
         // in `eval`:
-        (const { make_instruction_table::<SPEC>() })[opcode as usize](interpreter, host)
+        (const { make_instruction_table::<H, SPEC>() })[opcode as usize](interpreter, host)
         */
 
         /// Evaluates the opcode in the given context.
         #[inline(always)]
-        pub(crate) fn eval<SPEC: Spec>(opcode: u8, interpreter: &mut Interpreter, host: &mut dyn Host) {
+        pub(crate) fn eval<H: Host, SPEC: Spec>(opcode: u8, interpreter: &mut Interpreter, host: &mut H) {
```
rakita, line 61:
```suggestion
        pub fn eval<H: Host, SPEC: Spec>(opcode: u8, interpreter: &mut Interpreter, host: &mut H) {
```

Not related but would be good to be `public`

PR #741, opened 2023-09-22
docs: document everything, dedup existing docs

DaniPopes:
- Not actually everything, just adding one liner descriptions where I could.
- Adding `#[doc(inline)]` duplicates all docs so one has to be careful using it (#560)
- Having both `pub mod x;` and `pub use x::*;` is redundant and creates 2 paths from where an item can be imported, so I opted to keep the glob export.


crates/interpreter/src/interpreter/memory.rs
```diff
@@ -183,17 +179,9 @@ impl Memory {
     }
 }
 
-/// Rounds up `x` to the closest multiple of 32. If `x % 32 == 0` then `x` is returned.
-#[inline]
-pub(crate) fn next_multiple_of_32(x: usize) -> Option<usize> {
```
DaniPopes, line 188: Moved to `revm_primitives`

crates/precompile/Cargo.toml
```diff
@@ -1,16 +1,13 @@
...
+description = "Implementations of EVM precompiled contracts"
 edition = "2021"
 keywords = ["no_std", "ethereum", "evm", "revm", "precompiles"]
 license = "MIT"
 name = "revm-precompile"
 repository = "https://github.com/bluealloy/revm"
 version = "2.0.3"
 
-# Don't need to run build script outside of this repo
-exclude = ["build.rs", "src/blob/kzg_settings/*.txt"]
```
DaniPopes, line 12: This was moved to `primitives`

crates/revm/src/inspector.rs
```diff
@@ -47,70 +46,87 @@ pub trait Inspector<DB: Database> {
...
+    #[inline]
     fn call(
         &mut self,
-        _data: &mut EVMData<'_, DB>,
-        _inputs: &mut CallInputs,
+        data: &mut EVMData<'_, DB>,
+        inputs: &mut CallInputs,
     ) -> (InstructionResult, Gas, Bytes) {
+        let _ = data;
+        let _ = inputs;
```
mattsse, line 97: hmm, what's the reason for that?

crates/revm/src/inspector.rs
```diff
@@ -47,70 +46,87 @@ pub trait Inspector<DB: Database> {
...
-        _data: &Bytes,
+        evm_data: &mut EVMData<'_, DB>,
+        address: &B160,
+        topics: &[B256],
+        data: &Bytes,
     ) {
+        let _ = evm_data;
+        let _ = address;
+        let _ = topics;
+        let _ = data;
```
mattsse, line 68: same

mattsse (changes requested):
love the additional docs,

what's the reason for dropping the `_` from the unused references?

crates/revm/src/inspector.rs
DaniPopes: Mainly trait docs and editor code actions, when you auto complete a trait function it copies the same formatting, names etc as the definition, at least on vscode

DaniPopes: Before
image

After
image

DaniPopes: It's a very minor thing, idm if you want it the other way

mattsse: I see, TIL

I actually don't mind it then, defer to @rakita

mattsse (approved):

rakita:
As this one is a bigger PR, would leave it after the release and alloy merge.

Evalir (approved):
this is probably mergeable now!

crates/interpreter/src/interpreter/memory.rs
rakita: Related only to Interpreter so it was left there, does not make sense imo to move it to primitives

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -1,13 +1,10 @@
 use crate::{
-    primitives::{B256, U256},
+    primitives::{B256, STACK_LIMIT, U256},
     InstructionResult,
 };
 use alloc::vec::Vec;
 use core::fmt;
 
-/// The EVM stack limit, in number of items.
-pub const STACK_LIMIT: usize = 1024;
```
rakita, line 9: Related to the stack, soo imo `stack.rs` is a good place for it.

crates/interpreter/src/lib.rs
```diff
@@ -7,23 +11,22 @@ mod macros;
...
+mod instruction_result;
 pub mod instructions;
 mod interpreter;
 
 pub(crate) const USE_GAS: bool = !cfg!(feature = "no_gas_measuring");
 
 // Reexport primary types.
 pub use gas::Gas;
-pub use host::{DummyHost, Host};
+pub use host::*;
```
rakita, line 23: Intend to move all the things to name inclusion, `*` hides what is inside and was good for developments but specified names give better clarity to what is exported.

crates/precompile/Cargo.toml
```diff
@@ -1,16 +1,13 @@
 [package]
 authors = ["Dragan Rakita <dragan0rakita@gmail.com>"]
-description = "REVM Precompiles - Ethereum compatible precompiled contracts"
+description = "Implementations of EVM precompiled contracts"
```
rakita, line 3: Can you revert this

crates/revm/src/inspector.rs
rakita: Looks strange, but the outcome is better, i like it

crates/interpreter/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 authors = ["Dragan Rakita <dragan0rakita@gmail.com>"]
-description = "REVM Interpreter"
+description = "EVM bytecode interpreter"
```
rakita, line 3: Would be good to leave this

DaniPopes:
@rakita I moved the constants and util to primitives because there were some duplicates, I think they make more sense since it has a constants and utils module and I felt they don't really belong in random modules of other crates. WDYT?

rakita:
> @rakita I moved the constants and util to primitives because there were some duplicates, I think they make more sense since it has a constants and utils module and I felt they don't really belong in random modules of other crates. WDYT?

If we had only one lib sure, if maybe there is an intention to be used somewhere else sure.
but these are very specific constants and they are not put in the random modules but in the module that uses them.

Duplicates should ofcourse be removed.

PR #746, opened 2023-09-25
feat: add "kzg" as a separate feature

DaniPopes:
Separate kzg from std feature

### DIFF
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -46,7 +46,7 @@ std = [
     "secp256k1?/std",
 ]
 
-# The following features may not work on all platforms as they depends on C libraries.
+# The following features may not work on all platforms as they depend on C libraries.
 # Enables the KZG point evaluation precompile.
 c-kzg = ["dep:c-kzg", "revm-primitives/c-kzg"]
 
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -33,7 +33,7 @@ hex = "0.4"
 hex = "0.4"
 
 [features]
-default = ["std", "kzg", "secp256k1"]
+default = ["std", "c-kzg", "secp256k1"]
 std = [
     "revm-primitives/std",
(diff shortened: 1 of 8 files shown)


crates/revm/Cargo.toml
```diff
@@ -39,6 +39,12 @@ criterion = "0.5"
 
 [features]
 default = ["std", "secp256k1"]
```
Evalir, line 41: should `kzg` be turned on by default?

crates/primitives/Cargo.toml
```diff
@@ -86,6 +71,7 @@ serde = [
     "bytes/serde",
     "bitvec/serde",
     "bitflags/serde",
+    "c-kzg?/serde",
```
Evalir, line 74: TIL!

Evalir (approved):
nice! this should unblock snafus regarding building the foundry docker img

rakita:
> Separate kzg from std feature

Supporting, would just rename feature to `ckzg`, to tie the name with c-kzg lib.

mattsse (approved):
lgtm

crates/precompile/Cargo.toml
```diff
@@ -33,11 +33,22 @@ hex = "0.4"
...
+    "sha3/std",
+    "c-kzg?/std",
+    "secp256k1?/std",
+]
+
+# The following features may not work on all platforms as they depend on C libraries.
+# Enables the KZG point evaluation precompile.
+c-kzg = ["dep:c-kzg", "revm-primitives/c-kzg"]
+
+# Use `secp256k1` as a faster alternative to `k256`.
```
rakita, line 53:
```suggestion
# Use `secp256k1` as a faster alternative to `k256`.
# The problem that `secp256k1` has, is it fails to build for `wasm` target on Windows and Mac as it is c lib.
# In Linux it passes. If you don't require to build wasm on win/mac, it is safe to use it and it is enabled by default.
```

PR #748, opened 2023-09-25
feat(`deps`): Make `c-kzg` optional

Evalir:
There are some `c-kzg` binding snafus which mess with [foundry's docker build process](https://github.com/foundry-rs/foundry/actions/runs/6302981610/job/17111278694). While this gets solved, making `c-kzg` an optional dep/feature that can be enabled can also solve this issue.


Evalir:
nvm, #746 :D

PR #750, opened 2023-09-26
ci: concurrency for github actions

PaulRBerg:
The goal is to ensure that only a single job or workflow using the same concurrency group will run at a time.

E.g. if someone force pushes, CI won't run twice. The first workflow run will be stopped.


rakita (approved):
nice!

PR #751, opened 2023-09-26
fix: balance check disabled

Wodann:
Fixes the `disable_balance_check` option by setting the sender balance to the balance to be spent.


rakita (approved):
makes sense

PR #754, opened 2023-09-27
refactor: say "warm" instead of "hot"

PaulRBerg:
[EIP-2919](https://eips.ethereum.org/EIPS/eip-2929) does not make any reference to "hot". It only says "warm".

There's no need to introduce another term to refer to the same concept.

There are other references on the Internet, such as [EVM codes](https://www.evm.codes/about#accesssets), which also use only the "warm" terminology.


rakita (approved):
true!

PR #757, opened 2023-09-28
fix: use u128 for calc data fee result

Rjected:
A hive test runs with these parameters:
```
self.block.get_blob_gasprice = Some(136505023684851)
self.tx.get_total_blob_gas = 524288

multiply to get data fee
data fee returned: 16227713636554506240
data fee SHOULD be 71567945857683161088
```
The data fee overflowed in this case, causing an invalid state root in the test. This fixes the `calc_data_fee` method to compute a u128.


DaniPopes (approved):

crates/primitives/src/env.rs
```diff
@@ -424,10 +424,10 @@ impl Env {
     ///
     /// [EIP-4844]: https://eips.ethereum.org/EIPS/eip-4844
     #[inline]
-    pub fn calc_data_fee(&self) -> Option<u64> {
+    pub fn calc_data_fee(&self) -> Option<u128> {
```
rakita, line 427:
```suggestion
    pub fn calc_data_fee(&self) -> Option<U256> {
```

revm balance is represented as U256 so it would be good to return same type.

crates/primitives/src/env.rs
```diff
@@ -424,10 +424,10 @@ impl Env {
...
         self.block
             .get_blob_gasprice()
-            .map(|blob_gas_price| blob_gas_price * self.tx.get_total_blob_gas())
+            .map(|blob_gas_price| blob_gas_price as u128 * self.tx.get_total_blob_gas() as u128)
```
rakita, line 430:
```suggestion
            .map(|blob_gas_price| U256::from(blob_gas_price).saturating_mul(U256::from(self.tx.get_total_blob_gas()))
```

PR #759, opened 2023-09-28
Instruction table

rakita:
Put all instructions in the table and allow the table to be initialized. This is the first step towards introduction of custom opcodes.

Additionally, specify a Boxed instruction that is a little bit less performant (10-20%) but allows the use of closure over instructions.

An example of boxed instruction is used for `Inspector`, this is a step forward in having custom handles for some or all instructions.

The plain instruction table had the same performance as the previous eval fn that used `match` (0.01% invariance with cachegrind)

Removed generic of `INSTRUCTION` as benefits are neglectable, and code clarity is increased. Now EVMImpl has `Option` over instruction which is a lot nicer.

Removed `step` and `step_end` from `Host` trait as closure over instruction allows us the same behaviour.


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -829,6 +847,16 @@ const fn make_gas_table(spec: SpecId) -> [OpInfo; 256] {
     table
 }
 
+pub fn make_instruction_table<H: Host, SPEC: Spec>() -> [Instruction<H>; 256] {
```
DaniPopes, line 850: I think this and `instruction` can be made `const`

DaniPopes (approved):

crates/interpreter/src/instructions/opcode.rs
rakita: Tried it and the problem with `Host` error is something like "cant use parent generic" or something in that tone.

PR #760, opened 2023-09-28
docs: document non-zero amounts in State::increment_balances

Rjected:
Thought this would be good to add after debugging https://github.com/paradigmxyz/reth/pull/4840. This documents the same assumptions in `CacheAccount::increment_balance`, but in `State::increment_balances` as well. A `debug_assert` is added in `increment_balance`, to cause it to panic if the assumption is violated in tests.

### DIFF
--- a/crates/revm/src/db/states/cache_account.rs
+++ b/crates/revm/src/db/states/cache_account.rs
@@ -296,6 +296,7 @@ impl CacheAccount {
     /// Note: to skip some edge cases we assume that additional balance is never zero.
     /// And as increment is always related to block fee/reward and withdrawals this is correct.
     pub fn increment_balance(&mut self, balance: u128) -> TransitionAccount {
+        debug_assert!(balance > 0, "Balance is assumed to be positive");
         self.account_info_change(|info| {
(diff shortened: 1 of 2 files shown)


rakita:
> Thought this would be good to add after debugging [paradigmxyz/reth#4840](https://github.com/paradigmxyz/reth/pull/4840). This documents the same assumptions in `CacheAccount::increment_balance`, but in `State::increment_balances` as well. A `debug_assert` is added in `increment_balance`, to cause it to panic if the assumption is violated in tests.

Added a commit that would skip zero balance increment.

rakita (approved):

Rjected:
> > Thought this would be good to add after debugging [paradigmxyz/reth#4840](https://github.com/paradigmxyz/reth/pull/4840). This documents the same assumptions in `CacheAccount::increment_balance`, but in `State::increment_balances` as well. A `debug_assert` is added in `increment_balance`, to cause it to panic if the assumption is violated in tests.
> 
> Added a commit that would skip zero balance increment. If you are okay with it, we can merge it

Yep, that should work!

Issue #740, opened 2023-09-22
Document and implement formatting for errors


DaniPopes:
`InvalidTransaction`, `InvalidHeader` etc (`crates/primitives/src/result.rs`) do not implement `std::fmt::Display` or `std::error::Error`, which forces downstream crates to implement their own formatting/conversions (see [Foundry](https://github.com/foundry-rs/foundry/blob/adb526bc1db52db96ba6c69be178022f97e76ec5/crates/anvil/src/eth/error.rs#L203), [Reth](https://github.com/paradigmxyz/reth/blob/e83d3aa704f87825ca8cab6f593ab4d4adbf6792/crates/rpc/rpc/src/eth/error.rs#L371))

rakita:
done

PR #770, opened 2023-10-05
fix(primitives): Error Primitive Display + Error Implementations

refcell:
**Description**

Addresses #740 

Implements `std::error::Error + std::fmt::Display` for `InvalidTransaction` and `InvalidHeader`. If there are others that should be implemented, happy to do in this pr cc @DaniPopes

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -217,6 +217,68 @@ pub enum InvalidTransaction {
     DepositSystemTxPostRegolith,
 }
 
+impl std::error::Error for InvalidTransaction {}
+
+impl fmt::Display for InvalidTransaction {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            InvalidTransaction::PriorityFeeGreaterThanMaxFee => {
+                write!(f, "Priority fee is greater than max fee")
+            }
+            InvalidTransaction::GasPriceLessThanBasefee => {
+                write!(f, "Gas price is less than basefee")
+            }
+            InvalidTransaction::CallerGasLimitMoreThanBlock => {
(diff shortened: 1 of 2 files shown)


crates/primitives/src/result.rs
```diff
@@ -217,6 +218,68 @@ pub enum InvalidTransaction {
     DepositSystemTxPostRegolith,
 }
 
+impl error::Error for InvalidTransaction {}
```
rakita, line 221: You need to feature gate this with: `#[cfg(feature = "std")]` and use `std::error::Error` as `core::error::Error` is still experimental

https://doc.rust-lang.org/std/error/trait.Error.html

crates/primitives/src/result.rs
```diff
@@ -217,6 +218,68 @@ pub enum InvalidTransaction {
...
+                write!(f, "Lack of funds for max fee",)
+            }
+            InvalidTransaction::OverflowPaymentInTransaction => {
+                write!(f, "Overflow payment in transaction")
+            }
+            InvalidTransaction::NonceOverflowInTransaction => {
+                write!(f, "Nonce overflow in transaction")
+            }
+            InvalidTransaction::NonceTooHigh { .. } => write!(f, "Nonce too high"),
+            InvalidTransaction::NonceTooLow { .. } => write!(f, "Nonce too low"),
```
rakita, line 251: Would be good to include values from NonceTooLow inside `write!`.

you can just append it to macro:
https://doc.rust-lang.org/beta/core/macro.write.html

refcell: Ok added - wanted to avoid excess verbosity but I think this is very acceptable

rakita (approved):

PR #774, opened 2023-10-06
feat(revm): implement DatabaseRef trait for EthersDB

0xkr8os:
Provides a concrete implementation for the `DatabaseRef` trait instead of the `Database` trait. Implementation for the Database trait is simply a wrapper.


rakita (approved):
nice addon

PR #775, opened 2023-10-06
fix(primitives): Use Upstream Alloy Create Implementations

refcell:
**Description**

Since `create` and `create2_address` are used internally and external implementations can just use the methods directly on the re-exported `Address` type, this PR removes the utility methods and inlines the upstream alloy creation function calls.


rakita (approved):
Was left there to not have a lot of breaking changed in `3.5`. Fine to remove it now!

PR #783, opened 2023-10-09
use `asRef<Bytecode>` in fn contract in BundleBuilder

### LINKED ISSUE #782: `fn contract` in `bundle_state.rs` should accept `AsRef`
Related to https://github.com/paradigmxyz/reth/pull/4647

`contract` function in `BundleBuilder` should accept `AsRef<Bytecode>` so that in Reth it could be possible to directly use the function in this way:

```rust
bundle_builder.contract(address, bytecode);
```

and not being forced to do:
```rust
bundle_builder.contract(address, bytecode.0);
```

alessandromazza98:
Closes #782

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -141,8 +141,8 @@ impl BundleBuilder {
     }
 
     /// Collect contracts info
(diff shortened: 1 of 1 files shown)


rakita:
This is really strange to introduce

Issue #781, opened 2023-10-06
Optimism execution


Wollac:
When I tried to use the Optimism execution behind the `optimism` feature flag of version `v3.5.0`, I ran into the following problems:
- `Handler::optimism()` is never used. So even if `env.cfg.optimism` is `true`, the Optimism specific fee handling is not applied.
- The _Base Fee Vault_ calculation in the Optimism handler uses the L1 base, but according to the [L2 Execution Engine spec](https://github.com/ethereum-optimism/optimism/blob/develop/specs/exec-engine.md#base-fees-base-fee-vault), I was under the impression that `env.block.basefee * (gas.spend() - gas_refund)` should be used instead.
- When the `optimism` feature is enabled while  `env.cfg.optimism` is `false` for Ethereum transactions, the `tx_l1_cost` is still computed, which can lead to a panic as obviously the `enveloped_tx` is not provided in this case.

If the above observations are indeed correct and it would be helpful, I can provide a PR for these issues.

refcell:
Thank you for reporting this! Will take a look, reproduce, and get back to you!

As a side note, if you make a PR, it would be preferable to split it up to make reviewing and adding the relevant unit tests easier.

rakita:
@Wollac thank you!

`Handler::optimism()` should have been used if optimism flag is true here:
https://github.com/bluealloy/revm/blob/02d85ac751b79cd37d06627c1b5281a1d7f5b61d/crates/revm/src/evm_impl.rs#L418

For the future, the idea is to create the `RevmBuilder` that would allow setting either of the handler (Still need to flush some thing for it), but for now, just changing the code to `::optimism()` should do the work.

The Second point I will leave for @refcell, for the third point we definitely should not panic.

refcell:
Left a couple of nits, but besides those, you're pr looks great! Really exciting that you were able to test this against Optimism Blocks!!

On the topic of unit tests, I think pushing those into other prs is totally reasonable.

clabby:
Very exciting news that you guys are successfully executing Optimism blocks! Thanks a ton for fixing these up :)

Wollac:
Fixed with #789

PR #789, opened 2023-10-10
fix: Optimism execution

Wollac:
This PR fixes the issues with the Optimism execution described in #781:
- `Handler::optimism()` is used instead of `Handler::mainnet()` when the `optimism` feature is enabled and `env.cfg.optimism` is `true`.
- The base fee of a transaction is now calculated using  `env.block.basefee * (gas.spend() - gas_refund)`.
- The panic when executing Ethereum transactions has been fixed by calculating `tx_l1_cost` only when `env.cfg.optimism` is `true` *and* the corresponding transaction is not an Optimism deposit transaction. This supports regular Ethereum transactions when `env.cfg.optimism` is `false` and `env.tx.optimism.enveloped_tx` being `None` for non-deposit transactions otherwise.


crates/revm/src/optimism.rs
```diff
@@ -83,19 +77,16 @@ impl L1BlockInfo {
...
     /// Calculate the gas cost of a transaction based on L1 block data posted on L2
-    pub fn calculate_tx_l1_cost<SPEC: Spec>(&self, input: &Bytes, is_deposit: bool) -> U256 {
-        let rollup_data_gas_cost = self.data_gas::<SPEC>(input);
-
-        if is_deposit || rollup_data_gas_cost == U256::ZERO {
-            return U256::ZERO;
-        }
+    pub fn calculate_tx_l1_cost<SPEC: Spec>(&self, input: &Bytes) -> U256 {
+        // input must not be an deposit transaction
+        debug_assert!(!input.is_empty() && input[0] != 0x7E);
```
refcell, line 82: Can we just return `U256::ZERO` if either of these two conditions are met? This makes the function more abstract as opposed to it only being used for non-deposit, non-empty transactions.

crates/revm/src/optimism.rs
```diff
@@ -83,19 +77,16 @@ impl L1BlockInfo {
...
 
+        let rollup_data_gas_cost = self.data_gas::<SPEC>(input);
         rollup_data_gas_cost
             .saturating_add(self.l1_fee_overhead)
             .saturating_mul(self.l1_base_fee)
             .saturating_mul(self.l1_fee_scalar)
-            .checked_div(U256::from(1_000_000))
-            .unwrap_or_default()
+            / U256::from(1_000_000)
```
refcell, line 89: Would prefer if we kept this as a checked div just in case.

crates/revm/src/optimism.rs
```diff
@@ -160,17 +151,8 @@ mod tests {
...
-        assert_eq!(gas_cost, U256::ZERO);
-
-        let gas_cost = l1_block_info.calculate_tx_l1_cost::<RegolithSpec>(&input, false);
+        let gas_cost = l1_block_info.calculate_tx_l1_cost::<RegolithSpec>(&input);
         assert_eq!(gas_cost, U256::from(1048));
-
-        // Zero rollup data gas cost should result in zero for non-deposits
-        let input = bytes!("");
-        let gas_cost = l1_block_info.calculate_tx_l1_cost::<RegolithSpec>(&input, false);
-        assert_eq!(gas_cost, U256::ZERO);
```
refcell, line 174: It's not explicitly undefined behavior to pass empty payloads into the `calculate_tx_l1_cost` function so I lean towards having the function return `U256::ZERO` if the tx is a deposit or the bytes are empty.

Then we can add these two test cases: input with empty bytes and input with the first byte as `0x7E`.

crates/revm/src/optimism.rs
Wollac: `checked_div` only checks if the divisor is `0`. Since it is always hardcoded to one million, I don't think checked div would make any difference.
However, checked operations might make sense before multiplication and addition, as there should be no overflows there.

crates/revm/src/optimism.rs
Wollac: done

rakita (approved):
lgtm, waiting for you approval from @refcell @clabby

clabby (approved):
LGTM, thanks!

PR #794, opened 2023-10-11
feat(interpreter): add more helper methods to memory

DaniPopes:
These would be useful in Foundry.
Second commit reduces usage of inline always since it's almost always unnecessary, and sometimes can hinder performance when normal inline hint is enough.
In release mode in our case performance is the same.


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -151,30 +147,59 @@ impl SharedMemory {
...
-            Some(b) => *b = byte,
-            None => debug_unreachable!("set_byte OOB: {index}; len: {}", self.len()),
-        }
+    pub fn set_byte(&mut self, offset: usize, byte: u8) {
+        self.set(offset, &[byte]);
+    }
+
+    /// Sets the given 32-byte `value` to the memory region at the given `offset`.
+    ///
+    /// Panics on out of bounds.
```
rakita, line 192:
```suggestion
    /// # Panics
    ///
    /// Panics on out of bounds.
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -151,30 +147,59 @@ impl SharedMemory {
...
+    #[inline]
+    pub fn get_u256(&self, offset: usize) -> U256 {
+        self.get_word(offset).into()
+    }
+
     /// Sets the `byte` at the given `index`.
     ///
-    /// Panics when `index` is out of bounds.
-    #[inline(always)]
+    /// Panics on out of bounds.
```
rakita, line 183:
```suggestion
    /// # Panics
    ///
    /// Panics on out of bounds.
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -151,30 +147,59 @@ impl SharedMemory {
...
+    ///
+    /// Panics on out of bounds.
+    #[inline]
+    pub fn get_word(&self, offset: usize) -> B256 {
+        self.slice(offset, 32).try_into().unwrap()
+    }
+
+    /// Returns a U256 of the memory region at the given offset.
+    ///
+    /// Panics on out of bounds.
```
rakita, line 175:
```suggestion
    /// # Panics
    ///
    /// Panics on out of bounds.
```

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -151,30 +147,59 @@ impl SharedMemory {
...
+    #[inline]
+    pub fn get_byte(&self, offset: usize) -> u8 {
+        self.slice(offset, 1)[0]
+    }
+
+    /// Returns a 32-byte slice of the memory region at the given offset.
+    ///
+    /// Panics on out of bounds.
```
rakita, line 167:
```suggestion
    /// # Panics
    ///
    /// Panics on out of bounds.
```

rakita:
Jsut to talk about this a little:
General topic of inlining: https://matklad.github.io/2021/07/09/inline-in-rust.html

https://indico.cern.ch/event/386232/sessions/159923/attachments/771039/1057534/always_inline_performance.pdf
This slide from 2015 says that: 
for Intel compiler `inline is faster than others modifiers (or combination), On the worst case, inline is similar to always_inline`
For  GCC `always_inline with 10M is usually faster than other modifiers, in worst case, always_inline is similar to inline`

We do have `LTO` set to `true`.

It seems that the thing with inlining is to always measure

rakita (approved):

PR #796, opened 2023-10-12
chore: make memory-limit private

mattsse:
make memory limit field private and reorder impl order, adds current len to debug impl


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -250,6 +233,24 @@ impl SharedMemory {
     }
 }
 
+impl fmt::Debug for SharedMemory {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SharedMemory")
+            .field("current_len", &self.current_len)
+            .field(
+                "current_slice",
```
lorenzofero, line 241: Sorry, I forgot to change it when renaming the field
```suggestion
                "context_memory",
```

rakita (approved):

PR #800, opened 2023-10-15
fix: unused `step` and `step_end` results

makcandrov:
This PR fixes a bug introduced by #759:
- The result of `step` is read but not written to the interpreter result. If the function returns something other than `Continue`, it will lead to an infinite loop in the interpreter.
- The result of `step_end` is not being read at all, which means the execution cannot be halted by the inspector from this function.

### DIFF
--- a/crates/revm/src/inspector/instruction.rs
+++ b/crates/revm/src/inspector/instruction.rs
@@ -21,7 +21,9 @@ pub fn inspector_instruction<'a, SPEC: Spec + 'static, DB: Database>(
                 interpreter.instruction_pointer = interpreter.instruction_pointer.sub(1);
             }
             if let Some(inspector) = host.inspector.as_mut() {
-                if inspector.step(interpreter, data) != InstructionResult::Continue {
+                let result = inspector.step(interpreter, data);
+                if result != InstructionResult::Continue {
+                    interpreter.instruction_result = result;
                     return;
                 }
             }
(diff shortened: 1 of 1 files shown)


rakita:
Hey, I wanted to remove the return type from `step` and `step_end` as you can set that value directly inside `Inspector` `interpreter.instruction_result = result`.

Would be interested to get your feedback on that approach?

makcandrov:
Yes, I thought about it too - it definitely makes sense.

However, the current implementation is a confusing mix between the two and silently breaks some inspectors. That's why I suggested reverting to the original behavior.

If you want to take a look, I tried implementing the new behavior [here](https://github.com/bluealloy/revm/pull/804) (I wanted to open the PR in my fork it ended up in the main repo - sorry)

rakita:
> Yes, I thought about it too - it definitely makes sense.
> 
> However, the current implementation is a confusing mix between the two and silently breaks some inspectors. That's why I suggested reverting to the original behavior.
> 
> If you want to take a look, I tried implementing the new behavior [here](https://github.com/bluealloy/revm/pull/804) (I wanted to open the PR in my fork it ended up in the main repo - sorry)

Yeah, the current `main` approach is definitely a footgun. 

I like this other PR more: https://github.com/bluealloy/revm/pull/804 as it does not reuse the `InstructionResult` as returns (that felt bad in past, it should have been enum of some kind) and it is more straightforward on what changes inside aka `interpreter.instruction_result`.

makcandrov:
Yes, you can close this pull request and merge the other one then

makcandrov:
Closing in favor of #804

PR #804, opened 2023-10-16
refactor: remove `step` and `step_end` return result

makcandrov:
Alternative implementation of #800

### DIFF
--- a/crates/revm/src/inspector.rs
+++ b/crates/revm/src/inspector.rs
@@ -27,17 +27,12 @@ pub mod inspectors {
 pub trait Inspector<DB: Database> {
     /// Called before the interpreter is initialized.
     ///
-    /// If anything other than [InstructionResult::Continue] is returned then execution of the interpreter is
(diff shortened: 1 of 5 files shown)


rakita (approved):

makcandrov:
Just a little remark on this implementation: it is not entirely equivalent to the original one, as an inspector can now override a failed opcode execution in 'step_end' to resume the execution. The original implementation could only override a successful execution to a failed execution.

makcandrov:
> wdym? shouldn't it have the same behaviour:

I was talking about `step_end`:

https://github.com/bluealloy/revm/blob/d320ff8587d3c410f3803c64eaa3cb1525b23f80/crates/interpreter/src/interpreter.rs#L135-L143

In the original implementation, if `self.instruction_result` was `InstructionResult::Revert`, even if the inspector returned `InstructionResult::Continue`, the execution would halt, because `self.instruction_result` was returned without being updated.

PR #805, opened 2023-10-16
chore: Make Instruction fn const

rakita:
After merging [this PR](https://github.com/rust-lang/rust/issues/114994) some constraints got relaxed and we are able to create a const Instruction that contains `&mut` in its signature.

Currently, this works in the nightly version: `cargo 1.75.0-nightly (6fa6fdc76 2023-10-10)` and we should wait until this is in stable before merging.

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -64,7 +64,7 @@ macro_rules! opcodes {
         };
 
         /// Return the instruction function for the given opcode and spec.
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -86,6 +85,7 @@ pub fn make_instruction_table<SPEC: Spec, H: Host>() -> InstructionTable<H> {
     table
 }
 
+
```
rakita, line 88:
```suggestion
```

rakita:
Closing this, and will reopen in few months

PR #806, opened 2023-10-16
chore(SharedMemory): small internal refactoring

lorenzofero:
- `current_len` is redundant -> replaced with `last_checkpoint` with minor performance boost
- some renaming and refactoring done, the code is a little cleaner (imo)
- added a couple of tests

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -15,13 +15,12 @@ use core::{
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SharedMemory {
     /// The underlying buffer.
-    data: Vec<u8>,
+    buffer: Vec<u8>,
     /// Memory checkpoints for each depth.
     /// Invariant: these are always in bounds of `data`.
     checkpoints: Vec<usize>,
-    /// How much memory has been used in the current context.
-    /// Invariant: `checkpoints.last() + current_len <= data.len()`
-    current_len: usize,
+    /// Invariant: equals `self.checkpoints.last()`
+    last_checkpoint: usize,
     /// Memory limit. See [`crate::CfgEnv`].
     #[cfg(feature = "memory_limit")]
     memory_limit: u64,
@@ -30,7 +29,7 @@ pub struct SharedMemory {
 impl fmt::Debug for SharedMemory {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("SharedMemory")
-            .field("current_len", &self.current_len)
+            .field("current_len", &self.len())
             .field(
                 "context_memory",
                 &crate::primitives::hex::encode(self.context_memory()),
@@ -59,9 +58,9 @@ impl SharedMemory {
     #[inline]
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
-            data: Vec::with_capacity(capacity),
+            buffer: Vec::with_capacity(capacity),
             checkpoints: Vec::with_capacity(32),
-            current_len: 0,
+            last_checkpoint: 0,
             #[cfg(feature = "memory_limit")]
             memory_limit: u64::MAX,
         }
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -85,53 +84,44 @@ impl SharedMemory {
...
-
-        if let Some(available_memory) = self.data.get_mut(range) {
-            available_memory.fill(0);
-        } else {
-            self.data.resize(last_checkpoint + new_len, 0);
-        }
-
-        self.current_len = new_len;
+    pub fn resize(&mut self, new_size: usize) {
+        self.buffer.reserve(usize::max(new_size, 4 * 1024));
```
DaniPopes, line 123: `resize` allocates this, actually for `new_size > 4*1024` it would allocate twice.
Do you mean `usize::min`?

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -292,33 +276,19 @@ impl SharedMemory {
     /// Returns a reference to the memory of the current context.
     #[inline]
     fn context_memory(&self) -> &[u8] {
-        let last_checkpoint = self.last_checkpoint();
-        let current_len = self.current_len;
-        debug_assert!(last_checkpoint + current_len <= self.data.len());
-        // SAFETY: `last_checkpoint` and `current_len` are always in bounds of `self.data`
+        // safety: access bounded by buffer len
```
DaniPopes, line 279: Let's keep style consistent for `SAFETY:` comments

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -85,53 +84,44 @@ impl SharedMemory {
...
+            // safety: buffer len is less than or equal `old_checkpoint`
+            unsafe { self.buffer.set_len(old_checkpoint) };
         }
     }
 
     /// Returns the length of the current memory range.
     #[inline]
     pub fn len(&self) -> usize {
-        self.current_len
+        self.buffer.len() - self.last_checkpoint
```
DaniPopes, line 111: This makes more sense , but in this case can we not just keep using the `last().unwrap_or()` helper function for `last_checkpoint` instead of keeping it inline?

DaniPopes (changes requested):

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: My idea was to reserve a minimum of 4096 bytes to avoid many mini-allocations, but I agree that if `new_size > 4 * 1024` then this line is useless

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Isn't it faster to keep it in the stack instead of reading the same value from the heap multiple times?

crates/interpreter/src/interpreter/shared_memory.rs
DaniPopes: Yes, then it should be `min` not `max`

crates/interpreter/src/interpreter/shared_memory.rs
DaniPopes: Yeah I guess so

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Hmm I'm not sure: suppose `new_size` equals 32, if we use `min` then:
```rust
assert_eq!(usize::min(32, 4*1024), 32);
```
and we reserve 32 more bytes instead of 4096, which is not what we want.
I don't understand if I'm missing something here from what you said

DaniPopes: Ah you're right. This should be removed because `Vec` already does exponential allocations so this is redundant since we're starting already with 4*1024

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -85,53 +84,44 @@ impl SharedMemory {
...
 
     /// Prepares the shared memory for returning to the previous context.
     #[inline]
-    pub fn free_context_memory(&mut self) {
+    pub fn free_context(&mut self) {
         if let Some(old_checkpoint) = self.checkpoints.pop() {
-            let last_checkpoint = self.last_checkpoint();
-            self.current_len = old_checkpoint - last_checkpoint;
+            self.last_checkpoint = self.checkpoints.last().cloned().unwrap_or_default();
+            // safety: buffer len is less than or equal `old_checkpoint`
```
rakita, line 103:
```suggestion
            // SAFETY: buffer length is less than or equal `old_checkpoint`
```

SAFETY comment stile is from here: https://std-dev-guide.rust-lang.org/policy/safety-comments.html

crates/interpreter/src/interpreter/shared_memory.rs
rakita: To be precise it should be something like `. reserve(self.last_checkpoint+max(new_size,4*1024))` to allocate at least 4048, but as @DaniPopes said `resize` doubles capacity every time so it is good to rely on.

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Can I consider this conversation resolved then?

crates/interpreter/src/interpreter/shared_memory.rs
lorenzofero: Thanks for the reference! Soon I'll solve the comments

crates/interpreter/src/interpreter/shared_memory.rs
DaniPopes: yes

rakita (approved):
lgtm

PR #811, opened 2023-10-16
chore: review safety comments

DaniPopes:


crates/revm/src/inspector/instruction.rs
```diff
@@ -1,49 +1,39 @@
...
-            // step
-            let data = &mut host.data;
-            // Safety: as the PC was already incremented we need to subtract 1 to preserve the
-            // old Inspector behavior.
-            unsafe {
-                interpreter.instruction_pointer = interpreter.instruction_pointer.sub(1);
-            }
             if let Some(inspector) = host.inspector.as_mut() {
-                inspector.step(interpreter, data);
+                // Safety: as the PC was already incremented we need to subtract 1 to preserve the
```
lorenzofero, line 16:
```suggestion
                // SAFETY: as the PC was already incremented we need to subtract 1 to preserve the
```
According to https://std-dev-guide.rust-lang.org/policy/safety-comments.html#inside-safe-elements

crates/interpreter/src/instructions/system.rs
```diff
@@ -91,7 +90,6 @@ pub fn calldatacopy<H: Host>(interpreter: &mut Interpreter<'_>, _host: &mut H) {
     let data_offset = as_usize_saturated!(data_offset);
     shared_memory_resize!(interpreter, memory_offset, len);
 
-    // Safety: set_data is unsafe function and memory_resize ensures us that it is safe to call it
```
rakita, line 94: Can you revert this?

crates/interpreter/src/instructions/system.rs
```diff
@@ -45,7 +45,6 @@ pub fn codecopy<H: Host>(interpreter: &mut Interpreter<'_>, _host: &mut H) {
     let code_offset = as_usize_saturated!(code_offset);
     shared_memory_resize!(interpreter, memory_offset, len);
 
-    // Safety: set_data is unsafe function and memory_resize ensures us that it is safe to call it
```
rakita, line 48: Can you revert this?

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -40,7 +41,7 @@ impl Stack {
     #[inline]
     pub fn new() -> Self {
         Self {
-            // Safety: [`Self::push`] assumes that capacity is STACK_LIMIT
+            // Important: expansion functions assume that capacity is `STACK_LIMIT`.
```
rakita, line 44: Would still put `SAFETY` here

crates/revm/src/journaled_state.rs
```diff
@@ -183,7 +183,7 @@ impl JournaledState {
     /// 5. Increment nonce of created account if SpuriousDragon is active
     /// 6. Decrease balance of caller account.
     ///
-    /// Safety: It is assumed that caller balance is already checked and that
+    /// Note: it is assumed that caller balance is already checked and that
```
rakita, line 186: Would rephrase this as a `# Panic`

crates/interpreter/src/instructions/system.rs
DaniPopes: the function is not unsafe, but I'll keep the comment as a note

PR #812, opened 2023-10-16
refactor: rewrite `Stack::push_slice` to allow arbitrary lengths

DaniPopes:
Performance is the exact same. The current implementation is a hand-rolled version of this PR's.


rakita:
How did you measure performance for this? It looks great btw!

DaniPopes:
> How did you measure performance for this? 

Manually looked at generated assembly for `push{1,2,20,32}` and it's identical minus register names.

Also ran callgrind:

before (885c0cc81ae717d88126442917b5237ef28c0487): 416,256,326
after1 (cdd4754d9e27cb851b227161263323b3319c38e6): 416,256,326

PR #813, opened 2023-10-17
chore: introduce MSRV policy, set MSRV to 1.66

DaniPopes:


rakita:
As specified in readme only requirement for revm is to use the last stable rust, so closing this.

Issue #797, opened 2023-10-12
Comment output of cachegrind in PR


rakita:
Not sure how to do this but would be very beneficial

As i am on mac valgrind does not work and i have a Docker file:
```
FROM alpine:latest
RUN apk add g++ gcc clang-dev valgrind curl git llvm

# Install rust
RUN curl https://sh.rustup.rs -sSf | sh -s -- -y

# Add .cargo/bin to PATH
ENV PATH="/root/.cargo/bin:${PATH}"

# Check cargo is visible
RUN cargo --help
```

Run docker with:
` docker run -it -v $PWD:/tmp -w /tmp valgrind:1.0`

And I run snailtracer to get results like this:
```
cargo b -r -p revm-test --bin snailtracer
valgrind --tool=cachegrind target/release/snailtracer
```

0x0elliot:
Can i help out with this issue? Seems simple enough as an entrypoint for someone who's new to the EVM ecosystem.

rakita:
> Can i help out with this issue? Seems simple enough as an entrypoint for someone who's new to the EVM ecosystem.

Go for it, it is mostly github actions and how to output it in PR (not sure what github plugging are out there for it)

0x0elliot:
@rakita A bit confused as to what needs to be done. Do you mind giving me a bit of context? I reassure you, this would get lesser and lesser with the next few contributions.

If you wish, I can ping you somewhere else too (a discord server, the revm telegram group etc)

rakita:
> @rakita A bit confused as to what needs to be done. Do you mind giving me a bit of context? I reassure you, this would get lesser and lesser with the next few contributions.
> 
> If you wish, I can ping you somewhere else too (a discord server, the revm telegram group etc)

Idea is to run valgrind cachegrind on every new PR.

cachegrind is used to deduce the number of executed instructions and is a good metric to check how fast/slow the program is. In comparison, time measurement can be variable and hard to get one value, while num of the instruction is always the same for the same program so it is a good ballpark.

So executing this on every PR means adding github action: https://github.com/bluealloy/revm/tree/main/.github/workflows and using some (i hope it exists)  github plugin that outputs this to the PR as a comment.

Hope that this makes it clearer.

0x0elliot:
Thanks! That does. Let me play around with this.

rakita:
Still need to fix PAT and github token but issue is considered done

PR #816, opened 2023-10-18
Shared stack between calls

thedevbirb:
This is an experiment for a shared stack between calls. Let me know what you think about it and if it is feasible to bring it in!
The model is very similar to last one developed in #445 (with checkpoints), although there are some differences for the allocation strategy.

## Allocations

There are three different strategies here:
- do one single 32MB allocation -> we've seen with the shared memory that this approach is not good and overkill most of the times
- check for capacity at every `push` operation in the stack -> does the minimum amount of memory allocations, however it adds a small but non-negligible overhead for `push` operations, which results in a performance regression in a very stress test like the snailtracer bench
- ensure a capacity of `STACK_LIMIT` every time you enter new context -> Although this is very similar to what happens normally (i.e., on new context we allocate space for the stack), it still keeps peak memory usage allocated for next contexts and allows for faster `unsafe` operations, reducing `push` operations overhead. This is the approach I kept after all the experiments

## Performance

I managed to keep regression to the minimum, however benches don't really the shared stack at all. Even `bench_eval` on the snailtracer remains always on the same context, therefore the shared mechanism doesn't apply well.
There is of course a minimum of overhead because this abstraction ain't free, but it seems feasible.
```
analysis/transact/raw   time:   [8.2504 µs 8.3154 µs 8.4299 µs]
                        change: [+2.8469% +3.7488% +4.8213%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
analysis/transact/checked
                        time:   [8.2338 µs 8.3507 µs 8.4382 µs]
                        change: [+1.0517% +1.9726% +2.9546%] (p = 0.00 < 0.05)
                        Change within noise threshold.
analysis/transact/analysed
                        time:   [5.6613 µs 5.6720 µs 5.6838 µs]
                        change: [+0.6835% +2.5449% +4.1668%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

snailtracer/transact/analysed
                        time:   [67.868 ms 68.022 ms 68.191 ms]
                        change: [+12.395% +13.393% +14.409%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
snailtracer/eval        time:   [61.284 ms 61.386 ms 61.590 ms]
                        change: [+0.6965% +5.5300% +9.5800%] (p = 0.03 < 0.05)
                        Change within noise threshold.

transfer/transact/analysed
                        time:   [1.2470 µs 1.2484 µs 1.2505 µs]
                        change: [+6.3084% +6.6428% +6.9314%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) high mild
  6 (6.00%) high severe
```


crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,550 @@
...
+        }
+    }
+
+    /// Prepares the shared stack for a new context
+    #[inline]
+    pub fn new_context(&mut self) {
+        let buf_len = self.buffer.len();
+        self.checkpoints.push(buf_len);
+        self.last_checkpoint = buf_len;
+        self.buffer.reserve(STACK_LIMIT);
```
rakita, line 61: This reserve would copy the full buffer of the previous values, so it would be better to be more granular and introduce the Page of memory that would contain at least `4*1024` items. This value (as a 4) is very arbitrary but the idea is to reuse the current stack.

On `new_context` we would check if there are at least 1024 slots left on the present page, and if this is not the case would allocate a new page. `free_context` would do this in reverse.

rakita:
You can check performance with cachegrind and docker like this: https://github.com/bluealloy/revm/issues/797

and additionally, we should write the proper perf test for this. There is maybe bytecode that we can reuse from eth/tests: https://github.com/ethereum/tests/tree/develop/GeneralStateTests
There are fillers that are more readable: https://github.com/ethereum/tests/tree/develop/src/GeneralStateTestsFiller/stCallCodes

lorenzofero:
> You can check performance with cachegrind and docker like this: #797
> 
> and additionally, we should write the proper perf test for this. There is maybe bytecode that we can reuse from eth/tests: https://github.com/ethereum/tests/tree/develop/GeneralStateTests There are fillers that are more readable: https://github.com/ethereum/tests/tree/develop/src/GeneralStateTestsFiller/stCallCodes

I agree regarding a proper performance test for this. We need to choose something that goes up and down in call depths to have a proper picture of the gains of this setup.

In the meanwhile, I tried with cachegrind as you said (thanks, I'll keep that mind for the future). Here are the results on my machine:
- main https://github.com/bluealloy/revm/commit/1dcebc43f09893cd1520bf097b1e91ef9daf990b -- 482,459,263
- shared_stack https://github.com/bluealloy/revm/pull/816/commits/6c4acdf5fe780a5206980fe1310af4aade1797c1 -- 529,105,196

Therefore yes there is still some work to do to bring down regression when this is not used

crates/interpreter/src/interpreter/shared_stack.rs
lorenzofero: This seems exciting! Let me know if I got this right:
- instead of a single buffers we have multiple buffers (or "pages") of at least `4*1024` items -> which means a `Vec<Vec<U256>>` with initial value something like `vec![Vec::with_capacity(4 * STACK_LIMIT)]`
- on `new_context`, if there is enough space in the _last page_, we keep it for this new context, otherwise we allocate another page
- on `free_context`, we just mark that space of the _last page_ available in some way and then we make the necessary changes accordingly

In practice, this really resembles the model you proposed in https://github.com/ethereum/evmone/issues/481#issuecomment-1179088645. Since we know in advance how much we should expand the buffer (which was not possible with the shared memory)  this should play very nicely. It should have a little more overhead if you remain on the same context the whole tx since the setup is a little more complex than the previous one, but it should definitely be better for multi-contexts since we do not copy data anymore in order to get more capacity

rakita: Exactly, this would be a bit more complex to do with `SharedMemory`, but for the stack, it is a lot nicer and we should do it now.

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,775 @@
...
+            unsafe {
+                page.buffer.set_len(old_checkpoint);
+            };
+            self.context_len = old_checkpoint - last_checkpoint;
+        } else if self.page_idx > 0 {
+            // no more checkpoints means we need to move to the previous page
+            self.page_idx -= 1;
+            self.buffer = &mut self.page_mut().buffer;
+            // SAFETY: Going back to previous page implies non-empty checkpoints list
+            self.context_len = unsafe {
```
rakita, line 143: Why is this unsafe?

lorenzofero: Because of the `unwrap_unchecked` at the end of line 144. When going back to the previous page checkpoints length is greater than zero by construction, therefore checking the `Option<T>` is not needed.

Am I missing something?

lorenzofero:
Hey @rakita, I tried to task on the `ethereum/tests` repository about a stress test but the suggestion was to write a custom one. 

In the meanwhile I tried to implement this [approval transfer](https://github.com/ziyadedher/evm-bench/blob/main/benchmarks/erc20/approval-transfer/ERC20ApprovalTransfer.sol) bench: even if it does not reach a huge depth it loops over context changes, therefore I thought it could be good to see how it performs over those.
Here are the result (`cargo bench --all` against `main` 0d78d1eb304a2ce41ddac8f03206f5a316af247b):
```text
approval_transfer/transact/analysed
                        time:   [4.7078 µs 4.7261 µs 4.7508 µs]
                        change: [+10.900% +12.297% +13.797%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
approval_transfer/eval  time:   [55.583 ns 55.690 ns 55.817 ns]
                        change: [-84.656% -84.591% -84.545%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

Regarding the custom test is there something in particular you had in mind?

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+    pages: Vec<Page>,
+    /// Keeps track of the page used for the current context
+    /// Invariant: is a valid index of `self.pages`
+    page_idx: usize,
+    /// The pointer to the buffer used by the current page.
+    /// Needed for better performance and to avoid double
+    /// heap lookup for basic stack operations.
+    ///
+    /// Invariant: it a valid pointer to `self.pages[self.page_idx].buffer`
+    buffer: *mut Buffer,
```
rakita, line 60: Would put full `Page` here not just buffer. and rename `pages` field to `previous_pages` to represent pages in somekind of backlog (not currently active)

crates/interpreter/src/interpreter/shared_stack.rs
rakita: brainfart, ignore it I just saw `unwrap` and not `unwrap_unchecked`

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+        // SAFETY: by construction, `self.page_idx` is a valid
+        // index of `self.pages`
+        unsafe { self.pages.get_unchecked_mut(self.page_idx) }
+    }
+
+    #[inline]
+    fn buffer(&self) -> &Buffer {
+        // SAFETY: by construction, `self.buffer` is a valid
+        // pointer to the buffer of the current page
+        unsafe { &*self.buffer }
```
rakita, line 444: This would become `&mut self.page.buffer` after this change: https://github.com/bluealloy/revm/pull/816/files#r1385180054

crates/interpreter/src/interpreter/shared_stack.rs
rakita: hm i see the problem here, maybe we should not look at free pages as a stack. wdyt about idea of having two fields: `taken_pages` and `free_pages` and they are both Vecs.

 if the context is freed you put the Page to the `free_pages` vec and on `new_context` you pop page from `free_pages` or create a new one if there is none.

rakita: Another idea (not related to the first part) is related to the `Buffer` and the `context_len`, Buffer is a `Vec` here that has its own `len` and additionally we have `context_len` that we increment. 

We could work with `*mut U256` which would point us directly to the top of the stack, and we could have `context_len` that would tell us the bound of the stack. I like this idea, but we should be careful about the usage of the pointer.

rakita (changes requested):
left a few ideas that could improve the PR, I like the page system and new context, but I wanted to check how loop call will look so we can integrate this inside it so I apologize for the long overdue review.

crates/interpreter/src/interpreter/shared_stack.rs
lorenzofero: I like the first idea! It should make it easier to reason about and we can avoid the `page_idx` which can be annoying to work with.

Regarding the second idea let me know if I got this right: with a raw pointer to the top of the stack, all operations are not done on the buffer itself (which is preallocated with 1024 capacity) but rather on the pointer, by dereferencing and incrementing/decrementing it. If that's the case, do you think it provides a performance improvement over dealing with the `Buffer` pointer or it is a matter of ergonomics?

Lastly, but slighly unrelated: wdyt of putting both shared stack and shared memory under the `EVMContext` struct? Then when creating the interpreter we can pass a raw pointer to the current context buffer of stack and memory.

rakita: Usage of the pointer should be slightly faster, it should be less ergonomical as you directly handle the pointer.

Would be good to put it in `EvmContext` but I think borrowing is going to kill us there, and it would not matter a lot.

lorenzofero: Okay, I'll try to modify the shared stack with suggestions. It will be some work to do, I hope some wait is not a problem!

Regarding `EVMContext` maybe I can think of it in a separate PR after this and see if it is reasonable

lorenzofero:
> left a few ideas that could improve the PR, I like the page system and new context, but I wanted to check how loop call will look so we can integrate this inside it so I apologize for the long overdue review.

Thanks for the feedback and no problem for the overdue!

crates/interpreter/src/interpreter/shared_stack.rs
rakita: Take your time and pace yourself, I am fine to take over if you feel burdened to finish it, and I am fine with waiting. Both things work for me.

We can maybe saparate that pointer idea to another PR to not clog this.

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+        assert_eq!(shared_stack.page_mut().buffer.len(), STACK_LIMIT);
+        assert_eq!(shared_stack.len(), STACK_LIMIT);
+        assert_eq!(
+            shared_stack.push(U256::ZERO),
+            Err(InstructionResult::StackOverflow)
+        );
+    }
+
+    #[test]
+    fn push_slice() {
```
DaniPopes, line 625: Please keep the test cases from https://github.com/bluealloy/revm/pull/837
There's multiple things to test for this function

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+            return Err(InstructionResult::StackOverflow);
+        }
+        let buffer = self.buffer_mut();
+        let buf_len = buffer.len();
+        // SAFETY: the check above and the `new_context` method
+        // guarantee we have enough capacity
+        unsafe {
+            *buffer.get_unchecked_mut(buf_len) = value;
+            buffer.set_len(buf_len + 1);
+        }
```
DaniPopes, line 305: This is UB, please see #837

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+        } else {
+            let buffer = self.buffer_mut();
+            let buf_len = buffer.len();
+            // SAFETY: the check above and the `new_context`
+            // method guarantee we have enough capacity
+            unsafe {
+                let val = *buffer.get_unchecked(buf_len - N);
+                *buffer.get_unchecked_mut(buf_len) = val;
+                buffer.set_len(buf_len + 1);
+            };
```
DaniPopes, line 333: This is UB, please see #837

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+
+    /// Swaps the topmost value with the `N`th value from the top.
+    #[inline]
+    pub fn swap<const N: usize>(&mut self) -> Result<(), InstructionResult> {
+        if self.len() <= N {
+            return Err(InstructionResult::StackUnderflow);
+        }
+        let buffer = self.buffer_mut();
+        let last = buffer.len() - 1;
+        buffer.swap(last, last - N);
```
DaniPopes, line 347: avoid a local

```suggestion
        let last = buffer.len() - 1;
        self.buffer_mut().swap(last, last - N);
```

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+            debug_assert_eq!((i + 3) / 4, n_words, "wrote beyond end of stack");
+
+            // zero out upper bytes of last word
+            let m = i % 4; // 32 / 8
+            if m != 0 {
+                dst.add(i).write_bytes(0, 4 - m);
+            }
+
+            self.context_len = new_context_len;
+            self.buffer_mut().set_len(new_buffer_len);
```
DaniPopes, line 403: This also has to be updated to #837

DaniPopes (changes requested):

crates/interpreter/src/interpreter/shared_stack.rs
rakita: The thing is that this branch is not merged with the `main`, so it has old changes. This should be fixed after the git merge but it is good to be aware of it.

DaniPopes: Ah sorry, i didnt see that

lorenzofero:
Hey there, I pushed some changes regarding the free and taken pages model. Also, now both stack and memory are under a SharedContext struct. 
I still need to address some of the comments regarding UB and upstream sync, will do it shortly!

crates/interpreter/src/interpreter/shared_stack.rs
lorenzofero: Yeah if you want I'd be very happy to work on the pointer idea on a separate PR!

crates/interpreter/src/interpreter/shared_context.rs
```diff
@@ -0,0 +1,56 @@
+use crate::{SharedMemory, SharedStack, EMPTY_SHARED_MEMORY, EMPTY_SHARED_STACK};
+
+pub const EMPTY_SHARED_CONTEXT: SharedContext = SharedContext {
+    stack: EMPTY_SHARED_STACK,
+    memory: EMPTY_SHARED_MEMORY,
+};
```
DaniPopes, line 6: This should be a `const fn empty() -> Self { ... }` or `const EMPTY: Self = ...;` on all the related structs. Const item initializers are weird to me, so I'd prefer the former.

DaniPopes (changes requested):

crates/interpreter/src/interpreter/shared_stack.rs
```diff
@@ -0,0 +1,710 @@
...
+/// a vector of "pages" (buffers) or internal representation.
+///
+/// Each page includes a buffer of size `PAGE_SIZE` which
+/// is shared between calls, and when there is no more space left
+/// we move to a new page.
+///
+/// A [SharedStack] instance should always be obtained using
+/// the `new` static method to ensure memory safety.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SharedStack {
```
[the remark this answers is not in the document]
lorenzofero: I also agree that keeping the same name could be better since both are under the SharedContext struct now (we can also find better name for it)

crates/interpreter/src/interpreter/shared_context.rs
lorenzofero: Okay! I kept this for consistency with rakita's work, if he also agrees with this change for me it's perfectly fine

crates/interpreter/src/interpreter/shared_stack.rs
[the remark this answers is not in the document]
DaniPopes: If we keep the shared context idea, I think we should also add simple interpreter getters `fn {stack,memory}{,_mut}()` that just access from `self.shared_context`

crates/interpreter/src/interpreter.rs
```diff
@@ -39,13 +42,11 @@ pub struct Interpreter {
...
     pub gas: Gas,
-    /// Shared memory.
+    /// Shared stack.
     ///
     /// Note: This field is only set while running the interpreter loop.
-    /// Otherwise it is taken and replaced with empty shared memory.
-    pub shared_memory: SharedMemory,
-    /// Stack.
-    pub stack: Stack,
+    /// Otherwise it is taken and replaced with empty shared stack.
```
DaniPopes, line 48: s/stack/context

crates/interpreter/src/interpreter.rs
```diff
@@ -96,14 +97,18 @@ impl Interpreter {
             instruction_result: InstructionResult::Continue,
             is_static,
             return_data_buffer: Bytes::new(),
-            shared_memory: EMPTY_SHARED_MEMORY,
-            stack: Stack::new(),
+            shared_context: EMPTY_SHARED_CONTEXT,
```
DaniPopes, line 100: shouldn't this be `::new()` to pre-allocate?

crates/interpreter/src/instructions/macros.rs
```diff
@@ -63,117 +63,136 @@ macro_rules! shared_memory_resize {
...
+    )*);
+    ($interp:expr, $($x:expr),* $(,)?) => ($(
+        push!($interp, $interp.shared_context.stack, $x)
+        // match $interp.shared_context.stack.push($x) {
+        //     Ok(()) => {},
+        //     Err(e) => {
+        //         $interp.instruction_result = e;
+        //         return;
+        //     }
+        // }
```
lorenzofero, line 195: ops, will remove it

crates/interpreter/src/interpreter.rs
lorenzofero: Actually I don't think so, because we want to manually give the created context to then interpreter when it calls the `run` method of `EVMImpl`. See https://github.com/bluealloy/revm/pull/816/files/1de2425fed99f33422e0fa37911f7695a2a39c5b#diff-1d478ba44ccc56e3b1142bd3723bf97f3e254c25dd18323481aedadce0803e91R165-R170. Maybe I am missing something from what you said

lorenzofero:
Hey guys, I resolved some comments to clean this up: is there something left to do? More perf tests? Right now perf isn't ideal if you remain always in the same context. Maybe I can try to use the pointer to the top of the stack as @rakita said

lorenzofero:
Hey @rakita, I've seen that EvmBuilder PR is merged. Great work! 

I tried to take a look at the changes and I was wondering if it makes sense to take a look at this from scratch, if you feel a shared stack would be beneficial for this evm.

Both the pages model can be revisited (imo, it seems that a more complex strategy that allocates less is worse than a simpler one with more allocations) and also the `SharedContext` doesn't seem to play too nicely with the current setup. An example is passing manually `SharedContext` taken from the interpreter to `sub_create` and down to other functions such as `insert_create_output` which is currently used by inspector logic too.

rakita:
> Hey @rakita, I've seen that EVM Context-Builder PR is merged. Great work!
> 
> I tried to take a look at the changes and I was wondering if it makes sense to take a look at this from scratch, if you feel a shared stack would be beneficial for this evm.
> 
> Both the pages model can be revisited (imo, it seems that a more complex strategy that allocates less is worse than a simpler one with more allocations) and also the `SharedContext` doesn't seem to play too nicely with the current setup. An example is passing manually `SharedContext` taken from the interpreter to `sub_create` and down to other functions such as `insert_create_output` which is currently used by inspector logic too.

Thanks @lorenzofero!

My view of shared context is we want it is some way, I am open to new ideas, but just to say this is very low priority, as there is no impact by doing this.

thedevbirb:
Closing this as I think it should be re-visited from scratch. Might do that in the future but I'd prefer to focus on other contributions on revm, time permitting :)

thedevbirb:
Closing this as I think it should be re-visited from scratch. Might do that in the future but I'd prefer to focus on other contributions on revm, time permitting :)

PR #822, opened 2023-10-20
chore(eip4844): update kzg trusted setup

rakita:
Update to the official mainnet kzg trusted setup https://github.com/ethereum/c-kzg-4844/pull/377


rakita:
Need to update test test data

PR #823, opened 2023-10-20
chore: Refactor precompile list from Hash to vec

rakita:
Replace the precompile address check with something more robust. Precompiles are now saved inside a `Vec` and the search if account address is a precompile is done by binary search of that sorted Vec. This allows us to have arbitrary addresses inside precompiles.

Added a few comments and `inlined` `JournalState` functions.


crates/precompile/src/lib.rs
```diff
@@ -121,86 +118,77 @@ impl SpecId {
...
+                modexp::BERLIN,
+            ]);
+            precompiles.inner.sort_unstable_by_key(|i| i.0);
             precompiles
         })
     }
 
+    /// Returns precompiles for Cancun spec.
+    ///
     /// If `std` feature is not enabled KZG Point Evaluation precompile will not be included.
```
DaniPopes, line 192: Forgot this

```suggestion
    /// If the `c-kzg` feature is not enabled, the KZG Point Evaluation precompile will not be included.
```

PR #824, opened 2023-10-21
fix(op): Base Goerli `op-reth` sync patches

clabby:
## Overview

Contains several fixes for the Optimism state transition code.

### Context

After hooking up the OP execution changes to `op-reth` and performing a sync test on the Base Goerli network, a few bugs were found.

### Patches

1. The Optimism EL has a critical invariant that _deposit transactions may never fail_. This implies that even upon a total failure of the state transition, for any reason, the `mint` value should be persisted to the sender account's balance to prevent loss of funds, and the sender's nonce should be bumped as well.
    1. **Solution**: Pull out the default implementation of the `transact` function in the `Transact` trait, and define a catch-all failure handler that persists the `mint` value to and bumps the nonce of the caller account before returning a special `Halt` variant.
1. When a deposit `Halt`s, a few special rules apply. These happen to be identical to the logic in the above catch-all.
    1. **Solution**: Wipe the state of the journal and fall back to the catch-all that persists the mint value and bumps the nonce of the caller. This will return a `Halt::FailedDeposit` status.

Reference Branch (cherry-picked #781 onto `3.5.0` for integrating w/ `op-reth`: https://github.com/anton-rs/op-revm/tree/cl/patch-op)

### DIFF
--- a/crates/revm/src/evm_impl.rs
+++ b/crates/revm/src/evm_impl.rs
@@ -16,16 +16,19 @@ use alloc::vec::Vec;
 use auto_impl::auto_impl;
 use core::fmt;
 use core::marker::PhantomData;
-use revm_interpreter::opcode::make_boxed_instruction_table;
 use revm_interpreter::{
     gas::initial_tx_gas,
-    opcode::{make_instruction_table, InstructionTables},
+    opcode::{make_boxed_instruction_table, make_instruction_table, InstructionTables},
     SharedMemory, MAX_CODE_SIZE,
 };
 use revm_precompile::{Precompile, Precompiles};
 
 #[cfg(feature = "optimism")]
-use crate::optimism;
+use crate::{
+    optimism,
+    precompile::HashMap,
+    primitives::{Account, ExecutionResult, ResultAndState},
+};
 
 /// EVM call stack limit.
 pub const CALL_STACK_LIMIT: u64 = 1024;
@@ -87,11 +90,7 @@ pub trait Transact<DBError> {
     fn transact_preverified(&mut self) -> EVMResult<DBError>;
 
     /// Execute transaction by running pre-verification steps and then transaction itself.
-    #[inline]
-    fn transact(&mut self) -> EVMResult<DBError> {
-        self.preverify_transaction()
-            .and_then(|()| self.transact_preverified())
-    }
+    fn transact(&mut self) -> EVMResult<DBError>;
 }
 
 impl<'a, DB: Database> EVMData<'a, DB> {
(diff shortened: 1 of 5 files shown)


rkrasiuk (approved):
changes lgtm, pending @rakita

crates/revm/src/handler/optimism.rs
```diff
@@ -140,6 +144,133 @@ pub fn reward_beneficiary<SPEC: Spec, DB: Database>(
...
+        }
+        SuccessOrHalt::FatalExternalError => {
+            return Err(EVMError::Database(data.error.take().unwrap()));
+        }
+        SuccessOrHalt::InternalContinue => {
+            panic!("Internal return flags should remain internal {call_result:?}")
+        }
+    };
+
+    Ok(ResultAndState { result, state })
```
rakita, line 202: Would recommend doing something like this:
```
let ret = mainnet::main_return(..)?;
if let SuccessOrHalp::Halt(reason) = ret {


}
ret
```

This would reuse mainnet handler and just append optimism needed logic

crates/revm/src/handler.rs
```diff
@@ -43,7 +43,7 @@ pub struct Handler<DB: Database> {
     /// Calculate gas refund for transaction.
     /// Some chains have it disabled.
     pub calculate_gas_refund: CalculateGasRefundHandle,
-    /// Main return handle this handle output of the transact.
+    /// Main return handle, returns the output of the transaction.
```
rakita, line 46:
```suggestion
    /// Main return handle, returns the output of the transact.
```

crates/revm/src/evm_impl.rs
```diff
@@ -87,11 +85,7 @@ pub trait Transact<DBError> {
     fn transact_preverified(&mut self) -> EVMResult<DBError>;
 
     /// Execute transaction by running pre-verification steps and then transaction itself.
-    #[inline]
-    fn transact(&mut self) -> EVMResult<DBError> {
-        self.preverify_transaction()
-            .and_then(|()| self.transact_preverified())
-    }
+    fn transact(&mut self) -> EVMResult<DBError>;
```
rakita, line 88: Should stay as this and logic should be changes for used function.

crates/revm/src/evm_impl.rs
```diff
@@ -353,6 +347,19 @@ impl<'a, GSPEC: Spec + 'static, DB: Database> Transact<DB::Error> for EVMImpl<'a
...
+        #[cfg(not(feature = "optimism"))]
+        {
+            crate::handler::mainnet::default_transact(self)
+        }
+
+        #[cfg(feature = "optimism")]
+        {
+            crate::handler::optimism::default_transact(self)
+        }
+    }
```
rakita, line 362: This should not go here, but inside `transact_preverified` logic will be skipped otherwise.

rakita:
@clabby if you are okay with it, to speed things up i can create the PR on this branch

crates/revm/src/handler/optimism.rs
rakita: https://github.com/anton-rs/op-revm/pull/34

clabby: Yeah this is much cleaner, nice idea.

rakita:
failed test unrelated to PR

PR #825, opened 2023-10-21
feat: add data and into_data

mattsse:
adds back previous `Memory` data functions:

https://github.com/bluealloy/revm/pull/516/files

I don't fully understand if this is correct. idk what a context is or how the shared memory is laid out.

In reth inspectors we need to get the currently active memory, and I'm not sure this is still accurate:

https://github.com/paradigmxyz/reth/blob/9dd270e4e8095a4cab1d2118f8f965eec68f4db7/crates/revm/revm-inspectors/src/tracing/types.rs#L536-L537

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -26,25 +26,6 @@ pub struct SharedMemory {
     memory_limit: u64,
 }
 
-impl fmt::Debug for SharedMemory {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("SharedMemory")
-            .field("current_len", &self.len())
-            .field(
-                "context_memory",
-                &crate::primitives::hex::encode(self.context_memory()),
-            )
-            .finish_non_exhaustive()
-    }
-}
-
-impl Default for SharedMemory {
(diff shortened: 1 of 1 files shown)


rakita:
This is not oka. `.context_memory()` should be used to fetch current call memory frame

mattsse:
but this is exactly what this did?

and context_memory is private

rakita:
> but this is exactly what this did?
> 
> and context_memory is private

I didn't see that `context_memory` is not public, it should be.

`into_data` would still copy the `context_memory` with `split_off`, and using `into_data` for consuming shared memory for its last context and disregarding previous ones could introduce footguns.

mattsse:
> into_data would still copy the context_memory with split_off, and using into_data for consuming shared memory for its last context and disregarding previous ones could introduce footguns.

I see, yeah, makes sense, we need to clone that during tracing anyway,
will open another pr that makes context_memory pub then

rakita:
> > into_data would still copy the context_memory with split_off, and using into_data for consuming shared memory for its last context and disregarding previous ones could introduce footguns.
> 
> I see, yeah, makes sense, we need to clone that during tracing anyway, will open another pr that makes context_memory pub then

Sorry if i closed this abruptly, you can make the change here or a new PR sounds okay.

PR #828, opened 2023-10-21
chore(state): make bundle state non-optional

rkrasiuk:
## Description

There is no good reason to put `BundleState` in an `Option`. In the case of a missing pre-loaded bundle, the default can be used instead.


rakita (approved):
Yeah, the Idea for the option was to have optional bundle but this was flushed out inside `StateBuilder` and optional `transition_state` is used

PR #833, opened 2023-10-24
chore(opcodes): new_or_invalid helper

lorenzofero:
While doing an issue on reth (https://github.com/paradigmxyz/reth/issues/5067) I've found it can be helpful to have this method that simply converts to `INVALID` any unknown opcode, instead of having to do pattern matching

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -393,6 +393,17 @@ impl OpCode {
         }
     }
 
+    /// Instantiate a new opcode from a u8 or return
+    /// the `INVALID` opcode if missing from `OPCODE_JUMPMAP`
+    #[inline]
+    pub const fn new_or_invalid(opcode: u8) -> Self {
+        if let Some(op) = Self::new(opcode) {
+            op
+        } else {
+            Self(0xFE)
+        }
+    }
+
     /// Instantiate a new opcode from a u8 without checking if it is valid.
     ///
     /// # Safety


rakita:
This would not be great as any opcode is considered UNKNOWN and 0xFE is considered INVALID only if it is first in bytecode.

lorenzofero:
> This would not be great as any opcode is considered UNKNOWN and 0xFE is considered INVALID only if it is first in bytecode.

If you think it doesn't fit well in revm no problem! I'll close this PR. However, do you have any reference about what you said regarding the `INVALID` opcode? Thank you!

rakita:
> > This would not be great as any opcode is considered UNKNOWN and 0xFE is considered INVALID only if it is first in bytecode.
> 
> If you think it doesn't fit well in revm no problem! I'll close this PR. However, do you have any reference about what you said regarding the `INVALID` opcode? Thank you!

hm it was changed bcs of `EIP-3541` but maybe it is not correct in this case and `INVALID` should be removed so that we can have the same behaviour as other unknown opcode.

https://eips.ethereum.org/EIPS/eip-3541

lorenzofero:
> hm it was changed bcs of EIP-3541 but maybe it is not correct in this case and INVALID should be removed so that we can have the same behaviour as other unknown opcode.
[eips.ethereum.org/EIPS/eip-3541](https://eips.ethereum.org/EIPS/eip-3541)

Thanks for the reference. Anyway I think right now is better to close since it has no real utility here.

PR #835, opened 2023-10-25
basic_ref does not support async

leonardchinonso:
**This is a prototype of a suggested change.**

When `self.runtime` is `None`, the `futures::executor::block_on(f)` does not work with async.
A fix would be to use something that supports sync and async.

### DIFF
--- a/crates/revm/src/db/ethersdb.rs
+++ b/crates/revm/src/db/ethersdb.rs
@@ -53,13 +53,15 @@ impl<M: Middleware> DatabaseRef for EthersDB<M> {
     fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
         let add = eH160::from(address.0 .0);
 
-        let f = async {
-            let nonce = self.client.get_transaction_count(add, self.block_number);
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/ethersdb.rs
```diff
@@ -53,13 +53,15 @@ impl<M: Middleware> DatabaseRef for EthersDB<M> {
...
-        let (nonce, balance, code) = self.block_on(f);
+        let (nonce, balance, code) = tokio::task::block_in_place(|| {
+            let rt = tokio::runtime::Runtime::new().unwrap();
+            rt.block_on(async {
+                let nonce = self.client.get_transaction_count(add, self.block_number);
+                let balance = self.client.get_balance(add, self.block_number);
+                let code = self.client.get_code(add, self.block_number);
+                tokio::join!(nonce, balance, code)
+            })
+        });
```
rakita, line 64: This definitely is not the way. What is the problem, can you reproduce it?

rakita (changes requested):
Not sure what is the problem

rakita:
Closing this for now

PR #837, opened 2023-10-25
fix(interpreter): stack UB and `push_slice` impl

DaniPopes:
- `get_unchecked{,_mut}` is *always* UB if out of bounds. This is easy to overlook because it works correctly, but it's still UB. Use pointers instead.
- `Stack::push_slice` (#812) did not handle more than one word (32 bytes) correctly. Added simple unit tests to verify correct behavior. No performance changes.

### DIFF
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -203,7 +203,8 @@ impl Stack {
         } else {
             // Safety: check for out of bounds is done above and it makes this safe to do.
             unsafe {
-                *self.data.get_unchecked_mut(len) = *self.data.get_unchecked(len - N);
+                let data = self.data.as_mut_ptr();
+                core::ptr::copy_nonoverlapping(data.add(len - N), data.add(len), 1);
                 self.data.set_len(len + 1);
             }
             Ok(())
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -240,33 +240,49 @@ impl Stack {
         // SAFETY: length checked above.
         unsafe {
             let dst = self.data.as_mut_ptr().add(self.data.len()).cast::<u64>();
+            self.data.set_len(new_len);
+
             let mut i = 0;
 
             // write full words
-            let limbs = slice.rchunks_exact(8);
-            let rem = limbs.remainder();
-            for limb in limbs {
-                *dst.add(i) = u64::from_be_bytes(limb.try_into().unwrap());
+            let words = slice.chunks_exact(32);
+            let partial_last_word = words.remainder();
+            for word in words {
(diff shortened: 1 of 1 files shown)


rakita:
> * `get_unchecked{,_mut}` is _always_ UB if out of bounds. This is easy to overlook because it works correctly, but it's still UB. Use pointers instead.

The pointer would have the same `UB` behaviour, right? This is safe as to as the bounds are checked and we know the capacity of data vec.

DaniPopes:
No because `get_unchecked*` returns a reference, which has way more requirements for validity than pointers.

`get_unchecked` docs say:
> Calling this method with an out-of-bounds index is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) even if the resulting reference is not used.

Miri reports UB in `Stack::dup`:
image

rakita:
> No because `get_unchecked*` returns a reference, which has way more requirements for validity than pointers.
> 
> `get_unchecked` docs say:
> 
> > Calling this method with an out-of-bounds index is [undefined behavior](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) even if the resulting reference is not used.
> 
> Miri reports UB in `Stack::dup` on main: image

That seems like a false positive, reference is just a sugar here.

It seems that `Vec` uses `get_unchecked_mut` from the [slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked_mut) that works on [slice len](https://doc.rust-lang.org/src/core/slice/index.rs.html#249-254) so that is probably of reason for false positive as it does not know the capacity of the vector.

So just moving `self.data.set_len(len + 1);` above should fix miri

DaniPopes:
It's still uninitialized memory, so a reference to that is always undefined behavior.

"it works" because we have checked the bounds of the stack allocation and it happens that the implementation detail for `get_unchecked` works out of bounds too (i think), but it's still out of bounds so it shouldn't be relied upon.

rakita:
> It's still uninitialized memory, so a reference to that is always undefined behavior. It's not a false positive. Moving `set_len` might silence miri but I don't think it fixes the UB.
> 
> "it works" because we have checked the bounds of the stack allocation and it happens that the implementation detail for `get_unchecked` works out of bounds too (i think), but it's still out of bounds so it shouldn't be relied upon.

You are right, uninit memory should be dealt with, I have spent a lot of time in cpp land so this doesn't trigger me. Uninit can be a big problem and UB when you have a type that does the Drop or possibly the padding for U256 this is not the case so it is considered safe here and not a UB.

Good read here for exactly this: https://github.com/rust-lang/rust-clippy/issues/4483
It seems `MaybeUninit` was made to address this: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html

rakita (approved):
lgtm

PR #842, opened 2023-10-27
chore(state): simplify control flow in `CacheState::apply_evm_state`

rkrasiuk:
## Description

Simplify the control flow in `CacheState::apply_evm_state`. Extract account update to a separate `CacheState::apply_account_state` function.


rakita (approved):
lgtm

PR #844, opened 2023-10-27
chore(state): move account status transitions to `AccountStatus`

rkrasiuk:
## Description

- Move the complexity of account status transitions from `CacheAccount` to `AccountStatus`.
- Write additional docs for `AccountStatus` transitions.
- Deduplicate `AccountStatus` transitions in `CacheAccount::account_info_change` and `CacheAccount::change` since they were equivalent.

### DIFF
--- a/crates/revm/src/db/states/account_status.rs
+++ b/crates/revm/src/db/states/account_status.rs
@@ -27,7 +27,7 @@ impl AccountStatus {
 
     /// Account was destroyed by calling SELFDESTRUCT.
     /// This means that full account and storage are inside memory.
-    pub fn is_destroyed(&self) -> bool {
+    pub fn was_destroyed(&self) -> bool {
         matches!(
             self,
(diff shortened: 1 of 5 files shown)


crates/revm/src/db/states/account_status.rs
```diff
@@ -46,7 +27,7 @@ impl AccountStatus {
 
     /// Account was destroyed by calling SELFDESTRUCT.
     /// This means that full account and storage are inside memory.
-    pub fn was_destroyed(&self) -> bool {
+    pub fn is_destroyed(&self) -> bool {
```
rakita, line 30: `was_destroyed` is more accurate as it contains `DestroyedChanged`

rkrasiuk: ah, right, good point

rakita (approved):

PR #845, opened 2023-10-27
feat: decode KZG trusted setup at compile time

DaniPopes:


DaniPopes:
Hm this is pretty slow, 3.8s of const-eval straight on my machine

rakita:
> Hm this is pretty slow, 3.8s of const-eval straight on my machine

Idea is awesome btw, but why is it so slow?

DaniPopes:
I haven't looked into individual function performance (not sure how to do this for const-eval), but I would assume it's the combination that:
- const-eval code is interpreted without optimizations
- the file to decode is quite big
- and that we can't use `&mut`
  - every `\n` parsed iterates byte per byte, creating 2 new strings
  - every hex decode creates a new allocation for each word, then copies it into the output buffer 4000+ times, I assume this is the biggest offender

rakita:
is PR worth adding, not sure what we gain here?

DaniPopes:
Perf is unacceptable, will have to think about this more

rakita:
Will close PR, we can open it again if there is a change

PR #851, opened 2023-11-06
chore: Loop call stack

rakita:
This PR changes the recursive calls that we had to the loop call and interpreter stack.

This fully removes the possibility of problematic OS stack overflow that we had, it allows easier sharing of the shared context (As in memory, and removes lifetime from Interpreter) and better control over the call stack.

It allows easier the introduction of the handler for that part of the code and performance stayed the same

### DIFF
--- a/crates/interpreter/src/instructions/control.rs
+++ b/crates/interpreter/src/instructions/control.rs
@@ -54,9 +54,18 @@ fn return_inner(interpreter: &mut Interpreter<'_>, result: InstructionResult) {
     if len != 0 {
         let offset = as_usize_or_fail!(interpreter, offset);
         shared_memory_resize!(interpreter, offset, len);
-        interpreter.return_offset = offset;
+
+        // insert return data into `return_data_buffer`
+        interpreter.return_data_buffer = interpreter
+            .shared_memory
+            .as_mut()
+            .unwrap()
+            .slice(offset, len)
+            .to_vec()
+            .into()
+    } else {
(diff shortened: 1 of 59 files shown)


lorenzofero:
This is so exciting! I really like the way you managed the shared memory by taking it and replacing with an empty one.
It seems all the hard work has been done, however I'd really happy to give any help in case you might need it

rakita:
> This is so exciting! I really like the way you managed the shared memory by taking it and replacing with an empty one. It seems all the hard work has been done, however I'd really happy to give any help in case you might need it

Yeah, that is one of benefits of loop calls! Will circle back to your PR, as we now have this new mechanism

crates/interpreter/src/instructions/control.rs
```diff
@@ -34,50 +36,61 @@ pub fn jumpi<H: Host>(interpreter: &mut Interpreter<'_>, _host: &mut H) {
...
     gas!(interpreter, gas::BASE);
     // - 1 because we have already advanced the instruction pointer in `Interpreter::step`
     push!(interpreter, U256::from(interpreter.program_counter() - 1));
 }
 
 #[inline(always)]
-fn return_inner(interpreter: &mut Interpreter<'_>, result: InstructionResult) {
+fn return_inner(interpreter: &mut Interpreter, instruction_result: InstructionResult) {
     // zero gas cost
     // gas!(interpreter, gas::ZERO);
```
lorenzofero, line 52: Do we still need this comment for some reason?

PR #854, opened 2023-11-08
chore: use keccak256 for blockhash

minaminao:
As the discussion in https://github.com/foundry-rs/foundry/issues/6235 , currently, `blockhash(number)` returns `number`, but I think we should use `keccak256` for `blockhash`.

The main reason is:

>`blockhash`, even though it's a 256-bit value, can be represented by a much smaller number of bits  from a security point of view.
>For example, if a contract has a bug that the upper bits of the `blockhash` value are missing in inline-assembly or something, the bug would not be detected in tests.

This PR used it as we have done so before https://github.com/bluealloy/revm/pull/691 .

@DaniPopes This is a minor change and mds1 agreed that so I've sent this PR to facilitate it before I get your reply. :)

### DIFF
--- a/crates/revm/src/db/emptydb.rs
+++ b/crates/revm/src/db/emptydb.rs
@@ -1,7 +1,7 @@
 use core::{convert::Infallible, fmt, marker::PhantomData};
 use revm_interpreter::primitives::{
     db::{Database, DatabaseRef},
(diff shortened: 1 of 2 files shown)


rakita:
@DaniPopes waiting for your approval before merging

DaniPopes (approved):

PR #857, opened 2023-11-10
chore(precompiles): Make PrecompileWithAddress field public, from impl

0xmiri:
Make `PrecompileWithAddress` fields public so they can be accessible and add `From` impl for (address,precompile) tuple.


rakita (approved):
One conflict to fix, just merged one PR before this.

rakita:
I forgot to merge. Done!

PR #858, opened 2023-11-10
Fix error message for LackOfFundForMaxFee

bertmiller:
Instead of returning the fee for a transaction, the LackOfFundForMaxFee error currently returns gas_limit. This PR fixes that.

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -180,7 +180,7 @@ pub enum InvalidTransaction {
     RejectCallerWithCode,
     /// Transaction account does not have enough amount of ether to cover transferred value and gas_limit*gas_price.
     LackOfFundForMaxFee {
-        fee: u64,
+        fee: U256,
         balance: U256,
     },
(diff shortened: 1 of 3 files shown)


crates/primitives/src/result.rs
```diff
@@ -180,7 +180,7 @@ pub enum InvalidTransaction {
     RejectCallerWithCode,
     /// Transaction account does not have enough amount of ether to cover transferred value and gas_limit*gas_price.
     LackOfFundForMaxFee {
-        fee: u64,
+        fee: U256,
```
rakita, line 183: Could you make them as `Box<U256>` it will reduce the size of the type, and balance too?

rakita (approved):
lgtm, one nit if you are up to change it

rakita:
After changing the field to the Box i sow that optimism downcasted fee to the u64 so this PR even fixes that

rakita:
Will ignore fmt error, in local or CL logs it is not showing that anything is wrong

PR #859, opened 2023-11-12
chore: Fix rustdoc warnings

NianSol:
Solved rust doc warnings since the issue was open since september.
One part I was uncertain was
 _/// Takes `blob_excess_gas` saves it inside env
    /// and calculates `blob_fee` with [`BlobGasAndFee`]._
Because there's no mention of BlobGasAndFee in the code. I assumed it to be BlobExcessGasAndPrice that is what is used in the code below it

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -3,7 +3,6 @@ use alloc::vec::Vec;
 use core::fmt;
 
 
-
 /// Result of EVM execution.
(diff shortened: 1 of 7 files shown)


crates/primitives/src/result.rs
```diff
@@ -2,6 +2,8 @@ use crate::{Address, Bytes, Log, State, U256};
 use alloc::vec::Vec;
 use core::fmt;
 
+
+
```
rakita, line 6:
```suggestion
```

rakita (approved):
Thanks!

PR #862, opened 2023-11-15
perf: optimize gas overflow check

DaniPopes:
Partly reverts #592. Saturating add incurs more cost than just overflow check because we always branch if there is an overflow.

We can optimize the overflow check by using a bitwise or instead of logical one, so there is only one branch, although the compiler usually does this anyway.

Benches:
- Cachegrind: 430,444,571 to 421,462,297
- [Godbolt](https://godbolt.org/z/o5fvo33KW)

Note: I will be maintaining [a `cachegrind` branch](https://github.com/DaniPopes/revm/tree/cachegrind) on my fork where I will do these tests

### DIFF
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -90,11 +90,10 @@ impl Gas {
     /// Returns `false` if the gas limit is exceeded.
     #[inline(always)]
     pub fn record_cost(&mut self, cost: u64) -> bool {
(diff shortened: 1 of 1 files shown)


DaniPopes:
Well nevermind, looks like the compiler wins again (see llvm-mca output in [Godbolt](https://godbolt.org/z/xYPvb4b59)).

PR #864, opened 2023-11-16
refactor: simplify use statements

IaroslavMazur:


rakita (approved):
ty!

crates/interpreter/src/instructions/host.rs
```diff
@@ -1,10 +1,10 @@
-use crate::interpreter::InterpreterAction;
-use crate::primitives::{Address, Bytes, Spec, SpecId::*, B256, MAX_INITCODE_SIZE, U256};
+
```
rakita, line 1:
```suggestion
```

Rn new line

PR #869, opened 2023-11-17
refactor(primitives): don't duplicate the SpecId::enabled() logic

IaroslavMazur:
* remove unnecessary function/trait implementations

* fix docs errors

### DIFF
--- a/crates/primitives/src/specification.rs
+++ b/crates/primitives/src/specification.rs
@@ -96,22 +96,7 @@ pub trait Spec: Sized {
     /// Returns `true` if the given specification ID is enabled in this spec.
     #[inline]
     fn enabled(spec_id: SpecId) -> bool {
-        #[cfg(feature = "optimism")]
-        {
-            // If the Spec is Bedrock or Regolith, and the input is not Bedrock or Regolith,
-            // then no hardforks should be enabled after the merge. This is because Optimism's
-            // Bedrock and Regolith hardforks implement changes on top of the Merge hardfork.
-            let is_self_optimism =
-                Self::SPEC_ID == SpecId::BEDROCK || Self::SPEC_ID == SpecId::REGOLITH;
-            let input_not_optimism = spec_id != SpecId::BEDROCK && spec_id != SpecId::REGOLITH;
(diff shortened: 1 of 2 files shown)


crates/primitives/src/db.rs
```diff
@@ -61,13 +61,6 @@ pub trait DatabaseRef {
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct WrapDatabaseRef<T: DatabaseRef>(pub T);
 
-impl<F: DatabaseRef> From<F> for WrapDatabaseRef<F> {
-    #[inline]
-    fn from(f: F) -> Self {
-        WrapDatabaseRef(f)
-    }
-}
-
```
rakita, line 70: Revert this, in general, it is useful outside of lib

crates/primitives/src/specification.rs
```diff
@@ -35,11 +35,6 @@ pub enum SpecId {
 }
 
 impl SpecId {
-    #[inline]
-    pub fn try_from_u8(spec_id: u8) -> Option<Self> {
-        Self::n(spec_id)
-    }
```
rakita, line 41: Revert this, transitioning from u8 can be useful outside of lib

crates/primitives/src/specification.rs
```diff
@@ -111,7 +106,7 @@ pub trait Spec: Sized {
             }
         }
 
-        Self::SPEC_ID as u8 >= spec_id as u8
+        SpecId::enabled(Self::SPEC_ID, spec_id)
```
rakita, line 109: This is great, and you can even remove previous lines as they are checked inside `SpecId::enabled`

rakita (changes requested):
Mostly revert public function they should stay, and remove few more lines in `Spec::enabled`

crates/primitives/src/specification.rs
IaroslavMazur: Great finding!

rakita (approved):
lgtm

PR #871, opened 2023-11-18
feat: `Canyon` hardfork behind `optimism` feature flag

clabby:
## Overview

Adds the `Canyon` hardfork behind the `optimism` feature flag, and adds a macro to define the `SpecId` enum to separate the ordering of hardforks between L1 and `optimism`.

The `Canyon` hardfork does not implement any EVM changes, though it does implicitly enable `Shanghai`.

More information about `Canyon` can be found in the upstream tracking issue: https://github.com/ethereum-optimism/optimism/issues/7452

### DIFF
--- a/crates/primitives/src/specification.rs
+++ b/crates/primitives/src/specification.rs
@@ -2,32 +2,14 @@
 
 pub use SpecId::*;
 
-macro_rules! define_spec_id {
-    ($($name:ident = $value:expr),* $(,)?) => {
-        #[doc = "Specification IDs and their activation block. See: [Ethereum Execution Specs](https://github.com/ethereum/execution-specs)"]
-        #[repr(u8)]
-        #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, enumn::N)]
-        #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-        pub enum SpecId {
-            $($name = $value,)*
-        }
-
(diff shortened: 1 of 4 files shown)


crates/primitives/src/specification.rs
```diff
@@ -27,41 +46,34 @@ pub enum SpecId {
...
-                return false;
-            }
-        }
+);
 
-        our as u8 >= other as u8
-    }
-}
+#[cfg(feature = "optimism")]
+define_spec_id!(
```
rakita, line 53: I like the idea of having a separate SpecId structure, but it isn't best to wrap it inside macro.

You could do:
```
        /// Specification IDs and their activation block.
        /// 
        /// See: [Ethereum Execution Specs](https://github.com/ethereum/execution-specs)
        #[cfg(feature = "optimism")]
        #[repr(u8)]
        #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, enumn::N)]
        pub enum SpecId {
            ...
        }
```

clabby: Pulled it back out

rakita (approved):
nice!

rakita:
Just to note here, I created this change on the top of the https://github.com/bluealloy/revm/commit/1609e07c68048909ad1682c98cf2b9baa76310b5 commit. This "parent" commit is used inside reth and the idea is to freeze revm as there are some breaking revm changes so it would be less painful.

PR #875, opened 2023-11-21
temp: reth freeze branch

rakita:
Temporary branch for reth integration. It will not force reth to always bump potentially breaking changes from revm.

Currently containing:
* Optimism Canyon fork.
* alloy-primitives bump to v0.5
* fix: use maximum possible data fee for 4844 balance checks


gakonst:
Is this still needed?

rakita:
> Is this still needed?

yes, until foundry/reth transition to newest main

PR #876, opened 2023-11-21
feat: add Stack empty

mattsse:
Adds a helper function to get an empty Stack.

this can be beneficial during tracing where stack capturing is optional, and working with `Stack` over `Option<Stack>` is preferred

ref https://github.com/paradigmxyz/reth/pull/5521

### DIFF
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -46,6 +46,12 @@ impl Stack {
         }
     }
 
+    /// Instantiate a new stack without pre-allocating any memory.
+    #[inline]
+    pub fn empty() -> Self {
+        Self { data: Vec::new() }
+    }
+
     /// Returns the length of the stack in words.
     #[inline]
     pub fn len(&self) -> usize {


gakonst (approved):

rakita:
cc https://github.com/bluealloy/revm/pull/816 as this stack is going to be overwritten.

rakita (approved):

rakita (changes requested):
Wait, this would make functions invalid in Stack as they depend on STACK_LIMIT, so best not to include this.

rakita:
Additionally, I think you should just clone the stack slice as they have separate purpose that don't overlap, and when a shared stack comes this is going to be needed.

mattsse:
> Wait, this would make functions invalid in Stack as they depend on STACK_LIMIT, so best not to include this.

ah because the memory operations would be invalid, like ptr based copies?

makes sense, but then this type should not be Clone

gbrew:
> > Wait, this would make functions invalid in Stack as they depend on STACK_LIMIT, so best not to include this.
> 
> ah because the memory operations would be invalid, like ptr based copies? which require capacity?
> 
> makes sense, but then this type should not be Clone

Yeah, I was just looking at this, and I agree the #[derive(Clone)] on Stack is unsafe, as it won't preserve the invariant that the stack is fully allocated.

PR #881, opened 2023-11-24
chore: relax state generic

mattsse:
relax generic; remove trait bound

not required for struct def, as only the impl requires `Database`


DaniPopes (approved):

DaniPopes:
Don't know why Cargo.lock didn't get updated in #880

rakita (approved):
This is very interesting. It could potentially allow DB to be `DatabaseRef` so both of them can be used

PR #888, opened 2023-11-29
feat: External Contexts and simple EvmBuilder

rakita:
A lot of changes and Evm now contains just a `Context` and a `Handler`, context is divided into `EvmContext` and generic external context (Generic struct) and `Handler` contains a list of functions (handlers) that introduce logic.

`EvmBuilder` sets context and handler and allows creation of the `Evm`, where `Evm` is used in execution.

A few more handlers are added to `Handler`.

Precompile inner vec is replaced with HashSet as performance for 10 items is similar (the interesting thing is that linear search was faster than binary search for small numbers).


crates/revm/src/context.rs
```diff
@@ -1,30 +1,58 @@
...
+            evm: EvmContext::new_with_env(db, Box::new(Env::default())),
+            external: (),
+        }
+    }
+}
+
+/// EVM contexts contains data that EVM needs for execution.
 #[derive(Debug)]
-pub struct EvmContext<'a, DB: Database> {
+pub struct EvmContext<DB: Database> {
```
Wodann, line 48: I had a look at the PR to determine whether these changes would also resolve the lifetime issue that we reported in https://github.com/bluealloy/revm/pull/456/

Unfortunately, the presence of the `DB` generic in `EvmContext` makes it such that that generic type parameter also bleeds into the `Inspector` trait, thus tying the lifetime of `DB` to the `Inspector`.

This happens if `DB` is a reference type `&SomeDatabase`; a common pattern when the database is owned by someone else and is only temporarily passed into the `Evm` to run a transaction.

One way to resolve this would be the proposed solution in https://github.com/bluealloy/revm/pull/456/: The inspector receives a reference to a dynamic trait object `EvmContext` that has a generic type `DatabaseErrorT` instead of the `Database` generic type.

I'm happy to experiment with any solutions you have in mind or point you to a branch of EDR to experiment on

rakita: Nobody stops you from defining your version of Inspector and your handler register. You can even define the needed trait outside of revm and implement it on the EvmContext ([this trait](https://github.com/bluealloy/revm/pull/456/files#diff-43d7eb2000c0f76dbb285beb1d5aa9742c5ec725db8db4914fe4d6ffc06e7fcbR220-R233)) and you is in your own inspector.

This PR cuts the connection between Inspector and revm, and makes Inspector customisable in a way that anything can be used in its place.

Wodann: > This PR cuts the connection between Inspector and revm, and makes Inspector customisable in a way that anything can be used in its place.

Apologies, I missed that part. I thought they were still tied together.

I'll try defining my own inspector type to see whether that resolves the issue

rakita: One more observation (tbh not sure if it is valid) but the lifetime is removed from `EvmContext` so the `Inspector` just takes the Database there is no lifetime restriction.

so doesn't this mean you can specify the `Inspector` with db reference as.

```
impl<'a> Inspector<&'a DbType> for InspectorImpl {

}
```

`InspectorImpl` doesn't neet to have reference

rakita:
Valgrind Results:

```
==3781== Cachegrind, a cache and branch-prediction profiler
==3781== Copyright (C) 2002-2017, and GNU GPL'd, by Nicholas Nethercote et al.
==3781== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==3781== Command: target/release/snailtracer
==3781== 
--3781-- warning: L3 cache found, using its data for the LL simulation.
Snailtracer Host+Interpreter benchmark (2.1s) ...            not enough samples
==3781== 
==3781== I   refs:      455,346,299
==3781== I1  misses:          4,398
==3781== LLi misses:          2,885
==3781== I1  miss rate:        0.00%
==3781== LLi miss rate:        0.00%
==3781== 
==3781== D   refs:      237,388,032  (160,564,678 rd   + 76,823,354 wr)
==3781== D1  misses:        349,153  (    208,880 rd   +    140,273 wr)
==3781== LLd misses:        137,559  (      4,020 rd   +    133,539 wr)
==3781== D1  miss rate:         0.1% (        0.1%     +        0.2%  )
==3781== LLd miss rate:         0.1% (        0.0%     +        0.2%  )
==3781== 
==3781== LL refs:           353,551  (    213,278 rd   +    140,273 wr)
==3781== LL misses:         140,444  (      6,905 rd   +    133,539 wr)
==3781== LL miss rate:          0.0% (        0.0%     +        0.2%  )
```

PR #889, opened 2023-11-30
Fix/ec ops precompile

glitch003:


glitch003:
opened by accident - sorry

PR #892, opened 2023-12-02
fix(crates): Package Manifest Lint Reporting Levels

### LINKED ISSUE #887: Include cargo lints
On workspace level: https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

atm we still want to be compatible with 1.66 rust version as llvm 17 introduced invalid instruction: https://github.com/llvm/llvm-project/issues/69408 that is troubling for some projects. When rust bumps to new llvm this could go in.

Bug fix is included in 17.0.6: https://github.com/llvm/llvm-project/compare/llvmorg-17.0.5...llvmorg-17.0.6

refcell:
**Description**

Fixes #887 

Moves inline lint reporting levels to each crates' package manifest under the `[lints]` table.

### DIFF
--- a/crates/interpreter/Cargo.toml
+++ b/crates/interpreter/Cargo.toml
@@ -9,6 +9,12 @@ repository = "https://github.com/bluealloy/revm"
 version = "1.3.0"
 readme = "../../README.md"
 
+[lints.rust]
+unreachable_pub = "warn"
+unused_crate_dependencies = "warn"
+unused_must_use = "deny"
+rust_2018_idioms = "deny"
+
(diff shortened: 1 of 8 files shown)


rakita:
We should delay this as risc0 guys had breaking change with llvm17 so would wait couple of months for new rust release as it is not a big inconvenience for us: https://github.com/bluealloy/revm/issues/887

rakita:
Would leave it in draft as it is good reminder

rakita:
Opened alternative PR

PR #893, opened 2023-12-02
fix(bins): dependency bumps and manifest formatting

refcell:
**Description**

Small dependency bumps and package manifest formatting.


rakita:
Will close this as obsolete

PR #895, opened 2023-12-03
feat(examples): Generate Block Traces

refcell:
**Description**

Upstreams an example to generate block traces. (WIP)


refcell:
TODO: the bufwriter is currently not flushed before it is dropped as part of the eip 3155 tracer.

examples/generate_block_traces.rs
```diff
@@ -0,0 +1,173 @@
...
+                .map(|item| {
+                    let new_keys: Vec<U256> = item
+                        .storage_keys
+                        .into_iter()
+                        .map(|h256| U256::from_le_bytes(h256.0))
+                        .collect();
+                    (Address::from(item.address.as_fixed_bytes()), new_keys)
+                })
+                .collect();
+        }
```
rakita, line 136: else `env.tx.access_list = Default::default()`

PR #897, opened 2023-12-03
refactor(primitives): use the default value from AccountInfo

IaroslavMazur:


rakita (changes requested):
Would leave this as it is, it is more readable in that way

IaroslavMazur:
> Would leave this as it is, it is more readable in that way

With this change, I was, mainly, targeting a lower coupling and higher cohesion between `Account` and `AccountStatus`, such that `AccountStatus` is the one in control of its "default" value. Especially given that it already implements the `Default` trait.

It's both more logical like this, in general - and less error-prone (because if, for some reason, the "default" value of `AccountStatus` changes, it will have to be updated in just one place: the implementation of the `Default` trait for `AccountStatus`).

Having said this, I'm not insisting on integrating this, of course, but am just sharing my thought train. I'm fine with whatever you decide to do with this PR. 🙂

rakita:
> > Would leave this as it is, it is more readable in that way
> 
> With this change, I was, mainly, targeting a lower coupling and a higher cohesion between `Account` and `AccountStatus`, such that `AccountStatus` is the one in control of its "default" value. Especially given that it already implements the `Default` trait.
> 
> In my opinion, it's both more logical like this, in general - and less error-prone (because if, for some reason, the "default" value of `AccountStatus` changes, it will have to be updated in just one place: the implementation of the `Default` trait for `AccountStatus`).
> 
> Having said this, I'm not insisting on integrating this, of course, but am just sharing my thought train. I'm fine with whatever you decide to do with this PR. 🙂

It is more readable when you can see the exact type used, if `AccountStatus` was a bigger structure then I would agree, but as it is an Enum that represents a flag it is better to be explicit.

IaroslavMazur:
> > > Would leave this as it is, it is more readable in that way
> > 
> > 
> > With this change, I was, mainly, targeting a lower coupling and a higher cohesion between `Account` and `AccountStatus`, such that `AccountStatus` is the one in control of its "default" value. Especially given that it already implements the `Default` trait.
> > In my opinion, it's both more logical like this, in general - and less error-prone (because if, for some reason, the "default" value of `AccountStatus` changes, it will have to be updated in just one place: the implementation of the `Default` trait for `AccountStatus`).
> > Having said this, I'm not insisting on integrating this, of course, but am just sharing my thought train. I'm fine with whatever you decide to do with this PR. 🙂
> 
> It is more readable when you can see the exact type used, if `AccountStatus` was a bigger structure then I would agree, but as it is an Enum that represents a flag it is better to be explicit.

Alright, then, I'm closing the PR.

Thanks for considering the change!

PR #899, opened 2023-12-03
refactor(revm): leverage StorageSlot methods, where appropriate

IaroslavMazur:


crates/revm/src/db/states/bundle_account.rs
```diff
@@ -102,10 +102,10 @@ impl BundleAccount {
             match slot {
                 RevertToSlot::Some(value) => {
                     // Don't overwrite original values if present
-                    // if storage is not present set original values as current value.
+                    // if storage is not present set original value as current value.
                     self.storage
                         .entry(key)
-                        .or_insert(StorageSlot::new_changed(value, U256::ZERO))
+                        .or_insert(StorageSlot::new(value))
```
rakita, line 108: This is different behaviour

IaroslavMazur: Yes, it is (I hope you're referring to the differences between the logic of `new()` and `new_changed()`).

Why create a "changed" `StorageSlot`, if its `present_value` is being changed to the very same value as its `previous_or_original_value` after the execution of `.or_insert()` - practically, making the `StorageSlot` instance lose its "changed" status?

FWIW, it's still kinda weird to have `.present_value = value;` happen after `.or_insert()`, but this was weird before, as well. Now, `present_value` is, at least, not set to `U256::ZERO`, just to be re-set to `value` right after this.

Or am I missing something here?

rakita: I missed something, I just looked at the changed code and saw that it is different, but I didn't see that we override the `present_value` so it does not matter.

rakita (approved):

crates/revm/src/db/states/bundle_account.rs
IaroslavMazur: No problem. Glad we're on the same page now! 🙌🏻

PR #902, opened 2023-12-04
feat(examples): Custom Precompile Injection

### LINKED ISSUE #280: How to add a custom precompile contract
It seems that it is impossible to add a custom precompile when transact

refcell:
**Description**

Implements a very simple example to inject a custom precompile into the EVM Implementation.

Fixes #280


rakita:
This will be obsolete with `EvmBuilder` that should come in next weeks (pr: https://github.com/bluealloy/revm/pull/888)

So maybe it is not the best to include it

rakita:
Will close this is it will be different with `EvmBuilder`

Issue #904, opened 2023-12-05
Add `prepend_state` function to BundleState


rakita:
There was a need in reth to prepend state. This could be done with extend and removal of N reverts but cleaner solution is to add function directly to the BundleState.

So test is to add this to revm and cherry-pick it to reth_frozen branch/PR https://github.com/bluealloy/revm/pull/875 as this is currently used branch by reth/foundry and finally update it inside reth.

https://github.com/paradigmxyz/reth/pull/5683/files#diff-b80f08909adc9d005f5b75ef45b52d1f72fc935e96712f363513924e2cd34bb2R359-R368

peyha:
Seems like a really good first issue, I'll do it!

rakita:
@peyha gg

peyha:
I will try to cherry-pick the update to the reth_frozen branch and update reth accordingly this week

rakita:
> nch and update reth accordingly this w

Dont bother a lot with it, we expect to have transition in next few weeks (maybe month)

PR #907, opened 2023-12-08
feat(revm): implement prepend_state for BundleState

peyha:
For #904

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -658,6 +658,23 @@ impl BundleState {
             }
         }
     }
+
+
+    /// Prepends present the state with the given BundleState.
+    /// It adds changes from the given state but does not override any existing changes.
+    ///
+    /// Reverts are not updated.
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -658,6 +658,23 @@ impl BundleState {
...
+    pub fn prepend_state(&mut self, mut other: BundleState) {
+        let other_len = other.reverts.len();
+        // take this bundle
+        let this_bundle = std::mem::take(self);
+        // extend other bundle with this
+        other.extend(this_bundle);
+        // discard other reverts
+        other.take_n_reverts(other_len);
+        // swap bundles
+        std::mem::swap(self, &mut other)
```
rakita, line 676: This works, but we could be better with a small refactor. Make a `fn extend_state` and make `fn extend` calls it (it is just moving some code), and with this, you would not need to `take_n_reverts` and could extend only needed data.

rakita (approved):
lgtm! will merge after CI

PR #911, opened 2023-12-13
feat: add serde derives for `CacheDB` under "serde" flag

Autoparallel:
## Idea
In order to be able to take a snapshot of a `CacheDB` that can then be shared in many ways, having the `Serialize + Serialize` added is helpful. 

## Feedback
Change is minimal and is not set as default. Wrote simple unit test to check that this works. If this change is not wanted upstream, that is okay. Perhaps it can be under its own separate flag then, if need be. Please let me know!


rakita (approved):
lgtm!

PR #922, opened 2023-12-22
perf(primitives): optimize AccountInfo.is_empty()

IaroslavMazur:
This is a small optimization for AccountInfo.is_empty() with regards to the short-circuiting of bool operators in Rust.

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -352,12 +352,12 @@ impl CfgEnv {
         false
     }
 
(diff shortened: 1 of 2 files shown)


IaroslavMazur:
Reference: [Lazy boolean operators](https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators).

Quote: "the right-hand operand is only evaluated when the left-hand operand does not already determine the result of the expression"

rakita (approved):

PR #924, opened 2023-12-23
refactor(primitives): improve readability of Env.validate_tx()

IaroslavMazur:
The optimization comes from the fact that the respective variables are only created if they're, actually, going to be used after that (instead of being untouched due to an early return).

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -75,10 +75,6 @@ impl Env {
             }
         }
 
-        let gas_limit = self.tx.gas_limit;
-        let effective_gas_price = self.effective_gas_price();
-        let is_create = self.tx.transact_to.is_create();
-
         // BASEFEE tx check
         if SPEC::enabled(SpecId::LONDON) {
             if let Some(priority_fee) = self.tx.gas_priority_fee {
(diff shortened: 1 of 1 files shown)


rakita:
Would not merge this, the compiler is good enough to optimize it and not touching variables is a uncommon path.

IaroslavMazur:
> Would not merge this, the compiler is good enough to optimize it and not touching variables is a uncommon path.

Didn't know the compiler is looking for ways to make such optimizations, as well. Good to know!
In that case, these changes don't, actually, "optimize" things.

However, I believe that the proposed changes increase code readability.

IaroslavMazur:
> > e proposed changes increase code readability.
> 
> 
> 
> Would agree with readability, I would even remove the temp variable and use functions directly.

Should I, then, reopen the PR, removing the temporary variables?

rakita:
> > e proposed changes increase code readability.
> 
> Would agree with readability, I would even remove the temp variable and use functions directly.

whatever is easier for you, if you want to do it in this PR just edit the title

IaroslavMazur:
> whatever is easier for you, if you want to do it in this PR just edit the title

Done 🤝

rakita (approved):
lgtm!

PR #925, opened 2023-12-24
chore(interpreter): conditionally enable `optional_beneficiary_reward`

IaroslavMazur:
- enable the `optional_beneficiary_reward` feature automatically when the `dev` feature is enabled.

This modification seemed reasonable, being in line with how the `dev` features of the `primitives` and `revm` crates are defined.


rakita:
Most of those will be phased out gradually with Handler but it should be there

rakita (approved):

PR #935, opened 2023-12-28
feat(ethersdb): basic_ref return none  instead of panic

0xkr8os:
Instead of panicking whenever there's an error fetching the account data, return None. This behavior is similar to the EmptyDB and helpful in unit tests when using the mock provider. The tradeoff here is that any RPC client connection issues can surface as EVM execution revert errors but these can be easily mitigated earlier on in the middleware setup process. If the client connection is established, the behavior is essentially the same for callers.


rakita:
Silently failing is maybe even worse, as you would just get the wrong result and people wouldn't know if it is revm or connection.

The solution would be to propagate the error and set `Database::Error` correctly

rakita (approved):
lgtm

PR #939, opened 2023-12-30
refactor(interpreter): simplify the logic of calc.new_cost()

IaroslavMazur:


crates/interpreter/src/gas/calc.rs
```diff
@@ -311,19 +311,17 @@ fn xfer_cost(is_call_or_callcode: bool, transfers_value: bool) -> u64 {
...
+    // EIP-161: State trie clearing (invariant-preserving alternative)
+    if !SPEC::enabled(SPURIOUS_DRAGON) {
+        return NEWACCOUNT;
+    }
+
+    if transfers_value {
+        NEWACCOUNT
     } else {
         0
     }
```
rakita, line 327: In general it is very much nicer!

But would merge these two like: 
```suggestion
    // EIP-161: State trie clearing (invariant-preserving alternative)
    if SPEC::enabled(SPURIOUS_DRAGON) && !transfers_value {
        return 0;
    }

     NEWACCOUNT
```

rakita (approved):
one nit

crates/interpreter/src/gas/calc.rs
IaroslavMazur: Awesome finding!

PR #956, opened 2024-01-06
refactor(interpreter): local return_error! macro

IaroslavMazur:


crates/interpreter/src/instruction_result.rs
```diff
@@ -91,6 +91,50 @@ impl From<Halt> for InstructionResult {
...
+    };
+}
+
+#[macro_export]
+macro_rules! return_revert {
+    () => {
+        InstructionResult::Revert | InstructionResult::CallTooDeep | InstructionResult::OutOfFund
+    };
+}
+
```
rakita, line 110:
```suggestion

#[macro_export]
```

crates/interpreter/src/instruction_result.rs
```diff
@@ -237,20 +257,3 @@ impl From<InstructionResult> for SuccessOrHalt {
...
-            | InstructionResult::SelfDestruct
-    };
-}
-
-#[macro_export]
-macro_rules! return_revert {
-    () => {
-        InstructionResult::Revert | InstructionResult::CallTooDeep | InstructionResult::OutOfFund
-    };
-}
```
rakita, line 256: Would be good to add a test that covers all cases.

Something likes
```
let temp = InstructionResult::Continue;

match temp {
   return_ok!() => (),
   return_revert!() => (),
   return_error!() => (),
}
```

rakita (approved):
nits

crates/interpreter/src/instruction_result.rs
IaroslavMazur: I have considered exporting the macro to the whole crate, but decided that the macro should be kept local/private until/unless the need to access it from the outside appears in the future.

Wdyt?

crates/interpreter/src/instruction_result.rs
IaroslavMazur: Done

crates/interpreter/src/instruction_result.rs
rakita: It is a utility macro and could be useful. It is strange to half export `return_ok` and `return_revert` and not export the third one.

rakita (approved):

crates/interpreter/src/instruction_result.rs
IaroslavMazur: Included the suggested export into #961

crates/interpreter/src/instruction_result.rs
IaroslavMazur: It's only after your [add exhaustive check for return_ macros](https://github.com/bluealloy/revm/pull/956/commits/95f7e9d409329748caa222797388974004ccd2ce) commit that I've understood what exactly you wanted to be tested about the newly-added macros. 😬
My bad!

Also, thanks for fixing the name of the test function I've introduced! 🙌

PR #961, opened 2024-01-09
refactor(interpreter): consistency in all_results_are_covered()

IaroslavMazur:
While any of the 2 could've been used to achieve consistency, `{ }` has been chosen because, as far as I've read, it doesn't lead to any memory being allocated (being just a code block), while `( )`, representing the "unit type", does occupy some memory.

The latter isn't as important in the context of a test function, of course, but given that this would be the better option (over `( )`), in general, don't see why it wouldn't also be used throughout the test functions (to maintain the consistency with the production codebase functions).

Curious about your take on this, @rakita

### DIFF
--- a/crates/interpreter/src/instruction_result.rs
+++ b/crates/interpreter/src/instruction_result.rs
@@ -108,6 +108,7 @@ macro_rules! return_revert {
     };
 }
 
+#[macro_export]
 macro_rules! return_error {
     () => {
         InstructionResult::OutOfGas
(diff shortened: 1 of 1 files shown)


rakita (approved):
() does not take any memory, it is just a compiler sugar. In rust everything is a expression if you put `{}` it will compile it as `{()}` so there is no change here.

This look nicer so lgtm

IaroslavMazur:
> () does not take any memory, it is just a compiler sugar. In rust everything is a expression if you put `{}` it will compile it as `{()}` so there is no change here.

Good to know, thank you for the insight! 🙏

PR #965, opened 2024-01-10
ci(cachegrind): pass action always

rakita:
Always pass action as GITHUB_TOKEN is giving only read access from forked repositories: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#changing-the-permissions-in-a-forked-repository


0x0elliot:
nice!

PR #968, opened 2024-01-10
refactor: enhance readability

IaroslavMazur:


gakonst:
these seem unnecessary to me, personally

rakita:
would agree with @gakonst, `gas` to `change_gas` is not needed

gakonst:
Thanks @IaroslavMazur! Ideally let's try to agree on a change before implementing it, so we can make sure we do not waste your time. Appreciate it!!

rakita:
> Thanks @IaroslavMazur! Ideally let's try to agree on a change before implementing it, so we can make sure we do not waste your time. Appreciate it!!

@gakonst we are already chatting and will reopen this for the time being.

IaroslavMazur:
After some more talking to @rakita, I've just pushed a patch, removing the changes made to the names of the `gas!` and `gas_or_fail!` macros.

rakita (approved):
lgtm

PR #969, opened 2024-01-10
fix: dont calculate initcode keccak on CREATE

rakita:
Calculating keccak on CREATE opcode is unnecessary as it is only needed on CREATE2. `init_code_hash` is needed only for foundry so setting `B256::ZERO` can work.


gakonst (approved):
easy enough

mattsse (approved):
this looks reasonable

PR #975, opened 2024-01-13
refactor(log): use alloy_primitives::Log

slordua:


gakonst (changes requested):
pls remove `alloy_rlp` and fmt per clippy/fmt lints

slordua:
should be good now

slordua:
should be ready now

crates/revm/src/evm.rs
```diff
@@ -346,8 +346,8 @@ impl<EXT, DB: Database> Host for Evm<'_, EXT, DB> {
     fn log(&mut self, address: Address, topics: Vec<B256>, data: Bytes) {
         self.context.evm.journaled_state.log(Log {
             address,
-            topics,
-            data,
+            data: LogData::new(topics, data)
+                .expect("Invalid LogData: Number of topics should be <= 4"),
```
rakita, line 350: Would refactor `Host::log` function input to a `LogData `.

in essence change this: https://github.com/bluealloy/revm/blob/main/crates/interpreter/src/instructions/host.rs#L214

slordua: yea that's nicer

rakita: I didn't notice as the code was not changed but we could use `Log` directly in place of `LogData`. Other parts lgtm

slordua: could you clarify that last comment? not clear on what you're meaning

crates/interpreter/src/host.rs
```diff
@@ -48,8 +47,8 @@ pub trait Host {
     /// Set the transient storage value of `address` at `index`.
     fn tstore(&mut self, address: Address, index: U256, value: U256);
 
-    /// Emit a log owned by `address` with given `topics` and `data`.
-    fn log(&mut self, address: Address, topics: Vec<B256>, data: Bytes);
+    /// Emit a log owned by `address` with given `LogData`.
+    fn log(&mut self, address: Address, log_data: LogData);
```
rakita, line 51:
```suggestion
    fn log(&mut self, log: Log);
```

`Log` contains both `Address` and  `LogData`.

@slordua i meant this

slordua: ah ok i see it thx, will revise shortly

slordua:
is .expect ok? topics shouldn't exceed 4 right?

rakita:
> is .expect ok? topics shouldn't exceed 4 right? also wasn't sure if docs comment needed changing, still seemed appropriate to me
> 
> perhaps tho...
> 
> /// Emit a `Log` owned by `address` with given `LogData`.
> 
> would be better... idk

It is fine, lgtm

rakita (approved):

Issue #873, opened 2023-11-20
Suggesting specifc Create Inspector return type for `call` `call_end`


mattsse:
While it is possible and perhaps desired to override the address:

https://github.com/bluealloy/revm/blob/3608d2d893f973c9b86b6db60b3ffa11a61ef17c/crates/revm/src/inspector.rs#L134-L134

most of the time it is not needed.

Since we already made changes to the Inspector, I'm suggesting to make this return type a bit easier as well like

`struct CreateOutcome {result: InterpreterResult, address: Option<Address>}`

and some helper functions like `CreateOutcome::continue() -> Self`

what's the difference between the Option here:

https://github.com/bluealloy/revm/blob/3608d2d893f973c9b86b6db60b3ffa11a61ef17c/crates/revm/src/inspector.rs#L114-L118

and non optional here:

https://github.com/bluealloy/revm/blob/3608d2d893f973c9b86b6db60b3ffa11a61ef17c/crates/revm/src/inspector.rs#L129-L134

loocapro:
May I take this?

rakita:
Can you do it on top of the [external_field](https://github.com/bluealloy/revm/tree/external_field) branch? There are a few conflicting changes there

loocapro:
Sure!

mahmudsudo:
has this been taken already

loocapro:
> has this been taken already
> 
> 

Yes sorry.

loocapro:
HI @mattsse could you please clarify:

the title of the issue mentions call and call_end, while the description mentions create and create_end. Which ones should be worked?

rakita:
done

PR #980, opened 2024-01-15
Improve/inspector create

loocapro:
Related to https://github.com/bluealloy/revm/issues/873

I worked on top of [external_field branch](https://github.com/bluealloy/revm/tree/external_field) as asked by @rakita.

I defined a CreateOutcome struct and defined some helper methods to encapsulate some logic.

It's a first draft, @mattsse may I get some feedback please?


crates/interpreter/src/interpreter.rs
```diff
@@ -252,3 +271,67 @@ impl Interpreter {
...
+            | InstructionResult::Stop
+            | InstructionResult::Return
+            | InstructionResult::SelfDestruct => NextAction::Continue,
+            InstructionResult::Revert
+            | InstructionResult::CallTooDeep
+            | InstructionResult::OutOfFund => NextAction::Revert,
+            InstructionResult::FatalExternalError => NextAction::ExternalErr,
+            _ => NextAction::Other,
+        }
+    }
```
mattsse, line 336: I'm not convinced we want this because this introduced yet another intermediary type, and using the existing macros is more concise imo

I'd rather have `fn is_revert` that is a wrapper for revert!() macro

crates/interpreter/src/interpreter.rs
```diff
@@ -252,3 +271,67 @@ impl Interpreter {
...
+    pub fn result(&self) -> &InterpreterResult {
+        &self.result
+    }
+    pub fn address(&self) -> Option<Address> {
+        self.address
+    }
+    /// Retrieves the data buffer based on the interpreter's result.
+    ///
+    /// If the result is `Revert`, `CallTooDeep`, or `OutOfFund`, it returns the associated output.
+    /// Otherwise, it returns an empty `Bytes` object.
```
mattsse, line 310: hmm, looks like we only need this in the interpreter?

I think hiding this here just obfuscates the check, so I'd recommend removing this because I don't think it's very useful

crates/interpreter/src/interpreter.rs
```diff
@@ -93,26 +93,45 @@ impl Interpreter {
...
+    /// - May alter `instruction_result` in case of external errors.
+    pub fn insert_create_outcome(&mut self, create_outcome: CreateOutcome) {
+        self.return_data_buffer = create_outcome.data_buffer();
 
-        match result.result {
-            return_ok!() => {
-                push_b256!(self, address.unwrap_or_default().into_word());
-                self.gas.erase_cost(result.gas.remaining());
-                self.gas.record_refund(result.gas.refunded());
+        match create_outcome.next_action() {
```
rakita, line 124: Would leave this unchanged, it looks better, but macros are used so we don't have performance degradation. Would definitely consider this later but not as part of this PR.

mattsse (changes requested):
I like the wrapper type a lot, though I think the additional functions are not that useful

wdyt @rakita

loocapro:
> I like the wrapper type a lot, though I think the additional functions are not that useful
> 
> wdyt @rakita

Thanks for the feedback! I have followed your comments and have added a possible improvement over encapsulation of the `InstructionResult` structure.

What you guys think?

crates/interpreter/src/interpreter.rs
```diff
@@ -252,3 +280,69 @@ impl Interpreter {
...
+/// Represents the outcome of a create operation in an interpreter.
+///
+/// This struct holds the result of the operation along with an optional address.
+/// It provides methods to determine the next action based on the result of the operation.
+pub struct CreateOutcome {
+    // The result of the interpreter operation.
+    pub result: InterpreterResult,
+    // An optional address associated with the create operation.
+    pub address: Option<Address>,
+}
```
rakita, line 293: Would move this to its own file.

crates/interpreter/src/interpreter.rs
```diff
@@ -93,24 +93,52 @@ impl Interpreter {
...
+    ///
+    /// # Behavior
+    ///
+    /// The function updates the `return_data_buffer` with the data from `create_outcome`.
+    /// Depending on the `next_action` indicated by `create_outcome`, it performs one of the following:
+    ///
+    /// - `NextAction::Continue`: Pushes the address from `create_outcome` to the stack, updates gas costs, and records any gas refunds.
+    /// - `NextAction::Revert`: Pushes `U256::ZERO` to the stack and updates gas costs.
+    /// - `NextAction::ExternalErr`: Sets the `instruction_result` to `InstructionResult::FatalExternalError`.
+    /// - Other actions: Pushes `U256::ZERO` to the stack.
```
rakita, line 113: This comment needs to be modified

loocapro:
> two nits left.
> 
> Do you want to do similar thing for `CallOutcome`? It can be in this PR or the next one

Sure, i am gonna do in a new pr!

rakita:
Hey @loocapro, can you rebase these changed to `main`, `external_field` branch is merged to it?

rakita (approved):
lgtm

PR #981, opened 2024-01-16
fix: use maximum possible data fee for 4844 balance checks

Rjected:
ref https://github.com/paradigmxyz/reth/issues/6077

The balance checks shown in EIP-4844 use the maximum possible blob fee:
```python
        # modify the check for sufficient balance
        max_total_fee = tx.gas * tx.max_fee_per_gas
        if get_tx_type(tx) == BLOB_TX_TYPE:
            max_total_fee += get_total_blob_gas(tx) * tx.max_fee_per_blob_gas
        assert signer(tx).balance >= max_total_fee
```

This implements what is written in the spec, introducing a new method `calc_max_data_fee` to calculate `get_total_blob_gas(tx) * tx.max_fee_per_blob_gas`

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -47,6 +47,19 @@ impl Env {
         })
     }
 
+    /// Calculates the maximum [EIP-4844] `data_fee` of the transaction.
+    ///
+    /// This is used for ensuring that the user has at least enough funds to pay the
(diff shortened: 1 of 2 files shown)


rakita (approved):
lgtm, we should backport this to `reth_freeze` branch

Issue #967, opened 2024-01-10
Conver Optimism panic into graceful error


mattsse:
if the encoded tx field is missing transact currently panics:

https://github.com/bluealloy/revm/blob/b9ecc0034ff73f97c7cb25f4b3afb609b0fc0ea5/crates/revm/src/evm_impl.rs#L398-L398

https://github.com/bluealloy/revm/blob/b9ecc0034ff73f97c7cb25f4b3afb609b0fc0ea5/crates/revm/src/evm_impl.rs#L398-L398

this should be an error instead, perhaps like:

https://github.com/bluealloy/revm/blob/b9ecc0034ff73f97c7cb25f4b3afb609b0fc0ea5/crates/primitives/src/result.rs#L249-L250

ref https://github.com/paradigmxyz/reth/issues/6008

rakita:
As we want to phase out the feature but retain present optimism behaviour imo it would be good to add
`Error::Custom(String)` to main revm Error: https://github.com/bluealloy/revm/blob/main/crates/primitives/src/result.rs#L131

This could be temporary as we should introduce some generic for error to work with the custom handler, but atm I am not sure where to tie it so String will do the job.

PR #982, opened 2024-01-16
feat: Convert optimism panic into graceful error

kekonen:
In #967 there is described a temporary solution for a panic in optimism handler code.
So far, it is a simple extension for the `EVMError` enum, containing error explanation as a String.
All related panics are replaced with the code returning this error.

This is my first quick issue here, looking forward for feedback.

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -1,5 +1,5 @@
 use crate::{Address, Bytes, Log, State, U256};
-use alloc::{boxed::Box, vec::Vec};
+use alloc::{boxed::Box, string::String, vec::Vec};
 use core::fmt;
 
 /// Result of EVM execution.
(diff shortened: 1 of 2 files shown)


crates/primitives/src/result.rs
```diff
@@ -135,6 +135,11 @@ pub enum EVMError<DBError> {
     Header(InvalidHeader),
     /// Database error.
     Database(DBError),
+    /// Custom error.
+    ///
+    /// Currently implemented for the Optimism handler. This is an interim solution pending
+    /// the development of a more generic error mechanism for integration with custom handlers.
```
rakita, line 141: Would change this comment to reflect the present state and not a possible/unknown future.

```suggestion
    /// Useful for handler registers where custom logic would want to return their own custom error.
```

rakita (approved):
one nit on comment

Issue #872, opened 2023-11-20
Clarify Inspector::call retrun type


mattsse:
not obvious what the range of the return type represents

https://github.com/bluealloy/revm/blob/3608d2d893f973c9b86b6db60b3ffa11a61ef17c/crates/revm/src/inspector.rs#L82-L90

ref https://github.com/paradigmxyz/reth/pull/5502

rakita:
Done by adding structs

PR #985, opened 2024-01-17
refactor(Inspector) add CallOutcome to call/call_end

loocapro:
Related to #872

Added wrapper type on Inspector.call, refactored a bit the interpreter call_outcome function.

@rakita feel free to point any other place where I could contribute a bit!

Thanks a lot!


rakita (approved):
lgtm! 

Comments are a little bit verbose but it is fine.

PR #987, opened 2024-01-18
chore: fix serde std flags for no-std build

CeciliaZ030:


bins/revme/Cargo.toml
```diff
@@ -16,17 +16,20 @@ indicatif = "0.17"
...
     "c-kzg",
 ] }
 alloy-rlp = { version = "0.3", default-features = false, features = [
     "arrayvec", 
     "derive",
 ] }
-serde = { version = "1.0", features = ["derive", "rc"] }
-serde_json = { version = "1.0", features = ["preserve_order"] }
+serde = { version = "1.0", default-features = false, features = ["derive", "rc"] }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"]}
```
rakita, line 27: `preserve_order` was a nice feature, does it work with `no_std`?

bins/revme/Cargo.toml
```diff
@@ -16,17 +16,20 @@ indicatif = "0.17"
 plain_hasher = "0.2"
 revm = { path = "../../crates/revm", version = "3.5.0", default-features = false, features = [
     "ethersdb",
-    "std",
     "serde",
     "c-kzg",
```
rakita, line 20: `c-kzg` is hard to build in no_std environment, it is `c` lib. But c-kzg is needed for the test so it needs to be included.

I am fine with making `revme` std only as it is binary and more like a tool/utility.

mratsim: Maybe it can be a test only / bench only dependencies?

bins/revme/Cargo.toml
CeciliaZ030: it works, sry i added back at one place for miss the other :)

bins/revme/Cargo.toml
CeciliaZ030: I looked again, and it does make sense to keep this on std cuz it's a command-line app. I'm gonna revert changes for `revme`

rakita (approved):
lgtm!

bins/revme/Cargo.toml
leonardoalt: I think `preserve_order` doesn't really work with `no-std` because of https://github.com/serde-rs/json/blob/master/Cargo.toml#L58 .  That's why I had to remove it in https://github.com/bluealloy/revm/commit/cbbfe9c9f2f3e17d4e67e979f954cd9e67b45753 in order to make `revme` no-std. But that should be fine in this repo because as pointed out it's only required by `revme` which is a separate tool, also I don't think it would be possible to fully make `revme` `no-std` because of `c-kzg`.

crates/revm/Cargo.toml
```diff
@@ -22,8 +22,8 @@ revm-precompile = { path = "../precompile", version = "2.2.0", default-features
 auto_impl = { version = "1.1", default-features = false }
 
 # Optional
-serde = { version = "1.0", features = ["derive", "rc"], optional = true }
-serde_json = { version = "1.0", features = ["preserve_order"], optional = true }
+serde = { version = "1.0", default-features = false, features = ["derive", "rc"], optional = true }
+serde_json = { version = "1.0", default-features = false, features = ["alloc", "preserve_order"], optional = true }
```
leonardoalt, line 26: @rakita how critical is having `preserve_order` here?

rakita: It seems not at all critical, I thought it is there to format output, but after checking, this is not the case. I am okay with removing it.

leonardoalt: Nice, thanks!

rakita: Just curious why would you need binary to be `no_std` compatible?

leonardoalt: My need (and in this case @CeciliaZ030 's as well) is not the binary per se, but rather revm+serde. We want no-std revm to make ZK proofs for it with [powdr](https://github.com/powdr-labs/powdr) , which works fine. However it would be great to also be able to enable the `serde` feature in revm, in order to have better UX in message passing between the host and the guest when making ZK proofs.

rakita: Makes sense, I don't focus a lot of my attention on `revme` (binary) as it is used only for testing, was curious to see if I need to revise that view.

Okay, I just noticed that `preserve_order` is in both revm and revme. `preserved_order` is nice in `revm` as it formats trace json output that is easier to read, still okay with removing it.

Can you make a PR?

leonardoalt: Sure! We can also wait to see if https://github.com/serde-rs/json/pull/1101 gets merged, then we don't need to remove it.

rakita: Your call, if this blocks you or you can use this right now it would be best to make PR

PR #989, opened 2024-01-18
fix: optimism gas refunds

Wollac:
Prior to the Regolith update, deposit transactions should not receive gas refunds as ETH. For all other cases, non-deposit or post-regolith gas refunds are handled as in Mainnet.


Wollac:
I am a bit confused, why `test_consume_gas_with_refund` is failing. This should now exactly match the bevior before the API changes: https://github.com/bluealloy/revm/blob/b28c154759934e3577f959f1c237e5210c201777/crates/revm/src/handler/optimism.rs#L84-L89
and my full OP block execution tests with this PR pass while I was getting receipt mismatches before.

gakonst:
cc @clabby

Wollac:
Actually, the test was wrong.
With #888, the optimism tests were simply copied over, but before that, `handle_call_return` did *not* include the refund calculation, which happened in `calculate_gas_refund`. Now both happen in `first_frame_return` and the tests need to be adapted.

To make this a bit clearer, I also renamed `handle_call_return` to `first_frame_return`, which is the actual name in the handler.

rakita:
Makes sense! would want to check it up as I messed it up, but it will be merged today. Thank you!

rakita (approved):

PR #1002, opened 2024-01-22
chore: expose InstructionResult getters in Interpreter result

mattsse:
The `InterpreterResult` wraps an `InstructionResult`

This also exposes the `is_{error,success}` functions on the `InterpreterResult`

moves struct defs around and makes calls const.

### DIFF
--- a/crates/interpreter/src/interpreter.rs
+++ b/crates/interpreter/src/interpreter.rs
@@ -53,6 +53,35 @@ pub struct Interpreter {
     pub next_action: Option<InterpreterAction>,
 }
 
+/// The result of an interpreter operation.
+#[derive(Debug, Clone)]
+pub struct InterpreterResult {
+    /// The result of the instruction execution.
+    pub result: InstructionResult,
+    /// The output of the instruction execution.
+    pub output: Bytes,
+    /// The gas usage information.
+    pub gas: Gas,
+}
+
+#[derive(Debug, Clone)]
+pub enum InterpreterAction {
+    SubCall {
+        /// Call inputs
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/interpreter.rs
```diff
@@ -53,31 +53,6 @@ pub struct Interpreter {
...
-        /// This value must be ignored if `self.return_len` is 0.
-        return_memory_offset: Range<usize>,
-    },
-    Create {
-        inputs: Box<CreateInputs>,
-    },
-    Return {
-        result: InterpreterResult,
-    },
-}
```
rakita, line 79: Can you revert this? I liked more having structs first than impls

mattsse: sure, but I very much dislike this style:

```rust
struct A;
struct B;

impl A {}
impl B {}
```

imo this makes it a lot harder to navigate

mattsse: reverted

crates/interpreter/src/interpreter.rs
```diff
@@ -300,3 +304,23 @@ impl Interpreter {
...
+    /// Returns whether the instruction result is a success.
+    #[inline]
+    pub const fn is_ok(&self) -> bool {
+        self.result.is_ok()
+    }
+
+    /// Returns whether the instruction result is a revert.
+    #[inline]
+    pub const fn is_revert(&self) -> bool {
+        self.result.is_ok()
```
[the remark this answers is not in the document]
rakita: Good catch!

PR #1003, opened 2024-01-22
feat(inspector): Share call/create inputs in Inspector call_end/create_end

rakita:
Added `Rc<RefCell<>>` to call/create inspector handler so we can share inputs stack.

This functionality was removed with a call loop as the input is discarded after the frame is created and not available in *_end Inspector function, this removal became a problem for downstream libraries that use Inspector.

### DIFF
--- a/crates/revm/src/inspector/handler_register.rs
+++ b/crates/revm/src/inspector/handler_register.rs
@@ -1,3 +1,5 @@
+use core::cell::RefCell;
+
 use crate::{
     db::Database,
     handler::register::{EvmHandler, EvmInstructionTables},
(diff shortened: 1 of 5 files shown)


crates/revm/src/inspector/handler_register.rs
```diff
@@ -109,7 +119,11 @@ pub fn inspector_handle_register<'a, DB: Database, EXT: GetInspector<'a, DB>>(
                         return FrameOrResult::Result(output.interpreter_result);
                     }
                     // first call frame does not have return range.
-                    context.evm.make_call_frame(&call_inputs, 0..0)
+                    let out = context.evm.make_call_frame(&call_inputs, 0..0);
+                    call_input_stack_inner
+                        .borrow_mut()
+                        .push(Box::new(call_inputs));
```
DaniPopes, line 125: does this need to be boxed?

PR #1009, opened 2024-01-24
feat(op): Ecotone hardfork

clabby:
## Overview

Adds definitions and support for the upcoming Ecotone hardfork on Optimism.

### Details

The `Ecotone` hardfork is activated alongside `Cancun`, with few execution layer modifications of our own. The rollup will now be using 4844 DA, which involves some changes to the L1 fee calculation.

1. `Ecotone` L1 fee specifications: https://github.com/ethereum-optimism/specs/blob/main/specs/exec-engine.md#ecotone-l1-cost-fee-changes-eip-4844-da
2. `Ecotone` L1 fee reference implementation: https://github.com/ethereum-optimism/op-geth/blob/a4e85ec465e2de3f962d757d8f9b561b5461f861/core/types/rollup_cost.go

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -872,6 +872,11 @@ pub const fn spec_opcode_gas(spec_id: SpecId) -> &'static [OpInfo; 256] {
                     const TABLE: &[OpInfo;256] = &make_gas_table(SpecId::CANYON);
                     TABLE
                 }
+                #[cfg(feature = "optimism")]
+                SpecId::ECOTONE => {
+                    const TABLE: &[OpInfo;256] = &make_gas_table(SpecId::ECOTONE);
+                    TABLE
+                }
             }
         };
     }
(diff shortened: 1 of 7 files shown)


rakita (approved):
@clabby is `context.evm.env.cfg.optimism` only there to switch between mainnet and optimism transaction so both of them can be used? If this is the case, i want to remove it as `handler_register` in EvmBuilder allows having both optimism and mainnet together.

clabby:
@rakita Yep! The idea was to have runtime configuration for the optimism features, if `revm` is compiled with the `optimism` feature flag enabled. This was for tools like foundry to be able to swap between `optimism` and default `ethereum` execution at runtime, for example in fork tests.

This can be abstracted into the handler instead as well, but the ability to switch between the optimism and default execution at runtime is still desired.

rakita:
> @rakita Yep! The idea was to have runtime configuration for the optimism features, if `revm` is compiled with the `optimism` feature flag enabled. This was for tools like foundry to be able to swap between `optimism` and default `ethereum` execution at runtime, for example in fork tests.
> 
> This can be abstracted into the handler instead as well, but the ability to switch between the optimism and default execution at runtime is still desired.

Yeah, this can be abstracted with handler register! Good to know, will remove the flag.

Optimism feature is still needed for additional field (l1block) and for additional SpecIds. But will see to make those generic in future.

PR #1010, opened 2024-01-25
fix(State): Preserve original values on delete revert

rakita:
Delete an account from bundle state only if the original value is None, if we delete non original value we will lose this information.

Original account `None` means that account din't exist.

On revert if we want to revert on non existing account this means that it is not existing, so we revert on non existing account.


rakita:
closes https://github.com/paradigmxyz/reth/issues/5852 and https://github.com/paradigmxyz/reth/issues/5917

PR #1004, opened 2024-01-22
Burntpix criterion bench

loocapro:
Related to #986.

Hi @rakita ,

I tried to follow the criterion benches and implement the Burntpix bench from there.

I tried to find the cleanest way to initialise the REVM with empty db from some static genesis allocations, please tell me if you got a better idea :)

Waiting for your feedback !

Thanks a lot!

PR #1016, opened 2024-01-26
fix: const to static

loocapro:
Hi @DaniPopes , thanks for your suggestion. 

Here the changes requested.


DaniPopes (approved):
Thanks! For reference: https://github.com/bluealloy/revm/pull/1004#discussion_r1467839868

PR #1040, opened 2024-02-02
Add `ahash`'s `RandomState` for HashDoS prevention

jtriley-eth:
Adds `ahash` crate to `primitives`, exports `RandomState` from `primtives`, and adds `RandomState` to `TransientStorage` definition and instantiation.

This prevents HashDoS as documented in the [hashbrown crate](https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html#method.with_hasher).


jtriley-eth:
fixed in  #1041

PR #1059, opened 2024-02-06
chore: bump c-kzg and enable blst portable feature

Rjected:
This bumps `c-kzg-4844` and enables blst's `portable` feature. The fact that we have to add `blst` as an optional dependency to do this is a mistake on my part, since we should have introduced `portable` as a feature in `c-kzg-4844` directly. Added TODOs for that


rakita (approved):
lgtm! btw what was the problem without portable?

Rjected:
> lgtm! btw what was the problem without portable?

The issue was this in reth, probably because we built the release binaries on a machine with instructions the user didn't have

https://github.com/paradigmxyz/reth/issues/6220

Portable should allow for architecture or instruction-specific implementations to be selected at runtime

PR #1069, opened 2024-02-08
fix: typo on internal append_handle_register methods

Rjected:
Docs said `append handle register` so I just did a quick rename


rakita:
Would do other way, and rename everything to `append_handler_register`

Rjected:
> Would do other way, and rename everything to `append_handler_register`

sounds good, just renamed

PR #1073, opened 2024-02-10
refactor(precompile): make use of padding utilities, simplify secp256k1

DaniPopes:
... and some other minor refactors.

In secp256k1 both backends expect a 64-byte signature and the recovery ID in two separate arguments, therefor there is no point in packing them into a 65-byte array only to unpack them right after.

### DIFF
--- a/crates/precompile/src/hash.rs
+++ b/crates/precompile/src/hash.rs
@@ -1,14 +1,13 @@
 use super::calc_linear_cost_u32;
-use crate::{Error, Precompile, PrecompileResult, PrecompileWithAddress, StandardPrecompileFn};
-use sha2::*;
+use crate::{Error, Precompile, PrecompileResult, PrecompileWithAddress};
+use sha2::Digest;
 
-pub const SHA256: PrecompileWithAddress = PrecompileWithAddress(
-    crate::u64_to_address(2),
-    Precompile::Standard(sha256_run as StandardPrecompileFn),
-);
+pub const SHA256: PrecompileWithAddress =
+    PrecompileWithAddress(crate::u64_to_address(2), Precompile::Standard(sha256_run));
+
 pub const RIPEMD160: PrecompileWithAddress = PrecompileWithAddress(
     crate::u64_to_address(3),
-    Precompile::Standard(ripemd160_run as StandardPrecompileFn),
+    Precompile::Standard(ripemd160_run),
 );
 
 /// See: <https://ethereum.github.io/yellowpaper/paper.pdf>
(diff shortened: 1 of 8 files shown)


mattsse (approved):
changes lgtm

defer to @rakita re function arg changes

crates/precompile/src/hash.rs
```diff
@@ -32,8 +31,11 @@ fn ripemd160_run(input: &[u8], gas_limit: u64) -> PrecompileResult {
...
     } else {
-        let mut ret = [0u8; 32];
-        ret[12..32].copy_from_slice(&ripemd::Ripemd160::digest(input));
-        Ok((gas_used, ret.to_vec()))
+        let mut hasher = ripemd::Ripemd160::new();
+        hasher.update(input);
+
+        let mut output = [0u8; 32];
+        hasher.finalize_into((&mut output[12..]).into());
+        Ok((gas_used, output.to_vec()))
```
rakita, line 39: The previous version seems better

crates/precompile/src/lib.rs
```diff
@@ -181,8 +181,8 @@ impl Precompiles {
 
     /// Returns the precompile for the given address.
     #[inline]
-    pub fn get(&self, address: &Address) -> Option<Precompile> {
-        self.inner.get(address).cloned()
+    pub fn get(&self, address: &Address) -> Option<&Precompile> {
```
rakita, line 184: Precompile is just a function, as we know concrete type, it is okay to clone it

crates/precompile/src/modexp.rs
```diff
@@ -139,12 +137,7 @@ fn byzantium_gas_calc(base_len: u64, exp_len: u64, mod_len: u64, exp_highp: &U25
     let iter_count = U256::from(calculate_iteration_count(exp_len, exp_highp));
     // mul * iter_count bounded by 2^195 < 2^256 (no overflow)
     let gas = (mul * iter_count) / U256::from(20);
-
-    if gas.as_limbs()[1] != 0 || gas.as_limbs()[2] != 0 || gas.as_limbs()[3] != 0 {
-        u64::MAX
-    } else {
-        gas.as_limbs()[0]
-    }
+    gas.saturating_to()
```
rakita, line 140: Liking this a lot!

crates/precompile/src/hash.rs
DaniPopes: `::digest` returns a new array so this avoids the `copy_from_slice` by directly using the output buffer

PR #1074, opened 2024-02-10
perf/refactor(precompile): simplify bn128 precompile implementations

DaniPopes:
Reduce stack and heap allocations, factor out some common code to separate functions/closures.


rakita (approved):
lgtm! nice refactor

PR #1075, opened 2024-02-10
perf(precompile): don't allocate if padding is not needed

DaniPopes:
Return `Cow<[u8 (; LEN)]>` in padding functions when padding is not needed due to input being long enough.

~~First commits are https://github.com/bluealloy/revm/pull/1073 and https://github.com/bluealloy/revm/pull/1074.~~

~~Actual changes in the last commit.~~

### DIFF
--- a/crates/precompile/src/utilities.rs
+++ b/crates/precompile/src/utilities.rs
@@ -5,7 +5,7 @@ use core::cmp::min;
 ///
 /// Returns the first `LEN` bytes if it does not need padding.
 #[inline(always)]
-pub fn get_right_padded<const LEN: usize>(data: &[u8], offset: usize) -> Cow<'_, [u8; LEN]> {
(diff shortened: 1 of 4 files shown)


crates/precompile/src/utilities.rs
```diff
@@ -1,42 +1,148 @@
...
+///
+/// Returns the first `LEN` bytes if it does not need padding.
 #[inline(always)]
-pub fn get_right_padded<const LEN: usize>(data: &[u8], offset: usize) -> [u8; LEN] {
-    let mut padded = [0; LEN];
-    let start = min(offset, data.len());
-    let end = min(start.saturating_add(LEN), data.len());
-    padded[..end - start].copy_from_slice(&data[start..end]);
-    padded
+pub fn get_right_padded<const LEN: usize>(data: &[u8], offset: usize) -> Cow<'_, [u8; LEN]> {
```
rakita, line 8: As we have `right_pad`, would rename this to be more aligned, something like `right_par_with_offset`

rakita (approved):

Issue #1072, opened 2024-02-10
perf: make precompiles accept and return `Bytes` to reduce allocations


DaniPopes:
Some precompiles (like `identity`, KZG point evaluation) can avoid allocating completely by re-using the `CallInputs::input` `Bytes` structure's allocation, or by returning `Bytes::from_static` when the return value is statically known at compile time.

PR #1085, opened 2024-02-13
perf(precompile): use `Bytes` in precompile functions

DaniPopes:
Closes https://github.com/bluealloy/revm/issues/1072.

### DIFF
--- a/crates/precompile/src/identity.rs
+++ b/crates/precompile/src/identity.rs
@@ -1,5 +1,6 @@
 use super::calc_linear_cost_u32;
 use crate::{Error, Precompile, PrecompileResult, PrecompileWithAddress};
+use revm_primitives::Bytes;
 
 pub const FUN: PrecompileWithAddress =
     PrecompileWithAddress(crate::u64_to_address(4), Precompile::Standard(identity_run));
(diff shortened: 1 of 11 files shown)


crates/precompile/src/modexp.rs
```diff
@@ -108,15 +103,16 @@ where
     }
 
     // Padding is needed if the input does not contain all 3 values.
-    let base = right_pad_with_offset_vec(input, 0, base_len);
-    let exponent = right_pad_with_offset_vec(input, base_len, exp_len);
-    let modulus = right_pad_with_offset_vec(input, base_len.saturating_add(exp_len), mod_len);
+    let input = right_pad_vec(input, base_len + exp_len + mod_len);
+    let (base, input) = input.split_at(base_len);
+    let (exponent, modulus) = input.split_at(exp_len);
+    debug_assert_eq!(modulus.len(), mod_len);
```
DaniPopes, line 109: Made this allocate only once in case input is too short.

`modulus.len()` must be `mod_len` due to the padding above, this is the same as slicing the input 3 times.

crates/precompile/src/utilities.rs
```diff
@@ -26,8 +25,7 @@ pub fn right_pad<const LEN: usize>(data: &[u8]) -> Cow<'_, [u8; LEN]> {
         Cow::Borrowed(data.try_into().unwrap())
     } else {
         let mut padded = [0; LEN];
-        let end = min(LEN, data.len());
-        padded[..end].copy_from_slice(&data[..end]);
+        padded[..data.len()].copy_from_slice(data);
```
DaniPopes, line 28: I realized these `min` calls were unnecessary since `data.len()` is always less than the input length by definition.

crates/precompile/src/identity.rs
```diff
@@ -13,10 +14,10 @@ const IDENTITY_PER_WORD: u64 = 3;
...
 /// See: <https://ethereum.github.io/yellowpaper/paper.pdf>
 /// See: <https://etherscan.io/address/0000000000000000000000000000000000000004>
-fn identity_run(input: &[u8], gas_limit: u64) -> PrecompileResult {
+fn identity_run(input: &Bytes, gas_limit: u64) -> PrecompileResult {
     let gas_used = calc_linear_cost_u32(input.len(), IDENTITY_BASE, IDENTITY_PER_WORD);
     if gas_used > gas_limit {
         return Err(Error::OutOfGas);
     }
-    Ok((gas_used, input.to_vec()))
+    Ok((gas_used, input.clone()))
```
DaniPopes, line 22: This is the biggest win IMO, avoids malloc+memcpy.

rakita (approved):
lgtm!

rakita:
nit clippy

PR #1088, opened 2024-02-13
feat: expose builder stages

Rjected:
Exposes builder stages so we can write functions that return a builder type at a specific stage, for example:
```rust
    /// Returns new EVM builder
    fn evm(&self) -> EvmBuilder<'static, SetGenericStage, (), EmptyDB> {
        Default::default()
    }
```

### DIFF
--- a/crates/revm/src/lib.rs
+++ b/crates/revm/src/lib.rs
@@ -28,7 +28,7 @@ pub mod optimism;
 
 // Export items.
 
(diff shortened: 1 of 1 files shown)


onbjerg (approved):

Rjected:
the original idea was to have the processor configure the evm, instead of having args in the trait methods, but now we're going to just return the `Evm` directly. closing since we don't need it any more

PR #1096, opened 2024-02-16
fix: modify cfg spec_id (#1095)

### LINKED ISSUE #1095: SpecID not updated when modify_spec_id
https://github.com/bluealloy/revm/blob/main/crates/revm/src/handler.rs#L185-L199

```rust
    pub fn modify_spec_id(&mut self, spec_id: SpecId) {
        if self.cfg.spec_id == spec_id {
            return;
        }

        let registers = core::mem::take(&mut self.registers);
        // register for optimism is added as a register, so we need to create mainnet handler here.
        let mut handler = Handler::mainnet_with_spec(spec_id);
        // apply all registers to default handeler and raw mainnet instruction table.
        for register in registers {
            handler.append_handler_register(register)
        }
        handler.cfg = self.cfg();
        *self = handler;
    }
 ```
 
 Although the handler's behavior is updated, the cfg is still the old one with old SpecID.

publicqi:
This PR fixes #1095

### DIFF
--- a/crates/revm/src/handler.rs
+++ b/crates/revm/src/handler.rs
@@ -195,6 +195,7 @@ impl<'a, EXT, DB: Database> EvmHandler<'a, EXT, DB> {
             handler.append_handler_register(register)
         }
         handler.cfg = self.cfg();
+        handler.cfg.spec_id = spec_id;
         *self = handler;
     }
 }


rakita (approved):
Good catch

PR #1097, opened 2024-02-16
perf(state): avoid cloning full account

rkrasiuk:
## Description

Avoid cloning full plain account inside cache account. Previously, the plain account was cloned with storage which can get expensive.


mattsse (approved):

shekhirin (approved):

mattsse:
needs .as_ref

PR #1104, opened 2024-02-19
feat: split off serde_json dependency to its own feature

DaniPopes:
`serde_json` is not necessarily wanted when enabling `serde` impls.

Technically a breaking change.

### DIFF
--- a/crates/revm/Cargo.toml
+++ b/crates/revm/Cargo.toml
@@ -55,7 +55,8 @@ std = [
     "revm-interpreter/std",
     "revm-precompile/std",
 ]
-serde = ["dep:serde", "dep:serde_json", "revm-interpreter/serde"]
+serde = ["dep:serde", "revm-interpreter/serde"]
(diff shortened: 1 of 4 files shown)


mattsse (approved):
reasonable,

looks like serde_json is only used by the TracerEip3155 inspector
which is useful for testing, so this should probably stay here.

we could however feature gate inspector impls, which would make the json feature obsolete?

crates/revm/Cargo.toml
```diff
@@ -55,7 +55,8 @@ std = [
     "revm-interpreter/std",
     "revm-precompile/std",
 ]
-serde = ["dep:serde", "dep:serde_json", "revm-interpreter/serde"]
+serde = ["dep:serde", "revm-interpreter/serde"]
+json = ["serde", "dep:serde_json"]
```
rakita, line 59: would rename to `serde-json` as `json` can be misleading.

PR #1113, opened 2024-02-21
request: Remove DatabaseRef bound on CacheDB

jasalper:
The bound exists already in the implementations, no need to further constrain the struct itself. 

I removed the bound from all of the functions I could, but let me know if you'd like me to re-add the bound on any (like insert contract or insert account info)


rakita (approved):
lgtm

PR #1115, opened 2024-02-21
fix: replace tuple in sstore return with struct

### LINKED ISSUE #1114: Replace tuple in sstore return with struct
https://github.com/bluealloy/revm/blob/main/crates/revm/src/journaled_state.rs#L655
Replace this output with a struct that defines the names of variables

qiweiii:
fix #1114

### DIFF
--- a/crates/interpreter/src/host.rs
+++ b/crates/interpreter/src/host.rs
@@ -56,8 +56,12 @@ pub trait Host {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SStoreResult {
+    /// Value of the storage when it is first read
     pub original_value: U256,
+    /// Current value of the storage
     pub present_value: U256,
+    /// Values that that is set
     pub new_value: U256,
+    /// Is storage slot loaded from database
     pub is_cold: bool,
 }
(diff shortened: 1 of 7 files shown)


crates/interpreter/src/host.rs
```diff
@@ -56,3 +51,13 @@ pub trait Host {
     /// Mark `address` to be deleted, with funds transferred to `target`.
     fn selfdestruct(&mut self, address: Address, target: Address) -> Option<SelfDestructResult>;
 }
+
+/// Represents the result of an `sstore` operation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct SStoreResult {
```
qiweiii, line 58: not sure if this is the right file to place it

rakita: The place is fine, would add more comments.

original: `Value of the storage when it is first read`
present: `Current value of the storage`
new_value: `Values that that is set`
is_cold: `Is storage slot loaded from database`

rakita (approved):
nit on comments otherwise lgtm

crates/interpreter/src/host.rs
qiweiii: updated!

rakita:
Clippy not related

crates/interpreter/src/host.rs
```diff
@@ -56,3 +51,17 @@ pub trait Host {
...
+    /// Value of the storage when it is first read
+    pub original_value: U256,
+    /// Current value of the storage
+    pub present_value: U256,
+    /// Values that that is set
```
rakita, line 63:
```suggestion
    /// New value that is set
```

PR #1124, opened 2024-02-23
feat: add with_handler method to EvmBuilder

Rjected:
Added a method to make it possible to add a _specific_ handler, for example a handler other than optimism or mainnet, to the evm builder.

### DIFF
--- a/crates/revm/src/builder.rs
+++ b/crates/revm/src/builder.rs
@@ -200,7 +200,7 @@ impl<'a, EXT, DB: Database> EvmBuilder<'a, HandlerStage, EXT, DB> {
     ///
     /// # Example
     /// ```rust
-    ///
+    /// use revm::{EvmBuilder, Handler, primitives::{SpecId, HandlerCfg}};
     /// let builder = EvmBuilder::default();
     /// // set up empty handler cfg
     /// let handler = HandlerCfg::new(SpecId::LATEST);
--- a/crates/revm/src/builder.rs
+++ b/crates/revm/src/builder.rs
@@ -195,36 +195,6 @@ impl<'a, EXT, DB: Database> EvmBuilder<'a, HandlerStage, EXT, DB> {
         }
     }
 
-    /// This modifies the [EvmBuilder] to make it easy to construct an [`Evm`] with a _specific_
-    /// handler.
-    ///
-    /// # Example
-    /// ```rust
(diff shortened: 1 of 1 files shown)


crates/revm/src/builder.rs
```diff
@@ -195,6 +195,36 @@ impl<'a, EXT, DB: Database> EvmBuilder<'a, HandlerStage, EXT, DB> {
...
+    /// handler.
+    ///
+    /// # Example
+    /// ```rust
+    /// use revm::{EvmBuilder, Handler, primitives::{SpecId, HandlerCfg}};
+    /// use revm_interpreter::primitives::CancunSpec;
+    /// let builder = EvmBuilder::default();
+    /// // set up empty handler cfg
+    /// let handler = HandlerCfg::new(SpecId::LATEST);
+    /// let builder = builder.with_handler_cfg(handler);
```
rakita, line 208: Would remove this lines, they can be confusing
```suggestion
```

crates/revm/src/builder.rs
```diff
@@ -195,6 +195,36 @@ impl<'a, EXT, DB: Database> EvmBuilder<'a, HandlerStage, EXT, DB> {
...
+    ///
+    /// // get the desired handler
+    /// let mainnet = Handler::mainnet::<CancunSpec>();
+    /// let builder = builder.with_handler(mainnet);
+    ///
+    /// // build the EVM
+    /// let evm = builder.build();
+    /// ```
+    pub fn with_handler(
```
rakita, line 217: This function is inside:
`impl<'a, EXT, DB: Database> EvmBuilder<'a, HandlerStage, EXT, DB> {`

But it should be in:
`impl<'a, BuilderStage, EXT, DB: Database> EvmBuilder<'a, BuilderStage, EXT, DB> {`

Reason for this is that `with_handler` can be called from both `HandlerStage` and `SetGenericStage `

crates/revm/src/builder.rs
rakita: Second comment will address why you needed this previously: https://github.com/bluealloy/revm/pull/1124#discussion_r1501291862

crates/revm/src/builder.rs
Rjected: ah I see, yeah it makes sense to call it from either handler or setgeneric stages

rakita (approved):

Issue #798, opened 2023-10-13
Allow precompiles to capture state 


mattsse:
currently precompiles are limited to functions:

https://github.com/bluealloy/revm/blob/68f78ae0a4c2b351f2fe36236164d07933d0d4d5/crates/primitives/src/precompile.rs#L10-L11

it would be useful to allow them to capture state, by supporting `Arc< dyn Fn(&[u8], u64, env: &Env) -> PrecompileResult`

rakita:
Would add another enum item for this, and before opening this should made this extendable with handlers refactor

mattsse:
an enum would be the proper abstraction imo.

where the third variant is something like an `Arc<dyn PrecompileTrait>`

where precompile trait mirrors `fn(&[u8], u64, env: &Env) -> PrecompileResult`

perhaps we could also pass the evmdata as an additional parameter, in case a precompile needs access to the journalstate for some reason

PR #1131, opened 2024-02-27
feat: Add boxed precompile trait

rakita:
closes: #798 

Box is over the trait that has a `call` function. Trait is sync/send so it can be stored inside OnceBox.
And DynClone is added so precompiles can be cloned.

### DIFF
--- a/crates/primitives/src/precompile.rs
+++ b/crates/primitives/src/precompile.rs
@@ -1,6 +1,7 @@
 use crate::{Bytes, Env};
 use core::fmt;
 use dyn_clone::DynClone;
+use std::boxed::Box;
 
 /// A precompile operation result.
 ///
--- a/crates/primitives/src/precompile.rs
+++ b/crates/primitives/src/precompile.rs
@@ -1,7 +1,7 @@
 use crate::{Bytes, Env};
 use core::fmt;
 use dyn_clone::DynClone;
-use std::boxed::Box;
+use std::{boxed::Box, sync::Arc};
 
 /// A precompile operation result.
 ///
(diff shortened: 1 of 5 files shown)


crates/primitives/src/precompile.rs
```diff
@@ -10,6 +10,33 @@ pub type PrecompileResult = Result<(u64, Bytes), PrecompileError>;
 pub type StandardPrecompileFn = fn(&Bytes, u64) -> PrecompileResult;
 pub type EnvPrecompileFn = fn(&Bytes, u64, env: &Env) -> PrecompileResult;
 
+/// Clonable precompile trait. It is used to create a boxed precompile.
+pub trait ClonablePrecompileTrait: DynClone + Send + Sync {
```
mattsse, line 15:
```suggestion
pub trait StatefulPrecompile: DynClone + Send + Sync {
```

rakita:
@mattsse can you take a look? Is this going to work

crates/primitives/src/precompile.rs
```diff
@@ -10,6 +10,33 @@ pub type PrecompileResult = Result<(u64, Bytes), PrecompileError>;
...
+    fn call(&self, bytes: &Bytes, gas_price: u64, env: &Env) -> PrecompileResult;
+}
+
+dyn_clone::clone_trait_object!(ClonablePrecompileTrait);
+
+/// Box over clonable precompile trait.
+pub type BoxedPrecompileTrait = Box<dyn ClonablePrecompileTrait>;
```
mattsse, line 22: is this just `Arc<dyn>`?

bu dynclone also works, it's probably nicer for no_std? idk

crates/primitives/src/precompile.rs
```diff
@@ -10,6 +10,33 @@ pub type PrecompileResult = Result<(u64, Bytes), PrecompileError>;
...
+
+#[derive(Clone)]
+pub enum Precompile {
+    Standard(StandardPrecompileFn),
+    Env(EnvPrecompileFn),
+    BoxedEnv(BoxedPrecompileTrait),
+}
```
mattsse, line 29: we should add some docs here, and perhaps a From for T: StatefulPrecompile?

mattsse (changes requested):

crates/primitives/src/precompile.rs
rakita: It looks a lot like `Arc`, will need to check this statement but Box could allow mutability, this could be interesting. In this case we could even use arc.

mattsse: I think it doesn't matter, I envision this will be used like

```rust
#[derive(Clone)]
struct MyPrecompile {
  inner: Arc<Inner>
}
```

but I'd lean towards `Arc<dyn>`

rakita: Added both of them, `StatefulPrecompile` as `Arc` and `StatefulPrecompileMut` as `Box`

mattsse (approved):
lgtm

gakonst:
v exciting. could we get an example of what a stateful precompile could look like?

PR #1133, opened 2024-02-29
Add `db` and `db_mut` to evm

CorrM:
Old
```rust
evm.context.evm.db.clone();
evm.context.evm.db.insert_account_info(...);
```
New
```rust
evm.db().clone();
evm.db_mut().insert_account_info(...);
```


mattsse (approved):
I think this is reasonable

PR #1134, opened 2024-02-29
fix: make `feature = "optional_gas_refund"` work

yjhmelody:


mattsse (approved):
this looks right

rakita:
Would need to reintroduce this for mainnet. But this PR lgtm

Issue #1056, opened 2024-02-06
EthersDB deadlock


publicqi:
repro: https://github.com/publicqi/revm_hang

If a EthersDB is constructed in an async test function (or single-threaded executor) with no block number specified, the program will hang.

`#[tokio::test]` uses a single-threaded runtime. revm submits the `get_block_number` using `futures::executor::block_on`, but this task will never execute since the runtime is always polling the previous task (`test_hang`, in this case). 

Some refs when digging into the bug:
https://github.com/tokio-rs/tokio/issues/2603
https://github.com/tokio-rs/tokio/issues/2376

publicqi:
I think I encountered another scenario triggering the deadlock outside single-threaded executor. I haven't found the root cause yet. Will update here.

Currently my guess is that if this bug exists in single-threaded executor, it should also happen in multi-threaded but just at a smaller chance.

publicqi:
I added a min repro:

https://github.com/publicqi/revm_hang/commit/2b736a7f7622aff8ae2eccdb1fd2ca0559bf61dc

Otto-AA:
I just started to try out revm and found that the [generate_block_traces.rs](https://github.com/bluealloy/revm/blob/main/examples/generate_block_traces.rs) example gets stuck while executing the transactions.

From what I've found (as a Rust newbie):
- It's deterministic, ie in the normal setup it always gets stuck at tx 5, if I skip the first two transactions it gets stuck at tx 8. So it's not a problem of the transactions per se
- From the provider, the JsonRpcClient.request seems to wait infinitely for the response
- Further up the stacktrace, it's stuck in ethersdb `futures::executor::block_on(f)`

Am I correct, that this is another case of this issue?

rakita:
can you test if https://github.com/bluealloy/revm/pull/1089 fixes it?

Otto-AA:
No, using this branch it panics with following error:

> thread 'main' panicked at /home/oaie/.cargo/git/checkouts/revm-d03068627b7bf933/3c2af5f/crates/revm/src/db/ethersdb.rs:38:43:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.

<details>
<summary>stack backtrace:</summary>
```
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
[...]

PR #1135, opened 2024-02-29
Check runtime dynamically

wtdcode:
This PR removes the improper initialization of the runtime.

Explanation:

- Initializing a runtime when creating the struct doesn't make sense.

Users may create an `EthersDB` in a sync context but later use it in an async environment and vice versa. In this case, two runtimes will conflict. In addition, the current code, by default, creates a multithreaded runtime, which could be unexpected for users. For example, users may want to run many instances in parallel with separate cores pinned but the runtime created could interfere with this design.

Therefore, reasonable behavior is always checking if we are in an `async` environment by `Handle::try_current.` If we are in  `async`, we will pick the runtime created by users. Or else, we simply create a `current_thread` runtime to get things done. Note this is the recommended way by [tokio docs](https://tokio.rs/tokio/topics/bridging).

For `current_thread` runtime, a new thread is spawned to complete the futures as `tokio` doesn't allow us to `block_in_place`, which essentially does `spawn_blocking`. This could be removed once we have a `async` equivalent trait.

- Does it make things slower?

No. Because the current implementation only has __concurrency__ but not __parallelism__. Specifically, see:

```
        let f = async {
            let nonce = self.client.get_transaction_count(add, self.block_number);
            let balance = self.client.get_balance(add, self.block_number);
            let code = self.client.get_code(add, self.block_number);
            tokio::join!(nonce, balance, code)
        };
       let (nonce, balance, code) = self.block_on(f);
```

Either a `current_thread` or a `multi_thread` runtime will execute the three rpc calls one by one. Refer to [tokio::join!](https://docs.rs/tokio/latest/tokio/macro.join.html) documents for details.

- Any breaking changes?

No, as explained above.

### DIFF
--- a/crates/revm/src/db/ethersdb.rs
+++ b/crates/revm/src/db/ethersdb.rs
@@ -34,7 +34,11 @@ impl<M: Middleware> EthersDB<M> {
     /// internal utility function to call tokio feature and wait for output
     fn block_on<F: core::future::Future>(&self, f: F) -> F::Output {
         match Handle::try_current() {
-            Ok(handle) => handle.block_on(f),
+            Ok(handle) => {
+                tokio::task::block_in_place(move || {
+                    handle.block_on(f)
+                })
+            }
             Err(_) => Builder::new_current_thread()
                 .enable_all()
                 .build()
--- a/crates/revm/src/db/ethersdb.rs
+++ b/crates/revm/src/db/ethersdb.rs
@@ -39,6 +39,8 @@ impl<M: Middleware> EthersDB<M> {
     {
         match Handle::try_current() {
             Ok(handle) => match handle.runtime_flavor() {
+                // This essentially equals to tokio::task::spawn_blocking because tokio doesn't
+                // allow current_thread runtime to block_in_place
                 RuntimeFlavor::CurrentThread => std::thread::scope(move |s| {
                     s.spawn(move || {
                         Builder::new_current_thread()
--- a/crates/revm/src/db/ethersdb.rs
+++ b/crates/revm/src/db/ethersdb.rs
@@ -67,7 +67,7 @@ impl<M: Middleware> DatabaseRef for EthersDB<M> {
     fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
         let add = eH160::from(address.0 .0);
 
-        let f = async move {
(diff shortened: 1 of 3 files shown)


wtdcode:
This also solves #1056 as I tested.

crates/revm/src/db/ethersdb.rs
```diff
@@ -40,9 +33,13 @@ impl<M: Middleware> EthersDB<M> {
 
     /// internal utility function to call tokio feature and wait for output
     fn block_on<F: core::future::Future>(&self, f: F) -> F::Output {
-        match &self.runtime {
-            Some(runtime) => runtime.block_on(f),
-            None => futures::executor::block_on(f),
+        match Handle::try_current() {
+            Ok(handle) => handle.block_on(f),
```
mattsse, line 37: this will still panic if called from the main Runtime::block_on thread.

what we want here is

```rust
tokio::task::block_in_place(move || {
    handle.block_on(f)
});
```

mattsse (changes requested):

crates/revm/src/db/ethersdb.rs
wtdcode: Yes, I just noticed the problem. Actually, I think it looks impossible to run a future in a non-async function within a tokio runtime (?.

wtdcode: @mattsse

I investigated a bit. It looks like we can't support `current_thread` due to not being able to `blocking_in_place` without changing the signature of the `Database(Ref)` traits. I have two options here:

1. Document clearly that, `EthersDB` can't be used with `current_thread`.
2. Less preferably, we start another thread and create another runtime to do it.


Which one do you prefer?

wtdcode: Done with 2 so we can support all use cases, could you have a look? @mattsse

rakita: There is issue related to this, not sure if you have looked at it: https://github.com/bluealloy/revm/issues/1056

wtdcode: Yes, I tested this addresses #1056 cleanly.

mattsse (approved):
I think this should work

rakita:
Thanks!

PR #1140, opened 2024-03-01
perf(interpreter): evaluate instruction table constructor at compile time

DaniPopes:
Ideally this would also return `&'static` but since `InstructionTable` is generic over `H`, `&'static InstructionTable<H>` requires `H: 'static`, for which I haven't yet found a work-around.

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -28,9 +28,53 @@ pub type BoxedInstructionTable<'a, H> = [BoxedInstruction<'a, H>; 256];
 /// Note that `Plain` variant gives us 10-20% faster Interpreter execution.
 ///
 /// Boxed variant can be used to wrap plain function pointer with closure.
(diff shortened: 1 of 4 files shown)


gakonst (commented):
what's the perf change before/after?

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -28,11 +28,55 @@ pub type BoxedInstructionTable<'a, H> = [BoxedInstruction<'a, H>; 256];
...
+        const NEW: InstructionTable<H> = {
+            let mut tables: InstructionTable<H> = [control::unknown; 256];
+            let mut i = 0;
+            while i < 256 {
+                tables[i] = instruction::<H, SPEC>(i as u8);
+                i += 1;
+            }
+            tables
+        };
+    }
```
rakita, line 62: Didn't know this was possible. Nice one!

llvm got fixed in 1.75 so we could use newest rust.
But will merge this as it is.

rakita (approved):

PR #1141, opened 2024-03-01
Add VERBS to used by list

zombie-einstein:
Hi, we are using revm in an open source simulation library we built, this PR just adds an entry and link in the used by section


rakita (approved):
Welcome!

PR #1157, opened 2024-03-05
feat: use `impl` instead of `dyn` in `GetInspector`

DaniPopes:
Increases minimum Rust version to 1.75


mattsse (approved):
I like this

the version req bump should be fine imo

rakita:
Yep, we are fine with a bump, llvm bug is fixed in v1.75 https://github.com/bluealloy/revm/issues/887#issuecomment-1974197235

rakita (approved):
lgtm

Issue #1156, opened 2024-03-05
Add helper methods to `EvmContext`


DaniPopes:
Helper wrapper methods can be added to all of the methods in `JournaledState` that require a `DB` on `EvmContext` by just passing along all arguments to `self.journaled_state.<method>(..., &mut self.db)`.

E.g. `load_code`:
```rust
// JournaledState
impl JournaledState {
    pub fn load_code<DB: Database>(
        &mut self,
        address: Address,
        db: &mut DB,
    ) -> Result<(&mut Account, bool), EVMError<DB::Error>> { /* ... */ }
}

// Wrapper in EvmContext
impl<DB: Database> EvmContext<DB> {
    pub fn load_code(
        &mut self,
        address: Address
    ) -> Result<(&mut Account, bool), EVMError<DB::Error>> { // same result
        self.journaled_state.load_code(address, &mut self.db)
    }
}
```

PR #1158, opened 2024-03-05
feat: add more JournaledState methods to EvmContext

DaniPopes:
closes https://github.com/bluealloy/revm/issues/1156


crates/revm/src/journaled_state.rs
```diff
@@ -605,18 +605,19 @@ impl JournaledState {
 
     /// Load storage slot
     ///
-    /// # Note
+    /// # Panics
     ///
-    /// Account is already present and loaded.
+    /// Panics if the account is present in the state.
```
rakita, line 610:
```suggestion
    /// Panics if the account is not present in the state.
```

rakita (approved):

PR #1163, opened 2024-03-07
feat: implement DatabaseCommit for DatabaseComponents

Wodann:
These implementations were missing when working with a type that implements `DatabaseRef` instead of `Database`

### DIFF
--- a/crates/primitives/src/db/components.rs
+++ b/crates/primitives/src/db/components.rs
@@ -3,12 +3,15 @@ pub mod block_hash;
 pub mod state;
 
 pub use block_hash::{BlockHash, BlockHashRef};
+use hashbrown::HashMap;
 pub use state::{State, StateRef};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/primitives/src/db/components.rs
```diff
@@ -3,13 +3,16 @@ pub mod block_hash;
 pub mod state;
 
 pub use block_hash::{BlockHash, BlockHashRef};
+use hashbrown::HashMap;
```
rakita, line 6:
```suggestion
use crate::HashMap;
```
When `std` is enabled we are using std HashMap so in all places in code we include hashmap as `crate::HashMap`

Wodann: Addressed in https://github.com/bluealloy/revm/pull/1163/commits/9aca198e12bc82b7bc58fc3814db925f30535deb

rakita (approved):
lgtm

PR #1165, opened 2024-03-07
feat: add PrecompileError::Other

fgimenez:
Catch-all variant for custom precompiles failures.


mattsse (changes requested):
nit:
we can add a function:

```rust
impl PrecompileError {

  pub fn other(err: impl Into<String>) -> Self {
     Self::Other(err.into())
  }
}

fgimenez:
> we can add a function:

sure done

rakita (approved):

PR #1168, opened 2024-03-08
fix: reset tstorage on finalize

mattsse:
finalize is called via the output handler and is supposed to reset the journaledstate.

this ensures the transient storage is also cleared, restoring the initial state of JournaledState

### DIFF
--- a/crates/revm/src/journaled_state.rs
+++ b/crates/revm/src/journaled_state.rs
@@ -93,7 +93,7 @@ impl JournaledState {
 
     /// Does cleanup and returns modified state.
     ///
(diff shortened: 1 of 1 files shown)


rakita (approved):
good find!

crates/revm/src/journaled_state.rs
```diff
@@ -92,12 +92,27 @@ impl JournaledState {
     }
 
     /// Does cleanup and returns modified state.
+    ///
+    /// This resets the [JournalState] to its initial state in [Self::new]
```
rakita, line 96:
```suggestion
    /// This resets the [JournaledState] to its initial state in [Self::new]
```

PR #1170, opened 2024-03-08
feat(precompile): add a bool to bytes32 helper function

DaniPopes:
Useful for precompiles that return `bool` padded to 32 bytes.


mattsse (approved):

rakita (approved):
lgtm

PR #1171, opened 2024-03-08
feat(primitives): derive Default for SpecId

DaniPopes:


mattsse (approved):
I think that's reasonable

PR #1183, opened 2024-03-10
chore: improve error implementations

DaniPopes:
Make error messages consistent with Reth (starting lowercase, no full stops).


mattsse (approved):
reasonable

rakita (approved):

PR #1184, opened 2024-03-11
WIP: Implement EIP-5806

Amxx:
See https://github.com/Amxx/EIP-5806/tree/main to track the other implementation effort.

This PR provides `TransactTo::DelegateCall` that NomicFoundation's Hardhat EDR needs to implement EIP-5806. It also adds the restricted opcodes as documented in the EIP document.

I'd love to provide some testing, but I'm not very familiar with this projet. Can you guide me for that part?

### DIFF
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -152,6 +152,7 @@ pub fn sload<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
 
 pub fn sstore<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
     check_staticcall!(interpreter);
+    check_eip5806restricted!(interpreter, host);
 
     pop!(interpreter, index, value);
     let Some(SStoreResult {
@@ -229,6 +230,7 @@ pub fn log<const N: usize, H: Host>(interpreter: &mut Interpreter, host: &mut H)
 
 pub fn selfdestruct<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
     check_staticcall!(interpreter);
+    check_eip5806restricted!(interpreter, host);
     pop_address!(interpreter, target);
 
     let Some(res) = host.selfdestruct(interpreter.contract.address, target) else {
@@ -250,6 +252,7 @@ pub fn create<const IS_CREATE2: bool, H: Host, SPEC: Spec>(
     host: &mut H,
 ) {
     check_staticcall!(interpreter);
+    check_eip5806restricted!(interpreter, host);
 
     // EIP-1014: Skinny CREATE2
     if IS_CREATE2 {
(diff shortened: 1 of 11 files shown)


crates/interpreter/src/instructions/host.rs
```diff
@@ -229,6 +230,8 @@ pub fn log<const N: usize, H: Host>(interpreter: &mut Interpreter, host: &mut H)
 
 pub fn selfdestruct<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
     check_staticcall!(interpreter);
+    check_eip5806restricted!(interpreter, host);
+
```
Amxx, line 234:
```suggestion
```

crates/interpreter/src/instructions/macros.rs
```diff
@@ -17,6 +17,17 @@ macro_rules! check {
     };
 }
 
+macro_rules! check_eip5806restricted {
+    ($interp:expr, $host:expr) => {
+        if $host.env().tx.transact_to.is_delegatecall()
+            && $host.env().tx.caller != $interp.contract.address
```
Amxx, line 23:
```suggestion
            && $host.env().tx.caller == $interp.contract.address
```

rakita:
Hey, thank you for doing this, but revm should focus only on current mainnet EIP's, it is too early to merge this inside main repo.

For example, you can check the history of Transient Storage here, these things can span a year and more: https://github.com/bluealloy/revm/issues/154 from 

@Wodann ping on this as it is related to hardhat

Amxx:
I looked at the Inspector system, following [this](https://github.com/paradigmxyz/evm-inspectors) as an example.

However I'm not sure that would work:
- This PR is mostly about adding `TransactTo::DelegateCall(address)`, and I'm really not sure if that can be done without changes to revm.
- The "restricted opcode part" might be feasable through an Inspector, but I'm not sure how:
  - I can get everything I need from the `Interpreter` and the `EvmContext`, but if I want to write to the `interp.instruction_result`, I'm not sure how to get access to `InstructionResult`
  - Even if was able to set that, I'm not sure how a Inspector `step` can halt the execution

rakita:
> Following the discussion in telegram, I looked at the Inspector system, following [this](https://github.com/paradigmxyz/evm-inspectors) as an example.
> 
> However I'm not sure that would work:
> 
> * This PR is mostly about adding `TransactTo::DelegateCall(address)`, and I'm really not sure if that can be done without changes to revm.
> * The "restricted opcode part" might be feasable through an Inspector, but I'm not sure how:
>   
>   * I can get everything I need from the `Interpreter` and the `EvmContext`, but if I want to write to the `interp.instruction_result`, I'm not sure how to get access to `InstructionResult`
>   * Even if was able to set that, I'm not sure how a Inspector `step` can halt the execution

Probably will be hard, 

> Following the discussion in telegram, I looked at the Inspector system, following [this](https://github.com/paradigmxyz/evm-inspectors) as an example.
> 
> However I'm not sure that would work:
> 
> * This PR is mostly about adding `TransactTo::DelegateCall(address)`, and I'm really not sure if that can be done without changes to revm.
> * The "restricted opcode part" might be feasable through an Inspector, but I'm not sure how:
>   
>   * I can get everything I need from the `Interpreter` and the `EvmContext`, but if I want to write to the `interp.instruction_result`, I'm not sure how to get access to `InstructionResult`
>   * Even if was able to set that, I'm not sure how a Inspector `step` can halt the execution

Yeah, this will be more harder to add outside of Revm as it requires a new transaction type.

Will close this PR as it is not actively worked on.

PR #1185, opened 2024-03-11
Fix typo in readme

wtdcode:
Original cli gives:

```
error: target `generate_block_traces` in package `revm` requires the features: `std`, `serde-json`, `ethersdb`
Consider enabling them by passing, e.g., `--features="std serde-json ethersdb"`
```


rakita (approved):
ty

PR #1186, opened 2024-03-12
Handler rework

rakita:
experimental PR, removal of lifetime from handler


rakita:
Very hard to do (dont want to say impossible), the lifetimes of generics gets tied to the Box. So pattern where the inspector wraps the present handler to execute it is only possible with the box with lifetime `Box< T+'a>`. Another problem is with instructions that are in essence a generic `fn` but Rust ties lifetime of those together and does not implement HRTB (Higher Rank Trait Bound) on top of it. It is restrictive.

So API is not obvious but allows more features to be done. The recommendation is to create Evm when needed and dissolve the Evm into Context and return it back.

PR #1194, opened 2024-03-13
feat: add convert_boxed and insert_boxed for InstructionTable

Rjected:
This adds a test which serves as an example for adding custom instructions which require some context. This also adds `convert_boxed` and `insert_boxed` which should make it easier to do this using the revm builder.

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -81,12 +81,10 @@ impl<'a, H: Host + 'a> InstructionTables<'a, H> {
     pub fn convert_boxed(&mut self) {
         match self {
             Self::Plain(table) => {
-                let boxed_table = Self::Boxed(core::array::from_fn(|i| {
+                *self = Self::Boxed(core::array::from_fn(|i| {
                     let instruction: BoxedInstruction<'a, H> = Box::new(table[i]);
                     instruction
                 }));
-
-                *self = boxed_table;
             }
             Self::Boxed(_) => {}
         };
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -90,7 +90,6 @@ impl<'a, H: Host + 'a> InstructionTables<'a, H> {
         };
     }
 }
-}
 
 /// Make instruction table.
 #[inline]
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -46,16 +46,17 @@ impl<'a, H: Host + 'a> InstructionTables<'a, H> {
     ///
     /// Returns an error if the table is not boxed.
     #[inline]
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -42,6 +42,24 @@ impl<H: Host> InstructionTables<'_, H> {
...
+    ///
+    /// Returns an error if the table is not boxed.
+    #[inline]
+    pub fn insert_boxed(
+        &mut self,
+        opcode: u8,
+        instruction: BoxedInstruction<'a, H>,
+    ) -> Result<(), InsertError> {
+        match self {
+            Self::Plain(_) => Err(InsertError::NotBoxed),
```
mattsse, line 55: should we convert to boxed instead?

rakita: I would panic here. Silently converting can have a side-effect on performance

mattsse: this would make evm setup fallible and force unwrap with a preceding call to to_boxed, i think this makes it a lot less convenient to use, and since additional instructions are only supported by boxing all of them, converting them here seems reasonable to me

rakita: That is the point to fail and make it explicit to call, this would look something like.
```
table.convert_to_box();
table.insert_boxed(instr);
```
It is redundant but explicit on what is going on.

Not the point I would die on, so lets convert it to boxed inside `insert_boxed`.

Rjected: Changed to convert into the boxed variant inside `insert_boxed`. updated docs and the test which no longer needs a map

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -54,6 +74,28 @@ impl<'a, H: Host + 'a> InstructionTables<'a, H> {
...
+            Self::Boxed(_) => {}
+        };
+    }
+}
+
+/// An error that can occur when trying to insert instructions into the instruction table.
+#[derive(Debug)]
+pub enum InsertError {
+    /// Tried to insert a boxed instruction into an instruction table which was not boxed.
+    NotBoxed,
```
rakita, line 98:
```suggestion
```

Not used

crates/interpreter/src/instructions/opcode.rs
```diff
@@ -54,6 +74,22 @@ impl<'a, H: Host + 'a> InstructionTables<'a, H> {
...
+            Self::Plain(table) => {
+                *self = Self::Boxed(core::array::from_fn(|i| {
+                    let instruction: BoxedInstruction<'a, H> = Box::new(table[i]);
+                    instruction
+                }));
+            }
+            Self::Boxed(_) => {}
+        };
+    }
+}
```
rakita, line 92:
```suggestion
```

rakita (approved):

PR #1195, opened 2024-03-13
feat: optional nonce check

Wodann:
For Hardhat we need the ability to disable nonce checks. This PR follows the existing pattern for similar features and adds it to the `dev` feature flag.

### DIFF
--- a/crates/interpreter/Cargo.toml
+++ b/crates/interpreter/Cargo.toml
@@ -48,6 +48,7 @@ dev = [
     "optional_gas_refund",
     "optional_no_base_fee",
     "optional_beneficiary_reward",
+    "optional_nonce_check",
 ]
 memory_limit = ["revm-primitives/memory_limit"]
 optional_balance_check = ["revm-primitives/optional_balance_check"]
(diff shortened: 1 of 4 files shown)


Wodann:
The failing test is also present on `main`

rakita:
This pattern is going to be removed in future.

Would create a handler register that overrides the validation handler and sets Env nonce to the expected value: https://github.com/bluealloy/revm/blob/69e3a285db9feed1abf767414b7989edbac7e47a/crates/revm/src/handler/mainnet/validation.rs#L17-L36

In the future you can expect most of those features to be removed, as the same result can be done with handler register.

But as handler register are still new, would be okay to include this now. Up to you, how would you want to proceed?

Wodann:
As we need this for a bug fix in Hardhat, I'd appreciate if we can use this approach for now.

rakita:
btw you would get the same result if you set `tx.nonce = None`

rakita (approved):

rakita:
approving this as temporary fix.

Wodann:
> btw you would get the same result if you set `tx.nonce = None`

That's good to know. That's just as easy a fix, actually. Thanks!

Wodann:
The ability to set the `tx.nonce` to zero is just as good as the changes that I made here, so feel free to revert the commit.

Apologies; I should've realised.

PR #1201, opened 2024-03-15
chore: expose functionality for custom EVMs

fgimenez:
Aiming to help add custom instructions and precompiles

### DIFF
--- a/crates/precompile/src/secp256k1.rs
+++ b/crates/precompile/src/secp256k1.rs
@@ -40,7 +40,7 @@ mod secp256k1 {
 
 #[cfg(feature = "secp256k1")]
 #[allow(clippy::module_inception)]
-mod secp256k1 {
+pub mod secp256k1 {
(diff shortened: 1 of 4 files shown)


crates/precompile/src/secp256k1.rs
```diff
@@ -40,7 +40,7 @@ mod secp256k1 {
 
 #[cfg(feature = "secp256k1")]
 #[allow(clippy::module_inception)]
-mod secp256k1 {
+pub mod secp256k1 {
```
rakita, line 43: why do you need to be public? This mod is mad to hide the secp256k1 and k256 inclusion.

`ecrecover` should be made publish

fgimenez: makes total sense, with these changes https://github.com/bluealloy/revm/pull/1201/commits/746039da228d75a2fe9eb7bcb1e35046b4227477  `ecrecover` is reexported, PTAL

rakita (approved):

PR #1203, opened 2024-03-15
feat(interpreter): export utility macros

DaniPopes:
Extends https://github.com/bluealloy/revm/pull/1201

### DIFF
--- a/crates/interpreter/src/instructions/macros.rs
+++ b/crates/interpreter/src/instructions/macros.rs
@@ -62,9 +62,9 @@ macro_rules! gas_or_fail {
 /// Resizes the interpreter memory if necessary. Fails the instruction if the memory or gas limit
 /// is exceeded.
 #[macro_export]
-macro_rules! shared_memory_resize {
+macro_rules! resize_memory {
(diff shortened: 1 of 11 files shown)


mattsse (approved):
lgtm

crates/interpreter/src/instructions/macros.rs
```diff
@@ -29,12 +37,16 @@ macro_rules! gas {
...
+/// Records a `gas` refund.
+#[macro_export]
 macro_rules! refund {
     ($interp:expr, $gas:expr) => {
         $interp.gas.record_refund($gas)
     };
 }
 
+/// Same as [`gas!`], but with `gas` as an option.
+#[macro_export]
```
fgimenez, line 49: This macro calls `gas!`, if we want to consume it externally without importing also `gas!` we might need to change the call to `$crate::gas!`. Same for `as_usize_or_fail!` and `as_usize_or_fail_ret!`

rakita (approved):
lgtm! Pending @fgimenez comment

DaniPopes:
Done

fgimenez (approved):

PR #1208, opened 2024-03-18
Improve `EthersDB`

CorrM:
Improve EthersDB:
  - Rewrite(kinda) `new` function.
  - Add `set_block_number` function.
-----
Why i need to re-create `EthersDB` instance?
Your node will not querying data 10 minutes old. more info [here](https://ethereum.stackexchange.com/questions/149254/using-hardhat-to-fork-a-node-but-getting-missing-trie-node).

Also im fan of **reuse memory**, memory are cheap but i prefer to not waste it nor re-create instance.

example of the error:
```json
JsonRpcError { code: -32000, message: "error: missing trie node cd677575d733d1d9478ca4c8af9fc9991d98487964f595750c14a3675c1c86db (path 08) <nil>", data: None })
```


CorrM:
@rakita Why it fails to pass tests. it build successfully

rakita:
Don't spill the tokio outside the ethersdb as it is behind the feature flag. it fails as it want to build it as `no_std`.
That function does not make sense outside of `ethersdb.rs` file.

crates/revm/src/db/ethersdb.rs
```diff
@@ -1,10 +1,44 @@
...
 use ethers_providers::Middleware;
-use std::sync::Arc;
 use tokio::runtime::{Builder, Handle, RuntimeFlavor};
 
+use crate::primitives::{AccountInfo, Address, Bytecode, B256, KECCAK_EMPTY, U256};
+use crate::{Database, DatabaseRef};
+
+/// internal utility function to call tokio feature and wait for output
+#[inline]
+fn block_on<F>(f: F) -> F::Output
```
rakita, line 12: As this is only used by `EthersDB` there is no reason to be outside of it, can you revert back this change

PR #1211, opened 2024-03-19
feat: add benchmarks for ecpairing, kzg, ecrecover precompiles

Rjected:
Benchmarks runtime of a precompile instance, and prints out the gas used by the precompile call. This also exposes some methods and constants required for the precompile benchmarks


rakita (approved):
lgtm

PR #1215, opened 2024-03-21
feat: derive serde for OpCode, improve implementations

DaniPopes:


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -463,13 +470,8 @@ impl OpCode {
...
 
     /// Instantiate a new opcode from a u8 without checking if it is valid.
-    ///
-    /// # Safety
-    ///
-    /// All code using `Opcode` values assume that they are valid opcodes, so providing an invalid
-    /// opcode may cause undefined behavior.
     #[inline]
-    pub unsafe fn new_unchecked(opcode: u8) -> Self {
+    pub fn new_unchecked(opcode: u8) -> Self {
```
rakita, line 474: Made it unsafe as OpCode should not be possible to have an undefined byte

rakita:
I have refactored this file in EOF branch, removed deprecated things, etc. if serde is needed lets make just that change and leave other changes to that branch.

crates/interpreter/src/instructions/opcode.rs
DaniPopes: It's not memory unsafe and I'd like to use this to format any opcode, like `("{}",  Opcode::new_unchecked(op_byte))`

DaniPopes:
Serde and formatting are nice to have for general use

mattsse (approved):
lgtm,

perhaps we limit this pr to serde only if @rakita prefers this

crates/interpreter/src/instructions/opcode.rs
rakita: It is better to not have undefined byte in struct

rakita:
> Serde and formatting are nice to have for general use

> lgtm,
> 
> perhaps we limit this pr to serde only if @rakita prefers this

We can limit it to serde only.

rakita (approved):

PR #1219, opened 2024-03-22
feat(interpreter): expose mutable access methods on stack and memory

DaniPopes:


mattsse (commented):
those are very sensitive internals,
what's the use case you have in mind here?

DaniPopes:
We allow mutable access to everything else, we either lock everything or nothing

mattsse:
> We allow mutable access to almost everything else

fair, not opposed to this, so defer to @rakita here

mattsse (approved):

rakita (approved):
lgtm

PR #1221, opened 2024-03-24
perf(interpreter): keep track of remaining gas rather than spent

DaniPopes:
Removes a load and a comparison in the very hot `record_cost` function (see diff below) by using `overflowing_sub` instead of `saturating_add` + `< limit`.

Reduces `snailtracer` instructions from `2,768,629,456` to `2,661,673,460` (-3.86%).

I've also renamed `Gas::spend` to `Gas::spent` (with a deprecation warning).

image


mattsse (approved):
I think tracking remaining gas is more intuitive.

afict this merely flips cost recording logic

rakita (approved):
This is cool! We dont need to compare it with `gas.limit` anymore

PR #1228, opened 2024-03-25
chore(interpreter): unbox contract field

DaniPopes:
I don't see a reason for this to be boxed, it's 184 bytes big when the interpreter is 256 already, bringing it to 432 doesn't change much.

This doesn't affect the size of frames as they themselves are boxed.


mattsse (approved):
reasonable

rakita (approved):
This was useful when we had recursive loop calls. But now it can be removed

PR #1229, opened 2024-03-25
chore(deps): run `cargo update`

DaniPopes:
I'm not sure why dependabot doesn't do most of these. I needed to run this to fetch the latest `ruint`.


mattsse (approved):
should we even commit the lockfile here?

rakita:
> should we even commit the lockfile here?

It is good practice to lock on libraries and update them from time to time.

PR #1230, opened 2024-03-26
fix(TracerEip3155): clear Inspector data after transaction.

Pana:
Run multiple tx with a same TracerEip3155 will cause the gas_remaining incorrect.

### DIFF
--- a/crates/revm/src/inspector/eip3155.rs
+++ b/crates/revm/src/inspector/eip3155.rs
@@ -95,6 +95,30 @@ impl TracerEip3155 {
     pub fn set_writer(&mut self, writer: Box<dyn Write>) {
         self.output = writer;
     }
+
+    /// Resets the Tracer to its initial state of [Self::new].
+    /// This makes the inspector ready to be used again.
+    pub fn fuse(&mut self) {
+        let Self {
+            gas_inspector,
+            stack,
+            pc,
+            opcode,
+            gas,
+            refunded,
+            mem_size,
+            skip,
+            ..
+        } = self;
+        *gas_inspector = GasInspector::default();
+        stack.clear();
+        *pc = 0;
+        *opcode = 0;
+        *gas = 0;
+        *refunded = 0;
+        *mem_size = 0;
+        *skip = false;
+    }
 }
 
 impl TracerEip3155 {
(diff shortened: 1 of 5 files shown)


rakita (changes requested):
it is better to fix this inside Inspector.

Add `if context.journaled_state.depth() == 0 { /* clean data */` to
https://github.com/bluealloy/revm/blob/e7363d7dc5693550c72cd2773e4278f167b1cac9/crates/revm/src/inspector/eip3155.rs#L176

crates/revm/src/inspector/eip3155.rs
```diff
@@ -95,6 +98,30 @@ impl TracerEip3155 {
     pub fn set_writer(&mut self, writer: Box<dyn Write>) {
         self.output = writer;
     }
+
+    /// Resets the Tracer to its initial state of [Self::new].
+    /// This makes the inspector ready to be used again.
+    pub fn fuse(&mut self) {
```
rakita, line 104: `clear()` would be a better name

crates/revm/src/inspector/eip3155.rs
```diff
@@ -118,6 +145,28 @@ impl TracerEip3155 {
         self.output.write_all(b"\n")?;
         self.output.flush()
     }
+
+    fn print_summary<DB: Database>(
+        &mut self,
+        result: &InterpreterResult,
+        context: &mut EvmContext<DB>,
+    ) {
```
rakita, line 153: cool

crates/revm/src/inspector/eip3155.rs
```diff
@@ -173,19 +222,32 @@ impl<DB: Database> Inspector<DB> for TracerEip3155 {
...
 
-                time: None,
-                fork: Some(spec_name.to_string()),
-            };
-            let _ = self.write_value(&value);
+        self.print_summary(&outcome.result, context);
+
+        // clear the state if we are at the top level
+        if context.journaled_state.depth() == 0 {
+            self.fuse();
```
rakita, line 230: would include `print_summary` here and remove if check of depth from `print_summary`

rakita (approved):
looks good, left few nits.

PR #1234, opened 2024-03-27
feat: TracerEip3155 optionally traces memory

Otto-AA:
This adds the option to include the memory of each step in the EIP3155 trace.

If the option "include_memory" is true, the tracer will add an additional "memory: "0x..." field to each trace event. The EIP3155 seems to be ambiguous about the (optional) memory field. In the Output section it describes it as a "Array of Hex-Strings", however in the Test Cases it is only a Hex-String, not an Array. This PR implements the Hex-String version.

I think this would be a breaking change, as the TracerEIP3155 signature now has an additional "include_memory" flag. I've written the commit message accordingly.

Also note that this is my first contact with Rust, feel free to modify things that are not best practice :) The unit tests still ran through and I've manually tested the memory output. Also, the memory output makes the tracing much slower if it's enabled.

Example outputs:

```jsonl
{"pc":0,"op":96,"gas":"0x25c26","gasCost":"0x3","stack":[],"depth":1,"returnData":"0x","refund":"0x0","memSize":"0","opName":"PUSH1","memory":"0x"}
{"pc":2,"op":96,"gas":"0x25c23","gasCost":"0x3","stack":["0x80"],"depth":1,"returnData":"0x","refund":"0x0","memSize":"0","opName":"PUSH1","memory":"0x"}
{"pc":4,"op":82,"gas":"0x25c20","gasCost":"0xc","stack":["0x80","0x40"],"depth":1,"returnData":"0x","refund":"0x0","memSize":"0","opName":"MSTORE","memory":"0x"}
{"pc":5,"op":96,"gas":"0x25c14","gasCost":"0x3","stack":[],"depth":1,"returnData":"0x","refund":"0x0","memSize":"96","opName":"PUSH1","memory":"0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080"}
```

```jsonl                                                                                                                                                              
{"pc":2724,"op":241,"gas":"0x8e0a","gasCost":"0x238c","stack":["0x2e1a7d4d","0x264","0x16345785d8a0000","0xd9e1ce17f2641f24ae83637ab66a2cca9c378b9f","0x0","0x16345785d8a0000","0x60","0x0","0x60","0x0","0x60","0x16345785d8a0000","0xd9e1ce17f2641f24ae83637ab66a2cca9c378b9f","0x0"],"depth":2,"returnData":"0x","refund":"0x4dbc","memSize":"96","opName":"CALL","error":"CallOrCreate","memory":"0x000000000000000000000000d9e1ce17f2641f24ae83637ab66a2cca9c378b9f00000000000000000000000000000000000000000000000000000000000000030000000000000000000000000000000000000000000000000000000000000060"}                                               
{"pc":0,"op":96,"gas":"0x8fc","gasCost":"0x3","stack":[],"depth":3,"returnData":"0x","refund":"0x0","memSize":"0","opName":"PUSH1","memory":"0x"}                                                                  
{"pc":2,"op":96,"gas":"0x8f9","gasCost":"0x3","stack":["0x80"],"depth":3,"returnData":"0x","refund":"0x0","memSize":"0","opName":"PUSH1","memory":"0x"}
```

### DIFF
--- a/crates/revm/src/inspector/eip3155.rs
+++ b/crates/revm/src/inspector/eip3155.rs
@@ -168,13 +168,7 @@ impl<DB: Database> Inspector<DB> for TracerEip3155 {
             } else {
                 None
             },
-            memory: match &self.memory {
-                Some(memory) => {
-                    let hex_memory = hex::encode(memory);
-                    Some(format!("0x{hex_memory}"))
-                }
-                None => None,
-            },
+            memory: self.memory.as_ref().map(hex::encode_prefixed),
             storage: None,
             return_stack: None,
         };
--- a/crates/revm/src/inspector/eip3155.rs
+++ b/crates/revm/src/inspector/eip3155.rs
@@ -22,6 +22,8 @@ pub struct TracerEip3155 {
     refunded: i64,
     mem_size: usize,
     skip: bool,
+    include_memory: bool,
+    memory: Option<Vec<u8>>,
 }
 
 // # Output
@@ -58,7 +60,7 @@ struct Output {
     error: Option<String>,
     /// Array of all allocated values
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    memory: Option<Vec<String>>,
+    memory: Option<String>,
     /// Array of all stored values
     #[serde(default, skip_serializing_if = "Option::is_none")]
     storage: Option<HashMap<String, String>>,
@@ -98,12 +100,14 @@ impl TracerEip3155 {
 }
 
 impl TracerEip3155 {
-    pub fn new(output: Box<dyn Write>, print_summary: bool) -> Self {
+    pub fn new(output: Box<dyn Write>, print_summary: bool, include_memory: bool) -> Self {
         Self {
             output,
             gas_inspector: GasInspector::default(),
             print_summary,
+            include_memory,
             stack: Default::default(),
+            memory: Default::default(),
             pc: 0,
             opcode: 0,
             gas: 0,
@@ -128,6 +132,11 @@ impl<DB: Database> Inspector<DB> for TracerEip3155 {
     fn step(&mut self, interp: &mut Interpreter, context: &mut EvmContext<DB>) {
         self.gas_inspector.step(interp, context);
         self.stack = interp.stack.data().clone();
+        self.memory = if self.include_memory {
+            Some(interp.shared_memory.context_memory().to_owned())
+        } else {
+            None
+        };
         self.pc = interp.program_counter();
         self.opcode = interp.current_opcode();
         self.mem_size = interp.shared_memory.len();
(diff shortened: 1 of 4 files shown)


crates/revm/src/inspector/eip3155.rs
```diff
@@ -98,12 +100,14 @@ impl TracerEip3155 {
 }
 
 impl TracerEip3155 {
-    pub fn new(output: Box<dyn Write>, print_summary: bool) -> Self {
+    pub fn new(output: Box<dyn Write>, print_summary: bool, include_memory: bool) -> Self {
```
rakita, line 103: Wound not add include_memory here flag here  but a new function `with_memory(self) -> Self`

This would allow setting `TracerEip3155` as TracerEip3155::new(Box::new(std::io::stdout()), true).with_memory()`

Would be good to add same for `print_summary` but your choice if you want it in this PR

crates/revm/src/inspector/eip3155.rs
```diff
@@ -58,7 +60,7 @@ struct Output {
     error: Option<String>,
     /// Array of all allocated values
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    memory: Option<Vec<String>>,
+    memory: Option<String>,
```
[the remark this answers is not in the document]
rakita: Scratch this, it would be better to have `String` here. Conversion is going to slow down the tracer and memory cloning on every step is expensive, but we can't go around this.

rakita (changes requested):
Welcome to rust land!

One nit on tracer api but looks good.

crates/revm/src/inspector/eip3155.rs
Otto-AA: Thanks for the suggestion, I've implemented it for both now.
For the summary I've defaulted to true, as it is a MUST for the (current) EIP-3155. This implicitly adds it to the revme `execute_test_suite` stderr and stdout. I could change that if that's not wanted.

For the memory I've defaulted to false, as it's not mandatory and often not wanted due to performance issues.

Otto-AA:
Thanks for the nice welcome :)

I've update the code to use `hex::encode_prefixed` which makes it a bit faster, and to use the builder functions instead of bool parameters.

If you merge it, maybe squash it to a single commit.

crates/revm/src/inspector/eip3155.rs
```diff
@@ -22,6 +22,8 @@ pub struct TracerEip3155 {
     refunded: i64,
     mem_size: usize,
     skip: bool,
+    include_memory: bool,
+    memory: Option<Vec<u8>>,
```
rakita, line 26: Onee more thing, it would be good to have `String` here so we don't copy memory multiple times.

Now we are doing clone in `step` and clone into String in `step_end`

Otto-AA: I've changed it to String now, however my change did not improve the performance. I guess the `self.memory.take()` method also copies it internally. Without the `.take()` I've got the error "cannot move out of `self.memory` which is behind a mutable reference" and didn't manage to fix it in a better way.

rakita: `take()` does not copy full string, it is the right approach. Copying memory on every step is a heavy operation, not a lot we could do there. There are few optimisation that we could do but they require introspection of opcodes that change memory, definitely out of scope for this PR.

rakita (approved):
lgtm!

PR #1235, opened 2024-03-27
fix: build with warnings treated as errors

jtraglia:
This PR will build/test with warnings treated as errors. It also fixes the issues, all of which were redundant imports.

Also, rename `test-no-std` to `check-no-std` which is more accurate.


jtraglia:
Oops, still some warnings left to fix. Will get those.

jtraglia:
Oh I see, those imports are necessary with `--no-default-features`. Hmm.

jtraglia:
Going to close for now. Not exactly sure what to do. It's a low priority fix anyway.

PR #1236, opened 2024-03-28
fix(GasInspector): calculate correct remaining gas after call return

rakita:
After changing calls into loop, remaining gas calculation became simpler and this fixes the bug of incorrect remaining gas after returning from the CALL/CREATE opcodes.

Bug noticed here: https://github.com/paradigmxyz/reth/issues/6969
and here: https://github.com/paradigmxyz/evm-inspectors/pull/38#discussion_r1490933691


mattsse (approved):
this lgtm,

perhaps @DaniPopes can also take another look

DaniPopes (approved):

PR #1238, opened 2024-03-28
fix: use correct bn128 mul input length

jtraglia:
This is pretty minor, but `MUL_INPUT_LEN` was 128 when it only needed to be 96. So there was some unnecessary padding. I don't expect any performance difference from this, just fixing to be more correct. As part of this, I added comments explaining the size and splitting the value into each component.


rakita (approved):
lgtm, it is a lot more readable!
And confirming that it does not effect consensus, it does padding that is not used.

PR #1239, opened 2024-03-28
nit: remove unnecessary call to into_u256() for bn128 add

jtraglia:
This PR removes the intermediate call to `into_u256()` which appears to be unnecessary. I noticed that `ecmul` does it like this, and I think it's cleaner this way:

https://github.com/bluealloy/revm/blob/c3107ec3acf79119016acac203b1c2437005a90e/crates/precompile/src/bn128.rs#L157-L171

Also, rename `out` to `output` in `run_mul` for consistency.


rakita (approved):
lgtm

PR #1240, opened 2024-03-28
Add example pre-run instruction handler in EVM execution loop

yascha:
Add example of (unsafe) implementation of a handler that can be used to execute arbitrary EVM calls before a given evm instruction.  Useful in the new REVM model.

Example usage in EVM builder:
```
let old_handle = handler.execution_loop.pre_run_interpreter.clone();
            handler.execution_loop.pre_run_interpreter = Arc::new(
                move |evm: &mut Evm<'_, IE, DB>,
                      table: &[Box<dyn Fn(&mut Interpreter, &mut Evm<'_, IE, DB>)>; 256],
                      frame: &mut Box<CallStackFrame>|
                      -> Option<InterpreterAction> {
                    let engine_evm = EngineEVM {
                        evm,
                        table,
                        current_stack_frame: frame,
                    };
                    if let Some(pre_run) = IE::on_pre_run_interpreter(engine_evm) {
                        return Some(pre_run);
                    }
                    old_handle(evm, table, frame)
                },
            );
```

### DIFF
--- a/crates/revm/src/evm.rs
+++ b/crates/revm/src/evm.rs
@@ -208,11 +208,31 @@ impl<EXT, DB: Database> Evm<'_, EXT, DB> {
 
         loop {
             // run interpreter
-            let action = stack_frame
-                .interpreter
-                .run(shared_memory, instruction_table, self);
-            // take shared memory back.
-            shared_memory = stack_frame.interpreter.take_memory();
+            let action = if let Some(action) = {
+                // I'm so sorry draganrakita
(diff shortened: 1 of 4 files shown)


rakita:
Ty, will check it out

gakonst:
@yascha interesting, how do you think about the use case for this?

yascha:
@gakonst with the old implementation of revm, it was very easy to manipulate the evm (in a very custom manner) via rust code.  The new model has some incredibly nice changes, but it makes the evm manipulation trickier.
I've been using a variant of this hook for some time to enable executing arbitrary evm code anywhere within a callstack.

rakita:
Effort is moved to this PR: https://github.com/bluealloy/revm/pull/1369

PR #1242, opened 2024-03-28
fix: update/enable bn128 tests

jtraglia:
The tests in the bn128 module have been commented out for a while. They only test Byzantium, but I think they would still be useful to have. This PR will do the following:

* Define `*_GAS_COST` constants for add and multiply.
* Update `run_add` and `run_mul` to take `gas_cost` and return `PrecompileResult`.
* Uncomment & fix the tests.


rakita (approved):
lgtm

Issue #1220, opened 2024-03-24
Remove `SPEC` generic usage in gas cost calculation code


DaniPopes:
Usage of the `SPEC` generic is unnecessary in gas calculation code as the `SpecId` is already known at compile-time in the instructions and these functions are marked with `#[inline]`, so they would get optimized the same way.

By turning the `SPEC` generic into a `SpecId` parameter:
1. most gas calculation functions can be made `const`, allowing usage in compile-time gas tables and outside of `SPEC`-generic code
2. helps a bit with compile times by reducing monomorphization bloat

`SPEC::enabled(X)` becomes `SpecId::enabled(spec, X)`.

This likely applies to more places in `revm-interpreter`, but I haven't looked too deeply into it, and I'm confident that gas calculation is a good starting place.

Examples:
- `Gas::set_final_refund` (cannot be made const due to `&mut`, rest still applies): https://github.com/bluealloy/revm/blob/fda371f73aba2c30a83c639608be78145fd1123b/crates/interpreter/src/gas.rs#L89-L92
- `gas::call_gas`: https://github.com/bluealloy/revm/blob/c203f3d033c29d26a2d02e47ffd65efa21e6dc7e/crates/interpreter/src/gas/calc.rs#L274-L287

rakita:
> one note here is that I think we want a more flexible solution for the SpecId enum that would allow us to get rid of the duplicated op and ethereum SpecId enum. because we want to abstract away all the OP cfgs

@mattsse this is a different Issue, seems not related to this one. https://github.com/bluealloy/revm/issues/918

> Usage of the `SPEC` generic is unnecessary in gas calculation code as the `SpecId` is already known at compile-time in the instructions and these functions are marked with `#[inline]`, so they would get optimized the same way.
> 
> By turning the `SPEC` generic into a `SpecId` parameter:
> 
> 1. most gas calculation functions can be made `const`, allowing usage in compile-time gas tables and outside of `SPEC`-generic code
> 2. helps a bit with compile times by reducing monomorphization bloat
> 
> `SPEC::enabled(X)` becomes `SpecId::enabled(spec, X)`.
> 
> This likely applies to more places in `revm-interpreter`, but I haven't looked too deeply into it, and I'm confident that gas calculation is a good starting place.

In general supportive, it looks like a refactor without a lot of gains, but it can potentially be useful to const fn outside of Revm.

[...]

PR #1243, opened 2024-03-28
feat(interpreter): remove SPEC generic from gas calculation functions

DaniPopes:
Closes https://github.com/bluealloy/revm/issues/1220


crates/interpreter/src/gas/calc.rs
```diff
@@ -1,13 +1,34 @@
...
+            None => panic!("unwrap failed"),
+        }
+    };
+}
+
+/// `SSTORE` opcode refund calculation.
 #[allow(clippy::collapsible_else_if)]
-pub fn sstore_refund<SPEC: Spec>(original: U256, current: U256, new: U256) -> i64 {
-    if SPEC::enabled(ISTANBUL) {
+pub fn sstore_refund(spec_id: SpecId, original: U256, current: U256, new: U256) -> i64 {
```
rakita, line 28:
```suggestion
pub const fn sstore_refund(spec_id: SpecId, original: U256, current: U256, new: U256) -> i64 {
```

?

crates/interpreter/src/gas/calc.rs
```diff
@@ -85,13 +102,14 @@ fn log2floor(value: U256) -> u64 {
     l
 }
 
+/// `EXP` opcode cost calculation.
 #[inline]
-pub fn exp_cost<SPEC: Spec>(power: U256) -> Option<u64> {
+pub fn exp_cost(spec_id: SpecId, power: U256) -> Option<u64> {
```
rakita, line 107:
```suggestion
pub const fn exp_cost(spec_id: SpecId, power: U256) -> Option<u64> {
```

crates/interpreter/src/gas/calc.rs
```diff
@@ -160,45 +184,46 @@ pub fn keccak256_cost(len: u64) -> Option<u64> {
...
     } else {
         50
     }
 }
 
-#[allow(clippy::collapsible_else_if)]
-pub fn sstore_cost<SPEC: Spec>(
+/// `SSTORE` opcode cost calculation.
+#[inline]
+pub fn sstore_cost(
```
rakita, line 213:
```suggestion
pub const fn sstore_cost(
```

crates/interpreter/src/gas/calc.rs
```diff
@@ -344,7 +375,8 @@ pub fn memory_gas(a: usize) -> u64 {
 
 /// Initial gas that is deducted for transaction to be included.
 /// Initial gas contains initial stipend gas, gas for access list and input data.
-pub fn validate_initial_tx_gas<SPEC: Spec>(
+pub fn validate_initial_tx_gas(
```
rakita, line 378:
```suggestion
pub const fn validate_initial_tx_gas(
```

mattsse (approved):

crates/interpreter/src/gas/calc.rs
DaniPopes: Can't, U256 is missing a lot of const ops, most importantly Eq

crates/interpreter/src/gas.rs
```diff
@@ -94,8 +94,12 @@ impl Gas {
     ///
     /// Related to EIP-3529: Reduction in refunds
     #[inline]
-    pub fn set_final_refund<SPEC: Spec>(&mut self) {
-        let max_refund_quotient = if SPEC::enabled(LONDON) { 5 } else { 2 };
+    pub fn set_final_refund(&mut self, spec: SpecId) {
```
rakita, line 97: Would put here just a bool flag, `is_london` here

crates/interpreter/src/gas/calc.rs
rakita: So there is not a lot of point in half changing this, as we would have two types of function?

DaniPopes: There already is, dynamic and static gas, I tried to make all functions I could const

crates/interpreter/src/gas.rs
DaniPopes: Don't think that makes much sense if all callers will end up doing `spec.is_enabled_in(SpecId::LONDON)`

crates/interpreter/src/gas/calc.rs
rakita: And this would be the third, but lets go with the change `const fn` are better in general.

crates/interpreter/src/gas.rs
rakita: It is more code for upper layers, but it is easier to see that this function depends only on `SpecId::LONDON`. This is the only function in `Gas` that is hardfork related.

rakita (approved):
lgtm

rakita (approved):
lgtm

PR #1245, opened 2024-03-29
feat(interpreter): test Host object-safety, allow `dyn Host` in instructions

DaniPopes:
This allows running the interpreter with `&mut dyn Host`.

Useful to have in contexts where generics are not possible (e.g. dynamic callbacks / FFI, etc.) or testing/benchmarking code.

This was only limited by the type system requiring the maybe-sized (`+ ?Sized`) bound on the `Host` generics.


mattsse (approved):
this only affects the type system

I assume this makes things slightly easier when writing bindings for python etc.

rakita (approved):

PR #1248, opened 2024-03-29
nit: move div by zero check from smod to i256_mod

jtraglia:
This PR contains some minor nit fixes. Most notably, it moves the divisor zero-value check into the arithmetic function so that it behaves the same as `i256_div`, which will handle it by returning zero. Additionally:

* In `i256_div`, remove unnecessary `first_sign == Sign::Minus` check before `MIN_NEGATIVE_VALUE` check.
* In `i256_mod`, remove comment about "overflow checks" which aren't being done.
* In the tests, define constants for common values: one, two, three, etc.
* Add some new arithmetic tests and update old asserts to use new constants.

### DIFF
--- a/crates/interpreter/src/instructions/i256.rs
+++ b/crates/interpreter/src/instructions/i256.rs
@@ -77,7 +77,7 @@ pub fn i256_div(mut first: U256, mut second: U256) -> U256 {
     }
 
     let first_sign = i256_sign_compl(&mut first);
-    if first_sign == Sign::Minus && first == MIN_NEGATIVE_VALUE && second == U256::from(1) {
+    if first == MIN_NEGATIVE_VALUE && second == U256::from(1) {
         return two_compl(MIN_NEGATIVE_VALUE);
     }
 
@@ -105,9 +105,11 @@ pub fn i256_mod(mut first: U256, mut second: U256) -> U256 {
         return U256::ZERO;
     }
 
-    let _ = i256_sign_compl(&mut second);
+    let second_sign = i256_sign_compl(&mut second);
+    if second_sign == Sign::Zero {
+        return U256::ZERO;
+    }
 
-    // necessary overflow checks are done above, perform the operation
     let mut r = first % second;
 
     // set sign bit to zero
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/instructions/i256.rs
```diff
@@ -133,70 +187,86 @@ mod tests {
         assert_eq!(i8::MAX / -1, -i8::MAX);
 
         // Now the same calculations based on i256
-        let one = U256::from(1);
-        let one_hundred = U256::from(100);
         let fifty = U256::from(50);
-        let two = U256::from(2);
-        let neg_one_hundred = U256::from(100);
-        let minus_one = U256::from(1);
```
jtraglia, line 141: Worth pointing out that these variable names were misleading.

crates/interpreter/src/instructions/i256.rs
```diff
@@ -105,9 +105,11 @@ pub fn i256_mod(mut first: U256, mut second: U256) -> U256 {
         return U256::ZERO;
     }
 
-    let _ = i256_sign_compl(&mut second);
+    let second_sign = i256_sign_compl(&mut second);
+    if second_sign == Sign::Zero {
+        return U256::ZERO;
+    }
```
jtraglia, line 111: Also, just curious, what was the purpose of this before?

rakita: I am not sure, probably just an artefact of code moving

rakita (approved):
lgtm

crates/interpreter/src/instructions/i256.rs
```diff
@@ -125,6 +127,58 @@ mod tests {
...
+        0xffffffffffffffff,
+        0xffffffffffffffff,
+    ]);
+
+    const I256_MAX: U256 = U256::from_limbs([
+        0xffffffffffffffff,
+        0xffffffffffffffff,
+        0xffffffffffffffff,
+        0x7fffffffffffffff,
+    ]);
```
DaniPopes, line 180: FYI you can use `uint!` macro like
```rust
uint! {
assert_eq!(i256_div(100_U256, -1_U256), -100_U256);
}
```

PR #1254, opened 2024-04-02
feat: add tests for shift instructions

jtraglia:
This PR does the following:

* Add tests for shift operations (`shl`, `shr`, `sar`) from [EIP-145](https://eips.ethereum.org/EIPS/eip-145).
* Replace instance `U256::from_limbs` with `uint!{}` macro, for readability.
* Ever so slightly optimize `sar` short-circuit check. This can be 255.
* Add some comments, one about how `.unwrap()` is safe.

In the tests, is this the best way to execute an EVM instruction?

Also, I can remove the test comments if it's too much. I just copy/pasted the values into the structs.

### DIFF
--- a/crates/interpreter/src/instructions/bitwise.rs
+++ b/crates/interpreter/src/instructions/bitwise.rs
@@ -135,13 +135,8 @@ mod tests {
 
     #[test]
     fn test_shift_left() {
-        let contract = Contract {
-            input: Bytes::default(),
-            bytecode: BytecodeLocked::default(),
-            ..Default::default()
-        };
         let mut host = DummyHost::new(Env::default());
-        let mut interpreter = Interpreter::new(contract.clone(), u64::MAX, false);
+        let mut interpreter = Interpreter::new(Contract::default(), u64::MAX, false);
 
         struct TestCase {
             value: &'static str,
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -103,19 +104,560 @@ pub fn sar<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut H) {
...
+        let test_cases = [
+            /*
+            PUSH 0x0000000000000000000000000000000000000000000000000000000000000001
+            PUSH 0x00
+            SHR
+            ---
+            0x0000000000000000000000000000000000000000000000000000000000000001
+            */
+            TestCase {
+                value: "0x0000000000000000000000000000000000000000000000000000000000000001",
```
rakita, line 312: You can still use `uint` macro with hex `0xfffff_U256`

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -103,19 +104,560 @@ pub fn sar<H: Host, SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut H) {
...
+        }
+
+        let test_cases = [
+            /*
+            PUSH 0x0000000000000000000000000000000000000000000000000000000000000001
+            PUSH 0x00
+            SHR
+            ---
+            0x0000000000000000000000000000000000000000000000000000000000000001
+            */
```
rakita, line 310: Comment with value looks redundant

crates/interpreter/src/instructions/bitwise.rs
jtraglia: Hmm yeah. I thought about that. I thought it was cleaner as strings. I'll change it though.

crates/interpreter/src/instructions/bitwise.rs
jtraglia: Yes thanks, it's much cleaner without these.

rakita (approved):
lgtm

PR #1257, opened 2024-04-02
feat(`db`): Introduce `alloydb`

Evalir:
Adds an alloy-powered `Database` implementation, which uses the alloy providers to fetch account info. This is the direct alloy equivalent for `EthersDB`.

This is blocked on:
- [ ] We need to publish the used alloy crates (`alloy-provider`, `alloy-rpc-types`, `alloy-transport`), as right now these are patched to git.

### DIFF
--- a/crates/revm/src/db/alloydb.rs
+++ b/crates/revm/src/db/alloydb.rs
@@ -2,28 +2,32 @@ use crate::{
     db::{Database, DatabaseRef},
     primitives::{AccountInfo, Address, Bytecode, B256, KECCAK_EMPTY, U256},
 };
-use alloy_provider::Provider;
+use alloy_provider::{Network, Provider};
 use alloy_rpc_types::BlockId;
-use alloy_transport::TransportError;
+use alloy_transport::{Transport, TransportError};
 use tokio::runtime::{Builder, Handle};
 
 /// An alloy-powered REVM [Database].
 ///
 /// When accessing the database, it'll use the given provider to fetch the corresponding account's data.
 #[derive(Debug, Clone)]
-pub struct AlloyDB<P: Provider> {
+pub struct AlloyDB<T: Transport + Clone, N: Network, P: Provider<T, N>> {
     /// The provider to fetch the data from.
     provider: P,
     /// The block number on which the queries will be based on.
     block_number: Option<BlockId>,
+    _network: std::marker::PhantomData<N>,
+    _transport: std::marker::PhantomData<T>,
 }
 
-impl<P: Provider> AlloyDB<P> {
+impl<T: Transport + Clone, N: Network, P: Provider<T, N>> AlloyDB<T, N, P> {
     /// Create a new AlloyDB instance, with a [Provider] and a block (Use None for latest).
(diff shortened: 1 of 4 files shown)


crates/revm/src/db/alloydb.rs
```diff
@@ -0,0 +1,181 @@
...
+use std::sync::Arc;
+use tokio::runtime::{Builder, Handle};
+
+/// An alloy-powered REVM [Database].
+///
+/// When accessing the database, it'll use the given provider to fetch the corresponding account's data.
+#[derive(Debug, Clone)]
+pub struct AlloyDB<P: Provider> {
+    /// The provider to fetch the data from.
+    provider: Arc<P>,
```
gakonst, line 17: arc not needed

crates/revm/src/db/alloydb.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[test]
+    fn can_get_basic() {
+        let client = ProviderBuilder::new()
+            .on_reqwest_http(
+                "https://mainnet.infura.io/v3/c60b0bb42f8a4c6481ecd229eddaca27"
+                    .parse()
+                    .unwrap(),
+            )
+            .unwrap()
+            .boxed();
```
gakonst, line 169: boxed not needed i think

crates/revm/src/db/alloydb.rs
Evalir: ah sweet, the base provider is arc now!

crates/revm/src/db/alloydb.rs
Evalir: Ah yeah, this is an interesting byproduct of the `<T, N>` bounds on `Provider<T, N>`:

In _this_ case, this was needed because `AlloyDB<P: Provider>` was defaulting to `Provider<T = BoxTransport, N = Ethereum>`, therefore it was expecting a boxed transport. 

To make this fully generic we need to make `AlloyDb` be `pub struct AlloyDB<T: Transport + Clone, N: Network, P: Provider<T, N>>` So we can make `T: Transport + Clone, N: Network` and therefore accept any transport, and any network. The annoying issue about this is that rustc complains about `T` and `N` being "unused" because they're not in the body, so we have to mark them as used with phantom data markers. See https://github.com/rust-lang/rust/issues/23246

rakita (approved):
lgtm

crates/revm/src/db/alloydb.rs
```diff
@@ -0,0 +1,180 @@
...
+            balance,
+            nonce.to::<u64>(),
+            code_hash,
+            code,
+        )))
+    }
+
+    fn block_hash_ref(&self, number: U256) -> Result<B256, Self::Error> {
+        // Saturate usize
+        if number > U256::from(u64::MAX) {
```
yjhmelody, line 103: Wonder if need to determine block range. Since BLOCKHASH opcode semantics:
```
hash of the chosen block, or 0 if the block number is not in the valid range.
```

yjhmelody: @rakita

rakita: This is done inside instruction

PR #1267, opened 2024-04-06
perf(interpreter): use `pop_top!` where possible

DaniPopes:
Avoids a branch and a load/store of the stack length in `*LOAD` opcodes.

This is usually optimized out, but the code ends up being less than optimal as the optimizer does not recognize that we're writing into the same slot / it can't make that assumption for some reason.

I also removed slice bound checks in CALLDATALOAD as they were not optimized out, see inline comments.

Snailtracer:
- before: 1'326'516'806
- after: 1'315'075'774 (-0.86%)


rakita (approved):
lgtm

PR #1270, opened 2024-04-07
Add the modifies_memory macro

### LINKED ISSUE #1263: Opcode that modifies_memory
It would be great to have this helper functions inside revm: https://github.com/foundry-rs/foundry/blob/1281421e04144a8de0341f9cad9b623a012a74bf/crates/evm/core/src/opcodes.rs#L5-L25

h3lio5:
Port the opcode `modifies_memory` macro from https://github.com/foundry-rs/foundry/blob/1281421e04144a8de0341f9cad9b623a012a74bf/crates/evm/core/src/opcodes.rs#L5-L25

Fixes #1263

### DIFF
--- a/crates/interpreter/src/instructions/opcode.rs
+++ b/crates/interpreter/src/instructions/opcode.rs
@@ -496,28 +496,28 @@ impl OpCode {
     pub const fn get(self) -> u8 {
         self.0
     }
-}
 
-/// Returns true if the opcode modifies memory.
-/// <https://bluealloy.github.io/revm/crates/interpreter/memory.html#opcodes>
-/// <https://github.com/crytic/evm-opcodes>
-#[inline]
-pub const fn modifies_memory(opcode: OpCode) -> bool {
-    matches!(
-        opcode,
-        OpCode::EXTCODECOPY
-            | OpCode::MLOAD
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/instructions/opcode.rs
```diff
@@ -498,6 +498,28 @@ impl OpCode {
     }
 }
 
+/// Returns true if the opcode modifies memory.
+/// <https://bluealloy.github.io/revm/crates/interpreter/memory.html#opcodes>
+/// <https://github.com/crytic/evm-opcodes>
+#[inline]
+pub const fn modifies_memory(opcode: OpCode) -> bool {
```
mattsse, line 505: this should just be a function of the opcode type instead

mattsse (changes requested):

crates/interpreter/src/instructions/opcode.rs
h3lio5: sure, I moved the method into the struct.

mattsse (approved):
lgtm

rakita:
There will be some modification to that part of code when EOF PR is merged, but this lgtm!

PR #1277, opened 2024-04-08
feat(revme): add --keep-going to statetest command

DaniPopes:
And other minor improvements for debugging purposes.

`--keep-going` will not abort execution at the first failure and instead count all the errors.


onbjerg (approved):
lgtm, although maybe we should do `--no-fail-fast`? mirrors cargo

DaniPopes:
Already aliased to `--no-fail-fast`

rakita (approved):

PR #1284, opened 2024-04-09
feat: add flag to force hashbrown usage

Wodann:
This allows re-exporting of `hashbrown`, even if the `std` feature flag is enabled.

### DIFF
--- a/crates/interpreter/Cargo.toml
+++ b/crates/interpreter/Cargo.toml
@@ -25,6 +25,7 @@ serde = { version = "1.0", default-features = false, features = [
 [features]
 default = ["std"]
 std = ["serde?/std", "revm-primitives/std"]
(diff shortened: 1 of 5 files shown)


rakita (changes requested):
It would be good to propagate this flag to all crates.

Wodann:
> It would be good to propagate this flag to all crates.

Thanks. Done!

Wodann:
Could it be that this is a sporadic failure due to it requiring a remote connection?

https://github.com/bluealloy/revm/actions/runs/8632004425/job/23661622655?pr=1284#step:5:132

DaniPopes:
It's a spurious failure, RPCs like to suck, don't worry about it.

DaniPopes (approved):

rakita (approved):

PR #1301, opened 2024-04-12
Drops check for .json when testing a single file

### LINKED ISSUE #1247: revme statetest allow path to file that are not json
> If I rename it from ....json.min to foo.json, then it behaves differently. @rakita does reth treat non-json file suffixes differently? If so, I need to modify how my minimizer works.

Required here: https://github.com/ethereum/execution-specs/issues/917

Check if statest path is a folder or file and if it is file dont check postfix.

mw2000:
This allows statetest to be run on files that don't have a .json extension if test path is a single file path.

Fixes #1247


DaniPopes (approved):

mattsse (approved):

rakita (approved):
cool!

PR #1302, opened 2024-04-12
fix: correct some stack IO

DaniPopes:
CALLDATACOPY has no outputs, CREATE has 3 inputs (value, offset, size), CREATE2 has CREATE's inputs + 1 (salt).

Also the old opcode.rs file was not removed in EOF PR, so deleting it here.


mattsse (approved):

rakita (approved):
That is correct!

PR #1304, opened 2024-04-12
chore(interpreter): rename some macros

DaniPopes:
error on -> require


mattsse (commented):
alternative suggestion `ensure_<>!`

rakita:
> alternative suggestion `ensure_<>!`

require fits better here imo

PR #1306, opened 2024-04-12
chore(interpreter): rename wrapping_* opcodes

DaniPopes:
Wrapping doesn't make sense here, these should match the actual opcode name as defined in the yellow paper.


DaniPopes:
It's EVM opcodes, this is not the place for docs

rakita (approved):
I would like to expand the book with an instruction description. Fortunatly right now there are a lot of resources on opcodes: https://www.evm.codes/, but either way it is out of scope for this PR. Good rename!

PR #1307, opened 2024-04-12
refactor: shrink OpCodeInfo and add more methods

DaniPopes:
Based on https://github.com/bluealloy/revm/pull/1302

### DIFF
--- a/crates/interpreter/src/opcode.rs
+++ b/crates/interpreter/src/opcode.rs
@@ -286,8 +286,8 @@ impl OpCode {
 /// Information about opcode, such as name, and stack inputs and outputs.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct OpCodeInfo {
-    /// Invariant: `(name_ptr, name_len)` is a `&'static str`. It is a shorted variant of `str` as the name size is always 
-    /// less than 256 characters.
+    /// Invariant: `(name_ptr, name_len)` is a `&'static str`. It is a shorted variant of `str` as
+    /// the name length is always less than 256 characters.
     name_ptr: NonNull<u8>,
     name_len: u8,
     /// Stack inputs.
--- a/crates/interpreter/src/opcode.rs
+++ b/crates/interpreter/src/opcode.rs
@@ -3,7 +3,7 @@
 pub mod eof_printer;
 
 use crate::{instructions::*, primitives::Spec, Host, Interpreter};
-use core::fmt;
+use core::{fmt, ptr::NonNull};
 use std::boxed::Box;
 
 /// EVM opcode function signature.
@@ -137,7 +137,7 @@ impl fmt::Display for OpCode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let n = self.get();
         if let Some(val) = OPCODE_INFO_JUMPTABLE[n as usize] {
-            f.write_str(val.name)
+            f.write_str(val.name())
         } else {
             write!(f, "UNKNOWN(0x{n:02X})")
         }
@@ -186,7 +186,7 @@ impl OpCode {
         }
     }
 
-    /// Returns true if the opcode is a push instruction.
+    /// Returns true if the opcode is a `PUSH` instruction.
     #[inline]
     pub const fn is_push(self) -> bool {
         self.0 >= PUSH1 && self.0 <= PUSH32
@@ -214,9 +214,10 @@ impl OpCode {
     }
 
     /// Returns the opcode as a string.
+    #[doc(alias = "name")]
     #[inline]
     pub const fn as_str(self) -> &'static str {
-        self.info().name
+        self.info().name()
     }
 
     /// Returns the opcode name.
@@ -229,21 +230,26 @@ impl OpCode {
         }
     }
 
-    /// Returns inputs for the given opcode.
+    /// Returns the number of input stack elements.
+    #[inline]
     pub const fn inputs(&self) -> u8 {
-        self.info().inputs
+        self.info().inputs()
     }
 
-    /// Returns outputs for the given opcode.
+    /// Returns the number of output stack elements.
+    #[inline]
     pub const fn outputs(&self) -> u8 {
-        self.info().outputs
+        self.info().outputs()
     }
 
-    /// Returns a difference between input and output.
+    /// Calculates the difference between the number of input and output stack elements.
+    #[inline]
     pub const fn io_diff(&self) -> i16 {
         self.info().io_diff()
     }
 
+    /// Returns the opcode information for the given opcode.
+    #[inline]
     pub const fn info_by_op(opcode: u8) -> Option<OpCodeInfo> {
         if let Some(opcode) = Self::new(opcode) {
             Some(opcode.info())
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/opcode.rs
```diff
@@ -276,43 +284,132 @@ impl OpCode {
...
-    // If the opcode is return from execution. aka STOP,RETURN, ..
-    pub is_terminating_opcode: bool,
-    /// Size of opcode with its intermediate bytes.
-    ///
-    /// RJUMPV is special case where the bytes len is depending on bytecode value,
-    /// for RJUMV size will be set to one byte while minimum is two.
-    pub immediate_size: u8,
+    // TODO: make this a bitfield. This is OK for now since we have a few alignment bytes to spare.
+    // Invariant: `(name_ptr, name_len)` is a `&'static str`.
+    name_ptr: NonNull<u8>,
```
rakita, line 291: what is the benefit of this? Isn't `&'static str` exactly the same as `(ptr,len)`?

crates/interpreter/src/opcode.rs
```diff
@@ -276,43 +284,132 @@ impl OpCode {
...
+    name_len: u8,
+    inputs: u8,
+    outputs: u8,
+    not_eof: bool,
+    terminating: bool,
+    imm_size: u8,
```
rakita, line 297: Would revert to `immediate_size`, imm can be confusing. And revert comment's

crates/interpreter/src/opcode.rs
```diff
@@ -276,43 +284,132 @@ impl OpCode {
...
+        self.not_eof = true;
+        self
+    }
+
+    /// Sets the immediate bytes number.
+    ///
+    /// RJUMPV is special case where the bytes len is depending on bytecode value,
+    /// for RJUMPV size will be set to one byte while minimum is two.
+    #[inline]
+    pub const fn set_imm_size(mut self, n: u8) -> Self {
```
rakita, line 387: Would leave those three as standalone functions as their usage (at least for now) should be only related to `opcodes` macro.

crates/interpreter/src/opcode.rs
```diff
@@ -349,68 +447,45 @@ macro_rules! opcodes {
...
-    0x05 => SDIV       => arithmetic::sdiv           => stack_io<2, 1>;
-    0x06 => MOD        => arithmetic::rem            => stack_io<2, 1>;
-    0x07 => SMOD       => arithmetic::smod           => stack_io<2, 1>;
-    0x08 => ADDMOD     => arithmetic::addmod         => stack_io<3, 1>;
-    0x09 => MULMOD     => arithmetic::mulmod         => stack_io<3, 1>;
-    0x0A => EXP        => arithmetic::exp::<H, SPEC> => stack_io<2, 1>;
-    0x0B => SIGNEXTEND => arithmetic::signextend     => stack_io<2, 1>;
+    0x00 => STOP => control::stop => stack_io(0, 0), terminating;
+
+    0x01 => ADD        => arithmetic::wrapping_add   => stack_io(2, 1);
```
rakita, line 457: Nice one!

crates/interpreter/src/opcode.rs
DaniPopes: separated to make the structure smaller since length fits in u8

crates/interpreter/src/opcode.rs
DaniPopes: "imm" is a common abbreviation for "immediate" in bytecodes/VMs/compilers.
Comments were moved to doccomments in the related methods.

crates/interpreter/src/opcode.rs
DaniPopes: They're used in builder-style instantiation, so they make more sense as methods. I can however make them private if you want.

crates/interpreter/src/opcode.rs
```diff
@@ -655,34 +730,49 @@ mod tests {
...
+        expected[RJUMPV as usize] = 1;
+        expected[CALLF as usize] = 2;
+        expected[JUMPF as usize] = 2;
+        expected[DUPN as usize] = 1;
+        expected[SWAPN as usize] = 1;
+        expected[EXCHANGE as usize] = 1;
+        expected[EOFCREATE as usize] = 1;
+        expected[RETURNCONTRACT as usize] = 1;
+
+        for (i, opcode) in OPCODE_INFO_JUMPTABLE.iter().enumerate() {
```
DaniPopes, line 767: Updated this test since it was doing nothing before.

crates/interpreter/src/opcode.rs
rakita: That makes sense

crates/interpreter/src/opcode.rs
rakita: immediate is more readable for devs that are checking code.

crates/interpreter/src/opcode.rs
rakita: I would leave them as standalone, as in that way we dont clash with naming.

crates/interpreter/src/opcode.rs
```diff
@@ -276,43 +284,132 @@ impl OpCode {
...
-    pub immediate_size: u8,
+    // TODO: make this a bitfield. This is OK for now since we have a few alignment bytes to spare.
+    // Invariant: `(name_ptr, name_len)` is a `&'static str`.
+    name_ptr: NonNull<u8>,
+    name_len: u8,
+    inputs: u8,
+    outputs: u8,
+    not_eof: bool,
+    terminating: bool,
+    immediate_size: u8,
```
rakita, line 297:
```suggestion
    /// Invariant: `(name_ptr, name_len)` is a `&'static str`. It is a shorted variant of `str` as the name size is always 
    /// less than 256 characters.
    name_ptr: NonNull<u8>,
    name_len: u8,
    /// Stack inputs.
    inputs: u8,
    /// Stack outputs.
    outputs: u8,
    /// Number of intermediate bytes.
    ///
    /// RJUMPV is a special case where the bytes len depends on bytecode value,
    /// for RJUMV size will be set to one byte as it is the minimum immediate size.
    immediate_size: u8,
    /// Used by EOF verification. All not EOF opcodes are marked false.
    not_eof: bool,
    /// If the opcode stops execution. aka STOP, RETURN, ..
    terminating: bool,
``````

**Review by rakita — APPROVED** [review]
lgtm

File: crates/interpreter/src/opcode.rs [hunk 7]
```diff
@@ -276,43 +284,140 @@ impl OpCode {
...
-    pub name: &'static str,
-    pub inputs: u8,
-    pub outputs: u8,
-    // TODO make this a bitfield
-    pub is_eof: bool,
-    // If the opcode is return from execution. aka STOP,RETURN, ..
-    pub is_terminating_opcode: bool,
-    /// Size of opcode with its intermediate bytes.
+    /// Invariant: `(name_ptr, name_len)` is a `&'static str`. It is a shorted variant of `str` as the name size is always 
+    /// less than 256 characters.
```
rakita, line 290:
```suggestion
    /// Invariant: `(name_ptr, name_len)` is a `&'static str`. It is a shorted variant of `str` as
    /// the name length is always less than 256 characters.
```

not sure why fmt is failing here, checking doc length

PR #1327, opened 2024-04-17
Implement `with_chain_id` for `CfgEnv`

tcoratger:


crates/primitives/src/env.rs
```diff
@@ -338,6 +338,13 @@ pub struct CfgEnv {
 }
 
 impl CfgEnv {
+    pub fn with_chain_id(&self, chain_id: u64) -> Self {
```
mattsse, line 341: I think this should consume self?

mattsse (changes requested):

crates/primitives/src/env.rs
tcoratger: fixed

onbjerg (approved):

DaniPopes (approved):

rakita (approved):

PR #1344, opened 2024-04-19
chore: don't clone bytes in `Bytecode::bytes`

DaniPopes:
We can return a reference to `Bytes` instead of cloning it.
Because this change has not been released yet, I also renamed it for consistency to just `fn bytes`.

### DIFF
--- a/crates/primitives/src/bytecode.rs
+++ b/crates/primitives/src/bytecode.rs
@@ -91,17 +91,17 @@ impl Bytecode {
     }
 
     /// Returns a reference to the bytecode.
+    ///
     /// In case of EOF this will be the first code section.
     #[inline]
-    pub fn bytecode_bytes(&self) -> Bytes {
+    pub fn bytecode(&self) -> &Bytes {
         match self {
-            Self::LegacyRaw(bytes) => bytes.clone(),
-            Self::LegacyAnalyzed(analyzed) => analyzed.bytes(),
+            Self::LegacyRaw(bytes) => bytes,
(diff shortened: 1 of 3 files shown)


mattsse (approved):
although cheap to clone, we can still return a ref and leave it up to the caller.

no strong feelings about this

DaniPopes:
cheap != free :+1:
when taking slices &[u8] cloning is not great

crates/primitives/src/bytecode.rs
```diff
@@ -91,17 +91,17 @@ impl Bytecode {
     }
 
     /// Returns a reference to the bytecode.
+    ///
     /// In case of EOF this will be the first code section.
     #[inline]
-    pub fn bytecode_bytes(&self) -> Bytes {
+    pub fn bytes(&self) -> &Bytes {
```
rakita, line 97: Imo `bytecode` would be a better name

crates/primitives/src/bytecode/legacy.rs
```diff
@@ -42,8 +42,8 @@ impl LegacyAnalyzedBytecode {
     /// Returns bytes of bytecode.
     ///
     /// Bytes are padded with 32 zero bytes.
-    pub fn bytes(&self) -> Bytes {
-        self.bytecode.clone()
+    pub fn bytes(&self) -> &Bytes {
```
rakita, line 45: `.bytecode()`

rakita (approved):

PR #1345, opened 2024-04-19
feat: add helper methods to CallInputs

DaniPopes:
- It's not very intuitive how to update `CallInputs` address values, make this clearer in doc comments and some helper methods.
- The old `inputs.transfer.value` is not the same as the new `inputs.call_value()` due to DELEGATECALL, make this clearer by distinguishing this difference clearly.

Discovered in and helps with https://github.com/paradigmxyz/evm-inspectors/pull/97


crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -81,19 +132,61 @@ pub enum CallScheme {
...
+/// Call value.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub enum TransferValue {
-    /// Transfer value from caller to callee.
-    Value(U256),
-    /// For delegate call, the value is not transferred but
-    /// apparent value is used for CALLVALUE opcode
-    ApparentValue(U256),
+pub enum CallValue {
```
rakita, line 138: a lot better!

PR #1346, opened 2024-04-20
perf: remove bounds check in DUP, SWAP/EXCHANGE

DaniPopes:
It was introduced by making `n` dynamic instead of constant

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,11 @@ default-members = ["crates/revm"]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[profile.release]
+lto = true
+codegen-units = 1
+debug = true
+
 [profile.ethtests]
 inherits = "test"
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,11 +14,6 @@ default-members = ["crates/revm"]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
-[profile.release]
-lto = true
-codegen-units = 1
-debug = true
-
 [profile.ethtests]
 inherits = "test"
 opt-level = 3
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -251,30 +251,35 @@ impl Stack {
     /// # Panics
     ///
     /// Panics if `n` is 0.
-    #[inline]
+    #[inline(always)]
     #[cfg_attr(debug_assertions, track_caller)]
     pub fn swap(&mut self, n: usize) -> Result<(), InstructionResult> {
-        assume!(n > 0, "attempted to swap with 0");
-        let len = self.data.len();
-        if n >= len {
-            return Err(InstructionResult::StackUnderflow);
-        }
-        let last_index = len - 1;
-        self.data.swap(last_index, last_index - n);
-        Ok(())
+        self.exchange(0, n)
     }
 
-    /// Exchange two values on the stack. where `N` is first index and second index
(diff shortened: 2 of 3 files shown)


crates/interpreter/src/interpreter/stack.rs
```diff
@@ -222,8 +222,14 @@ impl Stack {
...
 
     /// Duplicates the `N`th value from the top of the stack.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `n` is 0.
     #[inline]
+    #[cfg_attr(debug_assertions, track_caller)]
     pub fn dup(&mut self, n: usize) -> Result<(), InstructionResult> {
+        assume!(n > 0, "attempted to dup 0");
```
rakita, line 232: What does `assume` do here, if we are checking `n >0`, as N is usize, it always will be more than 0?

DaniPopes: n == 0 doesn't make sense, there is no dup0 or swap0

onbjerg (approved):

crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -36,7 +36,7 @@ pub struct CallInputs {
     ///
     /// Previously `context.scheme`.
     pub scheme: CallScheme,
-    /// Whether the call is initiated inside a static call.
+    /// Whether the call is, or is initiated inside a static call.
```
onbjerg, line 39: think a word is missing here

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -232,36 +238,48 @@ impl Stack {
...
         let n_m_index = n + m;
         if n_m_index >= len {
             return Err(InstructionResult::StackUnderflow);
         }
-        let last_index = len - 1;
-        self.data.swap(last_index - n, last_index - n_m_index);
+        // SAFETY: `n` and `n_m` are checked to be within bounds, and they don't overlap.
+        unsafe {
+            // NOTE: `mem::swap` performs better than `slice::swap`/`ptr::swap`, as they cannot
+            // assume that the pointers are non-overlapping when we know they are not.
```
onbjerg, line 279: just to clarify: when we know they are not overlapping, right? this reads as "we know they are not non-overlapping" → "we know they are overlapping"

crates/interpreter/src/interpreter_action/call_inputs.rs
DaniPopes: is,
or
is initialized inside

I think this is correct

crates/interpreter/src/interpreter/stack.rs
DaniPopes: Yeah it's not great, chat gpt'd it

crates/interpreter/src/interpreter_action/call_inputs.rs
DaniPopes: rewrote it anyway

onbjerg: ah i see what this means now, yeah i think it's correct but it's hard to parse 😅

Cargo.toml
```diff
@@ -14,11 +14,6 @@ default-members = ["crates/revm"]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
-[profile.release]
-lto = true
-codegen-units = 1
```
rakita, line 19: I would leave `lto` and `codegen-units` as they can help with performance.

DaniPopes: remnants from debugging

Issue #1084, opened 2024-02-13
add more example in revm example crate 


Degen-Player:
make more usable example for revm

rakita:
@Degen-Player what examples would be usable for you?

Degen-Player:
> @Degen-Player what examples would be usable for you?

currently ethers-rs has a lot of very interesting examples. It would be great if revm had examples like this. It would be much easier to integrate revm

rakita:
Yep those examples are amazing. Don't have time to make them rn for Revm as there are few more priority tasks but will have this in mind

persondai:
add uniswap v3 example

Peponks9:
Hi @rakita from what i could follow from the activity looks like this issue might be closed or still open for more examples?

YXZ252426:
@rakita Hi, I’m thinking about opening a PR for a small `local_execution_tracer` example.

It would show a minimal custom `Inspector` that runs local bytecode in an in-memory DB and prints structured JSON traces: opcode steps, calls, logs, storage touches, and final state diffs.

I saw `block_traces` and `cheatcode_inspector`, but this would be a simpler RPC-free “build your own tracer” example.

Would this fit the examples directory?

PR #1353, opened 2024-04-22
Add uniswap V2 WETH-USDC swap example

pawurb:
Related to https://github.com/bluealloy/revm/issues/1084 . This example  shows common use cases not covered by other examples. 

Screenshot 2024-04-22 at 13 51 58

### DIFF
--- a/examples/uniswap_v2_usdc_swap.rs
+++ b/examples/uniswap_v2_usdc_swap.rs
@@ -145,7 +145,7 @@ async fn balance_of(
     token: Address,
     address: Address,
     cache_db: &mut CacheDB<EmptyDBTyped<Infallible>>,
-) -> anyhow::Result<u128> {
+) -> anyhow::Result<U256> {
     sol! {
         function balanceOf(address account) public returns (uint256);
     }
@@ -174,7 +174,7 @@ async fn balance_of(
         result => panic!("'balance_of' execution failed: {result:?}"),
     };
 
-    let balance: u128 = <u128>::abi_decode(&value, false)?;
+    let balance: U256 = <U256>::abi_decode(&value, false)?;
 
     Ok(balance)
 }
@@ -218,9 +218,9 @@ async fn get_amount_out(
         result => panic!("'get_amount_out' execution failed: {result:?}"),
     };
 
-    let result = <u128>::abi_decode(&value, false)?;
+    let result = <U256>::abi_decode(&value, false)?;
 
-    Ok(U256::from(result))
+    Ok(result)
 }
 
 async fn get_reserves(
@@ -254,10 +254,10 @@ async fn get_reserves(
         result => panic!("'get_reserves' execution failed: {result:?}"),
     };
 
-    let (reserve0, reserve1, _): (u128, u128, u32) =
-        <(u128, u128, u32)>::abi_decode(&value, false)?;
+    let (reserve0, reserve1, _): (U256, U256, u32) =
+        <(U256, U256, u32)>::abi_decode(&value, false)?;
 
-    Ok((U256::from(reserve0), U256::from(reserve1)))
+    Ok((reserve0, reserve1))
 }
 
 async fn swap(
--- a/examples/uniswap_v2_usdc_swap.rs
+++ b/examples/uniswap_v2_usdc_swap.rs
@@ -1,0 +1,357 @@
+use ethers_contract::BaseContract;
+use ethers_core::{
+    abi,
+    abi::parse_abi,
+    types::{Bytes, H160, U256},
+    utils::to_checksum,
+};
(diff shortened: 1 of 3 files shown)


examples/uniswap_v2_usdc_swap.rs
```diff
@@ -0,0 +1,357 @@
+use ethers_contract::BaseContract;
```
onbjerg, line 1: please use alloy, ethers is deprecated

onbjerg (changes requested):

examples/uniswap_v2_usdc_swap.rs
pawurb: I have very limited Rust experience, so probably I'm missing something. But is the project currently compatibly with alloy? Adding `alloy-primitives` crate causes dozens of compile errors across the project. I can look into rewriting this example but tips on which crates I should use/avoid would be helpful.

onbjerg: yes, it is compatible with alloy, it uses alloy internally: https://github.com/bluealloy/revm/blob/1ca3d39f6a9e9778f8eb0fcb74fe529345a531b4/crates/primitives/src/lib.rs

any crate should be fine, but it should use the same version as the crate itself, so currently 0.7: https://github.com/bluealloy/revm/blob/1ca3d39f6a9e9778f8eb0fcb74fe529345a531b4/crates/primitives/Cargo.toml#L20

pawurb:
@onbjerg Should be fixed. I think `ethers_providers` has to stay because of `EthersDB`, or is there some alloy alternative?

onbjerg:
@pawurb you can use `AlloyDB`! 😄

examples/uniswap_v2_usdc_swap.rs
```diff
@@ -0,0 +1,348 @@
...
+            output: Output::Call(value),
+            ..
+        } => value,
+        result => panic!("'get_reserves' execution failed: {result:?}"),
+    };
+
+    let (reserve0, reserve1): (u128, u128) = match <(u128, u128, u32)>::abi_decode(&value, false) {
+        Ok((reserve0, reserve1, _)) => (reserve0, reserve1),
+        Err(e) => panic!("'get_reserves' decode failed: {:?}", e),
+    };
```
onbjerg, line 257: you should be able to just use `?` her

pawurb: I'm getting:

```
the trait `StdError` is not implemented for `alloy_sol_types::Error`, which is required by `std::result::Result<(revm::revm_primitives::alloy_primitives::Uint<256, 4>, revm::revm_primitives::alloy_primitives::Uint<256, 4>), anyhow::Error>: FromResidual<std::result::Result<Infallible, alloy_sol_types::Error>>
```

I can wrap it in `anyhow` error, but it does not look much better than current panic.

onbjerg: if you want to panic, use `expect` instead

pawurb: In the end `?` works with `std` feature enabled.

examples/uniswap_v2_usdc_swap.rs
```diff
@@ -0,0 +1,343 @@
...
+    let client = ProviderBuilder::new()
+        .on_reqwest_http(
+            "https://mainnet.infura.io/v3/c60b0bb42f8a4c6481ecd229eddaca27"
+                .parse()
+                .unwrap(),
+        )
+        .unwrap();
+    let client = Arc::new(client);
+    let mut alloydb = AlloyDB::new(client, None);
+    let mut cache_db = CacheDB::new(EmptyDB::default());
```
onbjerg, line 25: you can just do `CacheDB::new(AlloyDB::new(client, None))`, then calls to e.g. `db.basic` will automatically first get pulled from the alloy provider, and then stored in the cache db! that avoids you needing to do what you are doing on eg L45, L48 etc

onbjerg (dismissed):

rakita (approved):

rakita:
pending comment https://github.com/bluealloy/revm/pull/1353#discussion_r1580970513

pawurb:
> pending comment [#1353 (comment)](https://github.com/bluealloy/revm/pull/1353#discussion_r1580970513)

@rakita I don't understand, I think I've already applied suggestions from this comment.

rakita:
> > pending comment [#1353 (comment)](https://github.com/bluealloy/revm/pull/1353#discussion_r1580970513)
> 
> @rakita I don't understand, I think I've already applied suggestions from this comment.

That is true, didn't see that is a outdated comment. lgtm, merging

PR #1354, opened 2024-04-22
feat: implement EIP-2935

onbjerg:
Implements the changes for `BLOCKHASH` as outlined in [EIP-2935], and adds EIP-2935 constants.

This requires an external processing step to fill the contract's storage with data. See https://github.com/paradigmxyz/reth/pull/7818

Additionally, since `BLOCKHASH` now does an `SLOAD` internally, I pulled the `sload` logic into a reusable `sload!` macro.

[EIP-2935]:  https://eips.ethereum.org/EIPS/eip-2935

### DIFF
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -5,7 +5,7 @@ use crate::{
     Host, InstructionResult, SStoreResult,
 };
 use core::cmp::min;
-use revm_primitives::{Address, BLOCK_HASH_HISTORY, HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS};
+use revm_primitives::{BLOCK_HASH_HISTORY, HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS};
 use std::vec::Vec;
 
 pub fn balance<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -5,7 +5,7 @@ use crate::{
     Host, InstructionResult, SStoreResult,
 };
 use core::cmp::min;
-use revm_primitives::{BLOCK_HASH_HISTORY, HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS};
+use revm_primitives::{Address, BLOCK_HASH_HISTORY, HISTORY_SERVE_WINDOW, HISTORY_STORAGE_ADDRESS};
 use std::vec::Vec;
 
 pub fn balance<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, host: &mut H) {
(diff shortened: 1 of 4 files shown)


crates/interpreter/src/instructions/host.rs
```diff
@@ -103,14 +103,28 @@ pub fn extcodecopy<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter,
...
-        if diff <= BLOCK_HASH_HISTORY && diff != 0 {
+        if SPEC::enabled(PRAGUE) && diff <= HISTORY_SERVE_WINDOW && diff != 0 {
+            let Some((value, is_cold)) = host.sload(
+                HISTORY_STORAGE_ADDRESS,
+                number.wrapping_rem(U256::from(HISTORY_SERVE_WINDOW)),
+            ) else {
+                interpreter.instruction_result = InstructionResult::FatalExternalError;
+                return;
+            };
+            gas!(interpreter, gas::sload_cost(SPEC::SPEC_ID, is_cold));
```
DaniPopes, line 124: Maybe extract the `SLOAD` logic from its function into a separate function where we also call here?

crates/primitives/src/constants.rs
```diff
@@ -1,12 +1,24 @@
+use alloy_primitives::address;
+
 use crate::Address;
```
DaniPopes, line 3:
```suggestion
use crate::{Address, address};
```

crates/interpreter/src/instructions/host.rs
```diff
@@ -103,14 +103,28 @@ pub fn extcodecopy<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter,
...
 
-    if let Some(diff) = host.env().block.number.checked_sub(*number) {
+    let block_number = host.env().block.number;
+
+    if let Some(diff) = block_number.checked_sub(*number) {
         let diff = as_usize_saturated!(diff);
+
         // blockhash should push zero if number is same as current block number.
-        if diff <= BLOCK_HASH_HISTORY && diff != 0 {
+        if SPEC::enabled(PRAGUE) && diff <= HISTORY_SERVE_WINDOW && diff != 0 {
```
DaniPopes, line 116: both branches have `diff != 0`, could be an extra branch or an if nest

DaniPopes (approved):

crates/interpreter/src/instructions/host.rs
onbjerg: yeah i had this in mind too, i just forgor

crates/interpreter/src/instructions/host.rs
onbjerg: it put it in a match

crates/interpreter/src/instructions/host.rs
onbjerg: i tried making a func but mut borrow of interpreter from `pop_top!` prevented me, so I made an `sload!` macro instead

DaniPopes (approved):

onbjerg:
Putting this in draft, unfortunately the current impl of EIP-2935 requires some changes to how we persist accounts and storage since it is exempt from state clearing. This will be changed in a later change to the EIP but the spec is currently frozen for devnet 0

onbjerg:
After some consulting in Eth R&D discord we are likely going to add a nonce > 1 or a balance of 1 wei to the EIP-2935 system account until the EIP is fixed, so this is RFR again.

rakita (approved):
lgtm with one nit on naming

crates/primitives/src/constants.rs
```diff
@@ -1,34 +1,54 @@
...
 
+/// EIP-2935: Serve historical block hashes from state
+///
+/// Number of block hashes the EVM can access in the past (Prague).
+pub const HISTORY_SERVE_WINDOW: usize = 8192;
+
+/// EIP-2935: Serve historical block hashes from state
+///
+/// The address where historical blockhashes are available.
+pub const HISTORY_STORAGE_ADDRESS: Address = address!("25a219378dad9b3503c8268c9ca836a52427a4fb");
```
rakita, line 19: `HISTORY_` can be ambiguous and the contract needs to be related to `BLOCKHASH` so maybe `BLOCK_HASH_STORAGE_ADDRESS`?

onbjerg: sgtm, i just copied over the constant names from the eip, will rename + refer ppl to the original constant name in the docs so its easy to compare with the eip

onbjerg:
@rakita Renamed the constants, PTAL

PR #1357, opened 2024-04-24
feat: add `sp1` precompile

ratankaliani:


crates/precompile/src/secp256k1.rs
```diff
@@ -8,7 +8,31 @@ pub const ECRECOVER: PrecompileWithAddress = PrecompileWithAddress(
...
+        // TODO: Can we clean up the interface for this logic?
+        let recovered_key = sp1_precompiles::secp256k1::ecrecover(&sig_plus_recid, msg)?;
+
+        let mut hash = keccak256(&recovered_key[1..]);
+
+        // truncate to 20 bytes
+        hash[..12].fill(0);
+        Ok(hash)
+    }
+}
```
gakonst, line 30: @mattsse @Rjected we should show to Succinct how to do this precompile override with EVMBuilder, similar to 
* https://github.com/paradigmxyz/alphanet/blob/main/crates/node/src/evm.rs#L66-L82
* https://github.com/paradigmxyz/alphanet/blob/main/crates/node/src/evm.rs#L114
cc @ratankaliani

gakonst (changes requested):
I don't think we'll want to upstream this feature flag in the revm repo, but we could add a feature flag I think cc @rakita that disables secp so that revm compiles, and we can use the handlers as I shared in the comment above to inject the "kosher" secp impl

rakita:
I will close this. Please look at examples on how to add a custom precompile: https://bluealloy.github.io/revm/crates/revm/builder.html#examples

And @ratankaliani If you need to fully remove the dependency, we can add the feature gate for it: https://github.com/bluealloy/revm/issues/1366

PR #1358, opened 2024-04-26
feat: parse opcodes from strings

DaniPopes:
Add `FromStr` to `OpCode` as the inverse of `fn as_str`. I put it under a default feature because we need `phf` for the map, but the dependency might not be desired if this feature is not used.

### DIFF
--- a/crates/interpreter/src/opcode.rs
+++ b/crates/interpreter/src/opcode.rs
@@ -125,6 +125,21 @@ where
     core::array::from_fn(|i| outer(table[i]))
 }
 
+/// An error indicating that an opcode is invalid.
+#[derive(Debug, PartialEq, Eq)]
+#[cfg(feature = "parse")]
+pub struct OpCodeError(());
+
+#[cfg(feature = "parse")]
+impl fmt::Display for OpCodeError {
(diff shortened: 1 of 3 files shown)


mattsse (approved):
reasonable

crates/interpreter/src/opcode.rs
```diff
@@ -125,6 +125,21 @@ where
     core::array::from_fn(|i| outer(table[i]))
 }
 
+/// An error indicating that an opcode is invalid.
+#[derive(Debug, PartialEq, Eq)]
+#[cfg(feature = "parse")]
+pub struct OpCodeError(());
```
onbjerg, line 131:
```suggestion
pub struct OpCodeError;
```

nit: should be enough?

onbjerg (approved):

crates/interpreter/src/opcode.rs
DaniPopes: that makes it non-instantiable from outside

DaniPopes:
Yeah they're not heavy, but I don't think most users will need this so might as well

PR #1359, opened 2024-04-28
fix: return the correct error in resize_memory

DaniPopes:
There is a duplicate `MemoryLimitOOG`, when it should be `MemoryOOG`.

Looks like this was never correct https://github.com/bluealloy/revm/commit/9b663bbccba6be2ba70dabb5479f994514d0f8c4


onbjerg (approved):
ok this took me a while, `MemoryLimitOOG` is if we specify a memory limit when the feature `memory_limit` is enabled and we expand memory more than the specified limit, while `MemoryOOG` is for regular OOG when interpreter memory is expanded and we run out of gas

`MemoryLimitOOG` is kind of a weird name

rakita (approved):

PR #1361, opened 2024-04-28
Memory expansion gas

DaniPopes:
Since we record normal gas way more than memory, remove the excess operation in `Gas::record_cost` by recalculating the current memory expansion cost when needed.

This has the only downside of not storing the last memory expansion cost, but this is currently not used in revm, reth or foundry (except for one cheatcode). It is still available by calculating it with `SharedMemory::current_expansion_cost`.

This is what evmone does: https://github.com/ethereum/evmone/blob/82407d06cbddae736fa329b2adb521de2912b8a1/lib/evmone/instructions.hpp#L80

Based on https://github.com/bluealloy/revm/pull/1360.

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -1,10 +1,5 @@
+use core::{cmp::min, fmt, ops::Range};
 use revm_primitives::{B256, U256};
-
-use core::{
-    cmp::min,
-    fmt,
-    ops::{BitAnd, Not, Range},
-};
 use std::vec::Vec;
 
 /// A sequential memory shared between calls, which uses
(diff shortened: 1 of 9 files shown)


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -312,39 +313,17 @@ impl SharedMemory {
...
-    x.saturating_add(r)
+pub const fn num_words(len: u64) -> u64 {
+    len.saturating_add(31) / 32
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
-    #[test]
```
onbjerg, line 327: add these back for num words

onbjerg (approved):

Cargo.toml
```diff
@@ -15,9 +15,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [profile.release]
-lto = true
 codegen-units = 1
-debug = true
```
rakita, line 20: revert those?

rakita (approved):
I worry a little about removal of  memory from Gas, as to not break something from lib users. But this is clearly better!

nit on lto/debug

PR #1367, opened 2024-04-30
Add `AuthCall` variant to `CallScheme`

tcoratger:
Related https://eips.ethereum.org/EIPS/eip-3074

### DIFF
--- a/crates/interpreter/src/interpreter_action/call_inputs.rs
+++ b/crates/interpreter/src/interpreter_action/call_inputs.rs
@@ -130,6 +130,8 @@ pub enum CallScheme {
     DelegateCall,
     /// `STATICCALL`
     StaticCall,
+    /// `AUTHCALL`
+    AuthCall,
 }
 
 /// Call value.


DaniPopes (approved):

rakita:
@Rjected does this enum need to be used in the handler in Alphanet to be useful?

Rjected:
> @Rjected does this enum need to be used in the handler in Alphanet to be useful?

It's not used in the handler, no, I'm not sure if this is needed actually

Rjected:
it looks like it exists in the branch of revm used in alphanet-foundry:
https://github.com/wevm/revm/blob/66a499fc281b493a34ee42069309cf7aed363052/crates/interpreter/src/inner_models.rs#L109

warning: old revm

Still need to understand how this is used to determine if it's necessary

rakita:
I will close this for now, and we can always revisit it in future.

There is possibility that EIP-3074 is not going in

PR #1369, opened 2024-05-01
feat: add a hook to optionally execute individual frames

DaniPopes:
Allows for substituting the default interpreter execution.

### DIFF
--- a/crates/revm/src/builder.rs
+++ b/crates/revm/src/builder.rs
@@ -488,7 +488,7 @@ mod test {
                     },
                 );
 
-                // need to make ensure the instruction table is a boxed instruction table so that we
+                // need to  ensure the instruction table is a boxed instruction table so that we
                 // can insert the custom instruction as a boxed instruction
                 handler
                     .instruction_table
--- a/crates/revm/src/builder.rs
+++ b/crates/revm/src/builder.rs
@@ -15,7 +15,7 @@ use std::boxed::Box;
 pub struct EvmBuilder<'a, BuilderStage, EXT, DB: Database> {
     context: Context<EXT, DB>,
     /// Handler that will be used by EVM. It contains handle registers
-    handler: Handler<'a, Evm<'a, EXT, DB>, EXT, DB>,
+    handler: Handler<'a, Context<EXT, DB>, EXT, DB>,
     /// Phantom data to mark the stage of the builder.
     phantom: PhantomData<BuilderStage>,
 }
@@ -269,7 +269,7 @@ impl<'a, BuilderStage, EXT, DB: Database> EvmBuilder<'a, BuilderStage, EXT, DB> 
     /// Creates the default handler.
     ///
     /// This is useful for adding optimism handle register.
-    fn handler(handler_cfg: HandlerCfg) -> Handler<'a, Evm<'a, EXT, DB>, EXT, DB> {
+    fn handler(handler_cfg: HandlerCfg) -> Handler<'a, Context<EXT, DB>, EXT, DB> {
         Handler::new(handler_cfg)
     }
 
@@ -291,7 +291,7 @@ impl<'a, BuilderStage, EXT, DB: Database> EvmBuilder<'a, BuilderStage, EXT, DB> 
     /// ```
     pub fn with_handler(
         self,
-        handler: Handler<'a, Evm<'a, EXT, DB>, EXT, DB>,
+        handler: Handler<'a, Context<EXT, DB>, EXT, DB>,
     ) -> EvmBuilder<'a, BuilderStage, EXT, DB> {
         EvmBuilder {
             context: self.context,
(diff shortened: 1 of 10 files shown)


crates/revm/src/evm.rs
```diff
@@ -274,19 +274,33 @@ impl<EXT, DB: Database> Evm<'_, EXT, DB> {
...
+                        next_action = a;
+                        break 'blk;
+                    }
+                }
+
+                // Otherwise, or if the hook returned `None`, run the interpreter.
+                let interpreter = stack_frame.interpreter_mut();
+                next_action = interpreter.run(shared_memory, instruction_table, self);
+                // Take the shared memory back.
+                shared_memory = interpreter.take_memory();
```
rakita, line 298: This doesn't look like a good flow. The handler should allow overriding logic, so the best option would be for `interpreter.run` to be inside handler logic.

For this to be possible we should input instruction_tables directly to the handler call. I can make those changes if you want?

crates/revm/src/handler/handle_types/execution.rs
```diff
@@ -3,20 +3,26 @@ use crate::{
...
 /// Handles first frame return handle.
 pub type LastFrameReturnHandle<'a, EXT, DB> = Arc<
     dyn Fn(&mut Context<EXT, DB>, &mut FrameResult) -> Result<(), EVMError<<DB as Database>::Error>>
         + 'a,
 >;
 
+/// Executes a single frame. Errors can be returned in the EVM context.
+/// If `None` is returned, the frame is instead executed normally through the interpreter.
+pub type ExecuteFrameHandle<'a, EXT, DB> = Arc<
+    dyn Fn(&mut Frame, &mut SharedMemory, &mut Evm<'a, EXT, DB>) -> Option<InterpreterAction> + 'a,
```
rakita, line 23: Do you need full `Evm` here? To be consistent it would be better to have `Context` here.

DaniPopes: `Host` is implemented on `Evm` so yes

crates/revm/src/evm.rs
DaniPopes: I was thinking to *optionally* override, with fallback to default execution.

rakita: It is a good alternative that you made, but in the future, i am building up to introduce generic call/create/eofcreate/auth (whatever) in that sense it should be called every time.

crates/revm/src/handler/handle_types/execution.rs
rakita: Yeah I do think it should be moved to `InnerEvmContext`. In recursion it was needed to be on Evm but this is not the case anymore.

rakita: Moved `Host` impl to `Context` as there is restriction inside Handler. Context is needed for instructions and other handles.

crates/revm/src/builder.rs
```diff
@@ -481,22 +481,18 @@ mod test {
...
-                    move |_interp: &mut Interpreter, _host: &mut Evm<'_, (), InMemoryDB>| {
+                    move |_interp: &mut Interpreter, _host: &mut Context<(), InMemoryDB>| {
                         // modify the value
                         let mut inner = custom_context.inner.borrow_mut();
                         *inner += 1;
                     },
                 );
 
-                // need to make esure the instruction table is a boxed instruction table so that we
+                // need to make ensure the instruction table is a boxed instruction table so that we
```
DaniPopes, line 491:
```suggestion
                // need to ensure the instruction table is a boxed instruction table so that we
```

crates/revm/src/evm.rs
```diff
@@ -74,6 +70,102 @@ impl<'a, EXT, DB: Database> Evm<'a, EXT, DB> {
...
+        let mut stack_frame = call_stack.last_mut().unwrap();
+
+        loop {
+            // Execute the frame.
+            let next_action =
+                self.handler
+                    .execute_frame(stack_frame, &mut shared_memory, &mut self.context);
+
+            // Take error and break the loop, if any.
+            // This error can be set in the Interpreter when it's interacting with the host (Self).
```
DaniPopes, line 98:
```suggestion
            // This error can be set in the Interpreter when it interacts with the context.
```

crates/revm/src/handler/handle_types/execution.rs
```diff
@@ -149,6 +164,17 @@ impl<'a, EXT: 'a, DB: Database + 'a> ExecutionHandler<'a, EXT, DB> {
 }
 
 impl<'a, EXT, DB: Database> ExecutionHandler<'a, EXT, DB> {
+    #[inline]
```
DaniPopes, line 167:
```suggestion
    /// Executes a single frame.
    #[inline]
```

crates/revm/src/handler/handle_types/execution.rs
```diff
@@ -5,18 +5,30 @@ use crate::{
...
 
+/// Executes a single frame. Errors can be returned in the EVM context.
+/// If `None` is returned, the frame is instead executed normally through the interpreter.
+pub type ExecuteFrameHandle<'a, EXT, DB> = Arc<
+    dyn Fn(
+            &mut Frame,
+            &mut SharedMemory,
+            &InstructionTables<'_, Context<EXT, DB>>,
+            &mut Context<EXT, DB>,
+        ) -> InterpreterAction
```
DaniPopes, line 28: Should this return a result?

crates/revm/src/handler/handle_types/execution.rs
```diff
@@ -5,18 +5,30 @@ use crate::{
...
+use std::{boxed::Box, sync::Arc};
 
 /// Handles first frame return handle.
 pub type LastFrameReturnHandle<'a, EXT, DB> = Arc<
     dyn Fn(&mut Context<EXT, DB>, &mut FrameResult) -> Result<(), EVMError<<DB as Database>::Error>>
         + 'a,
 >;
 
+/// Executes a single frame. Errors can be returned in the EVM context.
+/// If `None` is returned, the frame is instead executed normally through the interpreter.
```
DaniPopes, line 21:
```suggestion
```

DaniPopes (commented):
new changes lgtm

yascha:
🔥

PR #1374, opened 2024-05-02
perf(interpreter): branch less in as_usize_or_fail

DaniPopes:
`cmp ..., 0; jne` * 3 -> `mov; or; or; je`
The compiler already does this for the saturated but maybe because it has an early return it didn't do it for this as well

### DIFF
--- a/crates/interpreter/src/instructions/macros.rs
+++ b/crates/interpreter/src/instructions/macros.rs
@@ -300,14 +300,17 @@ macro_rules! push {
 /// Converts a `U256` value to a `u64`, saturating to `MAX` if the value is too large.
 #[macro_export]
 macro_rules! as_u64_saturated {
-    ($v:expr) => {{
-        let x: &[u64; 4] = $v.as_limbs();
-        if x[1] == 0 && x[2] == 0 && x[3] == 0 {
-            x[0]
-        } else {
-            u64::MAX
+    ($v:expr) => {
+        match $v.as_limbs() {
+            x => {
+                if (x[1] == 0) & (x[2] == 0) & (x[3] == 0) {
+                    x[0]
+                } else {
+                    u64::MAX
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/instructions/macros.rs
```diff
@@ -300,14 +300,17 @@ macro_rules! push {
...
 #[macro_export]
 macro_rules! as_u64_saturated {
-    ($v:expr) => {{
-        let x: &[u64; 4] = $v.as_limbs();
-        if x[1] == 0 && x[2] == 0 && x[3] == 0 {
-            x[0]
-        } else {
-            u64::MAX
+    ($v:expr) => {
+        match $v.as_limbs() {
```
onbjerg, line 304: why the match statement? optimizer trick?

onbjerg: ah, that makes a lot of sense

crates/interpreter/src/instructions/macros.rs
```diff
@@ -352,16 +355,15 @@ macro_rules! as_usize_or_fail_ret {
...
 
-    ($interp:expr, $v:expr, $reason:expr, $ret:expr) => {{
-        let x = $v.as_limbs();
-        if x[1] != 0 || x[2] != 0 || x[3] != 0 {
-            $interp.instruction_result = $reason;
-            return $ret;
+    ($interp:expr, $v:expr, $reason:expr, $ret:expr) => {
+        match $v.as_limbs() {
+            x => {
+                if (x[0] > usize::MAX as u64) | (x[1] != 0) | (x[2] != 0) | (x[3] != 0) {
```
rakita, line 361: Okay, this is cool, one less branch

rakita (approved):

PR #1375, opened 2024-05-02
add blob_count and max_blobs to `TooManyBlobs` err enum

yash-atreya:
## Motivation

`TooManyBlobs` enum does not have a blob count being processed in the tx and max blob count.

For example in anvil, ideally we'd like to show the error with these details here  - https://github.com/foundry-rs/foundry/blob/8766dc3e1f7a860bc75b4ebfc675dde3d02581fb/crates/anvil/src/eth/error.rs#L287

## Solution

Change enum to `TooManyBlobs(usize, usize)`.

### DIFF
--- a/crates/primitives/src/result.rs
+++ b/crates/primitives/src/result.rs
@@ -242,7 +242,10 @@ pub enum InvalidTransaction {
     /// `to` must be present
     BlobCreateTransaction,
     /// Transaction has more then [`crate::MAX_BLOB_NUMBER_PER_BLOCK`] blobs
-    TooManyBlobs(usize, usize),
+    TooManyBlobs {
+         max: usize,
+         have: usize,
+     },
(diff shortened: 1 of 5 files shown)


crates/primitives/src/result.rs
```diff
@@ -242,7 +242,7 @@ pub enum InvalidTransaction {
     /// `to` must be present
     BlobCreateTransaction,
     /// Transaction has more then [`crate::MAX_BLOB_NUMBER_PER_BLOCK`] blobs
-    TooManyBlobs,
+    TooManyBlobs(usize, usize),
```
rakita, line 245:
```suggestion
    TooManyBlobs {
         max: usize,
         have: usize,
     },
```

rakita (approved):
lgtm

PR #1376, opened 2024-05-03
feat: add trace option in `revme evm`

### LINKED ISSUE #1373: Add tracer option in `revme evm`
It would be great to have `--trace` cli option in revme evm command.

`revm evm 000000` command executed arbitrary bytecodes.

qiweiii:
resolve #1373

### DIFF
--- a/bins/revme/src/cmd/evmrunner.rs
+++ b/bins/revme/src/cmd/evmrunner.rs
@@ -98,8 +98,11 @@ impl Cmd {
             microbench::bench(&bench_options, "Run bytecode", || {
                 let _ = evm.transact().unwrap();
             });
-        } else if self.trace {
-            println!("Traces:");
+
+            return Ok(());
+        }
+
+        let out = if self.trace {
             let mut evm = evm
                 .modify()
                 .reset_handler_with_external_context(TracerEip3155::new(
@@ -108,16 +111,16 @@ impl Cmd {
                 .append_handler_register(inspector_handle_register)
                 .build();
 
-            let out = evm.transact().map_err(|_| Errors::EVMError)?;
-            println!("Result: {:#?}", out.result);
+            evm.transact().map_err(|_| Errors::EVMError)?
         } else {
-            let out = evm.transact().map_err(|_| Errors::EVMError)?;
-            println!("Result: {:#?}", out.result);
+            evm.transact().map_err(|_| Errors::EVMError)?
+        };
 
-            if self.state {
-                println!("State: {:#?}", out.state);
-            }
+        println!("Result: {:#?}", out.result);
+        if self.state {
+            println!("State: {:#?}", out.state);
         }
+
         Ok(())
     }
 }
--- a/bins/revme/src/cmd/evmrunner.rs
+++ b/bins/revme/src/cmd/evmrunner.rs
@@ -118,7 +118,6 @@ impl Cmd {
             out
         };
 
-        println!("Result: {:#?}", out.result);
         if self.state {
             println!("State: {:#?}", out.state);
         }
--- a/bins/revme/src/cmd/evmrunner.rs
+++ b/bins/revme/src/cmd/evmrunner.rs
@@ -113,7 +113,9 @@ impl Cmd {
 
             evm.transact().map_err(|_| Errors::EVMError)?
         } else {
-            evm.transact().map_err(|_| Errors::EVMError)?
+            let out = evm.transact().map_err(|_| Errors::EVMError)?;
+            println!("Result: {:#?}", out.result);
+            out
         };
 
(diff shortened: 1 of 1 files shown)


bins/revme/src/cmd/evmrunner.rs
```diff
@@ -99,6 +104,25 @@ impl Cmd {
             if self.state {
                 println!("State: {:#?}", out.state);
             }
+            if self.trace {
```
rakita, line 107: this `if self.trace {` should be moved few lines above.

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -99,6 +104,25 @@ impl Cmd {
...
+                println!("Traces:");
+                let mut evm = Evm::builder()
```
rakita, line 109: You don't need to build a new Evm you can modify the existing one with `evm.modify().with_external_context(...).append_...().build()`

rakita (changes requested):
good path, left few comments on refactor.

bins/revme/src/cmd/evmrunner.rs
rakita: Would put 102 line in `else` branch. If it is `--trace` we dont need to execute it twice.

And put line 103 println! after traces.

rakita: @qiweii

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -93,9 +98,22 @@ impl Cmd {
...
+                .append_handler_register(inspector_handle_register)
+                .build();
+
+            let out = evm.transact().map_err(|_| Errors::EVMError)?;
+            println!("Result: {:#?}", out.result);
         } else {
             let out = evm.transact().map_err(|_| Errors::EVMError)?;
             println!("Result: {:#?}", out.result);
+
             if self.state {
```
rakita, line 117: `self.trace` option should be independent of `self.trace`. So it should be put outside of else branch

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -93,9 +98,22 @@ impl Cmd {
             microbench::bench(&bench_options, "Run bytecode", || {
                 let _ = evm.transact().unwrap();
             });
+        } else if self.trace {
+            println!("Traces:");
```
rakita, line 102:
```suggestion
```

There is no need to printout this

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -93,9 +98,22 @@ impl Cmd {
...
+            let mut evm = evm
+                .modify()
+                .reset_handler_with_external_context(TracerEip3155::new(
+                    Box::new(std::io::stdout()),
+                ))
+                .append_handler_register(inspector_handle_register)
+                .build();
+
+            let out = evm.transact().map_err(|_| Errors::EVMError)?;
+            println!("Result: {:#?}", out.result);
```
rakita, line 112:
```suggestion
```
Or even this as the output of result is done inside `TracerEip3155`

bins/revme/src/cmd/evmrunner.rs
qiweiii: do you mean `independent of self.state`

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -93,13 +98,29 @@ impl Cmd {
...
+                .build();
+
+            evm.transact().map_err(|_| Errors::EVMError)?
         } else {
-            let out = evm.transact().map_err(|_| Errors::EVMError)?;
-            println!("Result: {:#?}", out.result);
-            if self.state {
-                println!("State: {:#?}", out.state);
-            }
+            evm.transact().map_err(|_| Errors::EVMError)?
```
rakita, line 116:
```suggestion
            let out = evm.transact().map_err(|_| Errors::EVMError)?;
            println!("Result: {:#?}", out.result);
            out
```

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -93,13 +98,31 @@ impl Cmd {
...
         } else {
             let out = evm.transact().map_err(|_| Errors::EVMError)?;
             println!("Result: {:#?}", out.result);
-            if self.state {
-                println!("State: {:#?}", out.state);
-            }
+            out
+        };
+
+        println!("Result: {:#?}", out.result);
```
rakita, line 121:
```suggestion
```

TracerEip3155 has its own output

rakita (approved):
lgtm.

Will merge after CI

Issue #1251, opened 2024-03-31
Review optimisation for `BYTE`, `SHL`, `SHR` and `SAR`


rakita:
Reference to this: https://github.com/imapp-pl/gas-cost-estimator/blob/master/docs/report_stage_iii.md#analysis-4

allnil:
hmm
want to take a look

rakita:
done in #1418

PR #1418, opened 2024-05-13
chore(interpreter): optimisation for BYTE, SHL, SHR and SAR

jpgonzalezra:
Hey guys, this is my first contribution here, so nice to meet you. I was working on this [issue 1251](https://github.com/bluealloy/revm/issues/1251) 

Motivation:
Investigate the performance of the BYTE, SHL, SHR and SAR OPCODEs and try to improvement 

Solution:
The improvements I've tried to apply in the resulting assembly were:
- reduced branching
- optimized memory access
- fewer instructions

I'm going to attach the before and after assembly for the opcode functions.

### Byte
[Before](https://github.com/bluealloy/revm/files/15299221/byte-before.log)
[After](https://github.com/bluealloy/revm/files/15299220/byte-after.log)

### SAR
[Before](https://github.com/bluealloy/revm/files/15299203/sar-before.log)
[After](https://github.com/bluealloy/revm/files/15299202/sar-after.log)

### SHL
[Before](https://github.com/bluealloy/revm/files/15299205/shl-before.log)
[After](https://github.com/bluealloy/revm/files/15299204/shl-after.log)

### SHR
[Before](https://github.com/bluealloy/revm/files/15299207/shr-before.log)
[After](https://github.com/bluealloy/revm/files/15299206/shr-after.log)

### DIFF
--- a/crates/interpreter/src/instructions/bitwise.rs
+++ b/crates/interpreter/src/instructions/bitwise.rs
@@ -1,11 +1,10 @@
-use super::i256::{i256_cmp, i256_sign_compl, two_compl, Sign};
+use super::i256::i256_cmp;
 use crate::{
     gas,
     primitives::{Spec, U256},
     Host, Interpreter,
 };
 use core::cmp::Ordering;
-use revm_primitives::uint;
 
 pub fn lt<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
     gas!(interpreter, gas::VERYLOW);
@@ -71,13 +70,9 @@ pub fn byte<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
 
-    let o1 = as_usize_saturated!(op1);
-    *op2 = if o1 < 32 {
-        // `31 - o1` because `byte` returns LE, while we want BE
-        U256::from(op2.byte(31 - o1))
-    } else {
-        U256::ZERO
-    };
+    let o1 = as_usize_saturated!(op1) % 32;
+    let byte_value = op2.byte(31 - o1);
+    *op2 = U256::from(byte_value);
 }
 
 /// EIP-145: Bitwise shifting instructions in EVM
@@ -85,7 +80,12 @@ pub fn shl<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &
     check!(interpreter, CONSTANTINOPLE);
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 <<= as_usize_saturated!(op1);
+    let shift = as_usize_saturated!(op1);
+    *op2 = if shift < 256 {
+        *op2 << shift
+    } else {
+        U256::ZERO
+    }
 }
 
 /// EIP-145: Bitwise shifting instructions in EVM
@@ -93,7 +93,12 @@ pub fn shr<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &
     check!(interpreter, CONSTANTINOPLE);
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 >>= as_usize_saturated!(op1);
+    let shift = as_usize_saturated!(op1);
+    *op2 = if shift < 256 {
+        *op2 >> shift
+    } else {
+        U256::ZERO
+    }
 }
 
 /// EIP-145: Bitwise shifting instructions in EVM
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -71,29 +70,35 @@ pub fn byte<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
...
-    let o1 = as_usize_saturated!(op1);
-    *op2 = if o1 < 32 {
-        // `31 - o1` because `byte` returns LE, while we want BE
-        U256::from(op2.byte(31 - o1))
-    } else {
-        U256::ZERO
-    };
+    let o1 = as_usize_saturated!(op1) % 32;
+    let byte_value = op2.byte(31 - o1);
+    *op2 = U256::from(byte_value);
```
DaniPopes, line 75: I don't think this is right, but also not sure if there is a test for this in statetests

DaniPopes (changes requested):

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -71,29 +70,35 @@ pub fn byte<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
...
     pop_top!(interpreter, op1, op2);
 
-    let o1 = as_usize_saturated!(op1);
-    *op2 = if o1 < 32 {
-        // `31 - o1` because `byte` returns LE, while we want BE
-        U256::from(op2.byte(31 - o1))
-    } else {
-        U256::ZERO
-    };
+    let o1 = as_usize_saturated!(op1) % 32;
```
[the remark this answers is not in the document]
jpgonzalezra: My bad, you are right

crates/interpreter/src/instructions/bitwise.rs
jpgonzalezra: This way when 'o1' exceeds 31, the function returns ZERO but to ensure the implementation works as expected, can you suggest any specific tests we should do just in case?

DaniPopes: I don't think so? 33 % 32 = 1 so it will be 31 - 1 byte instead of zero

jpgonzalezra: Yes, you're right. Sometimes we have things right in front of our eyes and they escape us. If you have any ideas on how to improve the method, please let me know.

jpgonzalezra: we can shifts the op2 right by calculated bits to align and isolate the desired byte using bitmask but I dont know if the result is better than byte(..)

something like this:

```rs
*op2 = if o1 < 32 {
      let shift = (31 - o1) * 8;
      U256::from((*op2 >> shift) & U256::from(0xFF))
  } else {
      U256::ZERO
  };
```

What do you think ?

DaniPopes: I believe `byte` is already optimal on since it will simply index into the value as a byte array. Thanks for the efforts anyway.

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -102,33 +107,32 @@ pub fn sar<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &
...
-    // more would mean that no original bits remain. The result depends on what the highest bit of
-    // the value is.
-    *op2 = if value_sign == Sign::Zero || op1 >= U256::from(255) {
-        match value_sign {
-            // value is 0 or >=1, pushing 0
-            Sign::Plus | Sign::Zero => U256::ZERO,
-            // value is <0, pushing -1
-            Sign::Minus => U256::MAX,
-        }
+    let shift = usize::try_from(op1).unwrap_or(256); // If op1 is greater than the maximum for usize, consider the shift complete.
```
DaniPopes, line 110: Please use `as_usize_saturated`, U256::try_into does not have great codegen.

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -102,33 +107,32 @@ pub fn sar<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &
...
+    if shift >= 256 {
+        // If the shift is 256 or more, the result depends on the sign of the last bit.
+        *op2 = if op2.bit(255) {
```
DaniPopes, line 113: Nit: we can move *op2 = to the outer if

DaniPopes (approved):
This looks fine to me, I have not verified all assembly but it looks like they would be improved.

crates/interpreter/src/instructions/bitwise.rs
jpgonzalezra: Thanks, I'm going to leave this method as it was =)

crates/interpreter/src/instructions/bitwise.rs
jpgonzalezra: Cool, let me do it

crates/interpreter/src/instructions/bitwise.rs
```diff
@@ -85,15 +84,25 @@ pub fn shl<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &
...
 }
 
 /// EIP-145: Bitwise shifting instructions in EVM
 pub fn shr<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut H) {
     check!(interpreter, CONSTANTINOPLE);
     gas!(interpreter, gas::VERYLOW);
     pop_top!(interpreter, op1, op2);
-    *op2 >>= as_usize_saturated!(op1);
+    let shift = as_usize_saturated!(op1);
+    *op2 = if shift < 256 {
```
rakita, line 101: I was looking at what `ruint` do 
https://github.com/recmo/uint/blob/b041f09be7035bb61f8e52e39194eb838e832483/src/bits.rs#L353-L356

`shift < 256` is better.

jpgonzalezra: cool. so, are we on the right track?

rakita (approved):
lgtm

PR #1420, opened 2024-05-14
feat(optimism): Implement new L1 cost function for Fjord

BrianBland:
* Add new Optimism Fjord hardfork spec
* Implement the Fjord L1 Cost function according to [this spec](https://github.com/ethereum-optimism/specs/pull/25/files#diff-7b73790400ced4f16890276af12f6d657235dce7613d9fa5ed9a7d65dcddf6dcR20)

Note 1: This adds a length-only FastLZ compression implementation under the `optimism` module. One alternative approach would be to depend on a battle-tested full FastLZ compression implementation, but this would be more computationally expensive and may require the use of C bindings. This could also be a test-only dependency for testing behavioral equivalence.

Note 2: The FastLZ size estimator is called in `L1BlockInfo.data_gas`, which itself is called within `L1BlockInfo.calculate_tx_l1_cost`. Callers may often need to perform this pseudo-compression twice per transaction unless we refactor these methods to reuse this computation.


crates/revm/src/optimism/fast_lz.rs
```diff
@@ -0,0 +1,315 @@
...
+    u32::from(input[idx as usize])
+        + (u32::from(input[(idx + 1) as usize]) << 8)
+        + (u32::from(input[(idx + 2) as usize]) << 16)
+}
+
+/// This is a more faithful adaptation of the solady implementation of FastLZ, returning only the compressed size.
+/// Unfortunately this is even harder to read due to the use of nearly-minified javascript.
+// https://github.com/Vectorized/solady/blob/5315d937d79b335c668896d7533ac603adac5315/js/solady.js
+#[allow(dead_code)]
+fn solady_flz_compress(ib: &[u8]) -> u32 {
```
BrianBland, line 110: Open to removing this implementation now that we have another EVM bytecode implementation below.

crates/revm/src/optimism/fast_lz.rs
```diff
@@ -0,0 +1,314 @@
...
...610313576000805b50508051604051600082901a600183901a60081b1760029290921a60101b91909117639e3779b9810260111c617ffc16909101805160e081811c878603811890911b9091189091528401908183039084841061026857506102a3565b600184019350611fff821161029d578251600081901a600182901a60081b1760029190911a60101b17810361029d57506102a3565b5061020c565b8383106102b1575050610313565b600183039250858311156102cf576102cc87878886036100a9565b96505b6102e3600985016003850160038501610079565b91506102f08782846100f7565b9650506103088461030386848601610186565b610186565b915050809350610200565b5050617fe061032884848589518601036100a9565b03925050506020820180820383525b81811161034e57617fe08101518152602001610337565b5060008152602001604052919050565b634e487b7160e01b600052604160045260246000fd5b60006020828403121561038657600080fd5b813567ffffffffffffffff8082111561039e57600080fd5b818401915084601f8301126103b257600080fd5b8135818111156103c4576103c461035e565b604051601f8201601f19908116603f011681019083821181831017156103ec576103ec61035e565b8160405282815287602084870101111561040557600080fd5b82602086016020830137600092810160200192909252509594505050505056fea264697066735822122000646b2953fc4a6f501bd0456ac52203089443937719e16b3190b7979c39511264736f6c63430008190033"));
```
BrianBland, line 275: cc @danyalprout as this is an adaptation of your parity fuzz test

crates/revm/src/optimism/fast_lz.rs
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    #[test]
+    fn test_flz_native_evm_parity() {
+        // This bytecode and ABI is for a contract, which wraps the LibZip library for easier fuzz testing.
+        // The source of this contract is here: https://github.com/danyalprout/fastlz/blob/main/src/FastLz.sol#L6-L10
+        let abi = BaseContract::from(
+            parse_abi(&["function fastLz(bytes) external view returns (uint256)"]).unwrap(),
+        );
```
gakonst, line 274: This shoudl be using alloy cc @DaniPopes 

```
alloy_sol_types::sol! {
    interface FaztLz {
       function fastLz(bytes) external view returns (uint256);
   }
}
```

And then `fastLzCall::abi_encode`

https://docs.rs/alloy-sol-types/latest/alloy_sol_types/

gakonst (dismissed):

crates/revm/src/optimism/fast_lz.rs
BrianBland: Thanks, I'll apply that refactor to the [examples](https://github.com/bluealloy/revm/blob/main/examples/fork_ref_transact.rs#L37-L45) as well if this is the preferred method for interacting with solidity types

crates/revm/Cargo.toml
```diff
@@ -54,20 +54,22 @@ ethers-core = { version = "2.0", optional = true }
...
 [dev-dependencies]
+alloy-provider = { git = "https://github.com/alloy-rs/alloy.git", rev = "44b8a6d", default-features = false, features = [
+    "reqwest",
+] }
 alloy-sol-types = { version = "0.7.0", default-features = false, features = [
     "std",
 ] }
+alloy-sol-macro = "0.7.0"
+# needed for enabling TLS to use HTTPS connections when testing alloy DB
+alloy-transport-http = { git = "https://github.com/alloy-rs/alloy.git", rev = "44b8a6d" }
```
rakita, line 65: as Alloy is not in crates this would block the publishing of revm to crates. We can't use it.

examples/fork_ref_transact.rs
```diff
@@ -1,5 +1,5 @@
-use ethers_contract::BaseContract;
-use ethers_core::abi::parse_abi;
+use alloy_sol_macro::sol;
+use alloy_sol_types::SolCall;
```
rakita, line 2: Would revert those, it is out of the scope for this PR.

crates/revm/src/optimism/fast_lz.rs
rakita: Should we move this inside the `test` mod?

rakita (changes requested):
left few nits

crates/revm/src/optimism/fast_lz.rs
```diff
@@ -0,0 +1,305 @@
...
+        i += 2;
+        a = i;
+    }
+    literals(b + 4 - a, a, ib, &mut ob);
+    ob.len() as u32
+}
+
+#[cfg(test)]
+mod tests {
+    use ethers_core::rand::RngCore;
```
DaniPopes, line 214: just import rand instead of using ethers

crates/revm/Cargo.toml
mattsse: I believe publishing should still work because these are dev deps which cargo removes before publishing afaik

BrianBland: I'll go ahead and undo this reordering of dependencies. Only the `rstest` line should be new

rakita: my bad, didn't see it was in dev deps

rakita (approved):
lgtm

crates/primitives/src/specification.rs
```diff
@@ -64,6 +64,7 @@ pub enum SpecId {
     CANCUN = 20,
     ECOTONE = 21,
     PRAGUE = 22,
+    FJORD = 23,
```
rakita, line 67: Should FJORD be after or before PRAGUE? In secp256r1 precompile PR it was before.

rakita: Put FJORD before PRAGUE as Prague is still not activated

PR #1426, opened 2024-05-16
Revert "Revert "feat: implement EIP-2935 (#1354)" (#1424)"

rakita:
This reverts commit f07c0ef1c27d7a7a6b160849f4f2c88dcd843cfe.


gakonst (commented):
ACK

Blockhash devnet-0 post [Pari's PR](github.com/ethereum/EIPs/pull/8564):
* 20 gas always
* has access to to 8192 blocks behind, which is unsound but it's fine for devnet-0
* does the backfill method

PR #1428, opened 2024-05-16
fix(precompile): BLS G2 MSM

shekhirin:
* Adds back the BLS Test Vectors [from AlphaNet](https://github.com/paradigmxyz/alphanet/tree/main/crates/precompile/test-vectors)
* Fixes bug introduced by the [newer](https://github.com/ethereum/EIPs/commit/57638c1f28cf3df3378886a534a19a566dff1a79) test vectors from the [EIP](https://eips.ethereum.org/assets/eip-2537/test-vectors) which involved a BLST footgun in the API that took us multiple hours to identify. This behavior MUST be clearly documented as it's dangerous for broader usage of BLST.

This was a process failure from our side:
1. We shipped 2537 in AlphaNet ~1 mo ago, and thought the EIP was "done"
2. We did not look at the updated test vectors, this would've surfaced this 2-3 weeks ago

Makes me think we may want to subscribe to specific EIP updates in an internal channel to ensure we don't miss anything.


gakonst:
Found & fixed the bug and updated PR description. Feel free to improve on my code (esp the empty slice check is ugly but it's 330am). This should be functional for Devnet-0. There might be better ways to work around this massive BLST footgun but don't have the time for this rn.

crates/precompile/src/bls12_381/g2_msm.rs
```diff
@@ -59,13 +65,26 @@ fn g2_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
...
+    };
     Ok((required_gas, out))
 }
+
+fn is_zero(buf: &[u8]) -> bool {
+    let (prefix, aligned, suffix) = unsafe { buf.align_to::<u128>() };
+
+    prefix.iter().all(|&x| x == 0)
+        && suffix.iter().all(|&x| x == 0)
+        && aligned.iter().all(|&x| x == 0)
```
DaniPopes, line 89: can't this just be one iter.all == 0 ?

crates/precompile/src/bls12_381/mod.rs
```diff
@@ -29,3 +29,105 @@ pub fn precompiles() -> impl Iterator<Item = PrecompileWithAddress> {
...
+                let expected_output = Bytes::from_hex(vector.expected).unwrap();
+                assert_eq!(
+                    expected_output, actual_output,
+                    "expected output: {expected_output}, actual output: {actual_output} in {test_name}");
+            }
+        }
+    }
+    //
+    // #[rstest]
+    // #[case::g1_empty(0, G1MUL_BASE, 0)]
```
DaniPopes, line 114: rm

shekhirin: uncommented and fixed

crates/precompile/src/bls12_381/g2_msm.rs
rakita: I have made this change, we can do it. Checking the infinity point is just checking if the slice is zero.

rakita (approved):
Pending some discussion, PR lgtm

crates/precompile/src/bls12_381/g1_add.rs
```diff
@@ -33,8 +33,11 @@ pub(super) fn g1_add(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         )));
     }
 
-    let a_aff = &extract_g1_input(&input[..G1_INPUT_ITEM_LENGTH])?;
-    let b_aff = &extract_g1_input(&input[G1_INPUT_ITEM_LENGTH..])?;
+    // NB: There is no subgroup check for the G2 addition precompile.
```
rakita:
```suggestion
    // NB: There is no subgroup check for the G1 addition precompile.
```

rakita (approved):
lgtm

PR #1430, opened 2024-05-17
chore: cleanup host blockhash fn

rakita:


crates/revm/src/context.rs
```diff
@@ -110,39 +110,37 @@ impl<EXT, DB: Database> Host for Context<EXT, DB> {
...
-                        .ok()
-                        .map(|v| v.into());
-                }
-            }
-            _ => {
-                // If blockhash is requested for the current block, the hash should be 0, so we fall
-                // through.
-            }
+        let block_number = as_usize_saturated!(self.env().block.number);
+        let requested_number = as_usize_saturated!(number);
```
DaniPopes, line 114: technically this breaks blockhash when block number is > u64::MAX because diff will always saturate to 0.

PR #1432, opened 2024-05-17
fix(precompile): ignore infinity points in G1 MSM

Rjected:
This applies the same fix we use for the G2 MSM infinity points, for the G1 MSM operation.


rakita (approved):
lgtm

PR #1436, opened 2024-05-17
feat(optimism): Add secp256r1 precompile for Fjord

BrianBland:
* Upstream the RIP-7212: `secp256r1::P256VERIFY` precompile from the [alphanet implementation](https://github.com/paradigmxyz/alphanet/blob/a7f66d66cc59033fe8b5364cbc3a98825384d549/crates/precompile/src/secp256r1.rs)
* Add new Optimism Fjord hardfork spec before Prague
* Add the `P256Verify` to the chain precompiles post-fjord via the Optimism handle register

Note: Introduces new optional dependency on the `p256` crate

### DIFF
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -45,7 +45,7 @@ secp256k1 = { version = "0.29.0", default-features = false, features = [
 blst = { version = "0.3.11", optional = true }
 
 # p256verify precompile
-p256 = { version = "0.13.2", features = ["ecdsa"] }
+p256 = { version = "0.13.2", default-features = false, features = ["ecdsa"] }
 
 [dev-dependencies]
 criterion = { version = "0.5" }
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -44,6 +44,9 @@ secp256k1 = { version = "0.29.0", default-features = false, features = [
 # BLS12-381 precompiles
 blst = { version = "0.3.11", optional = true }
 
+# p256verify precompile
+p256 = { version = "0.13.2", features = ["ecdsa"] }
+
 [dev-dependencies]
 criterion = { version = "0.5" }
 rand = { version = "0.8", features = ["std"] }
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -80,6 +80,7 @@ negate-optimism-default-handler = [
     "revm-primitives/negate-optimism-default-handler",
 ]
 
+# Enables the p256verify precompile.
 secp256r1 = ["dep:p256"]
 
 # These libraries may not work on all no_std platforms as they depend on C.
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -45,7 +45,7 @@ secp256k1 = { version = "0.29.0", default-features = false, features = [
 blst = { version = "0.3.11", optional = true }
 
 # p256verify precompile
-p256 = { version = "0.13.2", default-features = false, features = ["ecdsa"] }
+p256 = { version = "0.13.2", optional = true,  default-features = false, features = ["ecdsa"] }
 
 [dev-dependencies]
 criterion = { version = "0.5" }
@@ -70,7 +70,7 @@ std = [
 hashbrown = ["revm-primitives/hashbrown"]
 asm-keccak = ["revm-primitives/asm-keccak"]
 
-optimism = ["revm-primitives/optimism"]
+optimism = ["revm-primitives/optimism", "secp256r1"]
 # Optimism default handler enabled Optimism handler register by default in EvmBuilder.
 optimism-default-handler = [
     "optimism",
@@ -79,6 +79,8 @@ optimism-default-handler = [
 negate-optimism-default-handler = [
     "revm-primitives/negate-optimism-default-handler",
 ]
+
+secp256r1 = ["dep:p256"]
 
 # These libraries may not work on all no_std platforms as they depend on C.
 
--- a/crates/precompile/src/lib.rs
+++ b/crates/precompile/src/lib.rs
@@ -18,6 +18,7 @@ pub mod identity;
 pub mod kzg_point_evaluation;
 pub mod modexp;
 pub mod secp256k1;
+#[cfg(feature = "secp256r1")]
 pub mod secp256r1;
 pub mod utilities;
 
@@ -66,8 +67,6 @@ impl Precompiles {
             PrecompileSpecId::ISTANBUL => Self::istanbul(),
             PrecompileSpecId::BERLIN => Self::berlin(),
             PrecompileSpecId::CANCUN => Self::cancun(),
-            #[cfg(feature = "optimism")]
-            PrecompileSpecId::FJORD => Self::fjord(),
             PrecompileSpecId::PRAGUE => Self::prague(),
             PrecompileSpecId::LATEST => Self::latest(),
         }
@@ -161,6 +160,7 @@ impl Precompiles {
     }
 
     /// Returns precompiles for Fjord spec.
+    #[cfg(feature = "optimism")]
     pub fn fjord() -> &'static Self {
         static INSTANCE: OnceBox<Precompiles> = OnceBox::new();
         INSTANCE.get_or_init(|| {
@@ -266,8 +266,6 @@ pub enum PrecompileSpecId {
     ISTANBUL,
     BERLIN,
     CANCUN,
-    #[cfg(feature = "optimism")]
-    FJORD,
     PRAGUE,
     LATEST,
 }
@@ -289,9 +287,7 @@ impl PrecompileSpecId {
             #[cfg(feature = "optimism")]
             BEDROCK | REGOLITH | CANYON => Self::BERLIN,
             #[cfg(feature = "optimism")]
-            ECOTONE => Self::CANCUN,
-            #[cfg(feature = "optimism")]
-            FJORD => Self::FJORD,
+            ECOTONE | FJORD => Self::CANCUN,
         }
     }
 }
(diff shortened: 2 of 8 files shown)


rakita (changes requested):
Left few comments

crates/precompile/src/lib.rs
```diff
@@ -157,6 +160,19 @@ impl Precompiles {
         })
     }
 
+    /// Returns precompiles for Fjord spec.
+    pub fn fjord() -> &'static Self {
```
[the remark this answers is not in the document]
BrianBland: Removed this method in favor of using the Optimism handle register to load additional fjord precompiles

crates/precompile/Cargo.toml
```diff
@@ -44,6 +44,9 @@ secp256k1 = { version = "0.29.0", default-features = false, features = [
 # BLS12-381 precompiles
 blst = { version = "0.3.11", optional = true }
 
+# p256verify precompile
+p256 = { version = "0.13.2", default-features = false, features = ["ecdsa"] }
```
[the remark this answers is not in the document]
BrianBland: Gated behind a new `secp256r1` feature, which is added as a dependency for `optimism`

rakita (approved):
lgtm!

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,117 @@
...
+        return Err(PrecompileError::OutOfGas);
+    }
+    let result = verify_impl(input).is_some();
+    Ok((P256VERIFY_BASE, B256::with_last_byte(result as u8).into()))
+}
+
+/// Returns `Some(())` if the signature included in the input byte slice is
+/// valid, `None` otherwise.
+pub fn verify_impl(input: &[u8]) -> Option<()> {
+    if input.len() < 160 {
```
rakita, line 45: It is not concretely specified in EIP, but is mentioned as `Input data: 160 bytes of data including:` so should this be a `!=`?
```suggestion
    if input.len() != 160 {
```
This can be a consensus error if not set correctly, what do other implementations do?

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,117 @@
...
+const P256VERIFY_BASE: u64 = 3_450;
+
+/// Returns the secp256r1 precompile with its address.
+pub fn precompiles() -> impl Iterator<Item = PrecompileWithAddress> {
+    [P256VERIFY].into_iter()
+}
+
+/// [EIP-7212](https://eips.ethereum.org/EIPS/eip-7212#specification) secp256r1 precompile.
+pub const P256VERIFY: PrecompileWithAddress =
+    PrecompileWithAddress(u64_to_address(0x14), Precompile::Standard(p256_verify));
```
rakita, line 23: `elliptic curve at address PRECOMPILED_ADDRESS in 0x100 (indicates 0x0000000000000000000000000000000000000100).`
precompile is specified at the address, did Optimism changed this?
https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,117 @@
...
+//!
+//! The main purpose of this precompile is to verify ECDSA signatures that use the secp256r1, or
+//! P256 elliptic curve. The [`P256VERIFY`] const represents the implementation of this precompile,
+//! with the address that it is currently deployed at.
+use crate::{u64_to_address, Precompile, PrecompileWithAddress};
+use p256::ecdsa::{signature::hazmat::PrehashVerifier, Signature, VerifyingKey};
+use revm_primitives::{Bytes, PrecompileError, PrecompileResult, B256};
+
+/// Base gas fee for secp256r1 p256verify operation.
+const P256VERIFY_BASE: u64 = 3_450;
```
rakita, line 14:
```suggestion
const P256VERIFY_BASE: u64 = 3450;
```
It is fine without `_`, nowhere in the code underscore  is used and it is easier to search without it

crates/precompile/src/secp256r1.rs
BrianBland: Good catch. I think I may have been looking at an older version of the RIP/EIP. [This EIP source](https://eips.ethereum.org/EIPS/eip-7212) still references `0x0b`

crates/precompile/src/secp256r1.rs
BrianBland: op-geth appears to perform a != check ([source](https://github.com/ethereum-optimism/op-geth/blob/966c43537e49f7936bb57a426079fb0da9baf03b/core/vm/contracts.go#L1175))
Updated

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,117 @@
...
+    let sig: &[u8; 64] = input[32..96].try_into().unwrap();
+    // x, y: public key
+    let pk: &[u8; 64] = input[96..160].try_into().unwrap();
+
+    // append 0x04 to the public key: uncompressed form
+    let mut uncompressed_pk = [0u8; 65];
+    uncompressed_pk[0] = 0x04;
+    uncompressed_pk[1..].copy_from_slice(pk);
+
+    // Can fail only if the input is not exact length.
```
rakita, line 61: This is not true. Signature fields needs to be inside curve group.
This test case will panic:
```
        let input = [0xff; 64];
        let signature = Signature::from_slice(sig).unwrap();
```

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,117 @@
...
+    if input.len() < 160 {
+        return None;
+    }
+
+    // msg signed (msg is already the hash of the original message)
+    let msg: &[u8; 32] = input[..32].try_into().unwrap();
+    // r, s: signature
+    let sig: &[u8; 64] = input[32..96].try_into().unwrap();
+    // x, y: public key
+    let pk: &[u8; 64] = input[96..160].try_into().unwrap();
```
rakita, line 54: We can probably simplify this as the slices are mostly used as input inside p256 lib.
```suggestion
    let msg = &input[..32];
    // r, s: signature
    let sig = &input[32..96];
    // x, y: public key
    let pk = &input[96..160];
```

crates/precompile/src/secp256r1.rs
BrianBland: Thanks, fixed and added another test for this case

crates/precompile/src/secp256r1.rs
rakita: Yaeh that was a EIP while RIP should be/is a newer one

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    // append 0x04 to the public key: uncompressed form
+    let mut uncompressed_pk = [0u8; 65];
+    uncompressed_pk[0] = 0x04;
+    uncompressed_pk[1..].copy_from_slice(pk);
+
+    // Can fail only if the input is not exact length.
+    let signature = Signature::from_slice(sig).unwrap();
+    // Can fail if the input is not valid, so we have to propagate the error.
+    let public_key = VerifyingKey::from_sec1_bytes(&uncompressed_pk).ok()?;
```
rakita, line 67: Have checked this part of the spec:
`Note that many implementations use (0, 0) as the reference point at infinity, which is not on the curve and should therefore be rejected.`

I have feed the `VerifyingKey::from_sec1_bytes` with `pk = [0;64]` and it fails to get key, this is expected behaviour, just wanting to note it.

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,122 @@
...
+    // append 0x04 to the public key: uncompressed form
+    let mut uncompressed_pk = [0u8; 65];
+    uncompressed_pk[0] = 0x04;
+    uncompressed_pk[1..].copy_from_slice(pk);
+
+    // Can fail only if the input is not exact length.
+    let signature = match Signature::from_slice(sig) {
+        Ok(sig) => sig,
+        Err(_) => return None,
+    };
```
rakita, line 65:
```suggestion
    let signature = match Signature::from_slice(sig).ok()?;
```

rakita: I think this is last thing

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,119 @@
...
+    // x, y: public key
+    let pk = &input[96..160];
+
+    // append 0x04 to the public key: uncompressed form
+    let mut uncompressed_pk = [0u8; 65];
+    uncompressed_pk[0] = 0x04;
+    uncompressed_pk[1..].copy_from_slice(pk);
+
+    // Can fail only if the input is not exact length.
+    let signature = match Signature::from_slice(sig).ok()?;
```
rakita, line 62:
```suggestion
    let signature = Signature::from_slice(sig).ok()?;
```
and now to remove `match`

crates/precompile/src/secp256r1.rs
```diff
@@ -0,0 +1,119 @@
...
+///
+/// | signed message hash |  r  |  s  | public key x | public key y |
+/// | :-----------------: | :-: | :-: | :----------: | :----------: |
+/// |          32         | 32  | 32  |     32       |      32      |
+pub fn p256_verify(input: &Bytes, gas_limit: u64) -> PrecompileResult {
+    if P256VERIFY_BASE > gas_limit {
+        return Err(PrecompileError::OutOfGas);
+    }
+    let result = verify_impl(input).is_some();
+    Ok((P256VERIFY_BASE, B256::with_last_byte(result as u8).into()))
```
rakita, line 39: If the signature verification process succeeds, it returns 1 in 32 bytes format.
If the signature verification process fails, **it does not return any output data**.

So if it fails we should return `vec![]` and we are now returning `0x00000...`

crates/precompile/src/secp256r1.rs
BrianBland: Thanks, added a test case for this

PR #1445, opened 2024-05-20
fix: check for canonical Fp elements in MSM

Rjected:
This checks that the scalar values being used in precompiles are canonical, as in, they are valid Fr elements. A similar check is performed in `c-kzg-4844` here: https://github.com/ethereum/c-kzg-4844/blob/5b35b6147bf269b16c1bffd268fd03bac0c00b64/src/c_kzg_4844.c#L544

This could have been particularly bad, as geth (and other implementations) would have rejected certain scalar values that we would not have.


Rjected:
This is not a bug, geth does not actually check that scalars used are canonical Fr elements because it uses `SetBytes`:
https://github.com/Consensys/gnark-crypto/blob/564b6f724c3beac52d805e6e600d0a1fda9770b5/ecc/bls12-381/fr/element.go#L910-L934

PR #1446, opened 2024-05-20
chore: add docs for BLS scalar input decoding

Rjected:
After closing https://github.com/bluealloy/revm/pull/1445, I decided to add docs to make sure the fact that we do not perform any Fr canonical checks is well-reasoned. This is actually specified in the EIP, so I decided to link it and annotate the code with the relevant part of the spec.


onbjerg (approved):
great docs as always

rakita (approved):
lgtm

PR #1447, opened 2024-05-21
feat: load account should return db error

yjhmelody:
Since the db may be a remote db, it could happen a network error under the db error. 
And when remote db error happen, the application level maybe need to retry it. So it's better to keep db error original rather than be a custom string.

Or just return another db error with extra error info?

### DIFF
--- a/crates/revm/src/optimism/handler_register.rs
+++ b/crates/revm/src/optimism/handler_register.rs
@@ -240,33 +240,20 @@ pub fn reward_beneficiary<SPEC: Spec, EXT, DB: Database>(
         let l1_cost = l1_block_info.calculate_tx_l1_cost(enveloped_tx, SPEC::SPEC_ID);
 
         // Send the L1 cost of the transaction to the L1 Fee Vault.
-        let l1_fee_vault_account = match context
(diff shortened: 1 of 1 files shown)


crates/revm/src/optimism/handler_register.rs
```diff
@@ -240,29 +240,32 @@ pub fn reward_beneficiary<SPEC: Spec, EXT, DB: Database>(
...
             .journaled_state
             .load_account(optimism::L1_FEE_RECIPIENT, &mut context.evm.inner.db)
-        else {
-            return Err(EVMError::Custom(
-                "[OPTIMISM] Failed to load L1 Fee Vault account.".to_string(),
-            ));
+        {
+            Ok((l1_fee_vault_account, _)) => l1_fee_vault_account,
+            Err(err) => {
+                return Err(err);
```
DaniPopes, line 251: this looks like it can be just `?`

yjhmelody: fine

DaniPopes (approved):

rakita (approved):
seems resonable.
cc @refcell

rakita:
failed tests not related

PR #1450, opened 2024-05-24
feat(revm): revert EIP-2935 BLOCKHASH opcode changes

shekhirin:
As per https://github.com/ethereum/EIPs/pull/8577

Change explanation: https://github.com/bluealloy/revm/pull/1450#discussion_r1613301693

### DIFF
--- a/crates/revm/src/context.rs
+++ b/crates/revm/src/context.rs
@@ -8,15 +8,11 @@ pub use context_precompiles::{
 };
 pub use evm_context::EvmContext;
 pub use inner_evm_context::InnerEvmContext;
-use revm_interpreter::as_usize_saturated;
 
 use crate::{
     db::{Database, EmptyDB},
     interpreter::{Host, LoadAccountResult, SStoreResult, SelfDestructResult},
-    primitives::{
-        Address, Bytecode, EVMError, Env, HandlerCfg, Log, B256, BLOCKHASH_SERVE_WINDOW,
-        BLOCKHASH_STORAGE_ADDRESS, BLOCK_HASH_HISTORY, PRAGUE, U256,
-    },
+    primitives::{Address, Bytecode, Env, HandlerCfg, Log, B256, U256},
 };
 use std::boxed::Box;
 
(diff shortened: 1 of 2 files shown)


crates/revm/src/context.rs
```diff
@@ -110,38 +106,10 @@ impl<EXT, DB: Database> Host for Context<EXT, DB> {
...
-
-        if diff <= BLOCK_HASH_HISTORY {
-            return self
-                .evm
-                .block_hash(number)
-                .map_err(|e| self.evm.error = Err(e))
-                .ok();
-        }
-
-        if self.evm.journaled_state.spec.is_enabled_in(PRAGUE) && diff <= BLOCKHASH_SERVE_WINDOW {
```
DaniPopes, line 133: only this PRAGUE branch should be removed, the rest is still valid (diff == 0, diff <= HISTORY...)

shekhirin: are you sure? from the EIP:
> The `BLOCKHASH` opcode semantics remains the same as before.


https://github.com/bluealloy/revm/blob/eed27d92266776d653a1d08c46ffa9699d64d527/crates/revm/src/context.rs#L108-L113

DaniPopes: Yes, this is the before:
https://github.com/bluealloy/revm/blob/688c36ca4525cc44aa7c547b7b0f22a9490c4f2f/crates/interpreter/src/instructions/host.rs#L106-L123

shekhirin: oh, my bad, fixing

rakita (commented):
Logic can stay inside EvmContext, previously it was inside instruction.

This is old logic before EIP-2935: https://github.com/bluealloy/revm/blob/1640b8f6989d0d4a19dfeb343e7967d501f6e195/crates/interpreter/src/instructions/host.rs#L127-L139

rakita (approved):
lgtm, clippy pending

DaniPopes (approved):

PR #1454, opened 2024-05-24
chore: Update fork_ref_transact to use alloy instead of ethers

BrianBland:
Updates the `fork_ref_transact` example to use `alloy` instead of `ethers` for interfacing with ABIs


rakita (approved):
a lot better

PR #1459, opened 2024-05-27
chore(primitives): rename State/Storage to EvmState/EvmStorage

rakita:
Rename type so it does not collide with revm::db::State


DaniPopes (approved):

rakita:
With new EvmStorage type we now have clear separation from Storage used inside Evm and Storage used inside db::State.

PR #1460, opened 2024-05-27
non-allocating U256 mul_mod (ported from ruint)

impactdni2:
U256::mul_mod is one of the few things in revm that allocates, potentially in tight loops (UniswapV3 contracts).

This pulls in a non-allocating mul_mod from ruint (https://github.com/recmo/uint/commit/9f2a4beecaca69c8e43e3a0175671b4ed4e3594c) and uses it in place of the allocating version.

Since the original version was a self method on U256, this has been modified to be lhs/rhs (non-self), hardcoded for U256 rather than generic, and uses an unsafe block to access limbs (since its private).

Please let me know if you'd like some cleanup, or if this isn't something you'd like to add.

### DIFF
--- a/crates/interpreter/src/instructions/arithmetic.rs
+++ b/crates/interpreter/src/instructions/arithmetic.rs
@@ -99,8 +99,6 @@ pub fn signextend<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H
     }
 }
 
-
-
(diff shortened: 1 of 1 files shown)


DaniPopes:
Interesting, nice catch! I would prefer if we instead merged that PR upstream. Would you mind picking it up and adding some tests?

impactdni2:
Love it.  That's a much better solution you've got there (a bit of a hack for sure, but looks safe and self-explanatory).

DaniPopes:
This has been released in ruint 1.12.3, thanks for bringing this up!

PR #1469, opened 2024-05-31
feat(EOF): Add EOF to inspector handle register

rakita:
Implements TODO and adds integrates EOF handles in inspector


crates/revm/src/inspector/handler_register.rs
```diff
@@ -162,6 +162,43 @@ pub fn inspector_handle_register<DB: Database, EXT: GetInspector<DB>>(
     );
 
     // TODO(EOF) EOF create call.
```
fgimenez, line 164:
```suggestion
```

fgimenez (approved):

DaniPopes (approved):

PR #1470, opened 2024-05-31
fix(eof): add FrameResult if eof is in create tx

rakita:
Should be merged here: https://github.com/bluealloy/revm/pull/1467

EOFCreate call now can be done from create transaction.

Add ReturnContract as success return.


fgimenez (approved):

rakita:
Merged changes to https://github.com/bluealloy/revm/pull/1467

Issue #1473, opened 2024-05-31
EOF: change oob behavior of RETURNDATALOAD and RETURNDATACOPY


rakita:
Change is here: https://github.com/ipsilon/eof/pull/90

TODO put in this PR: https://github.com/bluealloy/revm/pull/1472

jpgonzalezra:
@rakita Can I take it ?

rakita:
Go for it

rakita:
Closed in #1476

PR #1476, opened 2024-05-31
feat(EOF): change oob behavior of RETURNDATALOAD and RETURNDATACOPY

jpgonzalezra:
[ISSUE](https://github.com/bluealloy/revm/issues/1473):

Change is here: https://github.com/ipsilon/eof/pull/90

RETURNDATACOPY:
    - same behavior as legacy, but changes the exceptional halt behavior to zero-padding behavior
 
RETURNDATALOAD:
    - if `offset + 32 > len(returndata buffer)` the result is zero-padded (same behavior as `CALLDATALOAD`). see matching behavior of `RETURNDATACOPY` in `Modified Behavior` section.


crates/interpreter/src/instructions/system.rs
```diff
@@ -123,23 +123,42 @@ pub fn returndatasize<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interprete
 
 /// EIP-211: New opcodes: RETURNDATASIZE and RETURNDATACOPY
 pub fn returndatacopy<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, _host: &mut H) {
-    check!(interpreter, BYZANTIUM);
+    check!(interpreter, PRAGUE);
```
jpgonzalezra, line 126: I'm not sure about this @rakita

rakita:
Old behaviour of RETURNDATACOPY needs to be preserved if it is not EOF bytecode

jpgonzalezra:
Thanks, @rakita! I'll start running the ethtests locally so I can detect these cases more easily.

jpgonzalezra:
@rakita I thought it would be easier to make it backwards compatible haha. Let me know what you think

rakita:
Hey, will take this over so it can be merged

crates/interpreter/src/instructions/system.rs
```diff
@@ -125,21 +125,46 @@ pub fn returndatasize<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interprete
...
-            &interpreter.return_data_buffer[data_offset..data_end],
+            &interpreter.return_data_buffer[data_offset..data_offset + copy_len],
         );
+
+        if interpreter.is_eof && copy_len < len {
+            interpreter
+                .shared_memory
+                .slice_mut(memory_offset + copy_len, len - copy_len)
+                .fill(0);
+        }
```
rakita, line 160: `else if {error}` is missing. Will restructure code a little bit here.

jpgonzalezra: You're right, go ahead 💪

rakita (approved):
lgtm, waiting for CI.

@jpgonzalezra can you take a look. in `returndatacopy` I used `shared_memory.set_data` and in `returndataload` i removed few lines.

jpgonzalezra:
@rakita awesome, the code is very clear now, thank you 💪

PR #1477, opened 2024-06-01
fix(EOF): panic on empty input range, and continue exec after eofcreate

rakita:
InstructionResult needs to be reset when eofcreate continues execution.

And Handle panic on empty range


rakita:
Note I assumed that there is wrong order of stack pop's for EXTCALL, but spec says that original impl was correct, and order is not same as CALL opcode. Order is: `EXTCALL (0xf8) with arguments (target_address, input_offset, input_size, value)`

PR #1478, opened 2024-06-01
chore(interpreter): use max gas limit in `impl Default for Interpreter`

DaniPopes:
Otherwise it's unusable because gas limit cannot be set after initialization.


rakita (approved):
lgtm

Issue #1464, opened 2024-05-28
`Bytecode::new_raw` should check first byte for EOF


rakita:
If the first byte is EOF prefix, the returned bytecode should be in EOF format.

rakita:
Done in https://github.com/bluealloy/revm/pull/1607

PR #1479, opened 2024-06-02
fix(primitives) : Check first byte for EOF

DoTheBestToGetTheBest:
Closes https://github.com/bluealloy/revm/issues/1464

### DIFF
--- a/crates/primitives/src/bytecode.rs
+++ b/crates/primitives/src/bytecode.rs
@@ -4,14 +4,7 @@ pub mod legacy;
 pub use eof::{Eof, EofDecodeError};
 pub use legacy::{JumpTable, LegacyAnalyzedBytecode};
 
-use crate::{keccak256, Bytes, B256, KECCAK_EMPTY};
-
-/// EOF prefix as per EIP-3540.
-const EOF_PREFIX: u8 = 0xEF;
-
-/// Version byte that immediately follows the EOF prefix in EIP-3540.
-// See https://github.com/bluealloy/revm/issues/1464
-const EOF_VERSION: u8 = 0x01;
+use crate::{keccak256, Bytes, B256, EOF_PREFIX, EOF_VERSION, KECCAK_EMPTY};
 
 /// State of the [`Bytecode`] analysis.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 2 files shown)


crates/primitives/src/bytecode.rs
```diff
@@ -66,10 +73,23 @@ impl Bytecode {
         matches!(self, Self::Eof(_))
     }
 
+    /// Checks if the bytecode starts with the EOF prefix followed by the correct version.
+    #[inline]
+    pub fn is_eof_format(bytes: &Bytes) -> bool {
+        bytes.len() >= 2 && bytes[0] == EOF_PREFIX && bytes[1] == EOF_VERSION
```
rakita, line 79: This is not correct. `EOF_MAGIC` is 0xEF00 while `EOF_VERSION` (At least version one) is 0x01. Would be good to move those in revm-primitives crate

prefix is `0xEF0001`

crates/primitives/src/bytecode.rs
```diff
@@ -66,10 +73,23 @@ impl Bytecode {
...
+    }
+
     /// Creates a new raw [`Bytecode`].
     #[inline]
     pub fn new_raw(bytecode: Bytes) -> Self {
```
rakita, line 84: Would return Error here as Legacy bytecode with 0xEF is considered invalid. And only valid EOF can have `0xEF` prefix.

Would additionally add `new_legacy` function for legacy codes.

crates/primitives/src/bytecode.rs
```diff
@@ -66,10 +66,35 @@ impl Bytecode {
...
+
+    /// Checks if the bytecode starts with the EOF prefix followed by the correct version.
+    #[inline]
+    pub fn is_eof_format(bytes: &Bytes) -> bool {
+        if bytes.len() >= 3 {
+            let prefix = u16::from_be_bytes([bytes[0], bytes[1]]);
+            let version = bytes[2];
+
+            prefix == EOF_PREFIX && version == EOF_VERSION
+        } else {
```
rakita, line 82: This can be missleading in tx create case. For EOF only first two bytes should be checked.

You can do this as:
`.get(0..2).filter(|&t| t == &EOF_MAGIC_BYTES).is_some()`

crates/primitives/src/constants.rs
```diff
@@ -60,3 +60,10 @@ pub const BLOB_GASPRICE_UPDATE_FRACTION: u64 = 3338477;
 
 /// First version of the blob.
 pub const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
+
+/// EOF prefix as per EIP-3540.
+pub const EOF_PREFIX: u16 = 0xEF00;
+
+/// Version byte that immediately follows the EOF prefix in EIP-3540.
+// See https://github.com/bluealloy/revm/issues/1464
+pub const EOF_VERSION: u8 = 0x01;
```
rakita, line 69: `EOF_MAGIC_BYTES` const were added to src/bytecode/eof.rs

crates/primitives/src/bytecode.rs
rakita: Would return error here

rakita:
Superseded by https://github.com/bluealloy/revm/pull/1607

PR #1486, opened 2024-06-04
perf: avoid cloning precompiles

DaniPopes:
`Precompiles::new().clone().into()` would first clone the static default precompiles, then collect them into a new hashmap again in `ContextPrecompiles`. This shows up as about ~3% in `Evm::transact` profiles.

Avoid both cloning and collecting by using a CoW structure internally. This requires us to remove the `Deref` and `DerefMut` implementations as the internal type will be different depending on whether we have the default static precompiles, or a custom map. This should be fine as we can still support all necessary operations without exposing the underlying map type.

### DIFF
--- a/crates/primitives/src/precompile.rs
+++ b/crates/primitives/src/precompile.rs
@@ -110,11 +110,25 @@ impl Precompile {
 
     /// Call the precompile with the given input and gas limit and return the result.
     pub fn call(&mut self, bytes: &Bytes, gas_price: u64, env: &Env) -> PrecompileResult {
-        match self {
+        match *self {
             Precompile::Standard(p) => p(bytes, gas_price),
             Precompile::Env(p) => p(bytes, gas_price, env),
-            Precompile::Stateful(p) => p.call(bytes, gas_price, env),
-            Precompile::StatefulMut(p) => p.call_mut(bytes, gas_price, env),
+            Precompile::Stateful(ref p) => p.call(bytes, gas_price, env),
+            Precompile::StatefulMut(ref mut p) => p.call_mut(bytes, gas_price, env),
+        }
+    }
+
+    /// Call the precompile with the given input and gas limit and return the result.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the precompile is a mutable stateful precompile.
+    pub fn call_ref(&self, bytes: &Bytes, gas_price: u64, env: &Env) -> PrecompileResult {
+        match *self {
+            Precompile::Standard(p) => p(bytes, gas_price),
+            Precompile::Env(p) => p(bytes, gas_price, env),
+            Precompile::Stateful(ref p) => p.call(bytes, gas_price, env),
+            Precompile::StatefulMut(_) => panic!("call_ref on mutable stateful precompile"),
         }
     }
 }
--- a/crates/primitives/src/precompile.rs
+++ b/crates/primitives/src/precompile.rs
@@ -120,9 +120,7 @@ impl Precompile {
 
     /// Call the precompile with the given input and gas limit and return the result.
     ///
-    /// # Panics
-    ///
-    /// Panics if the precompile is a mutable stateful precompile.
+    /// Returns an error if the precompile is mutable.
     pub fn call_ref(&self, bytes: &Bytes, gas_price: u64, env: &Env) -> PrecompileResult {
         match *self {
             Precompile::Standard(p) => p(bytes, gas_price),
--- a/crates/primitives/src/precompile.rs
+++ b/crates/primitives/src/precompile.rs
@@ -128,7 +128,9 @@ impl Precompile {
             Precompile::Standard(p) => p(bytes, gas_price),
             Precompile::Env(p) => p(bytes, gas_price, env),
             Precompile::Stateful(ref p) => p.call(bytes, gas_price, env),
-            Precompile::StatefulMut(_) => panic!("call_ref on mutable stateful precompile"),
+            Precompile::StatefulMut(_) => Err(PrecompileErrors::Fatal {
+                msg: "call_ref on mutable stateful precompile".to_string(),
+            }),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


rakita (approved):
good idea!

Added few nit changes:
* Return Fatal error to not panic.
* `PrecompileCow::Default` renamed to `PrecompileCow::StaticRef`
* Return Boxed iter from addresses and additionaly added HashMap for addresses in `Precompile`.

crates/revm/src/context/context_precompiles.rs
```diff
@@ -31,32 +31,25 @@ impl<DB: Database> Clone for ContextPrecompile<DB> {
...
     }
 }
 
 /// Precompiles context.
+#[derive(Clone)]
 pub struct ContextPrecompiles<DB: Database> {
     inner: PrecompilesCow<DB>,
 }
 
-impl<DB: Database> Clone for ContextPrecompiles<DB> {
```
DaniPopes: manual Clone impl is to avoid `DB: Clone`

crates/revm/src/context/context_precompiles.rs
```diff
@@ -144,29 +130,28 @@ impl<DB: Database> ContextPrecompiles<DB> {
...
-    fn to_mut_slow(&mut self) -> &mut HashMap<Address, ContextPrecompile<DB>> {
-        let PrecompilesCow::Default(precompiles) = self.inner else {
+        let PrecompilesCow::Owned(inner) = &mut self.inner else {
             unreachable!()
         };
+        inner
+    }
+
+    /// Mutates Self into Owned variant, or do nothing if it is already Owned.
+    /// Mutation will clone all precompiles.
```
DaniPopes, line 142: This is what `to_mut` does, this does not need to be public

crates/revm/src/context/context_precompiles.rs
```diff
@@ -144,29 +130,28 @@ impl<DB: Database> ContextPrecompiles<DB> {
...
-        match self.inner {
-            PrecompilesCow::Default(_) => self.to_mut_slow(),
-            PrecompilesCow::Owned(ref mut inner) => inner,
-        }
-    }
+        self.mutate_into_owned();
 
-    #[allow(clippy::wrong_self_convention)]
-    #[cold]
-    fn to_mut_slow(&mut self) -> &mut HashMap<Address, ContextPrecompile<DB>> {
```
DaniPopes: I don't understand this change

crates/revm/src/context/evm_context.rs
```diff
@@ -96,8 +96,7 @@ impl<DB: Database> EvmContext<DB> {
     #[inline]
     pub fn set_precompiles(&mut self, precompiles: ContextPrecompiles<DB>) {
         // set warm loaded addresses.
-        self.journaled_state.warm_preloaded_addresses =
-            precompiles.addresses().copied().collect::<HashSet<_>>();
+        self.journaled_state.warm_preloaded_addresses = precompiles.addresses_set();
```
DaniPopes, line 99: shouldn't this extend instead?

rakita: Extend is better, but will leave it as is in this PR.

crates/revm/src/context/context_precompiles.rs
rakita: True, will make it private

crates/revm/src/context/context_precompiles.rs
rakita: It is refactor of code so it reads better, the functionality is same

crates/primitives/src/precompile.rs
```diff
@@ -110,11 +110,27 @@ impl Precompile {
...
+            Precompile::Stateful(ref p) => p.call(bytes, gas_price, env),
+            Precompile::StatefulMut(ref mut p) => p.call_mut(bytes, gas_price, env),
+        }
+    }
+
+    /// Call the precompile with the given input and gas limit and return the result.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the precompile is a mutable stateful precompile.
```
DaniPopes, line 125: Outdated doc

crates/revm/src/context/context_precompiles.rs
DaniPopes: to_mut_slow is marked as cold because it's only done once, but to_mut is called in every other API, pls keep that, it's intentional

rakita: Cold does not do a lot here, that is why it is removed. `to_mut` is called only once (or few times) per transaction.

`mutate_into_owned` is going to be inlined and the current overhead is one pattern match (maybe it is even optimized), i dont see cold making any significant difference.

DaniPopes: Makes the fast path the one that's inlined, and the slow path that is called only once not inlined. Because this method is used in multiple places it makes sense, see `Cow::to_mut`

rakita: Is this significant at all? That is my question, there is not a lot of sense in optimizing every `if` statement.

rakita: It is used only to extend precompile list

DaniPopes: Sure, but if it was already done by me, why remove it? :D

rakita: Dont you think that? Optimisations are worth only on hot paths. Borrowed -> Owned path is slow, but check if it `Borrowed` is not, and besides, branch prediction does good job. The code now looks better and is nicer to read, in this case I optimised for that.

DaniPopes: No because it's not about speed but code size, you are inlining both the check and hot path at every `to_mut` callsite. This is the same situation as `OnceLock` for example, where 99% of the time it's called it will return the initialized value, otherwise it calls the initializer and returns the value from there

rakita: That is why i didn't understand you: "not about speed but code size.". `OnceLock` is a good example.

It is not going to matter a lot as `to_mut` is used only in extending precompiles, but it is a cool mechanism to have, so I will change it to that.

rakita: reintroduced cold function https://github.com/bluealloy/revm/pull/1486/commits/fa38da59a94ce43a8b508e0342a860a3ed81fa39 
@DaniPopes can you take a look

PR #1491, opened 2024-06-05
feat(op): use RollupCostData instead of full encoding

Rjected:
This is a WIP to ultimately replace the `tx_envelope` part of the `OptimismFields` with a new struct.

Currently it just introduces the struct, which basically mirrors the `RollupCostData` from op-geth, and uses it in a few places. The plan is to first integrate this internally, without breaking `OptimismFields`, and once it is properly tested, break `OptimismFields` to use this instead.

### DIFF
--- a/crates/revm/src/optimism/l1block.rs
+++ b/crates/revm/src/optimism/l1block.rs
@@ -1,4 +1,4 @@
-use crate::optimism::fast_lz::flz_compress_len;
+use crate::optimism::rollup_cost::RollupCostData;
 use crate::primitives::{address, db::Database, Address, SpecId, U256};
 use core::ops::Mul;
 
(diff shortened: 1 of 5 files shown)


crates/revm/src/optimism/rollup_cost.rs
```diff
@@ -0,0 +1,39 @@
...
+/// transaction pre-fjord.
+///
+/// The `fastlz_size` field is used to compute the data availability costs for a transaction
+/// post-fjord.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct RollupCostData {
+    /// The number of zeroes in the transaction.
+    pub(crate) zeroes: u64,
+    /// The number of ones in the transaction.
+    pub(crate) ones: u64,
```
DaniPopes, line 18: non-zero

crates/revm/src/optimism/rollup_cost.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// The size of the transaction after fastLZ compression.
+    pub(crate) fastlz_size: u32,
+}
+
+impl RollupCostData {
+    /// This takes bytes as input, creating a [`RollupCostData`] struct based on the encoded data.
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        let mut data = RollupCostData::default();
+        let mut i = 0;
+        while i < bytes.len() {
```
DaniPopes, line 28: the same functionality exists in gas costs, this should reuse that

Rjected: wdym / where? the counting for op data gas costs at least we want to remove, and put here instead

DaniPopes: https://github.com/bluealloy/revm/blob/b7b92aee232558ea48e9580bfac74d442a17babd/crates/interpreter/src/gas/calc.rs#L354-L374

crates/revm/src/optimism/l1block.rs
```diff
@@ -125,21 +125,20 @@ impl L1BlockInfo {
...
-            acc + if *byte == 0x00 {
-                ZERO_BYTE_COST
-            } else {
-                NON_ZERO_BYTE_COST
-            }
-        }));
+        let mut rollup_data_gas_cost = U256::from(cost_data.ones.get())
+            .saturating_mul(U256::from(NON_ZERO_BYTE_COST))
+            .saturating_add(
+                U256::from(cost_data.zeroes.get()).saturating_mul(U256::from(ZERO_BYTE_COST)),
```
DaniPopes, line 140: pretty sure all the arithmetic in this file can be scaled down to use u64

rakita (commented):
@Rjected this looks nice, can you check why tests are failing?

Rjected:
> @Rjected this looks nice, did you check why tests are failing?

nope, was busy with other things, can pick this back up this week

rakita:
Closing it as obsolete

PR #1495, opened 2024-06-07
Fix some nits in bls12_381 precompiles

jtraglia:
I was reviewing the new precompiles and noticed some stuff. This PR does the following:

* Fix a few typos.
* Replace g2 with g1 in `extract_g1_input`.
* Remove unnecessary type hint for `input_fps`.
* Rephrase comment about filtering points at infinity in the MSMs.
  * I found the part about returning bytes a little confusing.
  * We had to do the same thing in ckzg by the way 😢

These look great though. Didn't spot any real issues.

### DIFF
--- a/crates/precompile/src/bls12_381/g1.rs
+++ b/crates/precompile/src/bls12_381/g1.rs
@@ -85,7 +85,9 @@ pub(super) fn extract_g1_input(
         //
         // SAFETY: out is a blst value.
         if unsafe { !blst_p1_affine_on_curve(&out) } {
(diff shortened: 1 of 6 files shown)


Rjected (approved):
thanks! I like the new comments, I think this just needs a fmt fix

jtraglia:
> thanks! I like the new comments, I think this just needs a fmt fix

Thanks! And oops, should be good now.

DaniPopes (approved):

rakita (approved):
Thanks for review!
lgtm

Issue #1483, opened 2024-06-03
on EXTCODEHASH load bytecode and check if it is EOF


rakita:
if EXTCODEHASH target is an account with EOF bytecode, we need to return `0x9dbf3648db8210552e9c4f75c6a1c3057c0ca432043bd648be15fe7be05646f5`

This means that extcodehash would need to load account bytecode and check first byte of it.

https://github.com/ipsilon/eof/blob/main/spec/eof.md#modified-behavior
```
When executed from a legacy contract, if the target account of EXTCODEHASH is an EOF contract,
then it will return 0x9dbf3648db8210552e9c4f75c6a1c3057c0ca432043bd648be15fe7be05646f5
(the hash of EF00, as if that would be the code).
```

jpgonzalezra:
Can I tackle it?

PR #1501, opened 2024-06-08
feat: EXTCODEHASH load bytecode and check if it is EOF

jpgonzalezra:
if EXTCODEHASH target is an account with EOF bytecode, we need to return 0x9dbf3648db8210552e9c4f75c6a1c3057c0ca432043bd648be15fe7be05646f5

This means that extcodehash would need to load account bytecode and check first byte of it.

https://github.com/ipsilon/eof/blob/main/spec/eof.md#modified-behavior

When executed from a legacy contract, if the target account of EXTCODEHASH is an EOF contract,
then it will return 0x9dbf3648db8210552e9c4f75c6a1c3057c0ca432043bd648be15fe7be05646f5
(the hash of EF00, as if that would be the code).

[ISSUE](https://github.com/bluealloy/revm/issues/1483)


crates/interpreter/src/instructions/host.rs
```diff
@@ -72,6 +73,13 @@ pub fn extcodehash<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter,
     } else {
         gas!(interpreter, 400);
     }
+    if let Some((bytecode, _)) = host.code(address) {
```
rakita, line 76: This needs to go where `Host` is implemented here: https://github.com/bluealloy/revm/blob/561fbdbb117a5613475a8f099981a36fedd0104b/crates/revm/src/context.rs#L156

host.code_hash needs to return whatever is needed for instruction.

crates/primitives/src/constants.rs
```diff
@@ -60,3 +60,9 @@ pub const BLOB_GASPRICE_UPDATE_FRACTION: u64 = 3338477;
 
 /// First version of the blob.
 pub const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
+
+/// Precomputed hash for EOF bytecode used with the EXTCODEHASH opcode for EOF contracts.
+pub const EOF_BYTECODE_HASH: &[u8] = &[
+    0x9d, 0xbf, 0x36, 0x48, 0xdb, 0x82, 0x10, 0x55, 0x2e, 0x9c, 0x4f, 0x75, 0xc6, 0xa1, 0xc3, 0x05,
+    0x7c, 0x0c, 0xa4, 0x32, 0x04, 0x3b, 0xd6, 0x48, 0xbe, 0x15, 0xfe, 0x7b, 0xe0, 0x56, 0x46, 0xf5,
+];
```
rakita, line 68: You can use `&hex!("..")` here, it is more readable

crates/interpreter/src/instructions/host.rs
jpgonzalezra: makes sense, let me do it 💪

crates/revm/src/context.rs
```diff
@@ -154,10 +156,19 @@ impl<EXT, DB: Database> Host for Context<EXT, DB> {
...
-        self.evm
+        let code_hash = self
+            .evm
             .code_hash(address)
             .map_err(|e| self.evm.error = Err(e))
-            .ok()
+            .ok();
+
+        if let Some((bytecode, is_cold)) = self.code(address) {
+            if bytecode.is_eof() {
```
jpgonzalezra, line 166: @rakita What do you think about doing this instead of checking the first byte?

rakita: This is fine, as only EOF `0xEFXX` is `0xEF00`

rakita:
I took over this PR, as we need these kind of changes asap. Closing in favour of https://github.com/bluealloy/revm/pull/1504

crates/revm/src/context.rs
jpgonzalezra: thanks for the clarification

PR #1504, opened 2024-06-10
feat(EOF): EXTCODECOPY,EXTCODESIZE,EXTCODEHASH eof support

rakita:
Implement next lines of EOF spec: 

* When executed from a legacy contract, if the target account of EXTCODECOPY is an EOF contract, then it will copy up to 2 bytes from EF00, as if that would be the code.
* When executed from a legacy contract, if the target account of EXTCODEHASH is an EOF contract, then it will return 0x9dbf3648db8210552e9c4f75c6a1c3057c0ca432043bd648be15fe7be05646f5 (the hash of EF00, as if that would be the code).
* When executed from a legacy contract, if the target account of EXTCODESIZE is an EOF contract, then it will return 2.

https://github.com/ipsilon/eof/blob/main/spec/eof.md

closes https://github.com/bluealloy/revm/issues/1483

### DIFF
--- a/crates/revm/src/context/inner_evm_context.rs
+++ b/crates/revm/src/context/inner_evm_context.rs
@@ -10,7 +10,7 @@ use crate::{
         keccak256, Account, Address, AnalysisKind, Bytecode, Bytes, CreateScheme, EVMError, Env,
         Eof, HashSet, Spec,
         SpecId::{self, *},
-        B256, U256,
+        B256, EOF_MAGIC_ARRAY, EOF_MAGIC_HASH, U256,
     },
     FrameOrResult, JournalCheckpoint, CALL_STACK_LIMIT,
 };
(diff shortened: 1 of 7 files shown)


crates/revm/src/context/inner_evm_context.rs
```diff
@@ -160,21 +160,37 @@ impl<DB: Database> InnerEvmContext<DB> {
...
-    pub fn code(&mut self, address: Address) -> Result<(Bytecode, bool), EVMError<DB::Error>> {
+    pub fn code(&mut self, address: Address) -> Result<(Bytes, bool), EVMError<DB::Error>> {
         self.journaled_state
             .load_code(address, &mut self.db)
-            .map(|(a, is_cold)| (a.info.code.clone().unwrap(), is_cold))
+            .map(|(a, is_cold)| {
+                // SAFETY: safe to unwrap as load_code will insert code if it is empty.
+                let code = a.info.code.as_ref().unwrap();
+                if code.is_eof() {
+                    (EOF_MAGIC_ARRAY.into(), is_cold)
```
DaniPopes, line 174: this `.into()` will allocate when it's a constant, EOF_MAGIC_ARRAY should maybe be `&[u8; 2]` or `Bytes`?

DaniPopes (approved):

crates/revm/src/context/inner_evm_context.rs
rakita: Agree, made the change

PR #1505, opened 2024-06-11
refactor: replace U256 with u64 in BLOCKHASH

### LINKED ISSUE #1423: Don't use U256 operation in EIP-2935 BLOCKHASH
since the window is lower than u64::MAX we can use the lowest word and use u64 modulo operation instead of U256 for better performance.

TropicalDog17:
## Description

Change the type of the number parameter in the block_hash methods of Host and Database from U256 to u64.

Closes #1423 

\cc @rakita @DaniPopes

### DIFF
--- a/crates/revm/src/context.rs
+++ b/crates/revm/src/context.rs
@@ -110,7 +110,7 @@ impl<EXT, DB: Database> Host for Context<EXT, DB> {
 
     fn block_hash(&mut self, number: u64) -> Option<B256> {
         let block_number = as_usize_saturated!(self.env().block.number);
-        let requested_number = number as usize;
+        let requested_number = usize::try_from(number).unwrap_or(usize::MAX);
 
         let Some(diff) = block_number.checked_sub(requested_number) else {
             return Some(B256::ZERO);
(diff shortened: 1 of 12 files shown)


crates/revm/src/context.rs
```diff
@@ -108,9 +108,9 @@ impl<EXT, DB: Database> Host for Context<EXT, DB> {
         &mut self.evm.env
     }
 
-    fn block_hash(&mut self, number: U256) -> Option<B256> {
+    fn block_hash(&mut self, number: u64) -> Option<B256> {
         let block_number = as_usize_saturated!(self.env().block.number);
-        let requested_number = as_usize_saturated!(number);
+        let requested_number = number as usize;
```
rakita, line 113: This can panic. We should use `usize::try_from(number).unwrap_or(usize::MAX);`

rakita (changes requested):
cool change. One edge case pending

TropicalDog17:
@rakita thanks for your review, I've refactored accordingly.

rakita (approved):
lgtm

DaniPopes (approved):

PR #1512, opened 2024-06-11
perf: Removed .clone() in ExecutionHandler::call, and reusing output buffer…

impactdni2:
Reduces allocations in a two places.

1) In ExecutionHandler::call(), drops a .clone() on the inputs (allocated cloned Bytes)
2) In Interpreter::insert_call_outcome, re-uses the passed in call_outcome.result.output rather than cloning the Bytes into a the new location

All the tests pass as far as I can tell, but I'm getting a bit lower in the stack than I'm super comfortable in, so would love an extra set of eyes to tell me if I'm missing something.


rakita (approved):
lgtm

Issue #1488, opened 2024-06-05
Put `EOF` structure behind an `Arc`


DaniPopes:
`EOF` structure is present inside of `Bytecode`, and is cloned internally in `AccountInfo` and other places. It would improve performance to put behind an `Arc` to avoid cloning the underlying memory.

jpgonzalezra:
I'm taking a look at why the tests are failing, it seems to be something else and not the changes made here

PR #1517, opened 2024-06-13
feat(EOF): Put EOF bytecode behind an Arc

jpgonzalezra:
This is the [ISSUE](https://github.com/bluealloy/revm/issues/1488)

DaniPopes: EOF structure is present inside of Bytecode, and is cloned internally in AccountInfo and other places. It would improve performance to put behind an Arc to avoid cloning the underlying memory.

cc: @rakita @DaniPopes

### DIFF
--- a/crates/interpreter/src/instructions/control.rs
+++ b/crates/interpreter/src/instructions/control.rs
@@ -204,6 +204,8 @@ pub fn unknown<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
 
 #[cfg(test)]
 mod test {
+    use std::sync::Arc;
+
     use revm_primitives::{bytes, eof::TypesSection, Bytecode, Eof, PragueSpec};
 
     use super::*;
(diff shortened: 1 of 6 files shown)


jpgonzalezra:
I'm taking a look at why the tests are failing, it seems to be something else and not the changes made here

rakita:
> I'm taking a look at why the tests are failing, it seems to be something else and not the changes made here

It is a malformed json file https://github.com/ethereum/tests/pull/1384

It is not related to this PR.

rakita (approved):
lgtm!

btw there are keywords that github understands that are good to use https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

crates/interpreter/src/instructions/control.rs
```diff
@@ -204,6 +204,8 @@ pub fn unknown<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
 
 #[cfg(test)]
 mod test {
+    use std::sync::Arc;
+
```
rakita, line 208:
```suggestion
```

jpgonzalezra:
Thanks for the recommendation, I will start using it.

PR #1518, opened 2024-06-13
fix(revm): remove storage reset that clears is_cold flag

rakita:
Introduced here: https://github.com/bluealloy/revm/pull/1437
Found by @holiman
Fortunately, the bug was not included in Reth or in any release.

Done a few more renames.


rakita:
Slightly older ethereum/tests run locally.
Failure is due to json artefacts not related to PR.

PR #1522, opened 2024-06-14
chore: use const blocks

DaniPopes:
`const {}` blocks were stabilized yesterday as part of 1.79.

I would've liked to make the instruction table `&'static` but unfortunately that requires `H: 'static` even if it's through a function pointer.


onbjerg (approved):

DaniPopes:
Revm has no MSRV, but yes it would be a bump from 1.75 (I just tried `cargo msrv` on main b988a0f1253cd8b63a7dbf76f038758cc851bb6a)

rakita (approved):
lgtm!

rakita:
Will publish a new version of revm soon, so I will merge this PR after this.

rakita (approved):
lgtm

rakita:
Failed tests not related

PR #1524, opened 2024-06-14
feat: mutable access for all fields in BundleBuilder

### LINKED ISSUE #1519: Mutable access to `state_storage` in `BundleBuilder`
I think we could probably avoid this clone if we had access to a mutable reference of `state_storage` on the `BundleBuilder`

_Originally posted by @quentinv72 in https://github.com/paradigmxyz/reth/pull/8264#discussion_r1600570168_

jpgonzalezra:
This PR addresses Issue #1519 , let me know what do you think

Originally posted by @quentinv72 in https://github.com/paradigmxyz/reth/pull/8264#discussion_r1600570168
@rakita @gakonst 

Closes #1519

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -233,6 +233,53 @@ impl BundleBuilder {
     pub fn get_states(&self) -> &HashSet<Address> {
         &self.states
     }
+
+    /// Mutable getter for `states` field
+    pub fn get_states_mut(&mut self) -> &mut HashSet<Address> {
+        &mut self.states
+    }
+
+    /// Mutable getter for `state_original` field
+    pub fn get_state_original_mut(&mut self) -> &mut HashMap<Address, AccountInfo> {
+        &mut self.state_original
+    }
+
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -100,9 +100,14 @@ impl BundleBuilder {
     }
 
     /// Collect storage info of BundleState state
-    pub fn state_storage(mut self, address: Address, storage: HashMap<U256, (U256, U256)>) -> Self {
+    pub fn state_storage(
+        mut self,
+        address: Address,
+        storage: &mut HashMap<U256, (U256, U256)>,
```
rakita, line 106: This is not what the requirement was.

It is to add `get_state_storage_mut(` function

rakita: While we are at it, it would be good to add mut getters for all fields

jpgonzalezra: my bad, i'm going to fix it 💪

rakita (approved):
lgtm

PR #1527, opened 2024-06-16
db: add setters to `BundleBuilder` with `&mut self`

tcoratger:
Related https://github.com/paradigmxyz/reth/pull/8861/files#r1641783394

### DIFF
--- a/crates/revm/src/db/states/bundle_state.rs
+++ b/crates/revm/src/db/states/bundle_state.rs
@@ -79,30 +79,44 @@ impl BundleBuilder {
         }
     }
 
+    /// Apply a transformation to the builder.
+    pub fn apply<F>(self, f: F) -> Self
+    where
+        F: FnOnce(Self) -> Self,
+    {
+        f(self)
+    }
+
+    /// Apply a mutable transformation to the builder.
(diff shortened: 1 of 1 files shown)


crates/revm/src/db/states/bundle_state.rs
```diff
@@ -151,6 +151,61 @@ impl BundleBuilder {
         self
     }
 
+    /// Set address info of BundleState state.
+    pub fn set_state_address(&mut self, address: Address) {
```
mattsse, line 155: since this is a builder type, we can return &mut Self on these functions so they can be chained

the existing `self` functions should then reuse the &mut self ones so we don't duplicate the actual impl

mattsse (changes requested):
having these is very useful because usually we need to call them conditionally which because slightly ugly with with the consuming builder functions.

can we also add an `fn apply{_mut}<F: FnOnce(Self) -> Self>(self, f)`

crates/revm/src/db/states/bundle_state.rs
tcoratger: Fixed, let me know if this is ok for you

rakita (approved):

PR #1528, opened 2024-06-17
wip: Host associate type, ChainSpec refactor

rakita:
This adds ChainSpec as associate type inside Host and removed a need to modify instruction.

Small refactor on `ChainSpec` where associate types are moved to standalone traits.

### DIFF
--- a/crates/primitives/src/transaction.rs
+++ b/crates/primitives/src/transaction.rs
@@ -1,7 +1,19 @@
-use crate::{Address, Bytes, HashMap, TransactTo, B256, GAS_PER_BLOB, U256};
+use crate::{Address, Bytes, HashMap, InvalidTransaction, TransactTo, B256, GAS_PER_BLOB, U256};
+use cfg_if::cfg_if;
+use core::fmt::{Debug, Display};
 
 /// Trait for retrieving transaction information required for execution.
-pub trait Transaction {
+pub trait Transaction: Clone + Debug + Default + PartialEq + Eq {
+    cfg_if! {
+        if #[cfg(feature = "serde")] {
+            /// The error type that can be returned when validating a transaction.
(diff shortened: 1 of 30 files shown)


rakita:
Closing it as it is merged in Nomic repo

crates/primitives/src/transaction.rs
```diff
@@ -1,7 +1,19 @@
-use crate::{Address, Bytes, HashMap, TransactTo, B256, GAS_PER_BLOB, U256};
+use crate::{Address, Bytes, HashMap, InvalidTransaction, TransactTo, B256, GAS_PER_BLOB, U256};
+use cfg_if::cfg_if;
+use core::fmt::{Debug, Display};
 
 /// Trait for retrieving transaction information required for execution.
-pub trait Transaction {
+pub trait Transaction: Clone + Debug + Default + PartialEq + Eq {
```
Wodann, line 6: I was looking at this again and I'm not a fan a of adding these trait bounds on the `Transaction` trait itself.

This is saying that anything that implements `Transaction` should also have implemented `Clone`, `Debug`, `Default`, `PartialEq`, and `Eq`.

I think the trait should be further constrained to trait bounds at the usage site, not universally. This trait inherently doesn't have any of these requirements and thus is overly constrained, limiting reusability.

The same applies for the `type TransactionValidationError`. Maybe that should live on a secondary trait that's specifically used for the `trait ChainSpec`?

DaniPopes: No, latest stable is fine

Wodann: This is what I had in mind: https://github.com/bluealloy/revm/pull/1378/commits/4bb0f2726b3fa9a8c940c5097189b13193cc2ede

PR #1542, opened 2024-06-17
refactor: replace TransactTo with TxKind

Wodann:
This PR re-uses the `alloy_primitives` type `TxKind` instead of the custom REVM type `TransactTo`


crates/primitives/src/env.rs
```diff
@@ -658,41 +659,6 @@ pub struct OptimismFields {
     pub enveloped_tx: Option<Bytes>,
 }
 
-/// Transaction destination.
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub enum TransactTo {
```
DaniPopes, line 664: A type alias could be added here to ease users transition

Wodann: Addressed in https://github.com/bluealloy/revm/pull/1542/commits/e264643bcad9e5566ddac2d72615d81c65b131ac

rakita (approved):
lgtm

PR #1545, opened 2024-06-18
chore: remove DatabaseWithDebugError

DaniPopes:
Looks unused.

### DIFF
--- a/crates/primitives/src/db.rs
+++ b/crates/primitives/src/db.rs
@@ -67,12 +67,6 @@ impl<F: DatabaseRef> From<F> for WrapDatabaseRef<F> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


rakita:
It is here to be exported, usable for someone who wants a database with debug errors and does not need to redefine this trait. 

Would close this PR

mattsse:
this looks unimplemented though, so not really useful

DaniPopes:
if you're using it generically you would have either dbext like in foundry, or you can specify "Db<Error: Debug>" (now with rust 1.79 its even easier)

rakita:
Convinced. Merging PR

rakita (approved):

PR #1547, opened 2024-06-18
feat: add matter-labs based eip1962 precompile impls

Rjected:
This add the EIP 196 precompile implementation based on https://github.com/matter-labs/eip1962 instead of the substrate bn library.

This still needs:
- [ ] extensive testing
- [ ] potentially better error conversions
- [ ] benchmarking for ecMul and pairing precompiles

### DIFF
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -43,6 +43,9 @@ aurora-engine-modexp = { version = "1.1", default-features = false }
 # ecAdd, ecMul, ecPairing
 bn = { package = "substrate-bn", version = "0.6", default-features = false }
 
+# ecAdd (soon others)
+eth_pairings = { version = "0.2.0", default-features = false }
+
 # KZG point evaluation precompile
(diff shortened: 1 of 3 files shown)


rakita (changes requested):
I would want to leave bn behind a feature gate, in case of a need to switch back easily.

crates/precompile/Cargo.toml
```diff
@@ -63,6 +63,9 @@ serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
 
+# ecAdd, ecMul, ecPairing
+bn = { package = "substrate-bn", version = "0.6", default-features = false }
```
DaniPopes, line 67: looks unused?

Rjected:
The matter labs repo is not `no_std` compatible, working on it however, so I'll probably need to change from a `bn` feature to a `matter_labs` feature, where matter labs is only selected if std is also available

rakita:
Cleanup closing, if we go back to it we can always reopen

PR #1552, opened 2024-06-20
refactor: replace AccessList with alloy version

Wodann:
The replaces the custom `AccessListItem` definition with the `alloy` version. The alloy version uses a `B256` instead of a `U256`, which is conform the Ethereum spec.

This avoids type conversions from when interacting with revm, where currently `B256` storage indices need to be converted to `U256` values using big-Endian conversion.

I think it's worth considering changing the account storage to match this.

### DIFF
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -1,10 +1,11 @@
+use revm_primitives::AccessListItem;
+
 use super::constants::*;
 use crate::{
     num_words,
-    primitives::{Address, SpecId, U256},
+    primitives::{SpecId, U256},
     SelfDestructResult,
 };
-use std::vec::Vec;
 
 /// `const` Option `?`.
 macro_rules! tri {
@@ -357,7 +358,7 @@ pub fn validate_initial_tx_gas(
     spec_id: SpecId,
     input: &[u8],
     is_create: bool,
-    access_list: &[(Address, Vec<U256>)],
+    access_list: &[AccessListItem],
 ) -> u64 {
     let mut initial_gas = 0;
     let zero_data_len = input.iter().filter(|v| **v == 0).count() as u64;
@@ -375,9 +376,12 @@ pub fn validate_initial_tx_gas(
 
     // get number of access list account and storages.
     if spec_id.is_enabled_in(SpecId::BERLIN) {
-        let accessed_slots = access_list
-            .iter()
-            .fold(0, |slot_count, (_, slots)| slot_count + slots.len() as u64);
+        let accessed_slots =
+            access_list
+                .iter()
+                .fold(0, |slot_count, AccessListItem { storage_keys, .. }| {
+                    slot_count + storage_keys.len() as u64
+                });
         initial_gas += access_list.len() as u64 * ACCESS_LIST_ADDRESS;
         initial_gas += accessed_slots * ACCESS_LIST_STORAGE_KEY;
     }
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -378,7 +378,7 @@ pub fn validate_initial_tx_gas(
     if spec_id.is_enabled_in(SpecId::BERLIN) {
         let accessed_slots: usize = access_list
             .iter()
-            .map(|AccessListItem { storage_keys, .. }| storage_keys.len())
+            .map(|item| item.storage_keys.len())
             .sum();
         initial_gas += access_list.len() as u64 * ACCESS_LIST_ADDRESS;
         initial_gas += accessed_slots as u64 * ACCESS_LIST_STORAGE_KEY;
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -376,14 +376,12 @@ pub fn validate_initial_tx_gas(
 
     // get number of access list account and storages.
     if spec_id.is_enabled_in(SpecId::BERLIN) {
-        let accessed_slots =
-            access_list
-                .iter()
-                .fold(0, |slot_count, AccessListItem { storage_keys, .. }| {
-                    slot_count + storage_keys.len() as u64
-                });
+        let accessed_slots: usize = access_list
+            .iter()
+            .map(|AccessListItem { storage_keys, .. }| storage_keys.len())
+            .sum();
         initial_gas += access_list.len() as u64 * ACCESS_LIST_ADDRESS;
-        initial_gas += accessed_slots * ACCESS_LIST_STORAGE_KEY;
+        initial_gas += accessed_slots as u64 * ACCESS_LIST_STORAGE_KEY;
     }
 
     // base stipend
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -22,6 +22,7 @@ rust_2018_idioms = "deny"
 all = "warn"
 
 [dependencies]
+alloy-eips = { version = "0.1", default-features = false }
 alloy-primitives = { version = "0.7.2", default-features = false, features = [
     "rlp",
 ] }
@@ -53,6 +54,7 @@ hex = { version = "0.4", default-features = false }
 default = ["std", "c-kzg", "portable"]
 std = [
     "serde?/std",
+    "alloy-eips/std",
     "alloy-primitives/std",
     "hex/std",
     "bitvec/std",
@@ -61,6 +63,7 @@ std = [
 hashbrown = []
 serde = [
     "dep:serde",
+    "alloy-eips/serde",
     "alloy-primitives/serde",
     "hex/serde",
     "hashbrown/serde",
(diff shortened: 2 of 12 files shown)


crates/revm/src/journaled_state.rs
```diff
@@ -544,9 +545,10 @@ impl JournaledState {
             ),
         };
         // preload storages.
-        for slot in slots {
-            if let Entry::Vacant(entry) = account.storage.entry(*slot) {
-                let storage = db.storage(address, *slot).map_err(EVMError::Database)?;
+        for storage_key in storage_keys {
+            let slot = U256::from_be_bytes(storage_key.0);
+            if let Entry::Vacant(entry) = account.storage.entry(slot) {
+                let storage = db.storage(address, slot).map_err(EVMError::Database)?;
```
Wodann, line 551: Instead of converting to a `U256` should we do a backwards incompatible change that makes the account storage a `HashMap<B256, U256>`?

This would match the expected behaviour of the Ethereum protocol.

crates/primitives/src/bytecode/eof.rs
```diff
@@ -215,8 +215,10 @@ mod test {
         assert_eq!(eof.data_slice(0, 4), &[0x01, 0x02, 0x03, 0x04]);
         assert_eq!(eof.data_slice(0, 5), &[0x01, 0x02, 0x03, 0x04]);
         assert_eq!(eof.data_slice(1, 2), &[0x02, 0x03]);
-        assert_eq!(eof.data_slice(10, 2), &[]);
-        assert_eq!(eof.data_slice(1, 0), &[]);
-        assert_eq!(eof.data_slice(10, 0), &[]);
+
+        const EMPTY: &[u8] = &[];
```
Wodann, line 219: I'm unsure why, but these assertions were causing compiler errors locally due to underspecified Vec types

crates/primitives/src/bytecode/eof/header.rs
```diff
@@ -242,7 +242,7 @@ mod tests {
         let (header, _) = EofHeader::decode(&input).unwrap();
         assert_eq!(header.types_size, 4);
         assert_eq!(header.code_sizes, vec![1]);
-        assert_eq!(header.container_sizes, vec![]);
+        assert_eq!(header.container_sizes, Vec::<u16>::new());
```
Wodann, line 245: I'm unsure why, but these assertions were causing compiler errors locally due to underspecified Vec types

crates/primitives/Cargo.toml
```diff
@@ -22,6 +22,7 @@ rust_2018_idioms = "deny"
 all = "warn"
 
 [dependencies]
+alloy-eips = { version = "0.1", default-features = false }
```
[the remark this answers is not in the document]
rakita: It was strange that alloy implemented their own kzg thing and didn't import it from revm and add things they are missing.

This is a reason why there is revm-primitives, it is a low level crate.

rakita (changes requested):
I dont think this is good dependency to be added. And `U256` is there as the stack is in `U256` and we remove conversion between B256/U256 inside revm, this can be done outside of it.

I don't feel comfortable in including Alloy, and would reverse the dependency on alloy including revm-primitives.

`AccessListItem` is a good example of two slightly different usages. They need B256 for rpc, while revm prefers U256 for stack.

DaniPopes:
access list U256 does not touch the stack, it goes to state and state is u256

rakita:
> Access list U256 does not touch the stack, it goes to storage and storage is U256. The conversion is essentially free in the context of SLOAD/SSTORE where 99% of the time is spent in hashmaps and DB, I would not block this change on that.

True it touches storage and storage is U256 as the stack is U256. Again, it is strange to depend on alloy, and would like to see alloy including `revm-primitives` for example to add `From<>`

crates/primitives/src/bytecode/eof.rs
Wodann: Rust v1.79

crates/interpreter/src/gas/calc.rs
```diff
@@ -375,9 +376,12 @@ pub fn validate_initial_tx_gas(
 
     // get number of access list account and storages.
     if spec_id.is_enabled_in(SpecId::BERLIN) {
-        let accessed_slots = access_list
-            .iter()
-            .fold(0, |slot_count, (_, slots)| slot_count + slots.len() as u64);
+        let accessed_slots =
+            access_list
+                .iter()
+                .fold(0, |slot_count, AccessListItem { storage_keys, .. }| {
```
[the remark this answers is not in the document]
Wodann: Thanks. Addressed in https://github.com/bluealloy/revm/pull/1552/commits/c17c4f8a92f0d20458f50c9b95f482b18c893fe7

DaniPopes (approved):
Thanks, we resolved our disputes internally and decided to include this, one last nit for the test changes

Wodann:
> U256 is there as the stack is in U256 and we remove conversion between B256/U256 inside revm, this can be done outside of it.

I see your point w.r.t the SLOAD/SSTORE instructions popping a `U256` from the stack and wanting to use that to access `fn storage`. I'm happy to leave that as is, without conversion to `B256`.

> would like to see alloy including revm-primitives for example to add From<>

I'm also happy to move the `AccessList` and `AccessListItem` types to `revm-primitives`. Either is fine for us, as long as we can use the same type across the board.

The intent of this PR is to avoid the conversion `fn (Vec<(Address, Vec<B256>) -> Vec<(Address, Vec<U256>)>` to configure transactions.

Especially in combination with the [`trait Transaction`](https://github.com/bluealloy/revm/pull/1378/files#diff-07a440ede73610393cb59ae72a2f69037ee10bbe9ff0f46dff3187e8d2eba32aR36) proposed in the chain-specific configuration PR, this creates the ability to bypass the conversion of `From<ThirdPartyTransaction> for TxEnv`.

Instead we can directly use `ThirdPartyTransaction` in revm to query the transaction parameters for execution. This philosophy does require using the same types at the API level between Ethereum protocol and Ethereum VM. Otherwise people will always need to create two types: the chain transaction type and a chain "transaction environment" type.

Please let me know which changes are reasonable to get this PR across the finish line - if at all possible.

Wodann:
> Thanks, we resolved our disputes internally and decided to include this, one last nit for the test changes

Perfect. That voids the need for my previous message 😁

crates/primitives/src/bytecode/eof.rs
Wodann: I reverted the changes.

This might merely have been an artifact of me using devcontainers to develop. The devcontainer is built with Rust v1.76, after which I manually set a new `rustup` default of v1.79. `rust-analyzer` might not be picking up this change after the container has been built.

Wodann: The same error are occurring in CI: https://github.com/bluealloy/revm/actions/runs/9600615625/job/26477283846?pr=1552

Wodann: @DaniPopes I re-reverted the commit in https://github.com/bluealloy/revm/pull/1552/commits/cef93c7c5d3c054bb22d8b0623c620575ec35c5e

It seems to be a result of conflicting implementations of a trait between `core` and `serde_json` which prevents the `assert_eq!` macro from deducing the correct type.

DaniPopes: Interesting, looks like this https://github.com/serde-rs/json/issues/958

DaniPopes (approved):

crates/interpreter/src/gas/calc.rs
```diff
@@ -375,11 +376,12 @@ pub fn validate_initial_tx_gas(
 
     // get number of access list account and storages.
     if spec_id.is_enabled_in(SpecId::BERLIN) {
-        let accessed_slots = access_list
+        let accessed_slots: usize = access_list
             .iter()
-            .fold(0, |slot_count, (_, slots)| slot_count + slots.len() as u64);
+            .map(|AccessListItem { storage_keys, .. }| storage_keys.len())
```
mattsse, line 381:
```suggestion
            .map(|item| item.storage_keys.len())
```

examples/generate_block_traces.rs
```diff
@@ -130,12 +130,16 @@ async fn main() -> anyhow::Result<()> {
                         .0
                         .into_iter()
                         .map(|item| {
-                            let new_keys: Vec<U256> = item
+                            let storage_keys: Vec<B256> = item
                                 .storage_keys
                                 .into_iter()
-                                .map(|h256| U256::from_le_bytes(h256.0))
+                                .map(|h256| B256::new(h256.0))
```
mattsse, line 136: is this needed?

mattsse (changes requested):

examples/generate_block_traces.rs
DaniPopes: yeah it's from ethers

crates/revm/src/journaled_state.rs
rakita: This does not make sense

rakita:
I am still of the stance that AccessList should be inside this repo, and `alloy/eips` to depend on `revm-primitives`, revm-primitives is the lowest crate in the crate and can be published at any time. In the end Transaction (and its parts) without evm does not exist.

But having it in one or other repo is a maintenance problem, the user problem is having two different types and is something that would be good to solve. As I said internally I am okay to step back and include alloy/eips.

rakita (approved):
lgtm

rakita:
Failed tests not related to this PR

Issue #1534, opened 2024-06-17
Store a `tokio::Handle` in alloy/ethers DB


DaniPopes:
Store an instance of `tokio::Handle` by calling `Handle::try_current` in the constructor instead of calling it on every RPC call.

This also removes the need to create a tokio runtime for every RPC call as well by always requiring a tokio handle to be present at instantiation.

This change was also made in Foundry, see https://github.com/foundry-rs/foundry/pull/7860

rakita:
done

PR #1557, opened 2024-06-21
chore: store tokio::runtime::Handle in ethers/alloyDB

chirag-bgh:
ref #1534

### DIFF
--- a/crates/revm/src/db/ethersdb.rs
+++ b/crates/revm/src/db/ethersdb.rs
@@ -15,9 +15,18 @@ pub struct EthersDB<M: Middleware> {
 }
 
 impl<M: Middleware> EthersDB<M> {
-    /// create ethers db connector inputs are url and block on what we are basing our database (None for latest)
+    /// Create ethers db connector inputs are url and block on what we are basing our database (None for latest).
+    ///
+    /// Returns `None` if no tokio runtime is available or if the current runtime is a current-thread runtime.
     pub fn new(client: Arc<M>, block_number: Option<BlockId>) -> Option<Self> {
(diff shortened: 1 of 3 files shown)


DaniPopes (changes requested):
can you also return None if the runtime flavor is CurrentThread, and can you document it briefly in the fn new doc comments saying that it requires a tokio runtime?

otherwise looks good

crates/revm/src/db/ethersdb.rs
```diff
@@ -11,55 +11,45 @@ use crate::{Database, DatabaseRef};
...
             block_number,
+            handle,
         })
     }
 
-    /// internal utility function to call tokio feature and wait for output
+    /// Internal utility function to call tokio feature and wait for output
     #[inline]
-    fn block_on<F>(f: F) -> F::Output
+    fn block_on<F>(f: F, handle: &Handle) -> F::Output
```
mattsse, line 47: this function can now accept &self

mattsse (dismissed):

DaniPopes (approved):

rakita (approved):

wtdcode:
@rakita  This change is not appropriate. If you wish to enforce ethersdb/alloydb to always play with async world, you should mark `fn new` as `async fn new` even not using any `await` in the function. This removes the check and also presents the correct signature.

@DaniPopes I disagree with not allowing ethersdb/alloydb to create current thread runtime. This makes them totally incompatible with sync world and creating a current thread runtime is rather cheap.

wtdcode:
@DaniPopes Also, why would you disallow the current thread flavor? I can't understand.

PR #1565, opened 2024-06-24
feat(Prague): Add EIP-7702

rakita:
Waiting for https://github.com/alloy-rs/alloy/pull/950

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -1,6 +1,7 @@
+pub mod eip7702;
 pub mod handler_cfg;
 
-use alloy_primitives::TxKind;
+pub use eip7702::AuthorizationList;
 pub use handler_cfg::{CfgEnvWithHandlerCfg, EnvWithHandlerCfg, HandlerCfg};
 
 use crate::{
@@ -8,6 +9,7 @@ use crate::{
     Spec, SpecId, B256, GAS_PER_BLOB, KECCAK_EMPTY, MAX_BLOB_NUMBER_PER_BLOCK, MAX_INITCODE_SIZE,
     U256, VERSIONED_HASH_VERSION_KZG,
 };
+use alloy_primitives::TxKind;
 use core::cmp::{min, Ordering};
 use core::hash::Hash;
 use std::boxed::Box;
@@ -570,9 +572,7 @@ pub struct TxEnv {
     /// Set EOA account code for one transaction
     ///
     /// [EIP-Set EOA account code for one transaction](https://eips.ethereum.org/EIPS/eip-7702)
-    ///
-    /// TODO: include from alloy/eips crate.
-    pub authorization_list: Option<Vec<(Address, Address)>>,
+    pub authorization_list: Option<AuthorizationList>,
 
     #[cfg_attr(feature = "serde", serde(flatten))]
     #[cfg(feature = "optimism")]
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -572,7 +572,7 @@ pub struct TxEnv {
     /// [EIP-Set EOA account code for one transaction](https://eips.ethereum.org/EIPS/eip-7702)
     ///
     /// TODO: include from alloy/eips crate.
-    pub authorization_list: Option<Vec<(Address,Address)>>,
+    pub authorization_list: Option<Vec<(Address, Address)>>,
 
     #[cfg_attr(feature = "serde", serde(flatten))]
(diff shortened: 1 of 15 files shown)


crates/primitives/src/env.rs
```diff
@@ -92,6 +92,25 @@ impl Env {
     /// Return initial spend gas (Gas needed to execute transaction).
     #[inline]
     pub fn validate_tx<SPEC: Spec>(&self) -> Result<(), InvalidTransaction> {
+        // Check if the transaction's chain id is correct
+        if let Some(tx_chain_id) = self.tx.chain_id {
+            if tx_chain_id != self.cfg.chain_id {
```
onbjerg, line 97: per the spec: 

>Verify the chain id is either 0 or the chain’s current ID.

so we should also check if the chain id is 0 for 7702 tx's

crates/primitives/src/env.rs
```diff
@@ -128,65 +140,66 @@ impl Env {
...
+
+        // check if EIP-7702 transaction is enabled.
+        if !SPEC::enabled(SpecId::PRAGUE) && self.tx.authorization_list.is_some() {
+            return Err(InvalidTransaction::AuthorizationListNotSupported);
+        }
+
+        if self.tx.authorization_list.is_some() {
+            // Check if other fields are unset.
+            if self.tx.max_fee_per_blob_gas.is_some() || !self.tx.blob_hashes.is_empty() {
+                return Err(InvalidTransaction::AuthorizationListInvalidFields);
```
onbjerg, line 202: should this not be `BlobVersionedHashesNotSupported`? or something similar, this error is a bit vague since the error is that we have an auth list *and* blobs

crates/revm/src/context/inner_evm_context.rs
```diff
@@ -29,6 +29,8 @@ pub struct InnerEvmContext<DB: Database> {
     pub db: DB,
     /// Error that happened during execution.
     pub error: Result<(), EVMError<DB::Error>>,
+    /// EIP-7702 Authorization list of accounts that needs to be cleared.
+    pub valid_authorizations: Vec<Address>,
```
onbjerg, line 33: i think this doesn't need to be a vec, per the spec only 1 authorization is ever used per 7702 tx (the first valid one), at least as far as i understand it, and there is currently no way to do a 7702 call from a contract.

crates/revm/src/handler/mainnet/pre_execution.rs
```diff
@@ -47,6 +48,56 @@ pub fn load_accounts<SPEC: Spec, EXT, DB: Database>(
...
+                let code_hash = account.info.code_hash;
+
+                // 5. Set the code of authority to code associated with address.
+                context.evm.inner.journaled_state.set_code_with_hash(
+                    *authority,
+                    code.unwrap_or_default(),
+                    code_hash,
+                );
+
+                valid_auths.push(*authority);
```
onbjerg, line 94: per the spec:

>If any of the above steps fail, immediately stop processing that tuple and continue to the next tuple in the list. It will In the case of multiple tuples for the same authority, set the code specified by the address in the first occurrence.

so given the authorization here is valid, we should actually stop processing the next one. i.e. the first valid authorization is the one the tx is going to use (as far as i can parse)

onbjerg: ah, maybe i got it wrong, i mixed up the address used for the code and the authority. so if i understand correctly now, you can have *multiple* valid authorities, but if there are two entries for the same authority, only the first one is used?

crates/primitives/src/env.rs
rakita: That is for `authorization_list` where `chain_id` can be zero.

The transaction chain_id needs to be checked, and it is invalid if it is zero for Ethereum

crates/primitives/src/env.rs
rakita: Yeah, EnvTx sucks rn, as we have all fields in one struck. This error is here if it is AuthorizationList TX don't set other fields by mistake.

crates/revm/src/handler/mainnet/pre_execution.rs
rakita: > ah, maybe i got it wrong, i mixed up the address used for the code and the authority. so if i understand correctly now, you can have _multiple_ valid authorities, but if there are two entries for the same authority, only the first one is used?

True

onbjerg (approved):
lgtm

PR #1567, opened 2024-06-25
feat: add helper function to mape EVMError's Database error variant

mattsse:
makes it possible to transform the generic db error


rakita (approved):
lgtm

PR #1568, opened 2024-06-26
feat: add bytecode_address from CallInputs to Contract during construction.

vandenbogart:
It's completely possible I am missing something, but I couldn't find a good way to access the address of the account which owns the bytecode being executed while using the Inspector::step hook. It would be really convenient if the contract stored on the interpreter also contained the bytecode address like the CallInputs it is constructed from do.

### DIFF
--- a/crates/interpreter/src/interpreter/contract.rs
+++ b/crates/interpreter/src/interpreter/contract.rs
@@ -15,13 +15,13 @@ pub struct Contract {
     /// Bytecode contains contract code, size of original code, analysis with gas block and jump table.
     /// Note that current code is extended with push padding and STOP at end.
     pub bytecode: Bytecode,
-    /// Address of the account the bytecode was loaded from. This can be different from target_address
-    /// in the case of DELEGATECALL or CALLCODE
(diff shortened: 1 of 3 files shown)


rakita (commented):
This seems reasonable to add.

But would make `bytecode_address` a `Option` in a similar way as the `Contract` `hash` is, and would put it to `None` if it is a creation call.

crates/interpreter/src/interpreter/contract.rs
```diff
@@ -60,6 +65,7 @@ impl Contract {
             bytecode,
             hash,
             contract_address,
+            None,
```
rakita, line 68: This should be added from `env.tx.kind` if kind is `To` we should set that address.

rakita (commented):
One more nit

crates/interpreter/src/interpreter/contract.rs
vandenbogart: I think I get your meaning. I set it to None in the case of create and to the destination address in the case of Call

rakita (approved):

PR #1569, opened 2024-06-26
tests: migrate revme benchmarks to Criterion

JacekGlen:
This PR migrates the existing `revme evm --bench` tool to Criterion. The benefits of doing so include:
- more stable results
- detailed reports
- automatic performance degradation detection

The main motivation for this change is that the tool can be used in [gas-cost-estimator](https://github.com/imapp-pl/gas-cost-estimator) project, but its purpose goes beyond it. You can learn more about the project from this short [presentation](https://streameth.org/devconnect/watch?session=65b8f8d5a5b2d09b88ec1838), or visit the wiki [Stage 4 Scope](https://github.com/imapp-pl/gas-cost-estimator/wiki/Stage-4-Scope).

In the future, I'll extend the tool to run on a preloaded state, which is useful for testing scenarios with opcodes like `DELEGATECALL` or `SLOAD`.


rakita:
There is already a cli that allows you to run bytecode in `revme`: https://github.com/bluealloy/revm/blob/b3f31fc6f2fa0c15f7287ac83ba608a8513ffef1/bins/revme/src/cmd.rs#L21C5-L21C25

and even has a `--bench` flag. 
https://github.com/bluealloy/revm/blob/b3f31fc6f2fa0c15f7287ac83ba608a8513ffef1/bins/revme/src/cmd/evmrunner.rs#L94-L103

JacekGlen:
> There is already a cli that allows you to run bytecode in `revme`: https://github.com/bluealloy/revm/blob/b3f31fc6f2fa0c15f7287ac83ba608a8513ffef1/bins/revme/src/cmd.rs#L21C5-L21C25
> 
> and even has a `--bench` flag.
> 
> https://github.com/bluealloy/revm/blob/b3f31fc6f2fa0c15f7287ac83ba608a8513ffef1/bins/revme/src/cmd/evmrunner.rs#L94-L103

I missed that, this is cool!

I'll update my PR to use `criterion` in revme, as suggested. This gives much more stable results.

JacekGlen:
> Wouldn't mind replacing it with `criterion`

Hi @rakita , I've updates the existing `revme` benchmark with Criterion. Have a look at the change now, please.

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -125,3 +123,36 @@ impl Cmd {
...
+    let evm = Evm::builder()
+        .with_db(BenchmarkDB::new_bytecode(Bytecode::new()))
+        .modify_tx_env(|tx| {
+            tx.caller = address!("1000000000000000000000000000000000000000");
+            tx.transact_to = TxKind::Call(address!("0000000000000000000000000000000000000000"));
+        })
+        .build();
+
+    let host = DummyHost::new(*evm.context.evm.env.clone());
+    let instruction_table = make_instruction_table::<DummyHost, ShanghaiSpec>();
```
rakita, line 137: Having Spec in cli would be best, but for now, putting the latest spec is what would I expect, or at least CancunSpec

JacekGlen: You are right! Updated to the latest spec

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -125,3 +123,36 @@ impl Cmd {
...
+fn run_benchmark(bytecode_str: Cow<str>) {
+    let evm = Evm::builder()
+        .with_db(BenchmarkDB::new_bytecode(Bytecode::new()))
+        .modify_tx_env(|tx| {
+            tx.caller = address!("1000000000000000000000000000000000000000");
+            tx.transact_to = TxKind::Call(address!("0000000000000000000000000000000000000000"));
+        })
+        .build();
+
+    let host = DummyHost::new(*evm.context.evm.env.clone());
```
rakita, line 136:
```suggestion
    let mut env = Env::default();
    env.tx.caller = address!("1000000000000000000000000000000000000000");
    env.tx.transact_to = TxKind::Call(address!("0000000000000000000000000000000000000000"));

    let host = DummyHost::new(env);
```
Dont need builder here.

bins/revme/src/cmd/evmrunner.rs
```diff
@@ -125,3 +123,36 @@ impl Cmd {
...
+        )),
+        ..Default::default()
+    };
+    let mut criterion = criterion::Criterion::default();
+
+    let mut criterion_group = criterion.benchmark_group("revme");
+    criterion_group.bench_function("bytecode", |b| {
+        b.iter(|| {
+            Interpreter::new(contract.clone(), u64::MAX, false).run(
+                EMPTY_SHARED_MEMORY,
```
rakita, line 151:
```suggestion
                SharedMemory::new(); 
```

Shared memory is preallocated and shared between multiple calls, so having SharedMemory::new() will have an initialization hit but it will be slightly faster on memory expansion.

rakita (commented):
left few nits. lgtm otherwise

rakita:
Have made a switch here https://github.com/bluealloy/revm/pull/2476

PR #1576, opened 2024-06-28
Use HandleOrRuntime to allow alloydb/ethersdb to hold a custom runtime

wtdcode:
Following advice from @DaniPopes, this PR introduce `HandleOrRuntime` for both `EthersDB` and `AlloyDB` to avoid creating runtime for every call. Compared to #1557, this implementation also allows synchronous code to continue using both implementation by calling `AlloyDB::with_runtime(..., ..., new_current_thread())`, which reduces break changes.

This PR shall address the concerns of @DaniPopes while keeping the compatibility.

- If we are already in a tokio runtime, like foundry, the handle will be perfectly valid, and no runtime is created.
- If we are in synchronous code, users are responsible for creating the runtime only _once_ for `with_runtime`.

Note with this PR, we could also support current thread runtime now.

### DIFF
--- a/crates/revm/src/db/alloydb.rs
+++ b/crates/revm/src/db/alloydb.rs
@@ -6,7 +6,8 @@ use alloy_eips::BlockId;
 use alloy_provider::{Network, Provider};
 use alloy_transport::{Transport, TransportError};
 use std::future::IntoFuture;
-use tokio::runtime::Handle;
+
+use super::utils::HandleOrRuntime;
 
 /// An alloy-powered REVM [Database].
 ///
@@ -18,28 +19,42 @@ pub struct AlloyDB<T: Transport + Clone, N: Network, P: Provider<T, N>> {
     /// The block number on which the queries will be based on.
     block_number: BlockId,
     /// handle to the tokio runtime
-    handle: Handle,
+    rt: HandleOrRuntime,
     _marker: std::marker::PhantomData<fn() -> (T, N)>,
 }
 
 impl<T: Transport + Clone, N: Network, P: Provider<T, N>> AlloyDB<T, N, P> {
-    /// Create a new AlloyDB instance, with a [Provider] and a block (Use None for latest).
+    /// Create a new AlloyDB instance, with a [Provider] and a block.
     ///
     /// Returns `None` if no tokio runtime is available or if the current runtime is a current-thread runtime.
     pub fn new(provider: P, block_number: BlockId) -> Option<Self> {
-        let handle = match Handle::try_current() {
+        let rt = match Handle::try_current() {
             Ok(handle) => match handle.runtime_flavor() {
                 tokio::runtime::RuntimeFlavor::CurrentThread => return None,
-                _ => handle,
+                _ => HandleOrRuntime::Handle(handle),
             },
             Err(_) => return None,
         };
         Some(Self {
             provider,
             block_number,
-            handle,
+            rt,
             _marker: std::marker::PhantomData,
         })
+    }
+
+    // Create a new AlloyDB instance, with a provider and a block and a runtime.
+    //
+    // Refer to [tokio::runtime::Builder] how to create a runtime if you are in synchronous world.
+    // If you are already using something like [tokio::main], call AlloyDB::new instead.
+    pub fn with_runtime(provider: P, block_number: BlockId, runtime: Runtime) -> Self {
+        let rt = HandleOrRuntime::Runtime(runtime);
+        Self {
+            provider,
+            block_number,
+            rt,
+            _marker: std::marker::PhantomData,
+        }
     }
 
     /// Internal utility function that allows us to block on a future regardless of the runtime flavor.
@@ -49,7 +64,7 @@ impl<T: Transport + Clone, N: Network, P: Provider<T, N>> AlloyDB<T, N, P> {
         F: std::future::Future + Send,
         F::Output: Send,
     {
-        tokio::task::block_in_place(move || self.handle.block_on(f))
+        self.rt.block_on(f)
     }
 
     /// Set the block number on which the queries will be based on.
--- a/crates/revm/src/db/alloydb.rs
+++ b/crates/revm/src/db/alloydb.rs
@@ -6,13 +6,14 @@ use alloy_eips::BlockId;
 use alloy_provider::{Network, Provider};
 use alloy_transport::{Transport, TransportError};
 use std::future::IntoFuture;
+use tokio::runtime::{Handle, Runtime};
 
 use super::utils::HandleOrRuntime;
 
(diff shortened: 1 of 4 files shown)


wtdcode:
@rakita Can we have this before #1574 ?

crates/revm/src/db/utils.rs
```diff
@@ -0,0 +1,22 @@
...
+impl HandleOrRuntime {
+    #[inline]
+    pub(crate) fn block_on<F>(&self, f: F) -> F::Output
+    where
+        F: std::future::Future + Send,
+        F::Output: Send,
+    {
+        match self {
+            Self::Handle(handle) => tokio::task::block_in_place(move || handle.block_on(f)),
+            Self::Runtime(rt) => rt.block_on(f),
```
mattsse, line 19: this will panic if called within async execution context

crates/revm/src/db/ethersdb.rs
```diff
@@ -31,29 +33,51 @@ impl<M: Middleware> EthersDB<M> {
...
     }
 
+    // Create a new EthersDB instance, with a provider and a block (None for latest) and a runtime.
+    //
+    // Refer to [tokio::runtime::Builder] how to create a runtime if you are in synchronous world.
+    // If you are already using something like [tokio::main], call EthersDB::new instead.
+    pub fn with_runtime(
+        client: Arc<M>,
+        block_number: Option<BlockId>,
+        runtime: Runtime,
```
mattsse, line 58: so the caller is responsible for creating a new runtime?
in which case they could also pass in the handle?

crates/revm/src/db/utils.rs
wtdcode: Yes, that’s intended. Passing a runtime mostly would be the sync code. If user is within async execution context, they should use new directly instead.

crates/revm/src/db/ethersdb.rs
wtdcode: Make sense, we can add a new function and document the behavior.

crates/revm/src/db/utils.rs
wtdcode: As stated above, if user is in async context, they should call new as #1557 suggests. If user is in sync context, they should provide a runtime, mostly current_thread runtime.

If user intends to mix different contexts, they should be careful by themselves as it is known to be bad practice and cause problems hard to tackle.

crates/revm/src/db/alloydb.rs
```diff
@@ -6,50 +6,80 @@ use alloy_eips::BlockId;
...
+            block_number,
+            rt,
+            _marker: std::marker::PhantomData,
+        }
+    }
+
+    // Create a new AlloyDB instance, with a provider and a block and a runtime handle.
+    //
+    // Refer to [tokio::runtime::Builder] how to create a runtime if you are in synchronous world.
+    // If you are already using something like [tokio::main], call AlloyDB::new instead.
```
DaniPopes, line 64: these should be doc comments

DaniPopes (commented):
Makes sense

crates/revm/src/db/alloydb.rs
wtdcode: Thanks for pointing out. Done =)

DaniPopes: I mean you need 3 slashes

wtdcode: Done

rakita (approved):
lgtm

crates/revm/src/db/ethersdb.rs
```diff
@@ -2,14 +2,14 @@ use std::sync::Arc;
...
 use ethers_providers::Middleware;
-use tokio::runtime::Handle;
+use tokio::runtime::{Handle, Runtime};
 
 use crate::primitives::{AccountInfo, Address, Bytecode, B256, U256};
 use crate::{Database, DatabaseRef};
 
 use super::utils::HandleOrRuntime;
 
-#[derive(Debug, Clone)]
```
jafar75, line 10: @wtdcode would you explain why Clone has been removed? I'm slightly new in Rust and want to learn the technical concept behind this decision

wtdcode: Because runtime doesn’t implement Clone and it doesn’t make too much sense cloning the db. Or, you could wrap it with an Rc or Arc.
What’s your use case?

jafar75: thanks
In my case, I define an ethersDB instance and by using it, I define a cacheDB. I add some base data (deploy contract, change some slots, etc) to cacheDB and after that, by cloning it, I send the same copies to multiple threads. 

For now, I get an error (when cloning cacheDB) complains about ethersDB (as internal object of cacheDB) cannot be cloned. 

because inside separate threads, I need to write to db, I tested not cloning and use Arc/Mutex. but the performance was worse than cloning.

wtdcode: > thanks In my case, I define an ethersDB instance and by using it, I define a cacheDB. I add some base data (deploy contract, change some slots, etc) to cacheDB and after that, by cloning it, I send the same copies to multiple threads.
> 
> For now, I get an error (when cloning cacheDB) complains about ethersDB (as internal object of cacheDB) cannot be cloned.
> 
> because inside separate threads, I need to write to db, I tested not cloning and use Arc/Mutex. but the performance was worse than cloning.

It's not really make sense to clone `AlloyDB` or `EthersDB`. `Clone` was also added by me in a previous PR, but I recently found it was a mistake. Generally, it is due to:

1. Both `Provider` and `Middleware` are only safe to clone within the same runtime. If you send it across runtimes (or threads), many internal things and assumptions could break.
2. Even if ensure you are in the same async context all the time or just sync code, it just doesn't make sense to clone the EthersDB because it saves nothing.

To your use case which is similar to mine, my suggestion is:

1. Share a single `Rc<Mutex<CacheDB<EthersDB>>>` across all threads. This avoids duplicate caching for each thread and can speed up your application overall.
2. Clone `CacheDB` members (note they are public) except the `EthersDB` and create a new `EthersDB` instead. This generously is a manual "Clone" implementation.

jafar75: thanks, I have a challenge in using arc<mutex>.



in order to create evm and then transact(), I must do lock. but, about all of my perf. overhead is related to these steps, so it convert my program in a literally single-threaded system.

wtdcode: Note `with_db` accepts borrows.

jafar75: you mean `with_ref_db` ?  I want to update the cache in each `transact`

PR #1588, opened 2024-07-08
chore(revm): defer bytecode load

lightsing:
Current revm try to load bytecode just after depth check, but there are some conditions the code is not needed:
- insufficient balance to transfer
- call to a precompile address

As long as the code is not needed, the database provider may not have corresponding code, like in our implementation scroll-tech/stateless-block-verifier#15 use this as an optimization.

### DIFF
--- a/crates/revm/src/context/evm_context.rs
+++ b/crates/revm/src/context/evm_context.rs
@@ -166,21 +166,6 @@ impl<DB: Database> EvmContext<DB> {
             return return_result(InstructionResult::CallTooDeep);
         }
 
-        let (account, _) = self
-            .inner
-            .journaled_state
-            .load_code(inputs.bytecode_address, &mut self.inner.db)?;
-
-        let code_hash = account.info.code_hash();
-        let bytecode = account.info.code.clone().unwrap_or_default();
-
-        // ExtDelegateCall is not allowed to call non-EOF contracts.
-        if inputs.scheme.is_ext_delegate_call()
-            && bytecode.bytes_slice().get(..2) != Some(&EOF_MAGIC_BYTES)
-        {
-            return return_result(InstructionResult::InvalidExtDelegateCallTarget);
-        }
-
         // Create subroutine checkpoint
         let checkpoint = self.journaled_state.checkpoint();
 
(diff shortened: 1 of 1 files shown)


rakita:
Looks reasonable, will check it again to be sure

crates/revm/src/context/evm_context.rs
```diff
@@ -216,18 +201,39 @@ impl<DB: Database> EvmContext<DB> {
...
+                // Create interpreter and executes call and push new CallStackFrame.
+                Ok(FrameOrResult::new_call_frame(
+                    inputs.return_memory_offset.clone(),
+                    checkpoint,
+                    Interpreter::new(contract, gas.limit(), inputs.is_static),
+                ))
+            } else {
+                self.journaled_state.checkpoint_commit();
+                return_result(InstructionResult::Stop)
+            }
```
rakita, line 236:
```suggestion
            if bytecode.is_empty() {
                 self.journaled_state.checkpoint_commit();
                return_result(InstructionResult::Stop)
            }
            let contract = Contract::new_with_context(
                inputs.input.clone(),
                bytecode,
                Some(code_hash),
                inputs,
            );
            // Create interpreter and executes call and push new CallStackFrame.
            Ok(FrameOrResult::new_call_frame(
                inputs.return_memory_offset.clone(),
                checkpoint,
                Interpreter::new(contract, gas.limit(), inputs.is_static),
            ))
```

code style nit. Maybe it needs fmt

crates/revm/src/context/evm_context.rs
```diff
@@ -166,21 +166,6 @@ impl<DB: Database> EvmContext<DB> {
             return return_result(InstructionResult::CallTooDeep);
         }
 
-        let (account, _) = self
-            .inner
-            .journaled_state
-            .load_code(inputs.bytecode_address, &mut self.inner.db)?;
```
rakita, line 172: This is slightly an edge case that is not a consensus error but would fix it to be consistent. CALL from instruction loads account that makes account warm loaded, while if we do it from tx as in `TransactTo` this is not the case, this is really a edge case where it is noticeable only that account will not be returned from finalization and only if tx is calling precompile.

We can add here `load_account` in place of `load_code`, something like `let _ = self...load_acccount(..)?`

rakita (changes requested):
Left comment, and one nit

rakita (approved):
lgtm

PR #1601, opened 2024-07-10
fix: missing kzg_settings if kzg-rs feature enabled

jtguibas:
Makes sure `kzg_settings` is still in `CfgEnv` even if the `kzg-rs` feature is used.


rakita (approved):
lgtm

PR #1603, opened 2024-07-11
test(`GeneralState`): skip fewer specs

hai-rise:
- `SpecName::Unknown` previously only covered `Paris`, which should be mapped to `SpecId::MERGE` instead of `Merge` (not used in `GeneralStateTests`).
- Removed the skip on `SpecName::ByzantiumToConstantinopleAt5` as its tests are passing.
- Left a comment on why we skip `Constantinople`.


rakita (approved):
lgtm

PR #1605, opened 2024-07-11
fix(eip7702): Add tests and fix some bugs

rakita:
Few things added, still WIP:
* New temp command for eof validation
* Decoding of eip7702 tx for statetests (Authotization is not included as json field)
* Fix wrong gas per auth (It should be 2500)
* Fix: get code from `authorization.address` and not from `authority`


wtdcode:
Current revm 11.0 is conflicting with alloy-eips 0.1.5 (which includes https://github.com/alloy-rs/alloy/commit/5d8dce4b542a069d0aa2973e7705309422311de6) because of lacking this PR.

mattsse:
yanked the releases, should be done as minor, not patch

rakita:
Will merge it with alloy patches and we can make additional PR that uses crates version

PR #1606, opened 2024-07-12
fix(EOF): Use cfg code size limit for eofcreate

rakita:
Respect cfg code size for eofcreate output


crates/revm/src/context/inner_evm_context.rs
```diff
@@ -125,6 +125,11 @@ impl<DB: Database> InnerEvmContext<DB> {
         &mut self.env
     }
 
+    /// Returns referecne to [`CfgEnv`].
```
rakita, line 128:
```suggestion
    /// Returns reference to [`CfgEnv`].
```

DaniPopes (approved):

PR #1607, opened 2024-07-12
feat(EOF): Bytecode::new_raw supports EOF, new_raw_checked added

rakita:
`Bytecode::new_raw` now support EOF bytecode and will panic on invalid EOF code. In most cases, the validity of EOF is not questioned as validity is checked inside EVM.

An alternative safe function `new_raw_checked` was added that returns `EofDecodeError` if an invalid EOF bytecode can happen.

`Bytecode::new_legacy` to force legacy bytecode.

### DIFF
--- a/bins/revme/src/cmd/evmrunner.rs
+++ b/bins/revme/src/cmd/evmrunner.rs
@@ -24,21 +24,9 @@ pub enum Errors {
     #[error("EVM Error")]
     EVMError,
     #[error(transparent)]
-    Io(IoError),
+    Io(#[from] IoError),
     #[error(transparent)]
-    EofError(EofDecodeError),
-}
-
-impl From<EofDecodeError> for Errors {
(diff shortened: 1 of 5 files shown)


bins/revme/src/cmd/evmrunner.rs
```diff
@@ -25,6 +25,14 @@ pub enum Errors {
     EVMError,
     #[error(transparent)]
     Io(IoError),
+    #[error(transparent)]
+    EofError(EofDecodeError),
+}
+
+impl From<EofDecodeError> for Errors {
```
DaniPopes, line 32: i believe you can do EofError(#[from] EofDecodeError) in the enum

DaniPopes (approved):

bins/revme/src/cmd/evmrunner.rs
rakita: That is better

PR #1608, opened 2024-07-12
fix: allow non-static lifetime in HandleRegisterBox

Wodann:
When using `append_handler_register_box` Rust was always assigning the static lifetime to the `Box`, which is the default behaviour of Rust when a lifetime is not specified.

This change allows custom lifetimes for the box.


DaniPopes (approved):

rakita (approved):
lgtm

PR #1623, opened 2024-07-16
chore: bump alloy deps

rakita:
Backported the `RecoveredAuthorization` to minimize breaking changes.

Another option considered is to have: `Recovered(Vec<Option<RecoveredAuthorization>>), in place of `Recovered(Vec<RecoveredAuthorization>),` but this loses information on invalid Authorization that some hooks/handler/inspector maybe want to have.


mattsse (approved):
lgtm

PR #1627, opened 2024-07-16
fix(eof): don't fail if offset overflows `usize` in `RETURNDATALOAD`

DaniPopes:
As per the spec:
> if offset + 32 > len(returndata buffer) the result is zero-padded (same behavior as CALLDATALOAD)

Doesn't matter if 1 above or 2^255 above the length, it's always zero-padded


rakita (approved):
good catch

PR #1632, opened 2024-07-17
perf(eof): avoid some allocations

DaniPopes:
Plus some doc cleanups


crates/primitives/src/bytecode/eof.rs
```diff
@@ -21,13 +21,9 @@ pub const EOF_MAGIC: u16 = 0xEF00;
 /// EOF magic number in array form.
 pub static EOF_MAGIC_BYTES: Bytes = bytes!("ef00");
 
-/// EOF - Ethereum Object Format.
+/// Ethereum Object Format (EOF) container.
```
rakita, line 24:
```suggestion
/// EVM Object Format (EOF) container.
```

rakita (approved):

PR #1635, opened 2024-07-17
chore: simplify Inspector function

jackwener:
We don't need `let _`. We just use `_` + param is enough.

### DIFF
--- a/crates/revm/src/inspector.rs
+++ b/crates/revm/src/inspector.rs
@@ -28,17 +28,14 @@ pub mod inspectors {
 }
 
 /// EVM [Interpreter] callbacks.
-#[auto_impl(&mut, Box)]
+#[auto_impl(& mut, Box)]
 pub trait Inspector<DB: Database> {
     /// Called before the interpreter is initialized.
     ///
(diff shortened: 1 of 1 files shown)


crates/revm/src/inspector.rs
```diff
@@ -49,26 +46,18 @@ pub trait Inspector<DB: Database> {
...
-    }
+    fn step_end(&mut self, _interp: &mut Interpreter, _context: &mut EvmContext<DB>) {}
 
     /// Called when a log is emitted.
     #[inline]
-    fn log(&mut self, interp: &mut Interpreter, context: &mut EvmContext<DB>, log: &Log) {
-        let _ = interp;
-        let _ = context;
+    fn log(&mut self, _interp: &mut Interpreter, _context: &mut EvmContext<DB>, log: &Log) {
         let _ = log;
```
mattsse, line 61: can also remove this?

mattsse (approved):
iirc this was done, so implementers don't have to first rename the args, I think this is additional effort is negligible

rakita (commented):
This was done so peeople dont need to rename args

PR #1638, opened 2024-07-17
refactor: use `is_zero` for `U256` and `B256`

tcoratger:


crates/revm/src/context/evm_context.rs
```diff
@@ -2,6 +2,8 @@ use revm_interpreter::CallValue;
 use revm_precompile::PrecompileErrors;
 
 use super::inner_evm_context::InnerEvmContext;
+#[cfg(any(test, feature = "test-utils"))]
+use crate::primitives::U256;
 use crate::{
```
Rjected, line 7: why did this need to move?

Rjected (commented):
looks good, I just have one question about the U256 import moving

crates/revm/src/context/evm_context.rs
tcoratger: Because it is not used anymore in the main code but still being used under this cfg so that if we don't move it, it will trigger a clippy lint warning

crates/interpreter/src/instructions/arithmetic.rs
```diff
@@ -26,7 +26,7 @@ pub fn sub<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
 pub fn div<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H) {
     gas!(interpreter, gas::LOW);
     pop_top!(interpreter, op1, op2);
-    if *op2 != U256::ZERO {
+    if !(*op2).is_zero() {
```
DaniPopes, line 29:
```suggestion
    if !op2.is_zero() {
```

DaniPopes (commented):
and everywhere else
U256: Copy so auto deref takes place

tcoratger:
> and everywhere else U256: Copy so auto deref takes place

Oups yes just fixed

crates/revm/src/context/evm_context.rs
Rjected: ah I see

DaniPopes (approved):

crates/revm/src/context/evm_context.rs
rakita: Would move it inside `mod test-utils` as it is only used there.

rakita (approved):
one nit on `use crate::primitives::U256;` otherwise lgtm

crates/revm/src/context/evm_context.rs
tcoratger: This is also used in `mod tests` so that if I move the import inside `mod test-utils`, I've to do another equivalent import inside `mod tests` too. But as I just saw that it was done like that for other similar imports, I just changed the position, tell me what you think?

mattsse (approved):

PR #1643, opened 2024-07-18
feat: use batch bn256 pair operation

Rjected:
We are currently not taking advantage of the batch pair operation from the `bn` library for the pairing check precompile.

This yields a ~27% speedup on the existing bench:
```
Crypto Precompile benchmarks/precompile bench | ecpairing precompile
                        time:   [2.2389 ms 2.2441 ms 2.2495 ms]
                        change: [-27.689% -27.469% -27.227%] (p = 0.00 < 0.05)
                        Performance has improved.
```


DaniPopes (approved):
Nice! LGTM

rakita (approved):
lgtm.

PR #1645, opened 2024-07-19
Maybe it is PraGue? :D

ioterw:
Renamed some cool city a bit.


rakita (approved):
Still need to visit it :D

PR #1655, opened 2024-07-24
deduplication for COPY operations

### LINKED ISSUE #1637: Factor out `*COPY` common code
All `*COPY` operations have the same inputs, execute the same operation, and have the same gas cost (except for `EXTCODECOPY`); the only difference is which buffer they copy from. This code can be factored out for less code duplication.

harsh-ps-2003:
Ref #1637 

It reduces code duplication by creating a generic `copy_cost` function. It maintains the special logic for `extcodecopy_cost` while still using the generic function.

It will make adding new COPY operations easier in the future, as they can all use the same copy_cost function.

Also adds `copy_to_memory` method for the COPY opcodes to remove code duplication.


harsh-ps-2003:
can you take a look @DaniPopes!

DaniPopes:
This is fine but doesnt fix the issue, there is also the opcode implementation to dedup

harsh-ps-2003:
@DaniPopes what's up with this clippy thing, I didn't even touch pairings or utils! when I run `cargo clippy` only thing I get is `inline-const is experimental`! sorry, first time contributor :( Any dedup that I missed? Thanks for the patience

DaniPopes:
It's due to Rust 1.80 just being released, it should be fixed on main with https://github.com/bluealloy/revm/pull/1661

harsh-ps-2003:
So my PR is good for merge, once things get fixed? @DaniPopes

crates/interpreter/src/instructions/system.rs
```diff
@@ -38,24 +38,12 @@ pub fn codesize<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &mut H)
...
     // Inform the optimizer that the bytecode cannot be EOF to remove a bounds check.
     assume!(!interpreter.contract.bytecode.is_eof());
-    // Note: this can't panic because we resized memory to fit.
-    interpreter.shared_memory.set_data(
-        memory_offset,
-        code_offset,
-        len,
-        interpreter.contract.bytecode.original_byte_slice(),
-    );
+    let source = interpreter.contract.bytecode.original_byte_slice().to_vec();
```
DaniPopes, line 45: These `to_vec` clones can be removed by using raw pointers. Unfortunately for performance we have to use them to get around the aliasing issue with `&mut Interpreter`

harsh-ps-2003:
@DaniPopes The failing tests are all in the `optimism::fast_lz::tests::test_flz_native_evm_parity`rather than the core EVM instructions we modified in `system.rs`!

crates/interpreter/src/instructions/system.rs
```diff
@@ -179,6 +139,32 @@ pub fn returndataload<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &m
     *offset = B256::from(output).into();
 }
 
+// common logic for copying data from a source buffer to the EVM's memory
+fn copy_to_memory(
+    interpreter: &mut Interpreter,
+    memory_offset: U256,
+    data_offset: usize,
+    len: usize,
+    source: *const u8,
```
DaniPopes, line 148: Source's length is not `len`, this should be `*const [u8]`. This is why tests are failing

harsh-ps-2003:
Looks good? @DaniPopes

DaniPopes (approved):
Thanks, looks fine, will leave it up to @rakita

crates/interpreter/src/instructions/system.rs
```diff
@@ -179,6 +157,29 @@ pub fn returndataload<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &m
...
 }
 
+// common logic for copying data from a source buffer to the EVM's memory
+fn copy_to_memory(
+    interpreter: &mut Interpreter,
+    memory_offset: U256,
+    data_offset: usize,
+    len: usize,
+    source: *const [u8],
+) {
```
rakita, line 167: I would return a true/false here. Otherwise function silently fails, and can work only if it is in last line of instruction.

crates/interpreter/src/instructions/system.rs
```diff
@@ -179,6 +157,29 @@ pub fn returndataload<H: Host + ?Sized>(interpreter: &mut Interpreter, _host: &m
...
+    let memory_offset = as_usize_or_fail!(interpreter, memory_offset);
+    resize_memory!(interpreter, memory_offset, len);
+
+    // Note: this can't panic because we resized memory to fit.
+    unsafe {
+        interpreter
+            .shared_memory
+            .set_data(memory_offset, data_offset, len, &*source);
+    }
+}
```
rakita, line 181: I would remove this, and change behaviour of function to do gas, len check, memory resize.

As memory_offset is reused, we can return that from this function as `Option<usize>`

harsh-ps-2003: The problem with changing the functions behaviour as suggested by you is that we encountering is a classic case of Rust's borrow checker preventing simultaneous mutable and immutable borrows of the same data. 
When we do :
```rust
let source = interpreter.contract.input.as_ref();
    let memory_offset = copy_to_memory(interpreter, memory_offset, data_offset, len);
    if let Some(memory_offset) = memory_offset {
        // Note: this can't panic because we resized memory to fit.
        interpreter.shared_memory.set_data(memory_offset, data_offset, len, source);
    };
```
We create an immutable borrow of interpreter.contract.input and then we try to mutably borrow interpreter in the copy_to_memory function call. Finally, we try to use the immutable borrow source again in the set_data call.
The Rust borrow checker doesn't allow this because it can't guarantee that the mutable borrow in copy_to_memory won't invalidate the source reference.

How do you suggest I tackle this! @rakita

rakita: can this line: `let source = interpreter.contract.input.as_ref();` be moved inside `if let Some(memory_..` so you don't have a borrow?

harsh-ps-2003: Now when I return `Option<usize>` I get `expected enum `std::option::Option<usize>`
found unit type `()`` from the gas! macro. Thats why the tests are failing. How do you suggest I tackle this! @rakita

rakita: * There is a gas! version that takes Return. `gas!(10,None)` something similar should be added for `gas_or_fail!`
* Change `as_usize_or_fail` to `as_usize_or_fail_ret!(interpreter, memory_offset,None);`
* Change `resize_memory!` to `resize_memory!(interpreter, memory_offset, len, None);`

rakita:
Superseded by #1799

PR #1660, opened 2024-07-25
feat(EOF): Add EOF validation in revme bytecode cmd

rakita:
Show validation output after EOF bytecode gets decoded


rakita:
Error not related

PR #1675, opened 2024-08-01
fix(eip7702): Trigger selfdestruct if Loaded to empty change happens

rakita:
EIP-7702 introduces a new transition that an already Changed account can become empty after the transaction finishes. In this case, we have an empty account that is touched and the assumption is that it needs to be destroyed.

### DIFF
--- a/crates/revm/src/db/states/cache.rs
+++ b/crates/revm/src/db/states/cache.rs
@@ -1,5 +1,6 @@
 use super::{
-    plain_account::PlainStorage, transition_account::TransitionAccount, CacheAccount, PlainAccount,
+    plain_account::PlainStorage, transition_account::TransitionAccount, AccountStatus,
+    CacheAccount, PlainAccount,
 };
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
tested on devnet2 and this works

crates/revm/src/db/states/cache.rs
```diff
@@ -124,6 +125,16 @@ impl CacheState {
         let is_created = account.is_created();
         let is_empty = account.is_empty();
 
+        // EIP-7702 possible to have Changed/Loaded state that goes to empty state.
```
rakita, line 128:
```suggestion
        // first version of EIP-7702 makes it possible to have Changed/Loaded state that goes to empty state.
        // If Delegated contract sends all balance to other account.
```

rakita:
Closing this as new EIP-7702 version makes it obsolete https://github.com/bluealloy/revm/pull/1695

PR #1677, opened 2024-08-02
Add EOF Layout Fuzz Loop to `revme bytecode`

shemnon:
Update `revme` so that when `bytecode` is called without arguments it goes into the standard


rakita (approved):

rakita:
@shemnon I added a `--eof-initcode` and `--eof-runtime` flags for `bytecode` command, and refactored a code a little

PR #1686, opened 2024-08-08
Check for typos in CI

adria0:
If you found it interesting, this adds typo check in CI.

To run it manually:

`cargo install typos-cli` then

`typos` ( if you are ok with the suggestions, changes can be automatically written with `typos -w` )


rakita (approved):
lgtm

PR #1694, opened 2024-08-09
chore: use alloy type instead of re-defining

roynalnaruto:
The [`RecoveredAuthorization`](https://github.com/alloy-rs/alloy/blob/main/crates/eips/src/eip7702/auth_list.rs#L233) type is already defined in alloy-eips, the only difference being that the `into_parts` method returns a tuple of `(Authorization, RecoveredAuthority)` instead of `(Authorization, Option<Address>)`. However, the `RecoveredAuthority` itself exposes an `address(&self)` method that returns the `Option<Address>`.


rakita:
This is not great dependency as RecoveredAuthorization is made for Revm. So it should be inside the code that uses it.

PR #1713, opened 2024-08-20
deps: bump `derive_more` to 1.0

tcoratger:
Related https://github.com/paradigmxyz/reth/issues/10400


DaniPopes (approved):
derive_more dup will be removed with next alloy release
other dups come from reqwest dup which is due to ethers existing

rakita (approved):
lgtm

PR #1695, opened 2024-08-09
feat(eip7702): Impl newest version of EIP 

rakita:
Implementation of EIP-7702,

* Added StateLoad types to Host to structure is_cold field.
* Added new handler or eip7702 load
* Split refund logic from `last_frame_handler` to its own handler `post_exec.refund` it is more aligned with handling those outside of execution. Optimism handler updated
* Added last test suite for bls and eip7702 EIPs.

PR #1725, opened 2024-08-27
chore(deps): bump alloy and primitives

klkvr:
bumps alloy patch and alloy-primitives in https://github.com/bluealloy/revm/pull/1695 without changing dependency structure

we can do this later, I'll be just using this PR for consolidating changes in https://github.com/paradigmxyz/reth/pull/10537


Cargo.toml
```diff
@@ -24,4 +24,6 @@ inherits = "test"
 opt-level = 3
 
 [patch.crates-io]
-alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "fd159f6" }
\ No newline at end of file
+alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "05cc4e0" }
+alloy-provider = { git = "https://github.com/alloy-rs/alloy", rev = "05cc4e0" }
+alloy-transport = { git = "https://github.com/alloy-rs/alloy", rev = "05cc4e0" }
```
rakita, line 29: What about this version?

klkvr: > We should remove this patch right? And just go to latest version?

yep, removed patches in 84839ba

Issue #1724, opened 2024-08-27
EOFCREATE order of args


rakita:
ref: https://github.com/ipsilon/eof/pull/156

rakita:
Will gather all (small) changes through main EOF task: https://github.com/bluealloy/revm/issues/2012

PR #1732, opened 2024-08-29
eofcreate order of args

nkysg:
Closes https://github.com/bluealloy/revm/issues/1724


rakita:
This change is still pending inclusion in spec, the task made was a reminder for it.

PR #1733, opened 2024-08-29
 bump to newest eof eest tests pr-756

rakita:
Bump to this PR: https://github.com/ethereum/execution-spec-tests/pull/756

Generated by Besu. ty @shemnon


shemnon:
actually, generated by evmone.  I use both and diff regularly in EEST.

rakita:
> actually, generated by evmone. I use both and diff regularly in EEST.

Ah, i see the command now, would need to get myself a similar setup

PR #1746, opened 2024-09-04
fix(inspector): always call selfdestruct if entry is made

jsvisa:
fix https://github.com/paradigmxyz/reth/issues/10554

If EIP-6780 is enabled and selfdestruct is called not in the same create transaction, then no `AccountDestroyed` event will be returned: 
https://github.com/bluealloy/revm/blob/cabd1d1abbe9a1c58ded7316f3f4756eaacba559/crates/revm/src/journaled_state.rs#L510-L519

but actually in that case we also need to log the selfdestruct event.


rakita (approved):
good find!

PR #1748, opened 2024-09-04
Add test showing bug in calculation of l1 gas used for ecotone tx

emhane:
Adds test showing buggy computation of l1 gas used for ecotone transaction


rakita (approved):
lgtm

PR #1755, opened 2024-09-06
chore: make clippy happy

jsvisa:
when running with `cargo check --release` with warning as below:

```
warning: elided lifetime has a name
  --> crates/revm/src/context/context_precompiles.rs:98:72
   |
98 |     pub fn addresses<'a>(&'a self) -> Box<dyn ExactSizeIterator<Item = &Address> + 'a> {
   |                      -- lifetime `'a` declared here                    ^ this elided lifetime gets resolved as `'a`
   |
   = note: `#[warn(elided_named_lifetimes)]` on by default
```


DaniPopes (approved):

rakita (approved):
Thanks!

PR #1756, opened 2024-09-06
chore(revm): expose convenience cache state for preimages

0x00101010:
## Description

expose cache state for getting preimages, its initial user will be debug_execution_witness in reth. Reference PR here: https://github.com/paradigmxyz/reth/pull/10736/files

### DIFF
--- a/crates/interpreter/src/function_stack.rs
+++ b/crates/interpreter/src/function_stack.rs
@@ -51,9 +51,8 @@ impl FunctionStack {
 
     /// Pops a frame from the stack and sets current_code_idx to the popped frame's idx.
     pub fn pop(&mut self) -> Option<FunctionReturnFrame> {
-        self.return_stack.pop().map(|frame| {
+        self.return_stack.pop().inspect(|frame| {
             self.current_code_idx = frame.idx;
-            frame
         })
     }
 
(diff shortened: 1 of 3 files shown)


crates/interpreter/src/function_stack.rs
```diff
@@ -51,9 +51,8 @@ impl FunctionStack {
 
     /// Pops a frame from the stack and sets current_code_idx to the popped frame's idx.
     pub fn pop(&mut self) -> Option<FunctionReturnFrame> {
-        self.return_stack.pop().map(|frame| {
+        self.return_stack.pop().inspect(|frame| {
```
0x00101010, line 54: CI clippy error fixes

rakita:
I would rather not have this function as State would show wrong changes if CacheDB is taken without taking bundle state.

0x00101010:
> I would rather not have this function as State would show wrong changes if CacheDB is taken without taking bundle state.

Thanks for the review @rakita!

Any suggestions on how to take it out? Would it better to directly do it like https://github.com/paradigmxyz/reth/pull/10736/files#diff-7fbb0c1e65c92e8879aa308fee2538af673a2cd6d70f9b5b1103cf4d6554604dR378 or create a function called take_bundle_and_cache?

The background story here is that in order to implement `debug_execution_witness` API, we'll need to get CacheDB out so that we can retrieve preimages properly

Issue #1728, opened 2024-08-28
Use TestAuthorization and skip decoding of eip7702 tx


rakita:
In this PR we have decoded tx bytes to extract AuthorizationList.

a better way would be to use values from json as we are already parsing them, it is needed to cast them to RecoveredAuthorization and test/debug this.

royvardhan:
Hi @rakita I am down to attempt this but have some questions: 
1. I can read these tests at alloy-eips: https://github.com/alloy-rs/eips/blob/3136c2883e8d59cd1ac5b004dffb8bc826f0fc03/crates/eip7702/src/auth_list.rs#L302, would you like me to extend it with extracting values directing from json and then assert them?
2. Or would you like to add new tests in https://github.com/bluealloy/revm/blob/main/crates/primitives/src/eip7702/authorization_list.rs where we parse values from json and test them?

royvardhan:
@rakita Awaiting your response, thanks!

rakita:
We should remove this file: https://github.com/bluealloy/revm/blob/main/bins/revme/src/cmd/statetest/models/eip7702.rs
And use this struck: https://github.com/bluealloy/revm/blob/d38ff452714e1c4a17cf0fbdaa4913f6d61033c4/bins/revme/src/cmd/statetest/models/mod.rs#L139

PR #1757, opened 2024-09-07
refactor: use TestAuthorization and skip decoding of eip7702 tx

royvardhan:
Closes: https://github.com/bluealloy/revm/issues/1728


bins/revme/src/cmd/statetest/models/mod.rs
```diff
@@ -146,6 +128,29 @@ pub struct Authorization {
     signer: Option<Address>,
 }
 
+impl TestAuthorization {
+    pub fn signature(&self) -> Signature {
+        let parity = self.v != U256::ZERO;
```
rakita, line 133: This is not correct and `v` should stay as a big number.

rakita: Cast `v` to u64 and construct with `Parity::try_from(v)`, this parity is going to be checked inside Revm.

royvardhan: Getting overflow error trying to cast v to u64

rakita: Cast with `try_from` and use `unwrap_or(u64::MAX)` we don't care how big the number is but just that it is 0 or 1.

royvardhan: https://github.com/bluealloy/revm/actions/runs/10821550430/job/30023779288?pr=1757

royvardhan: using `u64::MAX` is literally assigned the max value of `u64` after having unwrap err, which is causing the `Parity` to trigger invalid parity error.

royvardhan: My bad: It is triggering `InvalidParity(value)` with value being `2`.

bins/revme/src/cmd/statetest/models/mod.rs
```diff
@@ -146,6 +127,30 @@ pub struct Authorization {
     signer: Option<Address>,
 }
 
+impl TestAuthorization {
+    pub fn signature(&self) -> Signature {
+        let v = u64::try_from(self.v).unwrap_or(u64::MAX);
+        let parity = Parity::try_from(v).unwrap();
```
rakita, line 133: You should do `unwrap_or(Parity::Eip155(36))` it will pass as valid Parity by Signature struct but it will be considered invalid in runtime consensus check in Revm as it expects `0/1` (And this is what this test is testing).

bins/revme/src/cmd/statetest/runner.rs
```diff
@@ -372,10 +372,17 @@ pub fn execute_test_suite(
                     .and_then(Option::as_deref)
                     .cloned()
                     .unwrap_or_default();
-                let Ok(auth_list) = test.eip7702_authorization_list() else {
+
+                if unit.transaction.authorization_list.is_empty() {
```
rakita, line 376: This is not good, it would skip all tests that do not have authorization_list

bins/revme/src/cmd/statetest/models/mod.rs
```diff
@@ -128,15 +109,15 @@ pub struct TransactionParts {
     #[serde(default)]
     pub access_lists: Vec<Option<AccessList>>,
     #[serde(default)]
-    pub authorization_list: Vec<Authorization>,
+    pub authorization_list: Vec<TestAuthorization>,
```
rakita, line 112: Can you try if parsing passes with removal of `#[serde(default)]` and with field looking like `Option<Vec<TestAuthorization>>`

bins/revme/src/cmd/statetest/runner.rs
royvardhan: So the current continue statement is not skipping the test which do not have authorization_list? https://github.com/bluealloy/revm/blob/9ded6908d7e81823f309251b7eb46513dbbe17b7/bins/revme/src/cmd/statetest/runner.rs#L376

bins/revme/src/cmd/statetest/models/mod.rs
royvardhan: Okay I made it optional and this is what the runner.rs snippet code is like, 
image
it still has the `continue` statement because the original code does in the main branch, Let me know if you wanted something else.

bins/revme/src/cmd/statetest/runner.rs
rakita: There are tests that does not have auth list, those should not be skipped but `env.tx.authorization_list` should be set to None

bins/revme/src/cmd/statetest/models/mod.rs
rakita: It should be something like `env.tx.authorization_list = auth.map(|list| list.iter().map(|auth| auth.into_recovered()).collect());`

royvardhan:
New pr, same work: https://github.com/bluealloy/revm/pull/1785
Had to close this because of new revm release.

bins/revme/src/cmd/statetest/runner.rs
royvardhan: thanks, applied changes!

PR #1767, opened 2024-09-13
feat(example): deploy bytecode from scratch

clearloop:
Introduces an example about deploying raw bytecode with revm with storage operations as example, it includes:

- [x] set up a basic REVM from scratch
- [x] concat [init_code, runtime_bytecode, constructor_params ] as bytecode
- [x] deploy the bytecode with REVM
- [x] verify the storage operation from init_code in the deployed contract
- [x] passing params to `constructor` with codecopy

### DIFF
--- a/examples/deploy.rs
+++ b/examples/deploy.rs
@@ -1,19 +1,41 @@
 use anyhow::{anyhow, bail};
 use revm::{
+    interpreter::opcode,
     primitives::{bytes::Bytes, hex, EthereumWiring, ExecutionResult, Output, TxKind, U256},
     Evm, InMemoryDB,
 };
 
-/// [ PUSH1, 0x01, PUSH1, 0x17, PUSH1, 0x1f, CODECOPY, PUSH0, MLOAD, PUSH0, SSTORE ]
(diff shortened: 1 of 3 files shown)


examples/deploy.rs
```diff
@@ -0,0 +1,65 @@
+use anyhow::{anyhow, bail};
+use revm::{
+    primitives::{bytes::Bytes, hex, EthereumWiring, ExecutionResult, Output, TxKind, U256},
+    Evm, InMemoryDB,
+};
+
+/// [ PUSH1, 0x01, PUSH1, 0x17, PUSH1, 0x1f, CODECOPY, PUSH0, MLOAD, PUSH0, SSTORE ]
+const INIT_CODE: &[u8] = &[
+    0x60, 0x01, 0x60, 0x17, 0x60, 0x1f, 0x39, 0x5f, 0x51, 0x5f, 0x55,
```
rakita, line 9: Nit, you could use exact notion from the comment if you include, `revm::interpreter::opcode::*`

rakita (approved):
Cool example, Left one nit but lgtm otherwise.

PR #1768, opened 2024-09-13
refactor: simplify SuccessOrHalt trait bound

Wodann:
This PR simplified the trait bound used in `SuccessOrHalt` to restrict it as little as possible.


rakita (approved):
lgtm

PR #1774, opened 2024-09-16
rfc: Lazy-load fee beneficiary in mainnet handler

clabby:
## Overview

Lazy-loads the fee beneficiaries like `geth`'s EVM does. In the case where no value is transferred to the fee beneficiary, the account is not actually loaded from state in `geth`. This is a small semantic difference between the two implementations that does not actually affect the outcome of the STF, but does make execution witnesses generated by `geth` incompatible with `reth` due to the lack of a MPT proof for the fee beneficiary account.

Because `load_account` hits the DB via `Database::basic`, a proof to the beneficiary account in the trie is required to complete execution.

This edge case shows up in empty L1/L2 blocks, specifically for the 4788 transaction, where `gas_price == 0`, no value is transferred, and no other transactions load the fee recipient.

Note that the fee recipients on Optimism are always loaded, so that logic remains untouched: https://github.com/ethereum-optimism/op-geth/blob/21010bb6b78b8478ff0121778b781411263e160d/core/state_transition.go#L585-L597

### Refs

https://github.com/ethereum/go-ethereum/blob/ae707445f54972f4fddc7c6cf8ba2de083636e50/core/state_transition.go#L468-L471

### DIFF
--- a/crates/revm/src/handler/mainnet/post_execution.rs
+++ b/crates/revm/src/handler/mainnet/post_execution.rs
@@ -2,7 +2,8 @@ use crate::{
     interpreter::{Gas, SuccessOrHalt},
     primitives::{
         Block, EVMError, EVMResult, EVMResultGeneric, ExecutionResult, ResultAndState, Spec,
-        SpecId, SpecId::LONDON, Transaction, U256,
+        SpecId::{self, LONDON},
+        Transaction, U256,
     },
     Context, EvmWiring, FrameResult,
 };
--- a/crates/revm/src/handler/mainnet/post_execution.rs
+++ b/crates/revm/src/handler/mainnet/post_execution.rs
@@ -41,19 +41,19 @@ pub fn reward_beneficiary<EvmWiringT: EvmWiring, SPEC: Spec>(
         effective_gas_price
     };
 
-    let coinbase_account = context
-        .evm
-        .inner
-        .journaled_state
-        .load_account(beneficiary, &mut context.evm.inner.db)
-        .map_err(EVMError::Database)?;
+    let reward = coinbase_gas_price * U256::from(gas.spent() - gas.refunded() as u64);
+    if reward.gt(&U256::ZERO) {
+        let coinbase_account = context
+            .evm
+            .inner
+            .journaled_state
+            .load_account(beneficiary, &mut context.evm.inner.db)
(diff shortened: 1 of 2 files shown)


crates/revm/src/handler/mainnet/post_execution.rs
```diff
@@ -36,11 +37,27 @@ pub fn reward_beneficiary<EvmWiringT: EvmWiring, SPEC: Spec>(
     // transfer fee to coinbase/beneficiary.
     // EIP-1559 discard basefee for coinbase transfer. Basefee amount of gas is discarded.
     let coinbase_gas_price = if SPEC::enabled(LONDON) {
-        effective_gas_price.saturating_sub(*context.evm.env.block.basefee())
+        let basefee = *context.evm.env.block.basefee();
+
+        // Skip the transfer if basefee and max priority fee per gas is zero, like geth.
+        if basefee.is_zero()
```
rakita, line 43: Why do we skip if basefee is zero? Shouldn't we skip only if `coinbase_gas_price == 0`

clabby: It looks like that's `geth`'s heuristic: https://github.com/ethereum/go-ethereum/blob/ae707445f54972f4fddc7c6cf8ba2de083636e50/core/state_transition.go#L468

crates/revm/src/handler/mainnet/post_execution.rs
```diff
@@ -49,11 +66,7 @@ pub fn reward_beneficiary<EvmWiringT: EvmWiring, SPEC: Spec>(
         .map_err(EVMError::Database)?;
 
     coinbase_account.data.mark_touch();
```
rakita, line 68: We are skipping touch on coinbase account.

Geth code looks like a edge case:
https://github.com/ethereum/go-ethereum/blob/ae707445f54972f4fddc7c6cf8ba2de083636e50/core/state/state_object.go#L452-L459

This can potentially be a bug for EIP-161 State clear. The trigger is if coinbase is empty_account (Created before EIP-161) and there is no fee for this tx.

clabby:
Going to go ahead and close this for the time being - will re-visit as we start to work on standardizing the execution witnesses generated by `op-reth` and `op-geth`.

PR #1778, opened 2024-09-16
feat: to_plain_state

### LINKED ISSUE #1777: Non-consuming conversion from `BundleState` to `StateChangeset`
Currently the API is `pub fn into_plain_state(self, is_value_known: OriginalValuesKnown) -> StateChangeset`. This means that reth DB interfaces must consume the bundlestate in order to write plain state to DB.

Given that all information in the `BundleState` is efficiently memcopy-able, we could make `pub fn to_plain_state(&self, is_value_known: OriginalValuesKnown) -> StateChangeset` with relatively little overhead (deref-and-copy instead of copy-to-stack-and-copy). This version would copy `AccountInfo` and rely on `bytes::Bytes` internal efficient `Clone` for contracts.

prestwich:
Closes #1777 by adding a non-consuming API for generating `StateChangeset`

There's some DRYing to be done.

### DIFF
--- a/crates/primitives/src/state.rs
+++ b/crates/primitives/src/state.rs
@@ -261,9 +261,13 @@ impl AccountInfo {
     }
 
     /// Returns account info without the code.
-    pub fn without_code(mut self) -> Self {
-        self.take_bytecode();
-        self
+    pub fn without_code(&self) -> Self {
+        Self {
+            balance: self.balance,
+            nonce: self.nonce,
+            code_hash: self.code_hash,
+            code: None,
+        }
     }
 
     /// Returns if an account is empty.
--- a/crates/primitives/src/state.rs
+++ b/crates/primitives/src/state.rs
@@ -260,14 +260,36 @@ impl AccountInfo {
         }
     }
 
-    /// Returns account info without the code.
-    pub fn without_code(&self) -> Self {
+    /// Returns a copy of this account with the [`Bytecode`] removed. This is
+    /// useful when creating journals or snapshots of the state, where it is
+    /// desirable to store the code blobs elsewhere.
+    ///
+    /// ## Note
+    ///
+    /// This is distinct from [`AccountInfo::without_code`] in that it returns
+    /// a new `AccountInfo` instance with the code removed.
+    /// [`AccountInfo::without_code`] will modify and return the same instance.
+    pub fn copy_without_code(&self) -> Self {
         Self {
             balance: self.balance,
             nonce: self.nonce,
             code_hash: self.code_hash,
             code: None,
         }
+    }
+
+    /// Strip the [`Bytecode`] from this account and drop it. This is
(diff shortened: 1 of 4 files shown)


crates/primitives/src/state.rs
```diff
@@ -261,9 +261,13 @@ impl AccountInfo {
     }
 
     /// Returns account info without the code.
-    pub fn without_code(mut self) -> Self {
-        self.take_bytecode();
-        self
+    pub fn without_code(&self) -> Self {
```
prestwich, line 264: This is a breaking change. Can make a separate function call to do this if necessary

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -592,7 +654,7 @@ impl BundleState {
             // append account info if it is changed.
             let was_destroyed = account.was_destroyed();
             if is_value_known.is_not_known() || account.is_info_changed() {
-                let info = account.info.map(AccountInfo::without_code);
+                let info = account.info.as_ref().map(AccountInfo::without_code);
```
prestwich, line 657: consequence of the breaking change above

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -581,6 +581,68 @@ impl BundleState {
         self.reverts.push(reverts);
     }
 
+    /// Generate a [`StateChangeset`] from the bundle state without consuming
+    /// it.
+    pub fn to_plain_state(&self, is_value_known: OriginalValuesKnown) -> StateChangeset {
```
DaniPopes, line 586: looks like there is no benefit of self ownership at all, so i would just remove the into method entirely, or deprecate + delegate to avoid breaking changes

crates/revm/src/db/states/bundle_state.rs
```diff
@@ -581,6 +581,68 @@ impl BundleState {
...
+                    storage: account_storage_changed,
+                });
+            }
+        }
+
+        let contracts = self
+            .contracts
+            .iter()
+            // remove empty bytecodes
+            .filter_map(|(b, code)| (*b != KECCAK_EMPTY).then_some((*b, code.clone())))
```
DaniPopes, line 637: `.filter(|| !KECCAK_EMPTY).map(|| clone)`

[`clippy::filter_map_bool_then`](https://rust-lang.github.io/rust-clippy/master/#/filter_map_bool_then)

crates/primitives/src/state.rs
prestwich: went ahead and removed this

prestwich:
I went ahead and gave the same treatment to a few other functions used in the reth DB api

DaniPopes (approved):
lgtm

rakita (approved):
lgtm

PR #1723, opened 2024-08-27
[wip] chore: rm alloy-eips and bump primitives

yash-atreya:
- [x] Bump alloy-primitives to `0.8`
- [x] Remove `alloy-eips` in favour of https://github.com/alloy-rs/eips
- [x] Bump alloy

PR #1789, opened 2024-09-19
chore: migrate off alloy-eips

prestwich:
Re-work of #1723 , now that #1695 has been merged. 

Removes the dep on `alloy-eips` in `revm-specification` in favor of depending on `alloy-eip2930` and `alloy-7702`.

This should be a change only to the dep graph, as the types are already re-exports


rakita (approved):
lgtm

PR #1791, opened 2024-09-19
refactor: make ChainSpec associated type on EvmWiring

Wodann:
This PR proposes a separation between the `EvmWiring` and `ChainSpec` such that the associated types of a `ChainSpec` can be used without the `ExternalContext` and `Database` that are only required for the `Evm`.

This allows re-usage of chain specs for other use cases.

### DIFF
--- a/crates/inspector/src/gas.rs
+++ b/crates/inspector/src/gas.rs
@@ -79,8 +79,9 @@ mod tests {
         bytecode::Bytecode,
         interpreter::{opcode, Interpreter},
         primitives::{address, Bytes, Log, TxKind},
-        wiring::EvmWiring as PrimitiveEvmWiring,
-        wiring::{DefaultEthereumWiring, EthereumWiring},
+        wiring::{
+            ChainSpec, DefaultEthereumWiring, EthereumWiring, EvmWiring as PrimitiveEvmWiring,
+        },
         Evm, EvmWiring,
     };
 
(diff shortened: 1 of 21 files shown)


rakita:
Something like this is probably possible, if you need to split the trait it can be done outside of Revm:
```
mod temp {

    pub trait ChainSpec {}

    pub trait ExtCtx {}

    pub trait Database {}

    pub trait PresentWiring {
        /// Chain specification type.
        type ChainSpec: ChainSpec;

        /// External context type
        type ExtCtx: Sized;

        /// Database type.
        type Database: Database;
    }

    pub trait NewChain {
        type ChainSpec: ChainSpec;
        type ExtCtx: ExtCtx;
    }

    pub trait NewWiring {
        type NC: NewChain;
        type Database: Database;
    }

    impl<T: NewWiring> PresentWiring for T {
        type ChainSpec = <<Self as NewWiring>::NC as NewChain>::ChainSpec;
        type ExtCtx = <<Self as NewWiring>::NC as NewChain>::ExtCtx;
        type Database = <Self as NewWiring>::Database;
    }
}
```

Wodann:
> Something like this is probably possible, if you need to split the trait it can be done outside of Revm:

That is possible indeed.

I was hoping to avoid redefinitions of traits, but we can define our own `ChainSpec` type instead and reimplement the REVM `EvmWiring` using that. Thanks!

rakita:
Closing this, if there are more things about integration lets talk about it in different issue.

PR #1795, opened 2024-09-23
chore: add ReentrancySentryOOG for SSTORE

jsvisa:
Currently reth's trace results 

Geth returns a special oog error when gasleft in sstore < 2300, https://github.com/ethereum/go-ethereum/blob/564b6161637fe7e7c28e7c10a1f1978a22861bd2/core/vm/gas_table.go#L184-L188, 

```go
func gasSStoreEIP2200(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (uint64, error) {
	// If we fail the minimum gas availability invariant, fail (0)
	if contract.Gas <= params.SstoreSentryGasEIP2200 {
		return 0, errors.New("not enough gas for reentrancy sentry")
	}
```

so the trace result of reth's is differ to geth's, eg this tx: https://etherscan.io/tx/0x87a19142a638f981a923a6b76859cd4a6d443c430435a1aedf5c9cb35c955768

diff as below: 
image

### DIFF
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -1,5 +1,5 @@
 use crate::{
-    gas::{self, warm_cold_cost, warm_cold_cost_with_delegation},
+    gas::{self, warm_cold_cost, warm_cold_cost_with_delegation, CALL_STIPEND},
     interpreter::Interpreter,
     Host, InstructionResult,
 };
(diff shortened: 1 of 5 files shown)


crates/interpreter/src/instructions/host.rs
```diff
@@ -136,15 +136,13 @@ pub fn sstore<H: Host + ?Sized, SPEC: Spec>(interpreter: &mut Interpreter, host:
...
-            state_load.is_cold,
-        )
-    });
+
+    // EIP-1706 Disable SSTORE with gasleft lower than call stipend
+    if SPEC::SPEC_ID.is_enabled_in(ISTANBUL) && interpreter.gas.remaining() <= CALL_STIPEND {
+        interpreter.instruction_result = InstructionResult::ReentrancySentryOOG;
+        return;
+    }
+    gas::sstore_cost(SPEC::SPEC_ID, &state_load.data, state_load.is_cold);
```
rakita, line 145: You would need `gas!(gas::sstore_cost...)` here

rakita (approved):
We don't strive to port all geth errors as return codes are not part of consensus and it does not make sense.

But I support this PR, as it cleans up sstore_cost and clarifies the logic flow.

One missing thing in comment, but lgtm

rakita (approved):
lgtm

PR #1798, opened 2024-09-24
isthmus: operator fee

yuwen01:
Adds the operator fee, as described in [this design doc](https://github.com/ethereum-optimism/design-docs/pull/81) and [this spec](https://github.com/ethereum-optimism/specs/pull/382/files)

Here, I propose that these changes be included in the Holocene hardfork instead of isthmus, like the spec says. This can easily be changed if necessary.


rakita:
Closing this as it is not planned, we can always open it when it gets included.

PR #1802, opened 2024-09-24
fix: make macro crate-agnostic

Wodann:
This allows usage of the macros in external crates without needing to include `InstructionResult` beforehand.


rakita (approved):
lgtm

PR #1805, opened 2024-09-25
chore(primitives): replace HashMap re-exports with alloy_primitives::map

DaniPopes:
The semantics of the "hashbrown" feature are the same, meaning that enabling it overrides std with hashbrown, as well as it being the default if std is not enabled.

See: https://github.com/alloy-rs/core/pull/743

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -28,7 +28,7 @@ alloy-primitives = { version = "0.8.5", default-features = false, features = [
 ] }
 
 [features]
-default = ["std", "alloy-primitives/default"]
+default = ["std"]
 std = ["alloy-primitives/std"]
 hashbrown = ["alloy-primitives/map-hashbrown"]
 serde = ["alloy-primitives/serde"]
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -22,7 +22,7 @@ rust_2018_idioms = "deny"
 all = "warn"
 
 [dependencies]
-alloy-primitives = { version = "0.8.4", default-features = false, features = [
+alloy-primitives = { version = "0.8.5", default-features = false, features = [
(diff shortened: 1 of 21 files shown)


crates/primitives/Cargo.toml
```diff
@@ -22,22 +22,16 @@ rust_2018_idioms = "deny"
...
-
-[build-dependencies]
-hex = { version = "0.4", default-features = false }
 
 [features]
-default = ["std"]
-std = ["alloy-primitives/std", "hex/std"]
-hashbrown = []
-serde = ["alloy-primitives/serde", "hex/serde", "hashbrown/serde"]
+default = ["std", "alloy-primitives/default"]
```
rakita, line 31: I would remove `alloy-primitives/default` to leave std map as a default. Other hashers could be enabled by alloy-primitives features

DaniPopes: Sure, do you want to expose these features in revm too?

DaniPopes: A small problem with this is that some methods only work with the std hasher (`new`, `with_capacity`, `From::from`...), which is fine if it's the default, but if you change hasher revm will fail to compile and it might not be caught in CI

rakita: Was surprised by `new`, `From::from` when I looked at the changes. We should have CI that switches to non standard hasher.

I want to experiment with this a little, features get unified if **same** version of dependency is included, as in include both `revm` and `revm-primitives`, so we could by enabling revm-primitives features enable them inside revm.

Need to check what happens if versions are not same, and how brittle is this pattern.

You don't need to do it now in this PR, but I will maybe add it later when I come to this.

rakita (approved):
lgtm, merging this first to not have conflict later.

Default hasher in primitives was put to `std`.

PR #1809, opened 2024-09-26
feat: add support for async database

morph-dev:
Inspired by [AlloyDB](https://github.com/bluealloy/revm/blob/f8d8ae9520025593cb49cb5974a310d1ffbef8fc/crates/database/src/alloydb.rs), I created `DatabaseAsync` and `DatabaseAsyncRef` traits that are async versions of `Database` and `DatabaseRef`. They are accompanied by `WrapDatabaseAsync`, that handles most of the logic.

Portal network already uses almost identical implementation: [link](https://github.com/ethereum/trin/blob/721a7e297c4b4fa1c088bbb5cd98eafde6263878/trin-evm/src/async_db.rs).

_Note:_ First time contributing to this repo. Let me know if I'm doing something wrong, have to run some tests, etc.

### DIFF
--- a/crates/database/interface/src/async_db.rs
+++ b/crates/database/interface/src/async_db.rs
@@ -1,0 +1,188 @@
+use core::future::Future;
+
+use primitives::{Address, B256, U256};
+use state::{AccountInfo, Bytecode};
+use tokio::runtime::{Handle, Runtime};
+
+use crate::{Database, DatabaseRef};
+
+/// The async EVM database interface.
+///
+/// Contains the same methods as [Database], but it returns [Future] type instead.
+///
+/// Use [WrapDatabaseAsync] to provide [Database] implementation for a type that only implements this trait.
+pub trait DatabaseAsync {
+    /// The database error type.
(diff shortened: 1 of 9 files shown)


rakita (approved):
nice! lgtm

crates/database/interface/src/async_db.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Use [WrapDatabaseAsync] to provide [Database] implementation for a type that only implements this trait.
+pub trait DatabaseAsync {
+    /// The database error type.
+    type Error: Send;
+
+    /// Get basic account information.
+    fn basic_async(
+        &mut self,
+        address: Address,
+    ) -> impl Future<Output = Result<Option<AccountInfo>, Self::Error>> + Send;
```
rakita, line 22: I was not up to date with what is canonical way to express this. I know that `async fn` is a possibility but after googling it seems it is hard to express Send/Sync/static requirements, and this looks like the best approach.

PR #1815, opened 2024-10-01
Bump EOF EEST tests to v1.1.1

cairoeth:
Bump EOF suite to https://github.com/ethereum/execution-spec-tests/releases/tag/eip7692%40v1.1.1

Generated with evmone. Allows to mark revm as up to date in the [testing readiness matrix](https://github.com/ipsilon/eof/blob/main/spec/implementation_matrix.md#testing-readiness-matrix), for both EEST and evmone I believe.

Fixtures can be checked against by running the command in the meta file


rakita (approved):
Cool, thanks!

rakita:
I think evemone suite has different tests, those too need to be imported?

PR #1816, opened 2024-10-02
chore: bump EOF evmone tests to v0.13.0

cairoeth:
Bump to tests from https://github.com/ethereum/evmone/releases/tag/v0.13.0


rakita (approved):
Few tests from evmone were moved to eest suite so this would explain less lines

PR #1819, opened 2024-10-09
chore: inline more `AccountInfo` fns and add docs

Rjected:
Adds more docs for `AccountInfo` constructors and inlines the rest of its pub functions


rakita (approved):
lgtm

PR #1820, opened 2024-10-11
feat(interpreter): impl Clone for Stack

jsvisa:
used in https://github.com/paradigmxyz/revm-inspectors/pull/220

### DIFF
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -7,7 +7,7 @@ use std::vec::Vec;
 pub const STACK_LIMIT: usize = 1024;
 
 /// EVM stack with [STACK_LIMIT] capacity of words.
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize))]
 pub struct Stack {
     /// The underlying data of the stack.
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/interpreter/stack.rs
```diff
@@ -7,7 +7,7 @@ use std::vec::Vec;
 pub const STACK_LIMIT: usize = 1024;
 
 /// EVM stack with [STACK_LIMIT] capacity of words.
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
mattsse, line 10: this is problematic, because breaks the invariant that this has type hash STACK_LIMIT capacity

DaniPopes: yeah shouldn't derive but Self::new with copy, pls also add a comment

jsvisa: How about we implement the clone manually, then we can use `clone()` as usual:

```rust
impl Clone for Stack {
    fn clone(&self) -> Self {
        let mut new_data = Vec::with_capacity(STACK_LIMIT);
        new_data.extend_from_slice(&self.data);
        Self { data: new_data }
    }
}
```

DaniPopes: that's what I mean

jsvisa: @DaniPopes applied, ptal

DaniPopes (approved):

mattsse (approved):
lgtm, pending @rakita

rakita (approved):
lgtm

jsvisa:
@rakita kindly ask when will this feature been released, maybe 19.0.0

rakita:
I hope in two weeks, but next release is going to be a bigger one.

rizkyy702:
> used in https://github.com/paradigmxyz/revm-inspectors/pull/220

PR #1823, opened 2024-10-15
chore: bump newest primitives

rakita:


DaniPopes (approved):
the duplicate hashbrown will go away with cargo update

PR #1827, opened 2024-10-17
feat(database): implement order-independent equality for Reverts

### LINKED ISSUE #1813: Manually impl `PartialEq` for `Reverts`
In reth's invalid block hooks, we first compare the invalid block's `BundleState` to a re-executed version, then use `serde_json` to serialize `BundleState` and save it as JSON. Because the reverts are unsorted, this leads to equivalent `BundleState`s being seen as not equal. Instead, there should be a comparison API for `Reverts` that returns `true` if the reverts have the same contents, regardless of ordering.

hoank101:
Fixes #1813

### DIFF
--- a/crates/database/src/states/reverts.rs
+++ b/crates/database/src/states/reverts.rs
@@ -239,25 +239,7 @@ impl AccountRevert {
 /// Implements partial ordering for AccountInfoRevert
 impl PartialOrd for AccountInfoRevert {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        match (self, other) {
-            (AccountInfoRevert::DoNothing, AccountInfoRevert::DoNothing) => Some(Ordering::Equal),
-            (AccountInfoRevert::DoNothing, _) => Some(Ordering::Less),
-            (AccountInfoRevert::DeleteIt, AccountInfoRevert::DoNothing) => Some(Ordering::Greater),
-            (AccountInfoRevert::DeleteIt, AccountInfoRevert::DeleteIt) => Some(Ordering::Equal),
-            (AccountInfoRevert::DeleteIt, AccountInfoRevert::RevertTo(_)) => Some(Ordering::Less),
-            (AccountInfoRevert::RevertTo(_), AccountInfoRevert::DoNothing) => {
-                Some(Ordering::Greater)
-            }
-            (AccountInfoRevert::RevertTo(_), AccountInfoRevert::DeleteIt) => {
-                Some(Ordering::Greater)
-            }
-            (AccountInfoRevert::RevertTo(a), AccountInfoRevert::RevertTo(b)) => Some(
-                a.balance
-                    .cmp(&b.balance)
-                    .then(a.nonce.cmp(&b.nonce))
-                    .then(a.code_hash.cmp(&b.code_hash)),
-            ),
-        }
+        Some(self.cmp(other))
     }
 }
 
--- a/crates/database/src/states/reverts.rs
+++ b/crates/database/src/states/reverts.rs
@@ -92,22 +92,18 @@ impl Reverts {
             if self_transition.len() != other_transition.len() {
                 return false;
             }
-
-            let mut self_map: HashMap<Address, &AccountRevert> = HashMap::new();
-            for (addr, account_revert) in self_transition {
-                self_map.insert(*addr, account_revert);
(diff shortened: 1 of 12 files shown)


hoank101:
hi @Rjected please help me review it, thank!

crates/database/src/states/reverts.rs
```diff
@@ -81,6 +82,39 @@ impl Reverts {
...
+                        self_map.remove(addr);
+                    }
+                    _ => return false,
+                }
+            }
+
+            if !self_map.is_empty() {
+                return false;
+            }
+        }
```
rakita, line 113: This does not look like a correct solution, there can be multiple reverts of same account.

Something like this is more preferred even if it is less performant:

```
let mut self_transition = self_transition.clone()
self_transition.sort()
other_transition.sort();
...
return self_transition == other_transition

```

hoank101: i just update, thank you

crates/database/src/states/reverts.rs
```diff
@@ -81,6 +82,35 @@ impl Reverts {
...
+            return false;
+        }
+
+        for (self_transition, other_transition) in self.0.iter().zip(other.0.iter()) {
+            if self_transition.len() != other_transition.len() {
+                return false;
+            }
+            if self_transition.len() != other_transition.len() {
+                return false;
+            }
```
rakita, line 97:
```suggestion
```

duplicated

crates/database/src/states/reverts.rs
```diff
@@ -81,6 +82,35 @@ impl Reverts {
...
+
+            let mut self_transition = self_transition.clone();
+            let mut other_transition = other_transition.clone();
+            // Sort both transitions
+            self_transition.sort_by(|(addr1, _), (addr2, _)| addr1.cmp(addr2));
+            other_transition.sort_by(|(addr1, _), (addr2, _)| addr1.cmp(addr2));
```
rakita, line 103: if addresses are equal, compare `AccountRevert`

crates/database/src/states/reverts.rs
```diff
@@ -198,6 +236,115 @@ impl AccountRevert {
...
+impl Ord for RevertToSlot {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match (self, other) {
+            (RevertToSlot::Some(a), RevertToSlot::Some(b)) => a.cmp(b),
+            (RevertToSlot::Some(_), RevertToSlot::Destroyed) => Ordering::Less,
+            (RevertToSlot::Destroyed, RevertToSlot::Some(_)) => Ordering::Greater,
+            (RevertToSlot::Destroyed, RevertToSlot::Destroyed) => Ordering::Equal,
+        }
+    }
+}
```
Rjected, line 308: Can these `PartialOrd` / `Ord` just be derives on the types?

crates/database/src/states/reverts.rs
```diff
@@ -198,6 +236,115 @@ impl AccountRevert {
...
+
+/// Implements total ordering for AccountRevert
+impl Ord for AccountRevert {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // First compare accounts
+        let account_ord = self.account.partial_cmp(&other.account);
+        if account_ord != Some(Ordering::Equal) {
+            return account_ord.unwrap();
```
Rjected, line 316: is this unwrap safe?

crates/database/src/states/reverts.rs
```diff
@@ -198,6 +236,115 @@ impl AccountRevert {
...
+impl PartialOrd for AccountInfoRevert {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Implements total ordering for AccountInfoRevert
+impl Ord for AccountInfoRevert {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other).unwrap()
```
Rjected, line 249: also wondering about this unwrap and whether or not we can just use `derive(PartialOrd, Ord)` on `AccountInfo Revert`

Rjected (dismissed):

hoank101:
hi @Rjected @rakita please help me review, thank

hoank101:
Hi @rakita can you merge this PR?

rakita (approved):
lgtm! And thank you for waiting!

PR #1829, opened 2024-10-22
chore: bump alloy-eip7702

klkvr:
alloy-eip7702 now has logic for authority recovery implemented as it should happen in EVM https://github.com/alloy-rs/eips/blob/e712799f4318634bc978d0749cfcd59ff9a59459/crates/eip7702/src/auth_list.rs#L234

This PR updates it to latest version and replaces `RecoveredAuthority` type with one from alloy-eip7702.

There's also now no need for authority list validity check as this is done on decoding

### DIFF
--- a/crates/revm/src/handler/mainnet/pre_execution.rs
+++ b/crates/revm/src/handler/mainnet/pre_execution.rs
@@ -115,11 +115,15 @@ pub fn apply_eip7702_auth_list<SPEC: Spec, EXT, DB: Database>(
     let mut refunded_accounts = 0;
     for authorization in authorization_list.recovered_iter() {
         // 1. Verify the chain id is either 0 or the chain's current ID.
-        if !authorization.chain_id() == 0
+        //println!("Chain ID: {}", authorization.chain_id());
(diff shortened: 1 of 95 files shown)


klkvr:
@rakita tests are failing on release/v47 branch even before my commits, so I guess this is fine?

rakita:
> @rakita tests are failing on release/v47 branch even before my commits, so I guess this is fine?

Let me check what is going on, i didn't check it previously as it was just a primitives bump

rakita (approved):
On second look, this look good

crates/revm/src/handler/mainnet/pre_execution.rs
```diff
@@ -115,9 +115,8 @@ pub fn apply_eip7702_auth_list<SPEC: Spec, EXT, DB: Database>(
     let mut refunded_accounts = 0;
     for authorization in authorization_list.recovered_iter() {
         // 1. Verify the chain id is either 0 or the chain's current ID.
-        if !authorization.chain_id() == 0
```
rakita, line 118: There was a bug here, `!chain_Id` would be a bitwise operation.

PR #1834, opened 2024-10-23
Chore: some no_std cleanup

prestwich:
- moves `printer` cfg flag to module level, instead of annotating every item in the module
- adds some missing imports to tests in `no_std`
- changes an impl from feature-gated `std::error::Error` to non-gated `core::error::Error`


rakita (approved):
lgtm

PR #1836, opened 2024-10-23
fix: EXTDELEGATECALL should fail on all precompiles

shemnon:
Since precompiles are not EOFv1 contacts all EXTDELEGATECALLs should fail with code 1.


shemnon:
Needed to get #1835 to pass

PR #1838, opened 2024-10-24
Activate Osaka Fork

shemnon:
* Accept Osaka in unit tests and eof validation tests.
* Update tests
  * remove evmone (they are all Prague based)
  * remove evmone exclusions
  * use eip7692@v2.0.0 for eest
  * remove references to test directories that are
    not valid with current specs 
* EXTDELEGATECALL fails when calling precompiles (needed to pass new tests)


rakita:
Backported here: https://github.com/bluealloy/revm/pull/1964 and here https://github.com/bluealloy/revm/pull/1963

PR #1840, opened 2024-10-25
refactor: make stricter database trait bounds for user ergonomics

### LINKED ISSUE #1674: Feature Request: Add Debug bound to Error types for Database trait
Currently downstream errors containing `Db::Error` have to add manual bounds for `Debug` if they want any printing strategy. However, `Debug` is derived or impld for all current errors used in revm. This would be a breaking change, however it is unlikely that any downstream implementors have non-Debug error types

This affects `EVMError<Db::Error>` because it derives `Debug` and therefore has no `Debug` impl unless the inner `Db::Error` has a `Debug` bound

affected traits:
- `Database`
- `DatabaseRef`
- `State`
- `StateRef`
- `BlockHash`
- `BlockHashRef`

What I want to do:

```
// broken code
#[derive(Debug, thiserror::Error)]
pub enum Error<Db> 
where
    Db: Database
{
    // results in an error 
    // `EvmError<Db::Error>` cannot be formatted using `{:?}` because it doesn't implement `Debug`
    #[error("{0:?}"]
    Evm(EVMError<Db::Error>),
    ...
}
```

What I have to do instead:
```
#[derive(Debug, thiserror::Error)]
pub enum Error<Db> 
where
    Db: Database
    // this bound holds for all current Database implementors, but must be manually written by consumers
    Db::Error: Debug,
{
    #[error("{0:?}"]
    Evm(EVMError<Db::Error>),
    ...
}
```

the same is true of `std::error::Error`, however:
- this approach does not dovetail with the crates `extern crate alloc as std` strategy for `no_std` support
- it would require an intermediary `trait ErrBound` to handle changing the bound for std/no_std
- there would need to be a `cfg(feature = "std")` implementation of `std::error::Error` for `DatabaseComponentError`

so while adding a `std::error::Error` it seems like supporting a bound of `std::error::Error` would be much more invasive, while adding `Debug` is non-invasive

prestwich:
Two quick ergonomics changes for downstream users. Intended to simplify bounds writing

This is a breaking change, as downstream implementors of these traits may need to be updated

----

## `Error` bounds for database errors

Adds an error bound for `Database::Error` and `DatabaseRef::Error`, and the async versions. This closes #1674. This prevents propagation of complex bounds for upstream wrapper structs, particularly error types that might wrap EVM output. 

motivation: 
simplify bounds writing for users by ensuring they don't need to manually write out error bounds. This is particulartly a problem when holding an instance of `EVMError<Db>`

Before:
```rust
use revm::{primitives::EVMError, Database};

pub struct Error<Db: Database<Error: core::error::Error>>(EVMError<Db::Error>);

impl<Db: Database<Error: core::error::Error>> From<EVMError<Db::Error>> for Error<Db>
{
    fn from(e: EVMError<Db::Error>) -> Self {
        Self(e)
    }
}

impl<Db: Database<Error: core::error::Error>> core::error::Error for Error<Db> {
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(&self.0)
    }
}
```

After:
```rust
use revm::{primitives::EVMError, Database};

pub struct Error<Db: Database>(EVMError<Db::Error>);

impl<Db> From<EVMError<Db::Error>> for Error<Db>
where
    Db: Database,
{
    fn from(e: EVMError<Db::Error>) -> Self {
        Self(e)
    }
}

impl<Db: Database> core::error::Error for Error<Db> {
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(&self.0)
    }
}
```

----

##  Add `Database` as a supertrait of `DatabaseCommit`

This does not close a current issue, but I'm happy to make one. This causes some bounds changes in the `database_components` examples

motivation: simplify bounds writing for users by allowing them to avoid specifying `Database`

alternate approach: a `trait CommitableDatabase: Database + DatabaseCommit` with a blanket `impl<T> CommittableDatabase for T where T: Database + DatabaseCommit`

Code before:
```rust
use revm::{Database, DatabaseCommit};
pub fn make_writable_evm<Db: Database + DatabaseCommit>(db: Db) -> Evm<'static, (), Db> { ... } 
```

Code after:
```rust
use revm::{DatabaseCommit};
pub fn make_writable_evm<Db: DatabaseCommit>(db: Db) -> Evm<'static, (), Db> { ... } 
```

----

## Add `WrapBlockHashRef` to example

Add a wrapper struct like `WrapDatabaseRef` to `BlockHashRef` to allow relaxing the bounds on the database components in the `database_components` example

----

Drive-by:
The asyncdb is cfged out when running `cargo c --all-features`. And running `cargo c -p revm-database-interface --features asyncdb` produced compilation errors due to disabled tokio features. I have enabled the necessary features. When running with `--workspace` it appears some other package enabled the features, and covered up the dep specification issue

### DIFF
--- a/crates/database/interface/src/async_db.rs
+++ b/crates/database/interface/src/async_db.rs
@@ -13,7 +13,7 @@ use crate::{Database, DatabaseRef};
 /// Use [WrapDatabaseAsync] to provide [Database] implementation for a type that only implements this trait.
 pub trait DatabaseAsync {
     /// The database error type.
-    type Error: Send;
+    type Error: core::error::Error + Send;
 
     /// Get basic account information.
     fn basic_async(
@@ -48,7 +48,7 @@ pub trait DatabaseAsync {
 /// Use [WrapDatabaseAsync] to provide [DatabaseRef] implementation for a type that only implements this trait.
 pub trait DatabaseAsyncRef {
     /// The database error type.
-    type Error: Send;
+    type Error: core::error::Error + Send;
 
     /// Get basic account information.
     fn basic_async_ref(
--- a/crates/database/interface/src/lib.rs
+++ b/crates/database/interface/src/lib.rs
@@ -21,7 +21,7 @@ pub use empty_db::{EmptyDB, EmptyDBTyped};
 #[auto_impl(&mut, Box)]
 pub trait Database {
     /// The database error type.
-    type Error;
+    type Error: core::error::Error;
 
     /// Get basic account information.
     fn basic(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error>;
@@ -38,7 +38,7 @@ pub trait Database {
 
 /// EVM database commit interface.
 #[auto_impl(&mut, Box)]
-pub trait DatabaseCommit {
+pub trait DatabaseCommit: Database {
     /// Commit changes to the database.
     fn commit(&mut self, changes: HashMap<Address, Account>);
 }
@@ -52,7 +52,7 @@ pub trait DatabaseCommit {
 #[auto_impl(&, &mut, Box, Rc, Arc)]
 pub trait DatabaseRef {
     /// The database error type.
-    type Error;
+    type Error: core::error::Error;
 
     /// Get basic account information.
     fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, Self::Error>;
--- a/examples/database_components/src/block_hash.rs
+++ b/examples/database_components/src/block_hash.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 #[auto_impl(&mut, Box)]
 pub trait BlockHash {
-    type Error;
+    type Error: core::error::Error + 'static;
 
     /// Get block hash by block number
     fn block_hash(&mut self, number: u64) -> Result<B256, Self::Error>;
@@ -15,7 +15,7 @@ pub trait BlockHash {
 
 #[auto_impl(&, &mut, Box, Rc, Arc)]
 pub trait BlockHashRef {
-    type Error;
+    type Error: core::error::Error + 'static;
 
     /// Get block hash by block number
     fn block_hash(&self, number: u64) -> Result<B256, Self::Error>;
@@ -40,5 +40,25 @@ where
 
     fn block_hash(&mut self, number: u64) -> Result<B256, Self::Error> {
         self.deref().block_hash(number)
+    }
+}
+
+/// Wraps a [`BlockHashRef`] to provide a [`BlockHash`] implementation.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct WrapBlockHashRef<T: BlockHashRef>(pub T);
+
+impl<F: BlockHashRef> From<F> for WrapBlockHashRef<F> {
+    #[inline]
+    fn from(f: F) -> Self {
+        WrapBlockHashRef(f)
+    }
+}
+
+impl<T: BlockHashRef> BlockHash for WrapBlockHashRef<T> {
+    type Error = T::Error;
+
+    #[inline]
+    fn block_hash(&mut self, number: u64) -> Result<B256, Self::Error> {
+        self.0.block_hash(number)
     }
 }
--- a/examples/database_components/src/block_hash.rs
+++ b/examples/database_components/src/block_hash.rs
@@ -42,23 +42,3 @@ where
         self.deref().block_hash(number)
     }
 }
-
-/// Wraps a [`BlockHashRef`] to provide a [`BlockHash`] implementation.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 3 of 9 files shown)


prestwich:
hmmmm. for some reason it seems not all packages are building on my local. so this may be more invasive than it appeared 🤔

prestwich:
Ok it seems to be mostly an example that was broken, so I've updated that :) 

I think this is ready for consideration

examples/database_components/src/state.rs
```diff
@@ -10,7 +10,7 @@ use std::sync::Arc;
 
 #[auto_impl(&mut, Box)]
 pub trait State {
-    type Error;
+    type Error: core::error::Error + 'static;
```
rakita, line 13: Why `'static`?

examples/database_components/src/lib.rs
```diff
@@ -46,7 +72,7 @@ impl<S: State, BH: BlockHash> Database for DatabaseComponents<S, BH> {
     }
 
     fn block_hash(&mut self, number: u64) -> Result<B256, Self::Error> {
-        self.block_hash
+        WrapBlockHashRef(&self.block_hash)
```
rakita, line 75: Why do we need `Wrap..` here can't we use `self.block_hash` directly?

examples/database_components/src/block_hash.rs
```diff
@@ -42,3 +42,23 @@ where
         self.deref().block_hash(number)
     }
 }
+
+/// Wraps a [`BlockHashRef`] to provide a [`BlockHash`] implementation.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct WrapBlockHashRef<T: BlockHashRef>(pub T);
```
rakita, line 48: hm, i can't remember it but would this code
```
impl<T: BlockHashRef> BlockHash for T {
```
solve all of this

examples/database_components/src/lib.rs
prestwich: tbh, i got caught up in following the pattern from the rest of the lib. you;re right this is unnecessary

examples/database_components/src/state.rs
prestwich: in this case, it's required by the `impl :Error for DatabaseComponentError<T, U>` [here](https://github.com/bluealloy/revm/pull/1840/files#diff-f3bbb2a6ce7d66ff5946b45cd64d22824d61ce4c6708b9f55278b5e186426924R41). The bound could be removed here, but it would require all downstream users to propagate 2 bounds every time they want to use that `impl Error for DatabaseComponentError` (which is pretty often in practice)

Given that we don't have any examples of errors borrowing (i.e. all our current databases would meet this bound) it seems pretty reasonble to put it here too. This is just an example lib, and I don't feel strongly about its ergonomics tho

I would consider adding a `+ 'static` bound to the `Error` assoc type in `trait Database` tho, for the same reason. We have no clear use case for errors that capture data instead of own it, and it causes a lot of downstream complexity

crates/database/interface/src/lib.rs
```diff
@@ -21,7 +21,7 @@ pub use empty_db::{EmptyDB, EmptyDBTyped};
 #[auto_impl(&mut, Box)]
 pub trait Database {
     /// The database error type.
-    type Error;
+    type Error: core::error::Error;
```
prestwich, line 24: I would consider adding a `'static` bound here, as described in https://github.com/bluealloy/revm/pull/1840/files#r1819000225

crates/database/interface/src/lib.rs
```diff
@@ -52,7 +52,7 @@ pub trait DatabaseCommit {
 #[auto_impl(&, &mut, Box, Rc, Arc)]
 pub trait DatabaseRef {
     /// The database error type.
-    type Error;
+    type Error: core::error::Error;
```
prestwich, line 55: I would consider adding a `'static` bound here, as described in https://github.com/bluealloy/revm/pull/1840/files#r1819000225

crates/database/interface/src/async_db.rs
```diff
@@ -13,7 +13,7 @@ use crate::{Database, DatabaseRef};
 /// Use [WrapDatabaseAsync] to provide [Database] implementation for a type that only implements this trait.
 pub trait DatabaseAsync {
     /// The database error type.
-    type Error: Send;
+    type Error: core::error::Error + Send;
```
prestwich, line 16: I would consider adding a `'static` bound here, as described in https://github.com/bluealloy/revm/pull/1840/files#r1819000225

crates/database/interface/src/async_db.rs
```diff
@@ -48,7 +48,7 @@ pub trait DatabaseAsync {
 /// Use [WrapDatabaseAsync] to provide [DatabaseRef] implementation for a type that only implements this trait.
 pub trait DatabaseAsyncRef {
     /// The database error type.
-    type Error: Send;
+    type Error: core::error::Error + Send;
```
prestwich, line 51: I would consider adding a `'static` bound here, as described in https://github.com/bluealloy/revm/pull/1840/files#r1819000225

prestwich:
CI failure seems to be related to #1844

prestwich:
closing as stale, may eventually resubmit

PR #1842, opened 2024-10-30
chore: bump alloy-eip7702 and remove `Parity` re-export

klkvr:
This enum will no longer exist after https://github.com/alloy-rs/core/pull/796

Also fixed some clippy lints


rakita (approved):
lgtm

PR #1843, opened 2024-11-01
perf(precompile): use secp256k1 global context for ecrecover

Rjected:
Initializing a new `Secp256k1` instance ever ecrecover call ends up dominating execution time, because of the randomization. This now uses the global context, avoiding randomization overhead.

Samply profiles from https://share.firefox.dev/4hwBrXj
Screenshot 2024-11-01 at 4 19 08 PM
Screenshot 2024-11-01 at 4 19 39 PM
Screenshot 2024-11-01 at 4 21 40 PM

### DIFF
--- a/crates/precompile/src/secp256k1.rs
+++ b/crates/precompile/src/secp256k1.rs
@@ -49,7 +49,7 @@ mod secp256k1 {
     use primitives::{alloy_primitives::B512, keccak256, B256};
     use secp256k1::{
         ecdsa::{RecoverableSignature, RecoveryId},
-        Message, Secp256k1,
+        Message, SECP256K1,
     };
 
     // Silence the unused crate dependency warning.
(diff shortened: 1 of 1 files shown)


crates/precompile/src/secp256k1.rs
```diff
@@ -59,7 +59,7 @@ mod secp256k1 {
         let recid = RecoveryId::from_i32(recid as i32).expect("recovery ID is valid");
         let sig = RecoverableSignature::from_compact(sig.as_slice(), recid)?;
 
-        let secp = Secp256k1::new();
+        let secp = &SECP256K1;
```
DaniPopes, line 62: can remove this variable too

DaniPopes (approved):

mattsse (approved):
we should also send this to [release/v50](https://github.com/bluealloy/revm/tree/release/v50) 

because this is the branch that's being released

onbjerg (approved):
nice catch

rakita (approved):
lgtm, test failing not related.

PR #1848, opened 2024-11-06
fix: fix incorrect html tags

pinglanlu:


rakita:
Test suite was copied from external source so it would be good for this to stay

PR #1852, opened 2024-11-13
feat: couple convenience functions for nested cache dbs

prestwich:
Adds a couple simple convenience functions for working with nested `CacheDb` instances. No particular motivation, just seemed like sensible things to have


rakita (approved):
lgtm

PR #1858, opened 2024-11-19
fix(eof-test): parse Osaka instead of Prague

hai-rise:
Fix CI: https://github.com/bluealloy/revm/actions/runs/11661887031/job/32467096293


rakita:
I will pause this after this lands: https://github.com/bluealloy/revm/pull/1865

PR #1859, opened 2024-11-19
perf(`CacheAccount`): remove unneeded clone

hai-rise:
Found it in the hot path here:
image


rakita (approved):
good find, lgtm

rakita:
test failing not related

PR #1861, opened 2024-11-19
perf: Add static Optimism precompiles for Fjord & Granite

hai-rise:
As `ContextPrecompiles::extend` is quite expensive:

image


rakita:
This is getting a major refactor in this PR https://github.com/bluealloy/revm/pull/1865

Will close this as obsolete

PR #1862, opened 2024-11-19
[v50] chore: refactor `L1BlockInfo::tx_estimated_size_fjord`

hai-rise:


rakita (approved):
lgtm.

I am focusing on the main (And should have the first release in a month) and performance work is not enough to justify a release on the side branch, so It is unknown when this is getting published.

rakita:
Okay will switch target to v51

PR #1864, opened 2024-11-19
perf: Add static Optimism precompiles for Fjord & Granite

hai-rise:
#1861.


rakita (approved):
lgtm.

I am focusing on the main (And should have the first release in a month) and performance work is not enough to justify a release on the side branch, so It is unknown when this is getting published.

PR #1865, opened 2024-11-19
feat: Restucturing Part7 Handler and Context rework

rakita:
Handle becomes list of traits.

Context:
 * goes full generic on every field (Tx/Cfg/Block/Journal/Db/Chain).
 * It is flattened. There is no InnerEvmContext and other things. It is easier to use

EvmWiring got removed.

`cargo depgraph --no-default-features --workspace-only --exclude example-block-traces,example-contract-deployment,example-database-components,example-database-ref,example-uniswap-get-reserves,example-uniswap-v2-usdc-swap  | dot -Tpng > graph.png`
Graph generated by the command above:
graph

Dependency graph took a good shape. Still needs a lot of cleanup but the major things look really good
* Split between context and context-interface and handler and handler-interface 
* How Handler uses context-interface and not the context. 
* Small pending things that would make this even better:
  * revm-state/revm-bytecode needs to be included from database-interface/context-interface
  * Interpreter needs to be split, dependency is on `InterpreterResult`
  * Precompiles needs to be split, dependency is on precompile result.
* Thing that i am unsure yet, should i introduce revm-ethereum and `revm-context` and `revm-handler` are ethereum specific. Not an important decision, and goal should be to clean/merge this asap as it is massive.


crates/handler/src/validation.rs
```diff
@@ -236,144 +266,110 @@ pub fn validate_tx_env<EvmWiringT: EvmWiring, SPEC: Spec>(
...
             .ok_or(InvalidTransaction::OverflowPaymentInTransaction)?;
     }
 
     // Check if account has enough balance for `gas_limit * max_fee`` and value transfer.
     // Transfer will be done inside `*_inner` functions.
-    if balance_check > account.info.balance {
-        if cfg.is_balance_check_disabled() {
-            // Add transaction cost to balance to ensure execution doesn't fail.
-            account.info.balance = account.info.balance.saturating_add(balance_check);
-        } else {
```
Wodann, line 317: This code was removed. I re-added it in https://github.com/bluealloy/revm/pull/2286

PR #1899, opened 2024-12-06
EIP-7742: Uncouple blob count between CL and EL

rakita:
The only meaningful change is that the `calc_excess_blob_gas` func now takes `target_blob_gas_per_block` as input.

Check to make sure that tx blob num is smaller than the block blob num is removed, this needs to be checked on block level.

### DIFF
--- a/crates/primitives/src/env.rs
+++ b/crates/primitives/src/env.rs
@@ -4,8 +4,8 @@ pub use handler_cfg::{CfgEnvWithHandlerCfg, EnvWithHandlerCfg, HandlerCfg};
 
 use crate::{
     calc_blob_gasprice, calc_excess_blob_gas, AccessListItem, Account, Address, AuthorizationList,
-    Bytes, InvalidHeader, InvalidTransaction, Spec, SpecId, B256, GAS_PER_BLOB,
-    MAX_BLOB_NUMBER_PER_BLOCK, MAX_CODE_SIZE, MAX_INITCODE_SIZE, U256, VERSIONED_HASH_VERSION_KZG,
+    Bytes, InvalidHeader, InvalidTransaction, Spec, SpecId, B256, GAS_PER_BLOB, MAX_CODE_SIZE,
+    MAX_INITCODE_SIZE, U256, VERSIONED_HASH_VERSION_KZG,
 };
 use alloy_primitives::TxKind;
 use core::cmp::{min, Ordering};
@@ -661,7 +661,7 @@ impl BlobExcessGasAndPrice {
     /// Calculate this block excess gas and price from the parent excess gas and gas used
     /// and the target blob gas per block.
     ///
-    /// This fields will be used to calculate `excess_blob_gas` with [`calc_excess_blob_gas`].
+    /// This fields will be used to calculate `excess_blob_gas` with [`calc_excess_blob_gas`] func.
     pub fn from_parent_and_target(
         parent_excess_blob_gas: u64,
         parent_blob_gas_used: u64,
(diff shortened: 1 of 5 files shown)


crates/primitives/src/env.rs
```diff
@@ -174,16 +174,6 @@ impl Env {
                     return Err(InvalidTransaction::BlobVersionNotSupported);
                 }
             }
-
-            // ensure the total blob gas spent is at most equal to the limit
-            // assert blob_gas_used <= MAX_BLOB_GAS_PER_BLOCK
-            let num_blobs = self.tx.blob_hashes.len();
-            if num_blobs > MAX_BLOB_NUMBER_PER_BLOCK as usize {
```
klkvr, line 181: should we still keep this check if we're between cancun and prague?

crates/primitives/src/env.rs
```diff
@@ -666,6 +657,22 @@ impl BlobExcessGasAndPrice {
...
+
+    /// Calculate this block excess gas and price from the parent excess gas and gas used
+    /// and the target blob gas per block.
+    ///
+    /// This fields will be used to calculate `excess_blob_gas` with [`calc_excess_blob_gas`] func.
+    pub fn from_parent_and_target(
+        parent_excess_blob_gas: u64,
+        parent_blob_gas_used: u64,
+        target_blob_gas_per_block: u64,
+    ) -> Self {
```
klkvr, line 669: I believe this should use parent's `target_blobs_per_block`?

crates/primitives/src/env.rs
rakita: It was unnecessary even in Prague. This check must be done on the whole block, not only one tx.

crates/primitives/src/env.rs
rakita: I don't think so, `target_blob_per_block` is what is received from CL when you want to build a new block, if this is something that we fetch from parent block there wouldn't be extensions to Engine API https://eips.ethereum.org/EIPS/eip-7742

rakita: `This value is set to the current target blob count. The Engine API is modified along with this EIP to provide the target_blobs_per_block with each payload and implementations can use this value to correctly set the block header field.` from EIP

klkvr: yep, it's received from CL to populate the current block 

but to calculate the excess blob gas I'd expect us to use the parent's target as the excess is calculated relative to the parent's usage

however I'm unsure about this

seems that in the current wip PR for geth the current target blob count is used https://github.com/lightclient/go-ethereum/pull/67/files#diff-1e0f75e568cb72ba19ee3b24dfd7f0ed05c9980e8fef61b14a95c4e09b80b08cR64

however, in one of the recent spec update PRs parent's target was used https://github.com/ethereum/EIPs/pull/8994/files#diff-fd7433dcf9e3427288829925409a7164a9b7b663a8d608c43a4d3c0c592ac4e2R89

klkvr: spec-tests repo version is also using parent's blob count https://github.com/ethereum/execution-spec-tests/blob/77552cc5c2fb6bcda1b207eec81f8e635fc37885/tests/prague/eip7742_uncouple_blob_count/spec.py#L139

rakita: Good resource, will ask in discord for clarification, maybe it is a parent

rakita: Tests are made with current value: https://discord.com/channels/595666850260713488/892088344438255616/1314676451089580032
Will merge this as is as the change is only in the variable name in Revm.

rakita: Besu uses parent, it feels a little bit nicer, so will rename it.

Issue #1782, opened 2024-09-18
Handler Example: use ERC-20 token to pay for gas 


rakita:
It is one of the interesting use cases where gas can be paid from ERC-20 account.

Handler that needs to be overwritten:
* `deduct_caller` from `PreExecution`: https://bluealloy.github.io/revm/crates/revm/handler.html#preexecutionhandler
* `reimburse_caller` and `reward_beneficiary` from post execution: https://bluealloy.github.io/revm/crates/revm/handler.html#postexecutionhandler
* validation check for Account here: `validate_tx_against_state` https://bluealloy.github.io/revm/crates/revm/handler.html#validationhandler.

mahmudsudo:
can i take on this ?

rakita:
@royvardhan has amazing amount of progress here: https://github.com/royvardhan/revm/pull/1 it needs to be ported to new handler on main

royvardhan:
> @royvardhan has amazing amount of progress here: [royvardhan#1](https://github.com/royvardhan/revm/pull/1) it needs to be ported to new handler on main

Working on the new handler example

PR #1901, opened 2024-12-09
feat: use custom handlers to pay for gas with erc-20

royvardhan:
Closes: #1782


examples/erc20_gas/src/main.rs
```diff
@@ -0,0 +1,198 @@
...
+    let mut evm = MainEvm::new(
+        Context::builder()
+            .with_db(cache_db)
+            .modify_tx_chained(|tx| {
+                tx.caller = from;
+                tx.transact_to = TxKind::Call(token);
+                tx.data = encoded.into();
+                tx.value = U256::from(0);
+            }),
+        EthHandler::default(),
```
royvardhan, line 181: @rakita So when I try to plug the custom handlers, it does not let me. As it expects the original handlers here with all the trait bounds. But the ones I have implemented also satisfies the trait bounds.

rakita: Where is the implementation of your Handler?

royvardhan: Here: https://github.com/royvardhan/revm/tree/1782-1/examples/erc20_gas/src/handlers

rakita: Will need to check errors

royvardhan: I am trying to do this: 
image

royvardhan: One of the errors: ```
arguments to this function are incorrect
expected struct `EthValidation<Context<BlockEnv, TxEnv, CfgEnv, &mut CacheDB<WrapDatabaseAsync<AlloyDB<..., ..., ...>>>>, ...>`
   found struct `Erc20Validation`
```

rakita: Hm it seems that EthHandler::new needs to be changed slightly.

royvardhan: There is no `new` method in `[Handler](https://github.com/bluealloy/revm/blob/45581c451a440776fd59576d7b27c366b1528724/crates/handler/interface/src/handler.rs#L3)`. So I did this, still not supporting: 
image

royvardhan: Due to this [MainEvm](https://github.com/bluealloy/revm/blob/45581c451a440776fd59576d7b27c366b1528724/crates/revm/src/evm.rs#L150) it is taking the default handler as `EthHandler`

FredCoen: did you try creating a type alias for your handler like its done for op here https://github.com/bluealloy/revm/blob/45581c451a440776fd59576d7b27c366b1528724/crates/optimism/src/handler.rs#L40

and then do a CustomEvm type alias similar to MainEvm but replaces the default Ethhandlert.
and then go CustomEvm::new(..

royvardhan: @FredCoen thanks man, fixed!

royvardhan:
@rakita I think we are good

examples/erc20_gas/src/main.rs
FredCoen: awesome 💪💪

rakita: Awesome. I will check it tomorrow

rakita (approved):
lgtm

rakita:
@royvardhan can you merge with newest main, just so that the CI passes?

royvardhan:
> @royvardhan can you merge with newest main, just so that the CI passes?

Yes ser. Will do

Issue #1890, opened 2024-12-05
Reimplement serde for Interpreter


rakita:
Commented here: https://github.com/bluealloy/revm/pull/1865

It should be a lot simpler now as only `ExtBytecode` would need to have a custom serde impl.

FredCoen:
wip here https://github.com/bluealloy/revm/pull/1909

rakita:
Done

PR #1909, opened 2024-12-11
implement serde for interpreter

FredCoen:
I tried to tackle https://github.com/bluealloy/revm/issues/1890

I am not sure this is correct. To my understanding the instruction_pointer is just a pointer and so i have replaced it with the program counter for serialization which i know is not the same but i also wonder what the value of that pointer is in a different context. To my understanding doing an absolute jump back to that offset should give us the correct instruction_pointer again when deserializing.

I probably need some guidance here, if this is too far from expecatiations please let me know i can also just let it go :)

In case this is correct i probably have to add a test where the instruction_pointer changes i.e a running execution. What is the best way to mock the host, do i have to implement a transaction, block etc. from scratch? @rakita thanks!

### DIFF
--- a/crates/interpreter/src/interpreter/ext_bytecode/serde.rs
+++ b/crates/interpreter/src/interpreter/ext_bytecode/serde.rs
@@ -1,0 +1,39 @@
+use crate::interpreter::Jumps;
+
+use super::ExtBytecode;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Serialize, Deserialize)]
+struct ExtBytecodeSerde {
+    base: bytecode::Bytecode,
+    program_counter: usize,
+}
+
+impl Serialize for ExtBytecode {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        ExtBytecodeSerde {
+            base: self.base.clone(),
+            program_counter: self.pc(),
+        }
+        .serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for ExtBytecode {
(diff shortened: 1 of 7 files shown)


crates/interpreter/src/interpreter/ext_bytecode/serde.rs
```diff
@@ -0,0 +1,39 @@
+use crate::interpreter::Jumps;
+
```
rakita, line 2:
```suggestion
```

fmt after this

crates/interpreter/src/interpreter/ext_bytecode/serde.rs
```diff
@@ -0,0 +1,39 @@
...
+    where
+        D: Deserializer<'de>,
+    {
+        let ExtBytecodeSerde {
+            base,
+            program_counter,
+        } = ExtBytecodeSerde::deserialize(deserializer)?;
+
+        let mut bytecode = Self::new(base);
+        bytecode.absolute_jump(program_counter);
```
rakita, line 36: Would put a small sanity check above this line, `if program_counter < bytecode.bytecode_len() { panic!("serde pc: {program_counter} is less than bytecode len");`

rakita (approved):
Few nits, othrwise lgtm

crates/interpreter/src/interpreter/ext_bytecode/serde.rs
FredCoen: added!

i believe you meant the message to be "\ is greater than or equal to bytecode len"

rakita:
```
i believe you meant the message to be " is greater than or equal to bytecode len"
```
Yeah, i edited my msg after i have reread it :D

rakita (approved):
lgtm

PR #1911, opened 2024-12-13
chore: put snailtracer and analysis contracts in files

Rjected:
This puts the `SNAILTRACER` and `ANALYSIS` contract bytecode in files, instead of including them as massive strings


rakita (approved):
Nice! Clippy pending

PR #1915, opened 2024-12-15
chore: integrate rust-secp256k1

### LINKED ISSUE #1914: Integrate rust-secp256k1
We should integrate it inside Precompiles under the feature flag.  https://github.com/tarcieri/rust-secp256k1-ecdsa-bench

stevencartavia:
should close #1914


crates/precompile/src/secp256k1.rs
```diff
@@ -45,20 +45,23 @@ mod secp256k1 {
 #[allow(clippy::module_inception)]
 mod secp256k1 {
     use primitives::{alloy_primitives::B512, keccak256, B256};
-    use secp256k1::{
+    use rust_secp256k1::{
```
rakita, line 48: Hey @stevencartavia, the idea is not to replace it but to add a new option with a feature flag for `rust_secp256k1`.

I would see to try `cfg_if` (there are a few examples of its usage in the repo)

crates/precompile/src/secp256k1.rs
```diff
@@ -45,20 +45,23 @@ mod secp256k1 {
...
-        Message, SECP256K1,
+        Message, Secp256k1,
     };
 
     // Silence the unused crate dependency warning.
     use k256 as _;
+    use secp256k1 as _;
+
+    pub fn ecrecover(sig: &B512, recid: u8, msg: &B256) -> Result<B256, rust_secp256k1::Error> {
+        let engine = Secp256k1::new();
```
rakita, line 58: does `rust_secp256k1` have something as a static context? for `secp256k1` this was one of the performance degradation and that is why we use `SECP256K1.recover_ecdsa` as it is faster

stevencartavia:
I haven't found I a way to fix that warning, if I add `use libsecp256k1 as _;` then it doesn't build it :(

stevencartavia:
I haven't found a way to fix that warning, if I add `use libsecp256k1 as _;` then it doesn't build it :(

crates/precompile/Cargo.toml
```diff
@@ -78,7 +79,7 @@ serde_json = "1.0"
 serde_derive = "1.0"
 
 [features]
-default = ["std", "c-kzg", "secp256k1", "portable", "blst"]
+default = ["std", "c-kzg", "secp256k1", "portable", "blst", "libsecp256k1"]
```
rakita, line 82:
```suggestion
default = ["std", "c-kzg", "secp256k1", "portable", "blst"]
```

rakita:
Have restructured the code a little, and left comments for visibility. Will merge after ci

crates/precompile/src/secp256k1.rs
```diff
@@ -84,8 +31,16 @@ pub fn ec_recover_run(input: &Bytes, gas_limit: u64) -> PrecompileResult {
...
-        .map(|o| o.to_vec().into())
-        .unwrap_or_default();
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "secp256k1")] {
+            let res = bitcoin_secp256k1::ecrecover(sig, recid, msg);
+        } else if #[cfg(feature = "libsecp256k1")] {
+            let res = parity_libsecp256k1::ecrecover(sig, recid, msg);
+        } else {
+            let res = k256::ecrecover(sig, recid, msg);
+        }
```
rakita, line 41: This is were switch is happening, priority is secp256k1 (bitcoin lib), parity (if enabled) and by default k256.

PR #1916, opened 2024-12-16
chore: use stderr for revme tracer. not panic on bytecode

rakita:


rakita:
Panic in Valgrind fixed here: https://github.com/bluealloy/revm/pull/1917

PR #1928, opened 2024-12-19
feat: Removed create address collision check

### LINKED ISSUE #1925: Remove create address collision check with precompile address
Rm this: https://github.com/bluealloy/revm/blob/ca85e67d1fd6122551bb52f8d2e4c6b8d2deadf1/crates/handler/src/frame.rs#L272
And this: https://github.com/bluealloy/revm/blob/ca85e67d1fd6122551bb52f8d2e4c6b8d2deadf1/crates/handler/src/frame.rs#L396

varun-doshi:
Closes #1925

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -267,12 +267,6 @@ where
             }
         };
 
-        // created address is not allowed to be a precompile.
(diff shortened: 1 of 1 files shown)


rakita (approved):
lgtm

PR #1930, opened 2024-12-19
docs: remove duplicate example from README.md

savvar9991:
- Removed duplicate entry of database reference example


rakita:
farming not allowed

PR #1951, opened 2024-12-25
feat: blst reprice, remove g1/g2 mul

rakita:


rakita:
Have run test suite from eest for devnet-5 1.0.0 locally.

Issue #1888, opened 2024-12-05
Add bytecode hash in Interpreter


rakita:
This was one of requirements from Foundry as recalculating a bytecode hash is expensive on every step.

FredCoen:
will take this one!

PR #1952, opened 2024-12-25
Add bytecode hash in interpreter #1888

FredCoen:
Resolves https://github.com/bluealloy/revm/issues/1888

### DIFF
--- a/crates/interpreter/src/interpreter/ext_bytecode.rs
+++ b/crates/interpreter/src/interpreter/ext_bytecode.rs
@@ -1,3 +1,5 @@
+use core::ops::Deref;
+
 use bytecode::{
     eof::TypesSection,
     utils::{read_i16, read_u16},
@@ -17,8 +19,10 @@ pub struct ExtBytecode {
     instruction_pointer: *const u8,
 }
 
-impl AsRef<Bytecode> for ExtBytecode {
-    fn as_ref(&self) -> &Bytecode {
+impl Deref for ExtBytecode {
+    type Target = Bytecode;
+
+    fn deref(&self) -> &Self::Target {
         &self.base
     }
 }
@@ -44,6 +48,14 @@ impl ExtBytecode {
         }
     }
 
+    /// Regenerates the bytecode hash.
+    pub fn regenerate_hash(&mut self) -> B256 {
+        let hash = self.base.hash_slow();
+        self.bytecode_hash = Some(hash);
+        hash
+    }
+
+    /// Returns the bytecode hash.
     pub fn hash(&mut self) -> Option<B256> {
         self.bytecode_hash
     }
--- a/crates/interpreter/src/interpreter/ext_bytecode.rs
+++ b/crates/interpreter/src/interpreter/ext_bytecode.rs
@@ -46,7 +46,6 @@ impl ExtBytecode {
 
     pub fn hash(&mut self) -> Option<B256> {
        self.bytecode_hash
-        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


FredCoen:
@rakita i am not sure if option is the best way to do this? Also let me know if we should set the value in the getter directly !

rakita:
Hey, check out the ExtBytecode, it is better place for the hash. And allow ExtBytecode to be initialized with the hash, if this is regenerated every time it is a performance hit.

FredCoen:
thanks! 

"allow ExtBytecode to be initialized with the hash" -> i introduced new_with_hash to allow this, otherwise calling the hash() method returns the hash and cashes it on the ExtByteCode. Is this what you had in mind?

For some reason it caused a regression on the ecrecover precompile perf. I am not sure why though. Do you have any idea @rakita ?

rakita:
thanks!

> "allow ExtBytecode to be initialized with the hash" -> i introduced new_with_hash to allow this, otherwise calling the hash() method returns the hash and cashes it on the ExtByteCode. Is this what you had in mind?

Yes! It would be good for `fn hash()` to return Option as there is already a function to generate hash in `Bytecode`.

> For some reason it caused a regression on the ecrecover precompile perf. I am not sure why though. Do you have any idea @rakita ?

It is not related, it can be ignored. The test is kinda flaky.

crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -29,6 +30,26 @@ impl ExtBytecode {
         Self {
             base,
             instruction_pointer,
+            bytecode_hash: None,
+        }
+    }
+
+    pub fn new_with_hash(base: Bytecode, hash: B256) -> Self {
```
rakita, line 37:
```suggestion
    /// Creates new `ExtBytecode` with the given hash.
    pub fn new_with_hash(base: Bytecode, hash: B256) -> Self {
```

crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -29,6 +30,27 @@ impl ExtBytecode {
...
+        }
+    }
+
+    pub fn hash(&mut self) -> B256 {
+        if let Some(hash) = self.bytecode_hash {
+            hash
+        } else {
+            let hash = self.base.hash_slow();
+            self.bytecode_hash = Some(hash);
+            hash
```
rakita, line 53:
```suggestion
    pub fn hash(&mut self) -> Option<B256> {
       self.bytecode_hash
```

crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -29,6 +30,22 @@ impl ExtBytecode {
...
+        Self {
+            base,
+            instruction_pointer,
+            bytecode_hash: Some(hash),
+        }
+    }
+
+    pub fn hash(&mut self) -> Option<B256> {
+       self.bytecode_hash
         }
```
rakita, line 49:
```suggestion
```

FredCoen:
hey i removed the cached hash test. Should i add anything else? CI passes where do you see it being flaky on your side locally?

rakita:
> hey i removed the cached hash test. Should i add anything else? CI passes where do you see it being flaky on your side locally?

Perfect!

The recovery test is flaky as nothing is changed with precompiles but it shows a diff, we can ignore it.

Have added the ExtBytecode::new_with_hash usege and will PR merge after CI

rakita (approved):
lgtm!

PR #1954, opened 2024-12-26
bump: alloy core and use PrimitiveSignature

rakita:
Bump alloy/core and use `PrimitiveSignature`


rakita:
Have locally run devnet-5 tests, 
```
Running tests in ../../tests/pectra_devnet5/state_tests/...
███...███ 170/170Finished execution. Total CPU time: 1.580341s
All tests passed!
```

PR #1957, opened 2024-12-26
feat: align Block trait

rakita:
In most cases, we would not need a reference but concrete type. Some block types (gas, basefee) were changes to smaller types.


rakita:
This change does not touch precompile, it is a flaky test

Issue #1879, opened 2024-12-05
Add helper functions that would create JournalInit from JournaledState


rakita:
This is needed from foundry. After this https://github.com/bluealloy/revm/pull/1865  JournaledState now contains the Database generic, and this will make it hard to clone this struct.

* `JournalInit` should be `JournaledState` with `EmptyDB` that is created from `JournaledState` with generic DB.
* And allow initializing `JournaledState` with `JournalInit`

PR #1961, opened 2024-12-31
Add helper functions that would create JournalInit from JournaledState #1879

FredCoen:
Resolves https://github.com/bluealloy/revm/issues/1879

### DIFF
--- a/crates/context/src/journal_init.rs
+++ b/crates/context/src/journal_init.rs
@@ -4,18 +4,32 @@ use database_interface::EmptyDB;
 /// A clonable version of JournaledState that uses EmptyDB.
 pub type JournalInit = JournaledState<EmptyDB>;
 
-impl<DB> From<&JournaledState<DB>> for JournalInit {
-    fn from(state: &JournaledState<DB>) -> Self {
-        Self {
+impl<DB> JournaledState<DB> {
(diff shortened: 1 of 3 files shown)


crates/context/src/journal_init.rs
```diff
@@ -0,0 +1,21 @@
+use super::journaled_state::JournaledState;
+use database_interface::EmptyDB;
+
+/// A clonable version of JournaledState that uses EmptyDB.
+pub type JournalInit = JournaledState<EmptyDB>;
```
rakita, line 5: This is cool!

crates/context/src/journal_init.rs
```diff
@@ -0,0 +1,21 @@
...
+
+impl<DB> From<&JournaledState<DB>> for JournalInit {
+    fn from(state: &JournaledState<DB>) -> Self {
```
rakita, line 8: This is an expensive operation that clones a lot of fields. would remove it and just add `into_init()` and `to_init()` functions to `JournaledState`

FredCoen: thx added that! so to_init() is still cloning similar to the rust from/into convention. I think that is what you wanted?

rakita (approved):
lgtm

PR #1962, opened 2025-01-01
fix: Documentation Corrections and Clarifications

savvar9991:
### 1. First change (line 6 in the first file):

**There are 6 precompiles implemented in REVM → There are 7 precompiles implemented in REVM**

> The number of precompiles was corrected from 6 to 7, which is a factual correction as the list indeed contains 7 precompiles.

### 2. In the kzg.md file, the following changes were made:

**a much more bigger scale → a much bigger scale**

> Justification: Fixed a grammatical error in the comparative form of the adjective. In English, you cannot use both "more" and the "-er" suffix simultaneously to form a comparative degree. It's correct to use either "more big" or "bigger", but not both. In this case, "a much bigger scale" is the correct form.

###  3. Changes in builder.md (lines 8 and 12):
**It allows handle registers → It allows handler registers**

> Fixed a typo: "handle" was replaced with "handler", which is the correct technical term in this context.

### 4. Article Addition (line 31):

**Functions from one stage are just renamed functions from other stage, it is made so that user is more aware of what underlying function does.
→ Functions from one stage are just renamed functions from other stage, it is made so that the user is more aware of what underlying function does.**

> Justification: Added the definite article "the" before "user". This is a grammatical improvement as when referring to a specific type of person (in this case, a user of the system), the definite article is required in English.

### 5. Handler Modification Change (line 160):

**They are used to add custom logic to the evm execution but as they are free to modify the Handler in any form they want.
→ They are used to add custom logic to the evm execution but they are free to modify the Handler in any form they want.**

> Justification: Removed the word "as" which was creating a redundant clause structure. The new version is more concise and clearer, while maintaining the same meaning.

### DIFF
--- a/documentation/src/crates/precompile.md
+++ b/documentation/src/crates/precompile.md
@@ -3,7 +3,7 @@
 The `precompile` crate contains the implementation of the Ethereum precompile opcodes in the EVM.
 Precompiles are a shortcut to execute a function implemented by the EVM itself, rather than an actual contract.
 Precompiled contracts are essentially predefined smart contracts on Ethereum, residing at hardcoded addresses and used for computationally heavy operations that are cheaper when implemented this way.
-There are 6 precompiles implemented in REVM, and they are: `blake2`, `bn128` curve, `identity`, `secp256k1`, `modexp`, and `sha256` and `ripemd160` hash functions.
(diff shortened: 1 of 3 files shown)


rakita (approved):
lgtm

PR #1970, opened 2025-01-06
fix(op): move l1block loading to verification

rakita:


mattsse (approved):
this looks correct now

crates/revm/src/optimism/handler_register.rs
```diff
@@ -70,13 +68,22 @@ pub fn validate_env<SPEC: Spec, DB: Database>(env: &Env) -> Result<(), EVMError<
...
         return Ok(());
     }
 
+    // storage l1 block info for later use. l1_block_info is cleared after execution.
+    if context.evm.inner.l1_block_info.is_none() {
+        // the L1-cost fee is only computed for Optimism non-deposit transactions.
+        let l1_block_info =
+            crate::optimism::L1BlockInfo::try_fetch(&mut context.evm.inner.db, SPEC::SPEC_ID)
+                .map_err(EVMError::Database)?;
+        context.evm.inner.l1_block_info = Some(l1_block_info);
```
rakita, line 82: @mattsse I added this part where the load is only done if `l1_block_info` field is `None`.

This can be a small optimization for the future where this is set before execution.

And every time after Revm finishes it will be set to `None` so it does not get that weird side effect. So it will function as before.

mattsse: makes sense

PR #1976, opened 2025-01-07
Fix naming for incrementing and decrementing balance in transfer function

FredCoen:
I think there was a naming confusion here. Fixed it with using decr for decremented balance and incr for incremented balance.


rakita (approved):
makes sense

PR #1993, opened 2025-01-13
feat(optimism): Re-use L1BlockInfoTx from Maili

refcell:
### Description

Kona as well as other libraries + applications use optimism types from [maili](https://github.com/op-rs/maili). [`maili-protocol`](https://docs.rs/maili-protocol/latest/maili_protocol/index.html) exposes the [`L1BlockInfoTx`](https://docs.rs/maili-protocol/latest/maili_protocol/enum.L1BlockInfoTx.html) type that provides all the functionality used in revm's custom `L1BlockInfo` type.

This PR removes the redundant [`L1BlockInfo`](https://github.com/bluealloy/revm/blob/main/crates/optimism/src/l1block.rs#L68) type from `revm`, using [`maili_protocol::L1BlockInfoTx`](https://docs.rs/maili-protocol/latest/maili_protocol/enum.L1BlockInfoTx.html) instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,18 +15,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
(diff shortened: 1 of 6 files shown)


rakita:
This dependency becomes hard to support as it ties `revm` release with `maili`. We had a similar discussion about alloy types and Revm, where nobody wanted to include each other.

For Optimism it seems more obvious what to do, wdyt about moving revm-optimism to a standalone repo, it can be in in `bluealloy/revm-optimism` or of your choosing?

mattsse (commented):
I'd also recommend not integrating this from mali because this makes the upgrade path more complex, unless l1blockinfo is coming from a standalone crate

refcell:
Will close this for now as to not clutter PR list.

Tracking this [in maili](https://github.com/op-rs/maili/issues/150) so I don't forget.

PR #2002, opened 2025-01-16
chore(precompiles): bump kzg.rs

nhtyy:
Succinct recently updated to v4, which includes breaking changes to the zkvm. This bump makes kzg-rs useable in the new v4 release.


rakita (approved):
lgtm!

Would note that it would be good if release notes for kzg-rs/bls12_381 were made in their repo, published code be merged on main and all of it tagged.

PR #2003, opened 2025-01-16
feat(examples): add cheatcode inspector example

agostbiro:
As discussed, adding an example that shows how to implement a Foundry-style Solidity test cheatcode inspector to make sure future changes account for this use-case.


rakita (approved):
lgtm

Issue #1997, opened 2025-01-14
Maximise reusability of APIs


Wodann:
While upgrading I ran into a lot of problems using prior APIs because a lot of APIs no longer expect individual pieces of data but rather receive the "wrapping" `CTX` type.

To increase reusability, I believe that - where possible - APIs should use the low-level types, rather than the `Context` generic type. E.g.

```
pub fn validate_initial_tx_gas<CTX, Error>(
	context: CTX,
	spec_id: SpecId,
) -> Result<InitialAndFloorGas, Error>
where
	CTX: TransactionGetter + CfgGetter,
	Error: From<InvalidTransaction>,
{
	// implementation
}
```
limits the reusability for locations where you don't have a context. By changing to:
```
pub fn validate_initial_tx_gas<ConfigT, ErrorT, TransactionT>(
	config: &ConfigT,
	transaction: &TransactionT,
) -> Result<InitialAndFloorGas, ErrorT>
where
	ConfigT: Cfg,
	TransactionT: Transaction,
```
we can call it when we have a context (`validate_initial_tx_gas(context.cfg(), context.tx()`) while also being able to call it when we just have a transaction and config type.

For convenience, you can provide helper methods that receive the context and internally call the low-level API that's more reusable.

rakita:
We could use even simpler fields: 
```rust
pub fn validate_initial_tx_gas<Error>(
	tx: &impl Transaction,
	spec_id: SpecId,
) -> Result<InitialAndFloorGas, Error>
	Error: From<InvalidTransaction>,
{
	// implementation
}
```

Wodann:
I have a local commit that does this for that function. I'll submit a PR

rakita:
done

PR #2006, opened 2025-01-17
fix: deduplicate validate_initial_tx_gas API

Wodann:
Instance of https://github.com/bluealloy/revm/issues/1997

### DIFF
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -74,10 +74,7 @@ where
         context: &Self::Context,
     ) -> Result<InitialAndFloorGas, Self::Error> {
         let spec = context.cfg().spec().into();
-        let initial_and_floor_gas =
-            validate_initial_tx_gas(context.tx(), spec).map_err(InvalidTransaction::from)?;
-
-        Ok(initial_and_floor_gas)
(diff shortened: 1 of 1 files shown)


crates/handler/src/validation.rs
```diff
@@ -74,8 +74,10 @@ where
         context: &Self::Context,
     ) -> Result<InitialAndFloorGas, Self::Error> {
         let spec = context.cfg().spec().into();
-        validate_initial_tx_gas::<&Self::Context, InvalidTransaction>(context, spec)
-            .map_err(Into::into)
+        let initial_and_floor_gas =
+            validate_initial_tx_gas(context.tx(), spec).map_err(InvalidTransaction::from)?;
+
+        Ok(initial_and_floor_gas)
```
rakita, line 80:
```suggestion
        validate_initial_tx_gas(context.tx(), spec).map_err(From::from)
```

should work

rakita (approved):
lgtm. One nit

crates/handler/src/validation.rs
Wodann: Addressed in https://github.com/bluealloy/revm/pull/2006/commits/fee24bc012334a633e73f6fa0574eb0dd81c2445

PR #2007, opened 2025-01-18
chore: bump alloy versions to match latest

0xAlcibiades:
The alloy ecosystem versions are fast paced, in this PR I propose an update 
to bring `revm` in line with `reth` and the current alloy version.

The motivation here is that reth is presently incompatible with the latest alloy
provider for simulation.


rakita (approved):
Thank you

PR #2009, opened 2025-01-19
fix: fix typos

rex4539:


rakita:
Thanks for pointing them. have made changes here: https://github.com/bluealloy/revm/pull/2009

PR #2023, opened 2025-01-25
Add essential EIP-7756 tracing fields

shemnon:
Add (for eof) section and funcion depth fields, and make pc container relative instead of code section relative.

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -153,6 +153,7 @@ where
 
         // ExtDelegateCall is not allowed to call non-EOF contracts.
         if is_ext_delegate_call && !bytecode.bytes_slice().starts_with(&EOF_MAGIC_BYTES) {
+            context.journal().checkpoint_revert(checkpoint);
             return return_result(InstructionResult::InvalidExtDelegateCallTarget);
         }
 
--- a/crates/interpreter/src/interpreter/ext_bytecode.rs
+++ b/crates/interpreter/src/interpreter/ext_bytecode.rs
@@ -30,7 +30,7 @@ impl Deref for ExtBytecode {
 impl ExtBytecode {
     /// Create new extended bytecode and set the instruction pointer to the start of the bytecode.
     pub fn new(base: Bytecode) -> Self {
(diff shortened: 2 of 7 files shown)


crates/handler/src/frame.rs
```diff
@@ -153,6 +153,7 @@ where
 
         // ExtDelegateCall is not allowed to call non-EOF contracts.
         if is_ext_delegate_call && !bytecode.bytes_slice().starts_with(&EOF_MAGIC_BYTES) {
+            context.journal().checkpoint_revert(checkpoint);
```
shemnon, line 156: This was causing the depth field to read incorrectly when this case triggered.  Not sure if here was any way to leverage it.  Also, EOF only.

rakita: Good find! Yeah it is good fix and it is EOF only

crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -92,6 +92,18 @@ impl Jumps for ExtBytecode {
...
         }
     }
+    #[inline]
+    fn trace_pc(&self) -> usize {
+        match &self.base {
+            Bytecode::Eof(_) => unsafe {
+                // SAFETY: `instruction_pointer` should be at an offset from the start of the bytecode.
+                // In practice this is always true unless a caller modifies the `instruction_pointer` field manually.
+                self.instruction_pointer
+                    .offset_from(self.base.eof().unwrap().raw.as_ptr()) as usize
```
rakita, line 102: I have already made a change so current `pc` is doing the same thing as `trace_pc`

https://github.com/bluealloy/revm/blob/be92e1db21f1c47b34c5a58cfbf019f6b97d7e4b/crates/interpreter/src/interpreter/ext_bytecode.rs#L87-L95

and bytecode() gives self.base.eof().raw here:
https://github.com/bluealloy/revm/blob/be92e1db21f1c47b34c5a58cfbf019f6b97d7e4b/crates/bytecode/src/bytecode.rs#L137-L143

PR #2024, opened 2025-01-25
Fix doc

cratiu222:
Hi
I managed to fix a few minor typo.


rakita:
the book is getting rewamped

PR #2032, opened 2025-01-28
chore: simplify some generics

klkvr:
Adds default generics for `EvmError`, `ResultAndState`, `ExecutionResult`.

Removes CTX generic from `InspectEvm` and `InspectCommitEvm` as it's unused.

Removes `Spec: Into<SpecId>` bound from `CfgEnv` struct definition

### DIFF
--- a/crates/inspector/src/exec.rs
+++ b/crates/inspector/src/exec.rs
@@ -22,7 +22,7 @@ use revm::{
 };
 use std::vec::Vec;
 
-pub trait InspectEvm<CTX, INTR: InterpreterTypes>: ExecuteEvm {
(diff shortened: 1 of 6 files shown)


crates/inspector/src/exec.rs
```diff
@@ -22,7 +22,7 @@ use revm::{
 };
 use std::vec::Vec;
 
-pub trait InspectEvm<CTX, INTR: InterpreterTypes>: ExecuteEvm {
+pub trait InspectEvm<INTR: InterpreterTypes>: ExecuteEvm {
```
rakita, line 25: Awesome! Didn't even notice

rakita (approved):
lgtm

PR #2036, opened 2025-01-28
chore: remove unnecessary `InspectorContext` generic

klkvr:
We don't need a third DB generic as it's always available as `CTX::Database`


rakita (approved):
True

PR #2042, opened 2025-01-30
chore: rename `OptimismHaltReason` -> `OpHaltReason`

klkvr:


mattsse (approved):
this follows the convention that we established across all of the op crates and (most of) reth

Op short for opstack

rakita (approved):
Intended to this for a while now. lgtm

PR #2043, opened 2025-01-31
feat(precompiles): `no_std` BLS12-381 Precompiles

refcell:
### Description

> [!WARNING]
>
> Make sure to remove the `bls-no-std` feature flag from revm (this was only used to test with revme).

Adds support to revm's BLS12-381 precompiles for `no_std` using the [bls12_381](https://docs.rs/bls12_381/0.8.0/bls12_381/index.html) crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -875,6 +875,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "bls12_381"
+version = "0.8.0"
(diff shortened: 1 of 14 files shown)


refcell:
Closing for now since all BLS precompiles will be accelerated, so we don't need `no_std` support atm.

PR #2044, opened 2025-01-31
feat: frame ctx inside Evm

rakita:


rakita:
Closing in favor of https://github.com/bluealloy/revm/pull/2049

PR #2050, opened 2025-02-05
fix: operator fee not correctly called

leruaa:
This PR fixes 2 bugs:

- The refund wasn't divided by OPERATOR_FEE_SCALAR_DECIMAL
- The operator fee was badly taken into account when the input is empty. I added similar checks as L1 fee.


rakita (approved):
lgtm

PR #2016, opened 2025-01-23
Include EIP7702 address for extcodecopy

elijahhampton:
closes https://github.com/bluealloy/revm/issues/2004

Issue #1896, opened 2024-12-05
Backport all changes from release branch


rakita:
List of changes
- [x] 2c9290a1 - (origin/release/v55) fix(op): add operator fee scalar scaling factor on refund (#2050) (2 days ago) <Aurélien> 
- [x] 14c5cc1d - (origin/release/v52) feat: Include EIP7702 address for extcodecopy (#2016) (11 days ago) <Elijah Hampton> 
- [x] 6d907985 - fix(op): isthmus compilation (#2022) (2 weeks ago) <rakita>
- [x] feat(optimism): implement BLS precompiles for isthmus https://github.com/bluealloy/revm/pull/2000 
- [x] 9dd9f9ea - chore: optimize tx l1 fetches (#1967) (2 hours ago) <rakita>
- [x] 226f0591 - feat: handle isthmus operator fee (#1960) (3 days ago) <Aurélien>
- [x] 811bc7d8 - fix: add proper part to devnet5 tests (10 days ago) <rakita>
- [x] d6b90a00 - (tag: v51, origin/release/v51) bump: alloy core and use PrimitiveSignature (#1954) (11 days ago) <rakita>
- [x] 66ee5069 - feat: blst reprice, remove g1/g2 mul, eest test bump (#1951) (12 days ago) <rakita>
- [x] 0c69ccd9 - chore: eip7702 chainid u256 change (#1950) (12 days ago) <rakita>
- [x] e30fe19c - feat: add Isthmus spec (#1948) (13 days ago) <Aurélien>
- [x] 12b0babc - (origin/release/v51) feat: eip7691 fraction update (#1900) (35 hours ago) <rakita>
- [x] d8ca097d - feat: apply latest EIP-7702 changes (#1850) (2 days ago) <Arsenii Kulikov>
- [x] 4e00f32d - EIP-7742: Uncouple blob count between CL and EL (#1899) (3 days ago) <rakita>
- [x] 6882e221 - feat(Prague): EIP-7623 Increase Calldata Cost (#1744) (3 days ago) <Roman Krasiuk>
- [x] a0024590 - (origin/rakita/l1block_cleanup_to_v512) chore: refactor L1BlockInfo::tx_estimated_size_fjord (#1862) (3 weeks ago) <Hai | RISE>
- [x] 4f762914 - [v50] perf: remove unneeded clones (#1863) (3 weeks ago) <Hai | RISE>
[...]

PR #2055, opened 2025-02-08
feat(eip7702): devnet6 changes and bump eest tests

rakita:
Backport from release PR https://github.com/bluealloy/revm/pull/2016
Related to https://github.com/bluealloy/revm/issues/1896


rakita:
Those two benches are very sensitive, we should probably introduce some loop for it

Issue #1884, opened 2024-12-05
Check performance of gas with i64


rakita:
https://ethereum-magicians.org/t/eip-4803-limit-transaction-gas-to-a-maximum-of-2-63-1/8296/7

Now we do:
```
        let (remaining, overflow) = self.remaining.overflowing_sub(cost);
        let success = !overflow;
        if success {
            self.remaining = remaining;
        }
        success
```

check assembly for diffs.

Ayushdubey86:
Hello @rakita  , can you check this pr and convo https://github.com/bluealloy/revm/pull/2062

rakita:
There is no noticeable performance improvement here

PR #2062, opened 2025-02-08
Check performance of gas with i64 #1884

Ayushdubey86:
### Summary  
- Replaced `overflowing_sub` with `checked_sub` for safer gas cost subtraction.  
- `overflowing_sub` required an explicit overflow check, while `checked_sub` directly prevents underflow without extra branching.  
- This avoids undefined behavior from potential underflow in `u64` arithmetic while maintaining performance.  

### Why Not Use `i64` Directly?  
- While `i64` allows for a simple `< 0` check, using `i64` would require changing `self.remaining` from `u64`, which is not currently feasible.  
- If we could use `i64`, it would simplify out-of-gas checks (`remaining - cost < 0`), which is CPU-friendly.  
- However, since `self.remaining` is `u64`, `checked_sub` is the best alternative for preventing underflow while keeping performance high.  

### Pipeline Considerations  
- Some test cases may need to be updated if they assume `overflowing_sub` behavior.  
- The logic remains functionally the same, but the error handling may differ slightly.

### DIFF
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -139,7 +139,7 @@ impl Gas {
     pub fn record_cost(&mut self, cost: u64) -> bool {
         if let Some(new_remaining) = self.remaining.checked_sub(cost) {
             self.remaining = new_remaining;
-            return true
+            return true;
         }
         false
     }
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -140,9 +140,8 @@ impl Gas {
         if let Some(new_remaining) = self.remaining.checked_sub(cost) {
             self.remaining = new_remaining;
             true
-        }
-        else {
-            false //OO Gas
+        } else {
+            false
         }
     }
 
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -139,10 +139,9 @@ impl Gas {
     pub fn record_cost(&mut self, cost: u64) -> bool {
         if let Some(new_remaining) = self.remaining.checked_sub(cost) {
             self.remaining = new_remaining;
-            true
-        } else {
-            false
+            return true
         }
+        false
     }
 
     /// Record memory expansion
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -136,12 +136,13 @@ impl Gas {
     /// Returns `false` if the gas limit is exceeded.
     #[inline]
     #[must_use = "prefer using `gas!` instead to return an out-of-gas error on failure"]
-      pub fn record_cost(&mut self, cost: u64) -> bool {
+    pub fn record_cost(&mut self, cost: u64) -> bool {
         if let Some(new_remaining) = self.remaining.checked_sub(cost) {
             self.remaining = new_remaining;
             true
-        } else {
-            false // OO Gas
+        }
+        else {
+            false //OO Gas
         }
     }
 


DaniPopes:
The diff looks fine, but the PR is not related to the linked issue (#1884), and the motivations are incorrect. The generated code is identical: https://godbolt.org/z/83hTeP3vM

Ayushdubey86:
@DaniPopes 

Thanks for the feedback! The goal was to improve performance by reducing unnecessary branching.

checked_sub removes the need for manual overflow handling, making the code cleaner and reducing conditional jumps. While the assembly may look similar, it's a safer and more efficient way to handle gas subtraction in u64.

For #1884, I considered using i64 to simplify the out-of-gas check (< 0 instead of overflow handling). But since self.remaining is u64, switching would introduce underflow issues.

If needed, I can change remaining to i64, but that would also require updates to functions like spent() and spent_sub_refunded() to handle negatives properly. Given that, checked_sub is the best balance of correctness and performance.

Let me know what you think!

rakita:
As @DaniPopes pointed out, it is the same assembly, so only semantics are changed.

Last your PR seemed more interesting: https://github.com/bluealloy/revm/pull/2060
Simplifying code gets us:
```
    pub fn record_cost(&mut self, cost: u64) -> bool {
        let remaining = self.remaining as i64 - cost as i64;
        self.remaining = remaining as u64;
        remaining > 0
    }
```
We don't have a branch here but `record_cost` generally has a branching in the parent to check the result of the bool. 
https://godbolt.org/z/54ExYhhsE

DaniPopes:
I've already tried this, you don't need to do i64, you can just always set `self.remaining` to the result of the overflowing operation which is the same. I did not measure any improvement as the branch still occurs in all callers to early return the gas overflow error regardless

crates/interpreter/src/gas.rs
```diff
@@ -137,12 +137,12 @@ impl Gas {
...
-        let (remaining, overflow) = self.remaining.overflowing_sub(cost);
-        let success = !overflow;
-        if success {
-            self.remaining = remaining;
+        if let Some(new_remaining) = self.remaining.checked_sub(cost) {
+            self.remaining = new_remaining;
+            true
+        } else {
+            false
         }
```
rakita, line 145:
```suggestion
        if let Some(new_remaining) = self.remaining.checked_sub(cost) {
            self.remaining = new_remaining;
            return true
        }
        return false
```

This looks nicer

rakita:
> I've already tried this here #1940 You don't need to do i64, you can just always set `self.remaining` to the result of the overflowing operation which is the same.
> 
> I did not measure any improvement as the branch still occurs in all callers to early return the gas overflow error regardless

Ack, thanks for checking it! It looks like there are no gains here.

Will close the original issue, and would rename this one to `chore:` as it does not improve performance, but makes a code little bit nicer.

Ayushdubey86:
No worries, i have made that committed suggested changes , if you want you can merge it ,if not its fine either way!

crates/interpreter/src/gas.rs
```diff
@@ -137,12 +137,11 @@ impl Gas {
     #[inline]
     #[must_use = "prefer using `gas!` instead to return an out-of-gas error on failure"]
     pub fn record_cost(&mut self, cost: u64) -> bool {
-        let (remaining, overflow) = self.remaining.overflowing_sub(cost);
-        let success = !overflow;
-        if success {
-            self.remaining = remaining;
+        if let Some(new_remaining) = self.remaining.checked_sub(cost) {
+            self.remaining = new_remaining;
+            return true
```
rakita, line 142:
```suggestion
            return true;
```

PR #2074, opened 2025-02-12
refactor: remove OpSpec

klkvr:
Removes `OpSpec` enum and uses `OpSpecId` directly 

Additionally derived `Default` on it

### DIFF
--- a/crates/optimism/src/spec.rs
+++ b/crates/optimism/src/spec.rs
@@ -17,7 +17,7 @@ pub enum OpSpecId {
 }
 
 impl OpSpecId {
-    /// Converts the [`OpSpec`] into a [`SpecId`].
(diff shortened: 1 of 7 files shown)


crates/optimism/src/spec.rs
```diff
@@ -35,27 +28,13 @@ impl OpSpecId {
     }
 
     pub const fn is_enabled_in(self, other: OpSpecId) -> bool {
-        self as u8 <= other as u8
+        other as u8 <= self as u8
```
rakita, line 31: good find

rakita (approved):

crates/optimism/src/spec.rs
klkvr: catched by tests below :)

Issue #2069, opened 2025-02-11
Untungle Inspector from .transact


rakita:
Allow Evm to call `transact`without specifying full Inspector trait requirement

PR #2075, opened 2025-02-12
Split Inspector trait from EthHandler into standalone crate

rakita:
Closes: https://github.com/bluealloy/revm/issues/2069


rakita:
Bench not related

Issue #2045, opened 2025-01-31
Add alloy/eips behind feature flag


rakita:
Make simple Auth and AccessList traits and add alloy/eips to implement those. `TxEnv<eip7702::Auth, eip2930::AccessList>` would allow usage of eips types inside TxEnv without exposing those traits

rakita:
Done

PR #2078, opened 2025-02-12
feat: integrate alloy-eips

klkvr:
Changes `access_list` and `authorization_list` fields of `TxEnv` to use types from alloy-eips.

I didn't touch the traits so they are still operating on tuples without locking to concrete eips types, allowing for future abstraction.


rakita (approved):
Thanks for doing this! lgtm

rakita:
Not exactly what is said here https://github.com/bluealloy/revm/issues/2045 but would be okay with traits

Issue #2081, opened 2025-02-13
Split Precompile Constants out of Individual Modules


refcell:
### Description

`revm-precompile` currently places BLS12-381 precompiles behind the `blst` feature flag. This restricts everything in the `bls12_381` module to only `std` environments. This ticket is to hoist the constants (for example, [bls12-381 g1 addition constants](https://github.com/bluealloy/revm/blob/main/crates/precompile/src/bls12_381/g1.rs#L6-L10)) into a module accessible from a `no_std` environment.

#### Constants Module Approach

One option to split these out is to move all the constants into a `consts.rs` module inside `bls12_381` that is not behind the `blst` feature flag.
Then, remove the `blst` feature flag on the whole `bls12_381` module and place it over only the precompiles within the `bls12_381` module.

Effectively, from a `no_std` environment the api would look like:
```
revm_precompile
- pub mod bls12_381
  - pub mod consts
``` 

And from an `std` environment, the api would look like:
```
revm_precompile
- pub mod bls12_381
  - pub mod consts;
  - pub fn precompiles()
  - pub mod g1_add
  - pub mod g1_msm
  - pub mod g2_add
  - pub mod g2_msm
  - pub mod map_fp2_to_g2
  - pub mod map_fp_to_g1
  - pub mod msm
  - pub mod pairing
```

#### Motivation

Downstream consumers of `revm` use `revm-precompile` to accelerate various precompiles for fault proof programs. For example, in [kona](https://github.com/op-rs/kona), the BLS12-381 precompiles are [accelerated](https://github.com/op-rs/kona/blob/main/bin/client/src/precompiles/mod.rs#L55-L63). Doing so, requires the [use of constants](https://github.com/op-rs/kona/blob/main/bin/client/src/precompiles/bls12_g1_add.rs#L17-L26) like input length, base fees, and precompile addresses that are already defined in revm, but are inaccessible due to `kona-client` having a strict `no_std` requirement.

rakita:
This makes sense

Ayushdubey86:
Hello [link](https://github.com/bluealloy/revm/pull/2083), linking this pr here, code is done, getting indentation problem, can someone check this out ! @refcell @rakita

Ayushdubey86:
whomsoever is free, can they check out this [pr](https://github.com/bluealloy/revm/pull/2085)!

refcell:
Nice work @Ayushdubey86 ! Closing this issue since PR #2085 implements the changes 😄

PR #2083, opened 2025-02-14
Split-Precompile-Constants-out-of-Individual-Modules-#2081 

Ayushdubey86:
Moved constants to reuse_const for no_std compatibility.


Ayushdubey86:
Hey @rakita , can you check this pr out ,  I am about to log off for the day! do let me know if there are any changes needed!

crates/precompile/src/bls12_381.rs
```diff
@@ -1,18 +1,27 @@
 use crate::PrecompileWithAddress;
 
 mod g1;
+#[cfg(feature = "blst")]
```
rakita, line 4: This does not look great. Can you create one `bls12_381_const.rs` file, scan all bls12 files for constant and move them to that file?

PR #2084, opened 2025-02-14
chore: Rename NameTrait to NameTr

rakita:
Slightly nicer names for trait that have type with same name


gakonst:
Maybe `NameT` instead of `Tr`? We do that in Reth

rakita:
> Maybe `NameT` instead of `Tr`? We do that in Reth

One letter abbreviations seems incorrect imo (Did try it in the previous commit), especially as this is the last letter in the name. Shortenings are usually few letters so you don't need to think about it.

PR #2088, opened 2025-02-17
refactor: re-export all crates from `revm`

klkvr:
Ideally consumers should be able to access any item from `revm` crate like it was before. Right now this is not possible due to `revm-inspector` depending on `revm` itself

Chnges in this PR:
- `ExecuteEvm` and `ExecuteCommitEvm` traits are moved to revm-handler
- Instead of depending on `revm`, `revm-inspector` now depends on specific crates
- `context-interface` is re-exported from `context`
- `database-interface` is re-exported from `database`
- `database` and `inspector` are re-exported from `revm`


crates/context/src/lib.rs
```diff
@@ -5,6 +5,8 @@
 #[cfg(not(feature = "std"))]
 extern crate alloc as std;
 
+pub use context_interface::*;
```
mattsse, line 8: this makes sense

mattsse (approved):
makes sense,

crates/context/src/lib.rs
rakita: This is cool

crates/handler/Cargo.toml
```diff
@@ -23,6 +23,7 @@ all = "warn"
 
 [dependencies]
 # revm
+database.workspace = true
```
rakita, line 26: We can use `database-interface` here

rakita (commented):
lgtm, left one nit

PR #2091, opened 2025-02-18
fix: call `Handler::clear`

klkvr:
Right now this is defined but never called which results in tx cached values not being reset for op


rakita (approved):
I would refactor it slightly, so clear is called every time even if a error happens

 But for now, lets merge it as it fixes the problem

PR #2095, opened 2025-02-19
Add Interop hardfork

SozinM:
Closes https://github.com/bluealloy/revm/issues/2096
For consistency https://github.com/paradigmxyz/reth/pull/14582#discussion_r1961513101
For https://github.com/paradigmxyz/reth/pull/14582


mattsse (approved):
this seems fine, even if interop doens't have any evm changes we should probably still introduce a dedicated variant?

would remove a few todos in reth and unblock work on interop which sin't blocked by isthmus

rakita (approved):
Maybe LATEST would be better but I dont mind having INTEROP

PR #2099, opened 2025-02-20
fix: update broken link secp256r1.rs

dramarereg:
Hi! I updates the broken link in secp256r1.rs, replacing EIP-7212 with RIP-7212. The change reflects the proposal's evolution from an Ethereum Improvement Proposal (EIP) to a Rollup Improvement Proposal (RIP).


mattsse (approved):
makes sense to link to the rips spec instead

rakita (approved):
thanks

PR #2100, opened 2025-02-20
G1_msm base gas fee CONST value correction

Ayushdubey86:


Rjected (approved):
ty, this is the right value, we should be testing this to prevent this type of regression

rakita (approved):

PR #2085, opened 2025-02-15
Split precompile constants out of individual modules #2081

Ayushdubey86:
This should do it, not much sure about indentations though. do let me know if it requires any further change @rakita!
pr being for this [issue](https://github.com/bluealloy/revm/issues/2081)

i have addressed this issue as well 
"This does not look great. Can you create one bls12_381_const.rs file, scan all bls12 files for constant and move them to that file?"

PR #2101, opened 2025-02-20
Revert "Split precompile constants out of individual modules #2081"

mattsse:
Reverts bluealloy/revm#2085

this change introduced a bug, we should start over here and do it right


rakita:
Will close this as bugs have been fixed here: https://github.com/bluealloy/revm/pull/2102 and CI for prague is enabled

PR #2105, opened 2025-02-22
chore: fix wrong comment & remove useless impl

jackwener:
When I read the source code of revm, I find some wrong comment. So I fix them.

- fix wrong comment
- remove useless impl

### DIFF
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -165,7 +165,7 @@ pub enum MemoryExtensionResult {
     Extended,
     /// Memory size stayed the same.
     Same,
-    /// Not enough gas to extend memory.s
+    /// Not enough gas to extend memory.
     OutOfGas,
 }
 
--- a/crates/interpreter/src/instruction_result.rs
+++ b/crates/interpreter/src/instruction_result.rs
@@ -110,8 +110,6 @@ impl From<TransferError> for InstructionResult {
     }
 }
 
-impl InstructionResult {}
-
 impl From<SuccessReason> for InstructionResult {
     fn from(value: SuccessReason) -> Self {
         match value {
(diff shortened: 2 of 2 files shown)


crates/state/src/account_info.rs
```diff
@@ -32,17 +32,13 @@ impl Default for AccountInfo {
...
+        (self.balance, self.nonce, self.code_hash) == (other.balance, other.nonce, other.code_hash)
     }
 }
 
 impl Hash for AccountInfo {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        self.balance.hash(state);
-        self.nonce.hash(state);
-        self.code_hash.hash(state);
+        (&self.balance, self.nonce, &self.code_hash).hash(state)
```
mattsse, line 41: this think this now copies the nonce, shouldn't matter for u64 I think but there's no reason to change the previous impl

mattsse (changes requested):
ty, this lgtm

smol nit

jackwener:
Thanks @mattsse review

crates/state/src/account_info.rs
```diff
@@ -32,9 +32,7 @@ impl Default for AccountInfo {
 
 impl PartialEq for AccountInfo {
     fn eq(&self, other: &Self) -> bool {
-        self.balance == other.balance
-            && self.nonce == other.nonce
-            && self.code_hash == other.code_hash
+        (self.balance, self.nonce, self.code_hash) == (other.balance, other.nonce, other.code_hash)
```
mattsse, line 35: this also technically copies into a tuple type, I don't think this is the same assembly, so we should keep this unchanged as well

mattsse (changes requested):

mattsse (approved):

PR #2111, opened 2025-02-25
chore: remove wrong `&mut`/`TODO`, and avoid useless `get_mut`

jackwener:
-  remove wrong `&mut`/`TODO`
- avoid useless `get_mut`
- delete empty file


crates/context/src/journaled_state.rs
```diff
@@ -601,7 +599,7 @@ impl<DB: Database> JournaledState<DB> {
                     state.get_mut(&address).unwrap().info.nonce -= 1;
                 }
                 JournalEntry::AccountCreated { address } => {
-                    let account = &mut state.get_mut(&address).unwrap();
```
jackwener, line 604: original type is `&mut &mut T`

Issue #2116, opened 2025-02-26
Add `fn derive_tx_type` to `TxEnv`


rakita:
To help derive tx type from set fields a helper function should be added

Something similar to this: https://github.com/bluealloy/revm/blob/1967753677ab6f394a5b77f9432ebb734488d927/crates/statetest-types/src/transaction.rs#L42

Ayushdubey86:
Hello @rakita,when free can you check this [pr](https://github.com/bluealloy/revm/pull/2118)! for above issue

PR #2118, opened 2025-02-26
Adding function derive_tx_type to TxEnv

Ayushdubey86:
@rakita 

This pr is for this issue->#2116 (Add fn derive_tx_type to TxEnv )

### DIFF
--- a/crates/context/src/tx.rs
+++ b/crates/context/src/tx.rs
@@ -94,6 +94,37 @@ impl Default for TxEnv {
         }
     }
 }
+impl TxEnv {
+    pub fn derive_tx_type(&self) -> Option<TransactionType> {
+        let mut tx_type = TransactionType::Legacy;
+
+        if !self.access_list.0.is_empty() {
+            tx_type = TransactionType::Eip2930;
+        }
+
+        if self.gas_priority_fee.is_some() {
+            tx_type = TransactionType::Eip1559;
+        }
+
+        if self.max_fee_per_blob_gas > 0 {
+            if let TxKind::Call(_) = self.kind {
+                tx_type = TransactionType::Eip4844;
+            } else {
+                return None; 
+            }
+        }
+
+        if !self.authorization_list.is_empty() {
+            if let TxKind::Call(_) = self.kind {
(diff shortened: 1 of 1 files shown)


crates/context/src/tx.rs
```diff
@@ -94,6 +96,37 @@ impl Default for TxEnv {
...
+
+        if !self.access_list.0.is_empty() {
+            tx_type = TransactionType::Eip2930;
+        }
+
+        if self.gas_priority_fee.is_some() {
+            tx_type = TransactionType::Eip1559;
+        }
+
+        if self.max_fee_per_blob_gas > 0 {
```
rakita, line 111: This condition should be `!self.blob_hashes.is_empty()`

crates/context/src/tx.rs
```diff
@@ -95,6 +96,44 @@ impl Default for TxEnv {
     }
 }
 
+pub enum DeriveTxTypeError {
```
rakita, line 99: Missing
```
#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

Ayushdubey86: I have pushed this , with all previous changes as well, do check it out !

crates/context/src/tx.rs
```diff
@@ -95,6 +96,46 @@ impl Default for TxEnv {
...
 
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum DeriveTxTypeError {
+    MissingTargetForEip4844,
+    MissingTargetForEip7702,
+}
+
+impl TxEnv {
+    pub fn derive_tx_type(&mut self) -> Result<(), DeriveTxTypeError> {
```
rakita, line 107:
```suggestion
    /// Derives tx type from transaction fields and sets it to `tx_type`.
    /// Returns error in case some fields were not set correctly.
    pub fn derive_tx_type(&mut self) -> Result<(), DeriveTxTypeError> {
```

PR #2119, opened 2025-02-26
chore: make OpEvm ContextSetters impl generic over precompiles

Rjected:
This was previously not generic, meaning we could not have this trait implementation with custom precompiles


rakita (approved):
good find

PR #2121, opened 2025-02-27
feat: TryDatabaseCommit

prestwich:
Adds a fallible version of `DatabaseCommit`. Idea came up while working [on this PR](https://github.com/init4tech/trevm/pull/85)

Basically, this would allow (e.g.) `Arc<ConcurrentState<Db>>` to be used concurrently for read-access across any number of EVMs (sharing memory cache resources), and then allow `Commit` access to the `BundleState` if and only if you have discarded all-but-one `Arc`. This would enable easier parallel simulation in rounds much easier

### DIFF
--- a/crates/database/interface/src/lib.rs
+++ b/crates/database/interface/src/lib.rs
@@ -44,11 +44,15 @@ pub trait DatabaseCommit {
 }
 
 /// EVM database commit interface that can fail.
+///
+/// This is intended for use with types that may fail to commit changes, e.g.
+/// because they are directly interacting with the filesystem, or must arrange
+/// access to a shared resource.
 pub trait TryDatabaseCommit {
-    /// Error type to be thrown when state accumulation fails.
+    /// Error type for when [`TryDatabaseCommit::try_commit`] fails.
     type Error: core::error::Error;
 
-    /// Commit changes to the database.
+    /// Attempt to commit changes to the database.
     fn try_commit(&mut self, changes: HashMap<Address, Account>) -> Result<(), Self::Error>;
 }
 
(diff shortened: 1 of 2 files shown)


prestwich:
incidentally, this can't be implemented effectively in trevm, as the blanket impl for `Db: DatabaseCommit` would have a potential conflict with the impl for `Arc<Db>` due to the fact that revm could `impl<Db> DatabaseCommit for Arc<Db>`. This is disallowed by the compiler 😮‍💨

crates/database/interface/src/lib.rs
```diff
@@ -43,6 +43,63 @@ pub trait DatabaseCommit {
     fn commit(&mut self, changes: HashMap<Address, Account>);
 }
 
+/// EVM database commit interface that can fail.
+///
+/// This is intended for use with types that may fail to commit changes, e.g.
+/// because they are directly interacting with the filesystem, or must arrange
+/// access to a shared resource.
+pub trait TryDatabaseCommit {
```
rakita, line 51: Can you move it to separate file?

crates/database/interface/src/lib.rs
```diff
@@ -43,6 +43,63 @@ pub trait DatabaseCommit {
...
+    type Error = core::convert::Infallible;
+
+    #[inline]
+    fn try_commit(&mut self, changes: HashMap<Address, Account>) -> Result<(), Self::Error> {
+        self.commit(changes);
+        Ok(())
+    }
+}
+
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
prestwich: ah yep, I misunderstood the no_std setup. will resolve

crates/database/interface/src/lib.rs
prestwich: done

PR #2125, opened 2025-02-27
Update debug

Ayushdubey86:


startup-dreamer:
I don't really see how is this adding something or solving the issue.

crates/context/src/evm.rs
```diff
@@ -22,6 +23,25 @@ impl<CTX> Evm<CTX, (), (), ()> {
     }
 }
 
+impl<CTX: Debug, INSP: Debug, I: Debug, P: Debug> Debug for Evm<CTX, INSP, I, P> {
```
rakita, line 26: Try to use `derive_where` crate. Similar to how `Context` uses it.

Issue #2115, opened 2025-02-26
Implement debug for library components


mattsse:
these, among others are all part of pub api

https://github.com/bluealloy/revm/blob/1967753677ab6f394a5b77f9432ebb734488d927/crates/context/src/evm.rs#L4-L10

all of these should have a debug impl, so that users can derive debug when using them in structs

rakita:
https://docs.rs/derive-where/latest/derive_where/ should help

startup-dreamer:
I'd like to work on this.

Ayushdubey86:
hey @mattsse @rakita , check this [pr](https://github.com/bluealloy/revm/pull/2125) for this issue

PR #2126, opened 2025-02-28
Feat: add the debug impl for Evm and EvmData type

startup-dreamer:
Closes: #2115 

Derives Debug to Evm and EvmData public api's as suggestede by @mattsse and @rakita.

Please review

### DIFF
--- a/crates/context/src/evm.rs
+++ b/crates/context/src/evm.rs
@@ -1,12 +1,15 @@
 use crate::setters::ContextSetters;
+use core::fmt::Debug;
 use core::ops::{Deref, DerefMut};
 
+#[derive(Debug)]
 pub struct Evm<CTX, INSP, I, P> {
     pub data: EvmData<CTX, INSP>,
     pub instruction: I,
     pub precompiles: P,
 }
 
+#[derive(Debug)]
 pub struct EvmData<CTX, INSP> {
     pub ctx: CTX,
     pub inspector: INSP,


crates/context/src/evm.rs
```diff
@@ -1,12 +1,16 @@
 use crate::setters::ContextSetters;
+use core::fmt::Debug;
 use core::ops::{Deref, DerefMut};
+use derive_where::derive_where;
 
+#[derive_where(Debug; CTX, INSP, I, P: Debug)]
```
rakita, line 6:
```suggestion
#[derive_where(Debug; CTX, INSP, I, P)]
```

Does it work without `: Debug`?

startup-dreamer: No It gives error.

crates/context/src/evm.rs
```diff
@@ -1,12 +1,16 @@
...
 pub struct Evm<CTX, INSP, I, P> {
     pub data: EvmData<CTX, INSP>,
     pub instruction: I,
     pub precompiles: P,
 }
 
+#[derive_where(Debug; CTX, INSP:  Debug)]
```
rakita, line 13:
```suggestion
#[derive_where(Debug; CTX, INSP)]
```

Check it like this

startup-dreamer: Compiler says:
```rust
this can be handled by standard `#[derive(..)]`, use a `skip` or `incomparable` attribute, implement `Default` on an enum, or different generic type parametersrust-analyzer[macro-error](https://rust-analyzer.github.io/manual.html#macro-error)
```

crates/context/src/evm.rs
rakita: what is the error?

startup-dreamer: Same as below:
```
this can be handled by standard `#[derive(..)]`, use a `skip` or `incomparable` attribute, implement `Default` on an enum, or different generic type parametersrust-analyzer[macro-error](https://rust-analyzer.github.io/manual.html#macro-error)
```

crates/context/src/evm.rs
rakita: oh, if it can be handled by standard, please use `#[derive(..)]`

startup-dreamer:
CI looks good as well thanks @rakita for quick response made the changes.

PR #2128, opened 2025-03-01
refactor: refactor file filtering for improved efficiency

mdqst:
I’ve made improvements to the file filtering logic for better performance and readability. Specifically, I replaced the use of `.to_string_lossy()` with `.to_str()` where possible, as it’s more efficient when dealing with UTF-8 file names. Additionally, I’ve switched to using `Path::extension()` for checking `.json` files, which is a more idiomatic and performant approach.

p.s. This update should help streamline the code.

### DIFF
--- a/bins/revme/src/dir_utils.rs
+++ b/bins/revme/src/dir_utils.rs
@@ -5,7 +5,7 @@ pub fn find_all_json_tests(path: &Path) -> Vec<PathBuf> {
     WalkDir::new(path)
         .into_iter()
         .filter_map(|e| e.ok())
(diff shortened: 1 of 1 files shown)


rakita (approved):
lgtm. fmt pending

rakita:
@mdqst fmt CI test is still failing, run `cargo fmt --all` or check https://github.com/bluealloy/revm/actions/runs/13607261589/job/38040188967?pr=2128 to see how it should look

rakita:
Will make a change in different PR

Issue #1870, opened 2024-11-27
Ergonomics request: encapsulate the `Cache` in `CacheDB<DB>` to allow it to be used independently


prestwich:
It's often useful during simulation to separate cached updates from the underlying DB, so that they can (e.g.) be absorbed into a higher-level cache.

Requested refactor would look something like this:
```
pub struct Cache {
    pub accounts: HashMap<Address, DbAccount>,
    pub contracts: HashMap<B256, Bytecode>,
    pub logs: Vec<Log>,
    pub block_hashes: HashMap<U256, B256>,
}

pub struct CacheDB<ExtDB> {
    cache: Cache,
    db: ExtDb
}
```

prestwich:
I can prep a PR if it wouldn't interfere with any ongoing refactor work

PR #2131, opened 2025-03-03
refactor(db): separate fields from `CacheDB` into `Cache`

lambda-0x:
resolves: #1870

### DIFF
--- a/crates/database/src/in_memory_db.rs
+++ b/crates/database/src/in_memory_db.rs
@@ -1,22 +1,24 @@
 use core::convert::Infallible;
 use database_interface::{Database, DatabaseCommit, DatabaseRef, EmptyDB};
-use primitives::{address, hash_map::Entry, Address, HashMap, Log, B256, KECCAK_EMPTY, U256};
+use primitives::{
+    address,
+    hash_map::{Entry, HashMap},
+    Address, Log, B256, KECCAK_EMPTY, U256,
+};
 use state::{Account, AccountInfo, Bytecode};
 use std::vec::Vec;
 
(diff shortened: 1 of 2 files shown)


crates/database/src/in_memory_db.rs
```diff
@@ -26,6 +28,31 @@ pub struct CacheDB<ExtDB> {
     pub logs: Vec<Log>,
     /// All cached block hashes from the [DatabaseRef].
     pub block_hashes: HashMap<U256, B256>,
+}
+
+impl Default for Cache {
+    fn default() -> Self {
+        let mut contracts = HashMap::new();
+        contracts.insert(KECCAK_EMPTY, Bytecode::default());
+        contracts.insert(B256::ZERO, Bytecode::default());
```
lambda-0x, line 37: should i keep it here or move it to `Default` implementation of `CacheDB`?

rakita: They should be here

crates/database/src/in_memory_db.rs
```diff
@@ -26,6 +28,31 @@ pub struct CacheDB<ExtDB> {
...
+        }
+    }
+}
+
+/// A [Database] implementation that stores all state changes in memory.
+///
+/// This implementation wraps a [DatabaseRef] that is used to load data ([AccountInfo]).
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct CacheDB<ExtDB> {
```
rakita, line 53: This is nice!

rakita (approved):
lgtm

PR #2134, opened 2025-03-03
refactor: add immutable gas API to LoopControl

Wodann:
This refactor allows borrowing either an immutable or mutable `Gas` for the `trait LoopControlTr`


rakita (approved):
lgtm

PR #2135, opened 2025-03-03
feat: add constructor with hardfork

Wodann:
Introduces an API that allows construction of a `CfgEnv` with a `SpecId`. This is useful when using a `SpecId` type that doesn't implement `Default`.

The changes avoid duplication of default parameters by calling `with_spec` from `fn default`.

### DIFF
--- a/crates/context/src/cfg.rs
+++ b/crates/context/src/cfg.rs
@@ -76,28 +76,23 @@ impl CfgEnv {
 }
 
 impl<SPEC> CfgEnv<SPEC> {
-    pub fn with_chain_id(mut self, chain_id: u64) -> Self {
-        self.chain_id = chain_id;
-        self
-    }
-
-    pub fn with_spec<OSPEC: Into<SpecId>>(self, spec: OSPEC) -> CfgEnv<OSPEC> {
-        CfgEnv {
-            chain_id: self.chain_id,
-            limit_contract_code_size: self.limit_contract_code_size,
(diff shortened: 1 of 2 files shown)


crates/context/src/cfg.rs
```diff
@@ -76,28 +76,23 @@ impl CfgEnv {
...
-    pub fn with_chain_id(mut self, chain_id: u64) -> Self {
-        self.chain_id = chain_id;
-        self
-    }
-
-    pub fn with_spec<OSPEC: Into<SpecId>>(self, spec: OSPEC) -> CfgEnv<OSPEC> {
-        CfgEnv {
-            chain_id: self.chain_id,
-            limit_contract_code_size: self.limit_contract_code_size,
+    pub fn with_spec(spec: SPEC) -> Self {
```
rakita, line 79:
```suggestion
    pub fn new_with_spec(spec: SPEC) -> Self {
```

Functions that create the object are more obvious if they start with `new_`

and would be good to leave previous functions

Wodann: Resolved by https://github.com/bluealloy/revm/pull/2135/commits/6d8ed5a96db261bc44e256f6b02fa9dbb33eae14

FYI, I chose the `with_*` function name as this naming convention is commonplace for constructors in the STL. E.g. [`Vec::with_capacity`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity).

rakita (approved):

PR #2148, opened 2025-03-05
fix: use correct HashMap import

klkvr:
we should always use `primitives::HashMap` vs `primitives::hash_map::HashMap` as `hash_map::HashMap` might have different default hasher generic

this was changed in https://github.com/bluealloy/revm/pull/2131


rakita (approved):
lgtm

Issue #2146, opened 2025-03-05
allow extending JournalEntry


omer-gen:
@rakita on my implementation of journal i need to add two new JournalEntry types and i am using an underlying JournledState<DB>, i was thinking to submit a PR that changes JournalEntry into a trait constrained bound by a fetch all the current types, and an implement to revert, this will make the current journal revert function a lot smaller since it will call the underlying revert method of the JournalEntry trait, and will move the code there into an implementation of the trait EthJournalEntry
before i do this would like your thoughts

rakita:
Hey, this is a good idea, journal in future will become more generic over state ( account/storage) so this is good step.

About the recent change that you made in Interpreter. I am making a simplification that would allow you to have only one trait.

omer-gen:
great stuff

omer-gen:
Thanks for the consideration on this! I'll give it a look and let you know (kind of accepted that I'll need to move some functionality from context to journal already, not the end of the world)

omer-gen:
@rakita having a look on the link "There isn’t anything to compare", might have changed

rakita:
@omer-gen here is PR: https://github.com/bluealloy/revm/pull/2147 the main trait is in `crates/interpreter/src/host.rs`

rakita:
Done

PR #2154, opened 2025-03-06
feat: make journal entries generic

omer-gen:
This pull requests moves revert functionality into the journal entry trait, and makes it possible to create custom journal entries which are revertible

closes issue:
https://github.com/bluealloy/revm/issues/2146


rakita:
> @rakita an idea for a future change: it could be easier to maintain if the revert and the change itself are symmetric, meaning that both the original state change and it's revert are handled in the entry, then it could be easier to unitest (checkpoint + do the change and revert, then check state is the same)

Yeah I assume something like this is going to be implemented in future, that both do the action and know how to revert it. But this is a good step forward

rakita (approved):
lgtm

Issue #2163, opened 2025-03-10
Replace deprecated `libsecp256k1` with `k256`


emhane:
Replace deprecated dep `libsecp256k1` with `k256`. `k256` is used among others by zeth and powdr and is at latest rust edition.

Soubhik-10:
Can I look into this?

rakita:
`libsecp256k1` is faster and used by default in std. Reason why we use k256 is for no_std support.

@Soubhik-10 this is not something we want to do.

emhane:
when is last time the speed difference was benchmarked? `k256` is being actively developed, may have improved

rakita:
It was a while ago https://github.com/tarcieri/rust-secp256k1-ecdsa-bench , either way `libsecp256k1` is a standard lib used by whole industry, both bitcoin and ethereum, so it is better for it to be the default one.

emhane:
5 years ago `k256` was better and `rust-secp256k1` was best according to those benches @rakita

rakita:
I hate those names, `[rust-secp256k1](https://github.com/rust-bitcoin/rust-secp256k1)` is standard for `secp256k1`,  parity lib `libsecp256k1 ` is enabled only with a feature flag (It is there as alternative) and `k256` is supported and enabled in `no_std` environment.

emhane:
right, haha, I bet some people feel that way about reth, zeta and geth.  I'm trying to figure out how feature `libsecp256k1` is enabled in reth...because it's defo being compiled but not straight forward why.

kayabaNerve:
I don't personally understand why maintaining support for Parity's `libsecp256k1` would be desirable. The repository itself was archived on March 14th, just a few days after the previous discussions. Even Parity won't maintain support for it.

rakita:
> I don't personally understand why maintaining support for Parity's libsecp256k1 would be desirable. The repository itself was archived on March 14th, just a few days after the previous discussions. Even Parity won't maintain support for it.

It was there in far fetched case if there is some bug and we need to quickly switch to a different lib. 

It was low effort to have it, but as it is deprecated lets remove it.

rakita:
Issue for removal https://github.com/bluealloy/revm/issues/2949

PR #2167, opened 2025-03-10
Chore: removed the deprecated `libsecp256k1` in favour of `k256`.

startup-dreamer:
Closes: #2163 

@emhane


rakita:
Left comment to original issue. We dont want to remove this lib

PR #2171, opened 2025-03-10
fix: switch order of publishing

rakita:
inspector is part of revm as of this release


rakita:
change in script it does not touch the code

PR #2177, opened 2025-03-10
fix: correctly propagate features

klkvr:
Right now `std` and `serde` are not always correctly propagated. This PR fixes it and adds `zepter` CI job checking features propagation as we do in alloy and reth


rakita (approved):
cool lgtm

PR #2178, opened 2025-03-10
refactor: remove CTX phantomdata from precompile providers

klkvr:
Right now `EthPrecompiles` and `OpPrecompileProvider` have a CTX generic and PhantomData. This is not neccesarry and is sometimes bad because we can't reuse the same provider object for different databases.

The PR adds `CTX` generic to `PrecompileProvider` trait instead and removes phantomdatas from those types.

Also, `OpPrecompileProvider` is renamed to `OpPrecompiles` to follow eth naming


mattsse (approved):
this makes sense to me

rakita (commented):
Makes sense

PR #2179, opened 2025-03-10
chore(precompile): Refactor out MSM Helper

refcell:
### Description

Refactor out the `msm_required_gas` method for re-use in kona. Pulling `msm_required_gas` into a non-`blst` feature flag gated module allows kona to use the method in a `no_std` environment instead of [having a duplicated method](https://github.com/op-rs/kona/blob/main/bin/client/src/precompiles/utils.rs#L3-L20).


rakita (approved):
lgtm

PR #2181, opened 2025-03-11
fix: remove wrong Clone Trait in WrapDatabaseRef

jackwener:
Removed the `Copy` traits from the `WrapDatabaseRef` struct's derive attributes.

`WrapDatabaseRef` shouldn't own `Copy` traits.


rakita (approved):
true

PR #2191, opened 2025-03-11
refactor: bubble up PrecompileError

Wodann:
Precompile errors are represented as a `String`, but there doesn't seem to be anything limiting us from bubbling up the `PrecompileError` type. This gives us better static typing for precompile errors.

### DIFF
--- a/bins/revme/src/cmd/statetest/runner.rs
+++ b/bins/revme/src/cmd/statetest/runner.rs
@@ -14,6 +14,7 @@ use revm::{
         Cfg,
     },
     database_interface::EmptyDB,
+    precompile::PrecompileError,
     primitives::{
         eip4844::TARGET_BLOB_GAS_PER_BLOCK_CANCUN, hardfork::SpecId, keccak256, Bytes, TxKind, B256,
     },
@@ -138,7 +139,10 @@ fn check_evm_execution(
     test: &Test,
     expected_output: Option<&Bytes>,
     test_name: &str,
-    exec_result: &Result<ExecutionResult<HaltReason>, EVMError<Infallible, InvalidTransaction>>,
+    exec_result: &Result<
+        ExecutionResult<HaltReason>,
+        EVMError<Infallible, PrecompileError, InvalidTransaction>,
+    >,
     db: &mut State<EmptyDB>,
     spec: SpecId,
     print_json_outcome: bool,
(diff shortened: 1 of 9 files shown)


rakita (changes requested):
PrecompileError is an internal error similar to Halt, so it is not possible to bubble it up to outside. It is converted to InterpreterResult.

Wodann:
> PrecompileError is an internal error similar to Halt, so it is not possible to bubble it up to outside. It is converted to InterpreterResult.

I could be wrong, but we actually do convert `PrecompileError` to an `EVMError` here:
https://github.com/bluealloy/revm/blob/main/crates/precompile/src/interface.rs#L69

It's also exposed in `EvmError::Precompile` as a `String`. So we are actually exposing the error, as far as I can tell.

Wodann:
> ## [CodSpeed Performance Report](https://codspeed.io/bluealloy/revm/branches/NomicFoundation%3Arefactor%2Fprecompile-error)
> ### Merging #2191 will **degrade performances by 4.78%**
> 
> Comparing `NomicFoundation:refactor/precompile-error` ([48f6528](https://github.com/bluealloy/revm/commit/48f65286d8122701dc84103b97d09aa889422098)) with `main` ([f2a2eb7](https://github.com/bluealloy/revm/commit/f2a2eb7fada832e8402fd6c1d375e26d271efea2))
> ### Summary
> 
> `❌ 1` regressions `✅ 7` untouched benchmarks
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/bluealloy/revm/branches/NomicFoundation%3Arefactor%2Fprecompile-error)._
> 
> ### Benchmarks breakdown
> 	Benchmark 	`BASE` 	`HEAD` 	Change
> ❌ 	`precompile bench | ecrecover precompile` 	194.7 µs 	204.5 µs 	-4.78%

I expect the performance regression to be a result of the usage of a nested `enum` with `String`. This could be avoided by using `enum` variants for the BLS12_381 precompile. Something like:

```rs#[derive(Debug, thiserror::Error)]
pub enum Error {
    ...
    #[error("G1ADD input should be {G1_ADD_INPUT_LENGTH} bytes, was {length}")]
    InvalidG1AddInputLength { length: usize },
    ...
}
```

rakita:
> I could be wrong, but we actually do convert PrecompileError to an EVMError here:
> https://github.com/bluealloy/revm/blob/main/crates/precompile/src/interface.rs#L69
> 
> It's also exposed in EvmError::Precompile as a String. So we are actually exposing the error, as far as I can tell.

That is a leftover and should be removed. I need to do cleanup and remove it.

rakita:
Closing it. Cleanup issue is here: https://github.com/bluealloy/revm/issues/2207

PR #2192, opened 2025-03-11
refactor: remove ContextSetters requirement from EVM execution traits

Wodann:
If you want to use the `ExecuteEvm`, `ExecuteCommitEvm`, or `InspectEvm` traits, you currently always need to implement `ContextSetters`. If you're using the `replay*` methods, you never actually need to use any of the setters though.

This PR removes the excessive constraints on the aforementioned traits by implementing the helper methods as helper functions of the module.

Additionally, I renamed the methods on `InspectEvm` to match the renamed methods on `ExecuteEvm` and `ExecuteCommitEvm`.

### DIFF
--- a/crates/handler/src/evm.rs
+++ b/crates/handler/src/evm.rs
@@ -43,20 +43,20 @@ pub trait EvmTr {
 }
 
 /// Execute EVM transactions. Main trait for transaction execution.
-pub trait ExecuteEvm: ContextSetters {
+pub trait ExecuteEvm {
     /// Output of transaction execution.
     type Output;
 
     /// Transact the transaction that is set in the context.
     fn replay(&mut self) -> Self::Output;
+}
 
-    /// Transact the given transaction.
-    ///
-    /// Internally sets transaction in context and use `replay` to execute the transaction.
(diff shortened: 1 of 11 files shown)


crates/handler/src/evm.rs
```diff
@@ -43,20 +43,20 @@ pub trait EvmTr {
 }
 
 /// Execute EVM transactions. Main trait for transaction execution.
-pub trait ExecuteEvm: ContextSetters {
+pub trait ExecuteEvm {
```
rakita, line 46: ContextSetters are part of API, idea is that you create one EVM and you can execute multiple transactions on top of it. That is why you need a Trait that knows the type of Transaction and Block.

The step forward for this would be to bring `Tx/Block` types to `ExecuteEvm` trait.

Wodann: Thanks. It seems like most of the changes included in this PR are covered in #2195 and #2194. Should anything else be preserved?

rakita: it seems that's it

rakita:
Few things backported in https://github.com/bluealloy/revm/pull/2195 and https://github.com/bluealloy/revm/pull/2194

PR #2196, opened 2025-03-12
test(state): bundle account edge case

hai-rise:


hai-rise:
Demonstration goal is done :pray:.

PR #2201, opened 2025-03-12
test: broken example when using database with lifetime

Wodann:
This example serves to illustrate a limitation when using the current API with (mutable) references.

The first commit shows the resulting compiler error - when using the current API.

The second commit tries to resolve this by introducing the `Context` as an associated type on `trait Inspector` such that we can introduce a `'context` lifetime for the context. This lifetime represents the duration during which the `Context` will be available to the inspector.

This results in the following compiler error:

```rs
error[E0308]: mismatched types
  --> examples/inspector_with_lifetime/src/main.rs:79:19
   |
79 |     let results = mine_block(&cfg, &mut db, transactions, &mut inspector)?;
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
   |
   = note: expected struct `revm::Context<_, _, _, &mut dyn revm::Database<Error = Infallible>, Journal<&mut dyn revm::Database<Error = Infallible>>, _>`
              found struct `revm::Context<_, _, _, &'context mut (dyn revm::Database<Error = Infallible> + 'context), Journal<&'context mut (dyn revm::Database<Error = Infallible> + 'context)>, _>`
note: the lifetime requirement is introduced here
  --> examples/inspector_with_lifetime/src/main.rs:33:41
   |
33 |     InspectorT: for<'context> Inspector<Context<'context> = ContextRef<'context>>,
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I've tried a lot of things to circumvent this but have been unable to. This might be a language limitation rather than an API limitation. See [this article](https://kazlauskas.me/entries/due-to-borrowck-limitations).

I still wanted to bring this to your attention, in case you had any ideas on how to solve it.

### DIFF
--- a/examples/inspector_with_lifetime/src/main.rs
+++ b/examples/inspector_with_lifetime/src/main.rs
@@ -23,15 +23,14 @@ pub trait DBSuperTrait: Database<Error = Infallible> + DatabaseCommit {}
 
 impl<T: Database<Error = Infallible> + DatabaseCommit> DBSuperTrait for T {}
 
-fn mine_block<'db, 'inspector, InspectorT>(
+fn mine_block<InspectorT>(
     cfg: &CfgEnv,
-    db: &'db mut dyn DBSuperTrait,
+    db: &mut dyn DBSuperTrait,
     transactions: Vec<TxEnv>,
-    inspector: &'inspector mut InspectorT,
+    inspector: &mut InspectorT,
 ) -> Result<Vec<ExecutionResult>, EVMError<Infallible>>
 where
-    'db: 'inspector,
-    InspectorT: Inspector<ContextRef<'inspector>>,
+    InspectorT: for<'context> Inspector<Context<'context> = ContextRef<'context>>,
 {
     let block = BlockEnv {
         prevrandao: Some(B256::random()),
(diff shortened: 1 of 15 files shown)


examples/inspector_with_lifetime/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    Ok(results)
+}
+
+fn run_transaction<InspectorT>(
+    block: &BlockEnv,
+    cfg: &CfgEnv,
+    db: &mut dyn DBSuperTrait,
+    tx: TxEnv,
+    inspector: &mut InspectorT,
```
rakita, line 55: Yeah, the Inspector takes the lifetime of the Context and it is impossible to use in the way we expect and it sucks. This was a reason why inspector field is part Evm struct, if you send full Evm here it will work, or even impl `InspectEvm`

Wodann: For some of our use cases that's possible, like building a block.

In other cases we construct the EVM ad-hoc. The flow is similar to this, e.g. for `eth_call`:

- Retrieve `state` at the requested block number from `blockchain`
- Construct a `BlockEnv` for the requested block number
- Construct a transaction for the `eth_call` request
- Construct EVM with DatabaseComponents (for `blockchain` and `state`), the requested `eth_call` transaction, `BlockEnv`, and `CfgEnv`
- Run `replay_transact`

I might just have to inline the contruction of an EVM in that case using an efficient builder pattern API.

rakita:
Cleanup

Issue #2189, opened 2025-03-11
Tests for op-revm


rakita:
Compared to Ethereum, op does not have statetest that we can use, and op-revm lacks testing vectors.

As op-revm rely on most of the Ethereum logic we should test just the difference, the tests should set up the tx, block and database and test for expected output be that error, gas or returned account balance.

List of things that require more tests:
* Tx calling precompile, If precompiles are properly activated or deactivated on every Op fork
* Deposit transaction call:
  * When successful
  * When it failed
  * When it is system tx.
  * check all cases when we have a change in logic inside op handler.
* Multi transaction calls. If l1block and enveloped bytes get cleaned up. On both error and successful calls.
* Does inspector works. Simple call when steps are called.

emhane:
can I be assigned pls? :)

Ayushdubey86:
Hi @emhane, I noticed there are many pull requests here, and some files have low code coverage. May I help by submitting a few pull requests? we can still keep this issue assigned to you!

emhane:
I'm good @Ayushdubey86 !

emhane:
```
Filename                                                    Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
optimism/src/api/builder.rs                                       2                 0   100.00%           2                 0   100.00%           9                 0   100.00%           0                 0         -
optimism/src/api/default_ctx.rs                                   2                 0   100.00%           2                 0   100.00%          15                 0   100.00%           0                 0         -
optimism/src/api/exec.rs                                         12                10    16.67%          10                 8    20.00%          42                34    19.05%           0                 0         -
optimism/src/evm.rs                                             210                22    89.52%          90                 2    97.78%         630                 6    99.05%           0                 0         -
optimism/src/fast_lz.rs                                          47                 1    97.87%          12                 1    91.67%         100                 5    95.00%           0                 0         -
optimism/src/handler.rs                                         246                24    90.24%          59                 1    98.31%         716                11    98.46%           0                 0         -
optimism/src/l1block.rs                                         115                38    66.96%          22                 2    90.91%         426                82    80.75%           0                 0         -
optimism/src/precompiles.rs                                      53                 4    92.45%          19                 1    94.74%         152                 3    98.03%           0                 0         -
[...]

PR #2204, opened 2025-03-13
test(op-precompiles): Check subset of l1 precompiles in op

emhane:

Checks that subset of l1 precompiles are part of op precompiles wrt hardfork

### DIFF
--- a/crates/optimism/src/evm.rs
+++ b/crates/optimism/src/evm.rs
@@ -110,7 +110,7 @@ mod tests {
         context::result::{ExecutionResult, OutOfGasError},
         context_interface::result::HaltReason,
         database::{BenchmarkDB, BENCH_CALLER, BENCH_CALLER_BALANCE, BENCH_TARGET},
-        precompile::bn128,
+        precompile::{bls12_381_const, bn128, u64_to_address},
         primitives::{hex::FromHex, Address, Bytes, TxKind, U256},
         state::Bytecode,
         Context, ExecuteEvm,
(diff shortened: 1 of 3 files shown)


emhane:
@rakita could you pls explain how tests pass locally without default features, but not in ci?

rakita:
Need to check to be sure, but bls does not have a no_std variant so one precompile will be missing.

rakita:
> Need to check to be sure, but bls does not have a no_std variant so one precompile will be missing.

Lets wait for https://github.com/bluealloy/revm/pull/2210, it should make that test pass.

blst are not supported for no_std but for kona they accelare all precompiles so they are fine without it.

rakita:
https://github.com/bluealloy/revm/pull/2249 is done so if we merge this, diff would be same. if there is no bls feature and this precompile is called by default it will throw Fatal error.

emhane:
found the issue here, when best is not enabled, we still add all the addresses to Prague with `bls12_381_precompiles_not_supported`, but not to isthmus

emhane:
odd , this still works locally, but in ci the output is different

emhane:
ah it's the precompile function pointers that are different in ci for Prague and isthmus sets...no easy way around that unfortunately. equality of function pointers is needed to see that granite overwrites the bn128 pair. will simplify the difference check to just check for address, to make it pass in ci. that granite overwrites bn128 pair is checked in another test already.

rakita (approved):
lgtm

rakita:
Testing that at least the address of the precompile is present is a good quick sanity test.

Issue #2172, opened 2025-03-10
no_std BLS12-381 Precompiles


rakita:
It is not audited but it can be useful for no_std environment: https://github.com/zkcrypto/bls12_381

It would be worth adding it with feature flag.

Ayushdubey86:
Hey @rakita , is it something similar to this [pr](https://github.com/bluealloy/revm/pull/2085) , or developer needs to add BLS 12-381 in cargo.toml, and make it accessible?  Also , can you assign this to me!

startup-dreamer:
Hi @Ayushdubey86, If you don't mind can I take this issue I have already started working on it, If my approach doesn't seem reasonable you can take it.

Ayushdubey86:
Hey @startup-dreamer,

I appreciate your enthusiasm! I have prior experience with this issue and have already started working on it, which is why I dropped a comment earlier. That said, there are several other open issues—feel free to pick any of those. Looking forward to collaborating!

rakita:
Hey, this is kinda a bigger effor and very sensitive code. Additional bls precompiles needs to be implemented to with this lib that is no_std compatible, old lib still needs to be the main one, but `[zkcrypto/bls12_381](https://github.com/zkcrypto/bls12_381)` can be used in zkvm. This is harder task, but if you can pass all statetest it would get merged

@Ayushdubey86 this is different from PR that you did.

Ayushdubey86:
Sure, looking into this! @rakita

rakita:
@startup-dreamer if you want to take it over, go for it. Just to be sure that were on same page this requires adding new precompiles with new lib, I am not sure if current and this lib have same API

kevaundray:
If I understand this issue correctly -- This should become easier with #2313 and #2314

kevaundray:
Can this be closed, given we merged https://github.com/bluealloy/revm/pull/2316 ?

rakita:
done

PR #2209, opened 2025-03-13
No std bls12 381 precompiles

Ayushdubey86:
hey @rakita , these are not complete changes but this pr is related to [issue](https://github.com/bluealloy/revm/issues/2172) ! 
I am having a conundrum , can you give me slight guidance on where i am going wrong! 

referenced by workspace at `\revm\Cargo.toml`

Caused by:
  failed to parse manifest at `revm\crates\optimism\Cargo.toml`

Caused by:
  optional dependency `bls12_381` is not included in any feature
  Make sure that `dep:bls12_381` is included in one of features in the [features] table.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ auto_impl = "1.2.0"
 bitflags = { version = "2.6.0", default-features = false }
 bitvec = { version = "1", default-features = false }
 blst = "0.3.13"
+bls12_381 = "0.8"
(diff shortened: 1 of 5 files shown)


rakita:
Maybe it would be best to leave this issue for someone more familiar with this precompile.

rakita:
Was close this PR, as it does not go in the direction of solving the issue

Issue #2208, opened 2025-03-13
Return Fatal error on bls precompiles if in no_std


rakita:
Blst does not have a no_std variant. Pending task for support is made here: https://github.com/bluealloy/revm/issues/2172

Until that task is done, we should return PrecompileError::Fatal error for any bls precompiles. Similar to this https://github.com/bluealloy/revm/blob/00d5ec474c15b351843da475f796f6e7101ab92e/crates/precompile/src/lib.rs#L147

All precompiles are here: https://github.com/bluealloy/revm/blob/00d5ec474c15b351843da475f796f6e7101ab92e/crates/precompile/src/bls12_381.rs#L15

Ayushdubey86:
looking into this! but will this be redundant after #2172 ?

rakita:
#2172 is bigger task, so until it is implemented, it would be good to have this PR.

Ayushdubey86:
Raised a dummy [pr](https://github.com/bluealloy/revm/pull/2210). do check this out

PR #2210, opened 2025-03-13
Return Fatal error on bls precompiles if in no_std #2208

Ayushdubey86:
As of now , i have added if condition globally to all precompile together, as i working on [issue](https://github.com/bluealloy/revm/issues/2172) but if you feel the need of more personal error like "c-kzg feature is not enabled" , I'll push it within this pr itself

### DIFF
--- a/crates/precompile/src/bls12_381.rs
+++ b/crates/precompile/src/bls12_381.rs
@@ -19,7 +19,7 @@ pub fn precompiles() -> impl Iterator<Item = PrecompileWithAddress> {
             vec![
                 PrecompileWithAddress(
                     u64_to_address(0x0A),
-                    |_,_| Err(PrecompileError::Fatal("no_std is not supported for BLS12-381 precompiles".into()))
+                    |_,_| Err(PrecompileError::Fatal("no_std is not supported for BLS12-381".into()))
                 )
             ].into_iter()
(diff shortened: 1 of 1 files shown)


crates/precompile/src/bls12_381.rs
```diff
@@ -13,14 +14,24 @@ mod utils;
...
-        g1_add::PRECOMPILE,
-        g1_msm::PRECOMPILE,
-        g2_add::PRECOMPILE,
-        g2_msm::PRECOMPILE,
-        pairing::PRECOMPILE,
-        map_fp_to_g1::PRECOMPILE,
-        map_fp2_to_g2::PRECOMPILE,
-    ]
-    .into_iter()
+    cfg_if! {
```
rakita, line 17: This is in the wrong place it should be in `fn prague` function

crates/precompile/src/bls12_381.rs
```diff
@@ -13,14 +14,24 @@ mod utils;
...
+        if #[cfg(not(feature = "std"))] {  // If no_std is enabled
+            vec![
+                PrecompileWithAddress(
+                    u64_to_address(0x0A),
```
rakita, line 21: All addresses should be covered, not just one

PR #2212, opened 2025-03-14
fix: update block hash serve window

mattsse:
while this is no longer used, the value is still incorrect

we can either remove this entirely or update the value


rakita:
Lets remove it, it is not useful.

rakita:
PR that removes it https://github.com/bluealloy/revm/pull/2228

PR #2215, opened 2025-03-14
feat: allow reuse of API for calculating initial tx gas for tx

Wodann:
This PR:
- adds rich error information for `InvalidTransaction::CallGasCostMoreThanGasLimit` and `InvalidTransaction::GasFloorMoreThanGasLimit`, such that we can inform the user - in the error message - about which values they should use to make their transaction valid.
- Splits off the `calculate_initial_tx_gas_for_tx` function so it can be reused

If approved, this PR should be applied after #2214, as it depends on the change made there.


rakita (approved):
lgtm, pending conflict

PR #2216, opened 2025-03-15
refactor: simplify single UT for OpSpecId compatibility.

jackwener:
simplify UT for OpSpecId compatibility

### DIFF
--- a/crates/database/src/states/bundle_state.rs
+++ b/crates/database/src/states/bundle_state.rs
@@ -6,10 +6,7 @@ use super::{
 };
 use bytecode::Bytecode;
 use core::{mem, ops::RangeInclusive};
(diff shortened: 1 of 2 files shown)


crates/database/src/states/bundle_state.rs
```diff
@@ -40,7 +37,7 @@ pub enum OriginalValuesKnown {
     /// If we don't expect parent blocks to be committed or unwinded from database, this option
     /// should be used.
     Yes,
-    /// Doesn't check original values, see the implementation of [BundleState::into_plain_state] for
```
jackwener, line 43: `into_plain_state` is desperated

rakita (approved):
Cool!

PR #2221, opened 2025-03-16
PrecompileError removal from PrecompileProvider

### LINKED ISSUE #2207: Remove PrecompileError from PrecompileProvider.
Few small things need to be done,
* EVMError should remove ::Precompile.
* PrecompileProvider should return String as a Err variant from `fn run` method. `PrecompileError` should be left as internal.
* `EthPrecompile` `run` method if it contains the `PrecompileError::Fatal(str)` should return Err(str).
* `From<PrecompileError>` should be removed from all errors.

SoarinSkySagar:
closes #2207


rakita:
superseded by https://github.com/bluealloy/revm/pull/2233

PR #2223, opened 2025-03-16
chore: add a safe blst wrapper

kevaundray:
This PR adds a safe blst wrapper and does a few miscellaneous cleanups, so leaving it in draft as it might be better to pull them out into separate PRs if the changes seem reasonable.


crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -39,26 +40,24 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
...
-        // filter points at infinity (and their corresponding scalars) from the input.
-        if slice.iter().all(|i| *i == 0) {
+        // Filter out the point at infinity from the MSM for two reasons:
+        // - The current library `blst` does not handle it gracefully. In particular, it will zero out all other
+        //    points in batch related methods such as `batch_normalization`, if even one of the points is the point at infinity.
+        // - This is an optimization that will allow us to do less work, since any scalar multiplied by the point at infinity is
+        //    essentially a no-op.
+        if encoded_g1_element.iter().all(|i| *i == 0) {
             continue;
         }
```
kevaundray, line 56: This may no longer be needed for safety, since we don't convert from affine to jacobian and then back to affine again, making it strictly an optimization. 

I need to check the blst code and check to see if tests fail to confirm

crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -49,9 +49,9 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         // Filter out points infinity as an optimization, since it is a no-op.
         // Note: Previously, points were being batch converted from Jacobian to Affine. In `blst`, this would essentially,
         // zero out all of the points. Since all points are in affine, this bug is avoided.
-        if encoded_g1_element.iter().all(|i| *i == 0) {
-            continue;
-        }
+        // if encoded_g1_element.iter().all(|i| *i == 0) {
+        //     continue;
+        // }
```
kevaundray, line 54: All tests are still passing for this commit, I'll revert it since it is still useful as an optimization to remove the points at infinity

crates/precompile/src/bls12_381/blst.rs
```diff
@@ -0,0 +1,248 @@
...
+
+use crate::bls12_381_const::SCALAR_LENGTH;
+use blst::{
+    blst_final_exp, blst_fp, blst_fp12, blst_fp12_is_one, blst_fp12_mul, blst_fp2, blst_map_to_g1,
+    blst_map_to_g2, blst_miller_loop, blst_p1, blst_p1_add_or_double_affine, blst_p1_affine,
+    blst_p1_from_affine, blst_p1_to_affine, blst_p2, blst_p2_add_or_double_affine, blst_p2_affine,
+    blst_p2_from_affine, blst_p2_to_affine, MultiPoint,
+};
+
+fn p1_to_affine(p: &blst_p1) -> blst_p1_affine {
```
rakita, line 11: Would be good to put `#[inline]` on all functions

rakita (approved):
lgtm

PR #2226, opened 2025-03-17
added host marker which allows context to implement custom host

omer-gen:


rakita:
@omer-gen can I look at the code I want to get a bigger picture here?

omer-gen:
sure, reached out on x dm

PR #2227, opened 2025-03-17
test(op-precompiles): Add test for g1 msm

emhane:
Add test for clalling g1 msm from op tx


rakita (approved):

rakita:
In general bls tests are covered by ethereum eest suite, so we are duplicating them a little here.

PR #2236, opened 2025-03-17
refactor: make str to SpecId conversion fallible

Wodann:
This matches the way string to `SpecId` conversion is handled to [`OpSpecId`](https://github.com/bluealloy/revm/blob/main/crates/optimism/src/spec.rs#L44), causing errors when invalid hardfork names are provided.

### DIFF
--- a/crates/optimism/src/spec.rs
+++ b/crates/optimism/src/spec.rs
@@ -1,5 +1,4 @@
 use core::str::FromStr;
-
 use revm::primitives::hardfork::{name as eth_name, SpecId, UnknownHardfork};
 
 #[repr(u8)]
--- a/crates/optimism/src/spec.rs
+++ b/crates/optimism/src/spec.rs
@@ -1,4 +1,6 @@
-use revm::primitives::hardfork::{name as eth_name, SpecId};
+use core::str::FromStr;
+
+use revm::primitives::hardfork::{name as eth_name, SpecId, UnknownHardfork};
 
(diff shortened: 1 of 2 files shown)


crates/primitives/src/hardfork.rs
```diff
@@ -73,31 +73,33 @@ pub mod name {
     pub const LATEST: &str = "Latest";
 }
 
-impl From<&str> for SpecId {
-    fn from(name: &str) -> Self {
+impl TryFrom<&str> for SpecId {
+    type Error = ();
```
mattsse, line 77: I think this is just `FromStr` now

can we introduce an error here because otherwise this can't be used with `.parse()?` for example

mattsse (dismissed):
let's also do this for opspecid if missing

crates/primitives/src/hardfork.rs
Wodann: Addressed by https://github.com/bluealloy/revm/pull/2236/commits/a866ac05c7c59a110b9bd2707ff0762220129979

rakita (approved):
lgtm

crates/optimism/src/spec.rs
```diff
@@ -1,4 +1,6 @@
-use revm::primitives::hardfork::{name as eth_name, SpecId};
+use core::str::FromStr;
+
```
rakita, line 2:
```suggestion
```

PR #2237, opened 2025-03-17
fix: DepositTx Identifier is 126 (`0x7e`) not `0x7f`

refcell:
### Description

Fixes deposit tx identifier to be the correct `0x7e` instead of `0x7f`.


rakita (approved):
lgtm

PR #2242, opened 2025-03-18
test(op-precompiles): clean up op tx tests

emhane:
Remove redundancy in precompile test on op tx


crates/precompile/src/secp256r1.rs
```diff
@@ -12,8 +12,11 @@ use crate::{
 use p256::ecdsa::{signature::hazmat::PrehashVerifier, Signature, VerifyingKey};
 use primitives::{Bytes, B256};
 
+/// Address of secp256r1 precompile.
+pub const P256VERIFY_ADDRESS: u64 = 256;
```
rakita, line 16: `: Address = u64_to_address(256);` would work here

rakita: Lets do this in follow up

rakita (approved):

PR #2243, opened 2025-03-18
test(op-precompiles): remove redundancy g1 and g2 add tests

emhane:

Based on https://github.com/bluealloy/revm/pull/2253

Removes redundancy in g1 add and g2 add tests

### DIFF
--- a/crates/optimism/src/evm.rs
+++ b/crates/optimism/src/evm.rs
@@ -429,22 +429,39 @@ mod tests {
         ));
     }
 
-    #[test]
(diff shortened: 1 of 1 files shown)


emhane:
gas for g1 add should be computable as `initial-gas` + `g1-add-gas-base-fee`, however, when tx input is non-empty, these tests fail with too low gas error before calling precompile:
```
Transaction(Base(GasFloorMoreThanGasLimit { gas_floor: 31200, gas_limit: 25471 }))
```

PR #2246, opened 2025-03-18
chore: add documentation for the gas related constants for EIP2537

kevaundray:


crates/precompile/src/bls12_381_const.rs
```diff
@@ -1,19 +1,66 @@
...
+pub const G1_MSM_BASE_GAS_FEE: u64 = 12000;
+/// MSM_MULTIPLIER specifies the division constant that is used to determine the
+/// gas needed to compute an MSM.
+///
+/// The cost to do an MSM is determined by the formula:
+///    (k * MSM_BASE_GAS_FEE * DISCOUNT\[k\]) // MSM_MULTIPLIER
+/// where k is the number of point-scalar pairs.
+///
+/// Note: If `k` is more than the size of the discount table, then
+/// the last value in the discount table is chosen.
```
kevaundray, line 32: This is also a coincidence imo -- Ideally we have a MAX_DISCOUNT constant like in the EIP instead of relying on the last value in the discount table to match the MAX_DISCOUNT constant

rakita (approved):

PR #2247, opened 2025-03-18
chore: Add invariant test for G1/G2 Scalar multiplication

kevaundray:
The G1/G2 scalar multiplication gas costs are specified in two different places in the EIP in two different ways.

They are equal when the first value in the discount table is equal to the MSM_MULTIPLIER. It is not the most robust way to check this invariant because you can compute the cost for G1/2_MUL in a more convoluted way. 

If a more convoluted way is chosen, then this test should start to fail.

What this means is that if we wanted to just change the cost for G1/2_MUL we would need to change the MSM_BASE_GAS_FEE. It was done this way because there is _some_ relationship between the cost for a scalar multiplication and the cost to do an MSM which is multiple scalar multiplications.


kevaundray:
Blocking as I am not sure that the EIP-2537 tests are running

rakita:
This is a unit test and it does not touch the eth state tests, it is safe to merge imo. We had problem with CI eof-verification that is fixed on the main

PR #2248, opened 2025-03-18
chore: special case G1 and G2 scalar multiplication

kevaundray:
When an G1_MSM or G2_MSM is called with a single input, we can us the simpler scalar multiplication algorithm instead of the MSM algorithm.


kevaundray:
Same comment as here: https://github.com/bluealloy/revm/pull/2247#issuecomment-2734713967

rakita:
Part of the test that runs statetest is passing, EOF validation that is failing is fixed on main

PR #2249, opened 2025-03-18
feat: Return Fatal error on bls precompiles if in no_std

### LINKED ISSUE #2208: Return Fatal error on bls precompiles if in no_std

nemnem7:
closes #2208

### DIFF
--- a/crates/precompile/src/bls12_381_utils.rs
+++ b/crates/precompile/src/bls12_381_utils.rs
@@ -1,5 +1,8 @@
-use crate::bls12_381_const::MSM_MULTIPLIER;
-
+use crate::bls12_381_const::{
+    G1_ADD_ADDRESS, G1_MSM_ADDRESS, G2_ADD_ADDRESS, G2_MSM_ADDRESS, MAP_FP2_TO_G2_ADDRESS,
+    MAP_FP_TO_G1_ADDRESS, MSM_MULTIPLIER, PAIRING_ADDRESS,
+};
+use crate::{u64_to_address, PrecompileError, PrecompileWithAddress};
 /// Implements the gas schedule for G1/G2 Multiscalar-multiplication assuming 30
 /// MGas/second, see also: <https://eips.ethereum.org/EIPS/eip-2537#g1g2-multiexponentiation>
 #[inline]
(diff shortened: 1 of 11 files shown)


crates/precompile/src/lib.rs
```diff
@@ -162,14 +162,37 @@ impl Precompiles {
...
+                    PrecompileWithAddress(
+                        u64_to_address(PAIRING_ADDRESS),
+                        |_,_| Err(PrecompileError::Fatal("no_std is not supported for BLS12-381 precompiles".into()))),
+                    PrecompileWithAddress(
+                        u64_to_address(MAP_FP_TO_G1_ADDRESS),
+                        |_,_| Err(PrecompileError::Fatal("no_std is not supported for BLS12-381 precompiles".into()))),
+                    PrecompileWithAddress(
+                        u64_to_address(MAP_FP2_TO_G2_ADDRESS),
+                        |_,_| Err(PrecompileError::Fatal("no_std is not supported for BLS12-381 precompiles".into())))
+                    )
```
rakita, line 192: Move this to the separate function `bls12_381_precompiles_not_supported() -> Vec<_>` so it can be reused and it makes things cleaner

crates/precompile/src/lib.rs
```diff
@@ -162,14 +162,37 @@ impl Precompiles {
...
-
-            // Don't include BLS12-381 precompiles in no_std builds.
-            #[cfg(feature = "blst")]
-            let precompiles = {
-                let mut precompiles = precompiles;
-                precompiles.extend(bls12_381::precompiles());
-                precompiles
-            };
+            cfg_if! {
+                if #[cfg(not(feature = "std"))] {  // If no_std is enabled
```
rakita, line 166:
```suggestion
                if #[cfg(not(feature = "blst"))] {  // if blst is enabled {
                        bls12_381::precompiles()
                } else {
                        bls12_381_utils:: bls12_381_precompiles_not_supported()
                }
```

rakita: It should look something like this

nemnem7: Done

crates/precompile/src/lib.rs
nemnem7: Done

crates/precompile/src/lib.rs
```diff
@@ -161,15 +161,17 @@ impl Precompiles {
...
-            let precompiles = {
-                let mut precompiles = precompiles;
-                precompiles.extend(bls12_381::precompiles());
-                precompiles
-            };
+            let mut precompiles = Self::cancun().clone();
+
+            let bls;
+            cfg_if! {
+                      if #[cfg(not(feature = "blst"))] {  // if blst is enabled {
```
rakita, line 168:
```suggestion
                      if #[cfg(feature = "blst")] {  // if blst is enabled {
```

crates/precompile/src/bls12_381_utils.rs
```diff
@@ -13,3 +16,48 @@ pub fn msm_required_gas(k: usize, discount_table: &[u16], multiplication_cost: u
 
     (k as u64 * discount * multiplication_cost) / MSM_MULTIPLIER
 }
+
+pub fn bls12_381_precompiles_not_supported() -> Vec<PrecompileWithAddress> {
+    vec![
+        PrecompileWithAddress(u64_to_address(0x0a), |_, _| {
```
rakita, line 22: 0x0a missing, add constant.

nemnem7: Done

crates/precompile/src/lib.rs
nemnem7: Done

nemnem7:
I updated the addresses' type (they are now of type Address, not u64) and function arguments in optimism/src/evm.rs where the addresses are used as function arguments

crates/precompile/src/lib.rs
```diff
@@ -161,15 +161,16 @@ impl Precompiles {
...
-            };
+            let mut precompiles = Self::cancun().clone();
+
+            cfg_if! {
+                      if #[cfg(feature = "blst")] {  // if blst is enabled {
+                              let bls = bls12_381::precompiles();
+                      } else {
+                              let bls = bls12_381_utils:: bls12_381_precompiles_not_supported();
+                      }
+            }
```
rakita, line 172:
```suggestion
            cfg_if! {
                  if #[cfg(feature = "blst")] {  // if blst is enabled {
                      let bls = bls12_381::precompiles();
                  } else {
                      let bls = bls12_381_utils:: bls12_381_precompiles_not_supported();
                  }
            }
```

rakita: lint format

PR #2257, opened 2025-03-19
fix(optimism): Operator Fee Calculation & Deposit Identifier

refcell:
### Description

- Fixes deposit tx identifier to be `0x7E` instead of `0x7F`.
- Fixes the operator fee calculation. h/t @leruaa 🎉 

Action tests against the golang implementation are demonstrated to pass in https://github.com/op-rs/kona/pull/1277

### DIFF
--- a/crates/revm/src/optimism/handler_register.rs
+++ b/crates/revm/src/optimism/handler_register.rs
@@ -68,11 +68,6 @@ pub fn validate_env<SPEC: Spec, DB: Database>(env: &Env) -> Result<(), EVMError<
 pub fn validate_tx_against_state<SPEC: Spec, EXT, DB: Database>(
     context: &mut Context<EXT, DB>,
 ) -> Result<(), EVMError<DB::Error>> {
-    // No validation is needed for deposit transactions, as they are pre-verified on L1.
-    if context.evm.inner.env.tx.optimism.source_hash.is_some() {
-        return Ok(());
-    }
-
     // storage l1 block info for later use. l1_block_info is cleared after execution.
     if context.evm.inner.l1_block_info.is_none() {
         // the L1-cost fee is only computed for Optimism non-deposit transactions.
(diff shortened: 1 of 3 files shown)


refcell:
Looks like there's 1 more action test left to fix before we land these changes

crates/revm/src/optimism/l1block.rs
```diff
@@ -192,21 +192,27 @@ impl L1BlockInfo {
     /// Introduced in isthmus. Prior to isthmus, the operator fee is always zero.
     pub fn operator_fee_charge(&self, input: &[u8], gas_limit: U256, spec_id: SpecId) -> U256 {
         // If the input is a deposit transaction or empty, the default value is zero.
-        if input.is_empty() || input.first() == Some(&0x7F) {
+        if input.is_empty() || input.first() == Some(&0x7E) {
```
clabby, line 195: Why are we considering the operator fee to be zero based off of the input size? Isn't that not super important for what succinct is trying to charge for (compute units, rather than DA cost?)

crates/revm/src/optimism/handler_register.rs
```diff
@@ -279,14 +279,13 @@ pub fn reimburse_caller<SPEC: Spec, EXT, DB: Database>(
...
-        let caller_account = context
-            .evm
-            .inner
-            .journaled_state
-            .load_account(context.evm.inner.env.tx.caller, &mut context.evm.inner.db)?;
+    if !is_deposit {
+        mainnet::reimburse_caller::<SPEC, EXT, DB>(context, gas)?;
+    }
+
+    if SPEC::SPEC_ID.is_enabled_in(SpecId::ISTHMUS) {
```
clabby, line 288: This doesn't quite make sense; We don't charge the operator fee on deposit transactions, so why would there be a refund?

crates/revm/src/optimism/l1block.rs
```diff
@@ -192,21 +192,27 @@ impl L1BlockInfo {
...
             .operator_fee_scalar
             .expect("Missing operator fee scalar for isthmus L1 Block");
         let operator_fee_constant = self
             .operator_fee_constant
             .expect("Missing operator fee constant for isthmus L1 Block");
 
-        let product = gas_limit.saturating_mul(operator_fee_scalar)
-            / (U256::from(OPERATOR_FEE_SCALAR_DECIMAL));
+        let product =
+            gas.saturating_mul(operator_fee_scalar) / (U256::from(OPERATOR_FEE_SCALAR_DECIMAL));
```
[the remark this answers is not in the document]
clabby: Ah I see - nice, my mistake.

crates/revm/src/optimism/l1block.rs
leruaa: @yuwen01 Do you remember why you added `input.is_empty()`?

crates/revm/src/optimism/handler_register.rs
refcell: I'm not convinced there is any check in the operator cost func for deposit txs

rakita (approved):

PR #2258, opened 2025-03-19
chore: Adds a wrapper around substrate-bn for EIP196

kevaundray:


crates/precompile/src/bn128/substrate.rs
```diff
@@ -0,0 +1,144 @@
...
+/// and creates a G2 point.
+///
+/// Note: The point at infinity which is represented as (0,0) is
+/// handled specifically because `AffineG2` is not capable of
+/// representing such a point.
+/// In particular, when we convert from `AffineG2` to `G2`, the point
+/// will be (0,0,1) instead of (0,1,0)
+#[inline]
+fn new_g2_point(x: Fq2, y: Fq2) -> Result<G2, PrecompileError> {
+    // TODO : Check whether or not we need these zero checks
```
kevaundray, line 68: This TODO was there before, we can remove it since the Note in the comments specify that we do need the check

crates/precompile/src/bn128/substrate.rs
```diff
@@ -0,0 +1,144 @@
...
+    let ba = read_fq2(&input[0..FQ2_LEN])?;
+    let bb = read_fq2(&input[FQ2_LEN..2 * FQ2_LEN])?;
+    new_g2_point(ba, bb)
+}
+
+/// pairing_check performs a pairing check on a list of G1 and G2 point pairs and
+/// returns true if the result is equal to the identity element.
+///
+/// Note: If the input is empty, this function returns true.
+/// This is different to EIP2537 which disallows the empty input.
```
kevaundray, line 137: I guess this might be confusing to users since bn128 does something on empty inputs and bls12-381 does a different thing

crates/precompile/src/bn128/substrate.rs
rakita: Makes sense lets remove it. That comment is probably there from few years ago :)

crates/precompile/src/bn128/substrate.rs
rakita: The comment you left is explanatory.

rakita (approved):
cool, good cleanup!

crates/precompile/src/bn128/substrate.rs
rakita:
```suggestion
```

kevaundray:
This should hopefully make it a bit easier to swap out the bn128 library in the future because _most_ of the API that a library needs to expose is contained in that substrate.rs file :)

rakita:
> This should hopefully make it a bit easier to swap out the bn128 library in the future because most of the API that a library needs to expose is contained in that substrate.rs file :)

Nice, this is something we need. We tried to integrate https://github.com/matter-labs/eip1962 behind the feature flag for std env, as it is faster, this would make this easier.

kevaundray:
> > This should hopefully make it a bit easier to swap out the bn128 library in the future because most of the API that a library needs to expose is contained in that substrate.rs file :)
> 
> Nice, this is something we need. We tried to integrate https://github.com/matter-labs/eip1962 behind the feature flag for std env, as it is faster, this would make this easier.

I can take a look at this -- we probably need to expose more methods to the wrapper

PR #2264, opened 2025-03-20
chore: Add `g1_mul`, `g1_add` and `read_scalar` methods into substrate wrapper for bn128

kevaundray:
This adds point_mul, point_add and read_scalar. They are implicitly in the public API of the bn254 curve.

### DIFF
--- a/crates/precompile/src/bn128/substrate.rs
+++ b/crates/precompile/src/bn128/substrate.rs
@@ -1,4 +1,4 @@
-use super::{FQ2_LEN, FQ_LEN, G1_LEN};
+use super::{FQ2_LEN, FQ_LEN, G1_LEN, SCALAR_LEN};
 use crate::PrecompileError;
 use bn::{AffineG1, AffineG2, Fq, Fq2, Group, Gt, G1, G2};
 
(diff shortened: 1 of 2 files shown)


crates/precompile/src/bn128/substrate.rs
```diff
@@ -129,6 +129,33 @@ pub(super) fn read_g2_point(input: &[u8]) -> Result<G2, PrecompileError> {
...
+#[inline]
+pub(super) fn read_scalar(input: &[u8]) -> bn::Fr {
+    assert_eq!(
+        input.len(),
+        SCALAR_LEN,
+        "unexpected scalar length. got {}, expected {SCALAR_LEN}",
+        input.len()
+    );
+    // `Fr::from_slice` can only fail when the length is not `SCALAR_LEN`.
+    bn::Fr::from_slice(input).unwrap()
```
kevaundray, line 144: Originally this was returning an error, however we would need to add a new enum variant into PrecompileError for ScalarLength invalid. This would be a breaking change for those who are matching on enums so I chose to do what was already happening (panicking when length is not 32)

crates/precompile/src/bn128/substrate.rs
```diff
@@ -129,6 +129,33 @@ pub(super) fn read_g2_point(input: &[u8]) -> Result<G2, PrecompileError> {
     new_g2_point(ba, bb)
 }
 
+/// Reads a scalar from the input slice
+///
+/// Note: The scalar does not need to be canonical.
```
rakita, line 134:
```suggestion
/// Note: The scalar does not need to be canonical.
///
/// # Panics
///
/// If `input.len()` is not equal to [`SCALAR_LEN`].
```

PR #2267, opened 2025-03-20
test(op-test-cov): Add test for serializing deposit transaction parts

emhane:
Improves test coverage for optimism crate

### DIFF
--- a/crates/optimism/src/transaction/deposit.rs
+++ b/crates/optimism/src/transaction/deposit.rs
@@ -31,3 +31,26 @@ impl DepositTransactionParts {
         }
     }
 }
+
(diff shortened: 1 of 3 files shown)


rakita (approved):

crates/optimism/src/transaction/deposit.rs
```diff
@@ -31,3 +31,26 @@ impl DepositTransactionParts {
         }
     }
 }
+
+#[cfg(all(test, feature = "serde"))]
+mod tests {
+    use super::*;
+    use alloy_primitives::b256;
```
rakita, line 38: You can use `revm::primitives:b256` for this

PR #2268, opened 2025-03-20
refactor: issue #2213

### LINKED ISSUE #2213: `Option<B256>` for `DepositTransaction::source_hash`
The [`DepositTransaction`](https://github.com/bluealloy/revm/blob/f742477abcac1baaf2608c59096bdb09bf29a521/crates/optimism/src/transaction/deposit.rs#L8) trait contains a `fn source_hash(&self) -> B256;` to retrieve the source hash of an OP deposit transaction.

This trait is implemented for the overall [`OpTransaction`](https://github.com/bluealloy/revm/blob/main/crates/optimism/src/transaction/abstraction.rs#L121), which means that I can also call `transaction.source_hash()` if `transaction` is an `OpTransaction` for e.g. a blob transaction.

This made me wonder whether it wouldn't be better for the API to be:

```rs
fn source_hash(&self) -> Option<&B256>
```

since that would signal whether the transaction has a source hash (only true for deposit transaction).

WDYT, @rakita? If you agree, I can implement this change.

Wodann:
Resolves #2213

### DIFF
--- a/crates/optimism/src/handler.rs
+++ b/crates/optimism/src/handler.rs
@@ -2,10 +2,7 @@
 use crate::{
     api::exec::OpContextTr,
     constants::{BASE_FEE_RECIPIENT, L1_FEE_RECIPIENT, OPERATOR_FEE_RECIPIENT},
-    transaction::{
-        deposit::{DepositTransaction, DEPOSIT_TRANSACTION_TYPE},
-        OpTransactionError, OpTxTr,
-    },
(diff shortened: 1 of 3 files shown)


rakita:
It was implemented here: https://github.com/bluealloy/revm/pull/2265

Wodann:
My PR also made the `is_system_tx` optional to signal it's a deposit-transaction-specific field.

Is that something you'd still want to include in `main`, @rakita? If so, I can rebase and update my branch.

PR #2266, opened 2025-03-20
feat: Add a wrapper for matter-labs-eip1962 for EIP196

kevaundray:
This PR does not change make any breaking changes as the default library is still sustrate-bn.

The main difference now is that the substrate-bn library is optional (this follows the same strategy that the repository uses when there are two libraries for a given functionality). 

This means that substrate-bn is now explciitly specified as in a crates default features in order to enable it. Moreover, config macros were added to various parts of the code in order to pass cargo checks and tests that run with `--no-default-features`.

PR #2270, opened 2025-03-20
feat!: Make matter-labs library the default

kevaundray:
This is based off of #2266 -- Opening this PR:

- To run the benchmarks on the matter-labs impl in CI
- See the difference in performance

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1556,6 +1556,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "eth_pairings"
+version = "0.6.0"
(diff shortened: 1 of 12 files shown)


rakita:
This is awesome, I am even thinking of leaving it as default. Would like to do this after v20.0.0 gets released (in a few days)

kevaundray:
I guess one issue with releases is that https://github.com/matter-labs/eip1962 has not made a release in a while: The latest release is 0.2 on crates.io whereas the version on their main branch is 0.6.0

kevaundray:
Closing as we plan to use arkworks instead

PR #2271, opened 2025-03-20
chore: add early exit for no-ops in EIP1962 and EIP2537

kevaundray:
This PR avoids doing redundant work when the computation for the pairing or MSM will result in a no-op


rakita (approved):
lgtm

PR #2276, opened 2025-03-21
chore: remove wrong `&mut` and duplicated spec

jackwener:
- `&mut` and spec is useless

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -316,10 +316,12 @@ where
         let caller_balance = context
             .journal()
             .load_account(inputs.caller)?
-            .map(|a| a.info.balance);
+            .data
+            .info
+            .balance;
 
         // Check if caller has enough balance to send to the created contract.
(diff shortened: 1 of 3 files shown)


crates/handler/src/frame.rs
```diff
@@ -316,10 +316,12 @@ where
         let caller_balance = context
             .journal()
             .load_account(inputs.caller)?
-            .map(|a| a.info.balance);
+            .data
+            .info
+            .balance;
```
mattsse, line 321: hmm, this went from `info.balance.data` to `data.info.balance`?

or am I missing something

mattsse (commented):
ty, all of these lgtm

one q re balance @rakita

crates/handler/src/frame.rs
rakita: This is fine, `.map` is from `StateLoad` struck that maps data to a different type.

rakita (approved):

PR #2277, opened 2025-03-21
fix: sstore should make account touched

jackwener:
* Added a call to `Self::touch_account` to ensure accounts are touched, which is important for pre-SpuriousDragon accounts.


rakita (changes requested):
This is not needed as sstore instruction is called only from account that are executing and they are always touched

jackwener:
> This is not needed as sstore instruction is called only from account that are executing and they are always touched

Thanks for your explanation

PR #2285, opened 2025-03-21
feat: (Try) Add arkworks wrapper for EIP1962

kevaundray:
This PR is not mergeable because I have essentially just removed the bn wrapper and not implemented the arkworks wrapper behind a feature flag.

Mainly looking to see if tests vectors pass in CI and the performance increase or decrease with arkworks

### DIFF
--- a/crates/precompile/src/bn128.rs
+++ b/crates/precompile/src/bn128.rs
@@ -4,11 +4,11 @@ use crate::{
 };
 use std::vec::Vec;
 
-use substrate::{
+use arkworks::{
     encode_g1_point, g1_point_add, g1_point_mul, pairing_check, read_g1_point, read_g2_point,
     read_scalar,
 };
-mod substrate;
+mod arkworks;
 
 pub mod add {
     use super::*;
--- a/crates/precompile/src/bn128/arkworks.rs
+++ b/crates/precompile/src/bn128/arkworks.rs
@@ -173,9 +173,7 @@ pub(super) fn read_scalar(input: &[u8]) -> Fr {
         "unexpected scalar length. got {}, expected {SCALAR_LEN}",
         input.len()
     );
-    let mut input = input.to_vec();
-    input.reverse();
-    Fr::deserialize_uncompressed_unchecked(&input[..]).unwrap() // TODO: Check the reasons as to why this can fail with unchecked
+    Fr::from_be_bytes_mod_order(input)
 }
 
 /// Performs point addition on two G1 points.
--- a/crates/precompile/src/bn128/arkworks.rs
+++ b/crates/precompile/src/bn128/arkworks.rs
@@ -18,7 +18,7 @@ use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 /// Panics if the input is not at least 32 bytes long.
 #[inline]
 fn read_fq(input_be: &[u8]) -> Result<Fq, PrecompileError> {
-    // TODO: maybe assert input_be is 32 bytes, then change panic note to _exactly_ 32 bytes?
+    assert_eq!(input_be.len(), FQ_LEN, "input must be {FQ_LEN} bytes");
 
     let mut input_le = [0u8; FQ_LEN];
     input_le.copy_from_slice(input_be);
--- a/crates/precompile/src/bn128/arkworks.rs
+++ b/crates/precompile/src/bn128/arkworks.rs
@@ -1,6 +1,6 @@
 use super::{FQ2_LEN, FQ_LEN, G1_LEN, SCALAR_LEN};
 use crate::PrecompileError;
-use std::vec::Vec; // TODO: Remove this import, `cargo clippy --no-default-features` needs it to disambiguate
+use std::vec::Vec;
 
 use ark_bn254::{Bn254, Fq, Fq2, Fr, G1Affine, G1Projective, G2Affine};
 use ark_ec::{pairing::Pairing, AffineRepr, CurveGroup};
--- a/crates/precompile/src/bn128/arkworks.rs
+++ b/crates/precompile/src/bn128/arkworks.rs
@@ -121,25 +121,25 @@ pub(super) fn read_g1_point(input: &[u8]) -> Result<G1Affine, PrecompileError> {
 #[inline]
 pub(super) fn encode_g1_point(point: G1Affine) -> [u8; G1_LEN] {
     let mut output = [0u8; G1_LEN];
+    let Some((x, y)) = point.xy() else {
+        return output;
+    };
 
-    if let Some((x, y)) = point.xy() {
-        // TODO: check if serialize_uncompressed ever writes less than FQ_LEN
-        let mut x_bytes = [0u8; FQ_LEN];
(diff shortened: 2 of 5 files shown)


crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
+use super::{FQ2_LEN, FQ_LEN, G1_LEN, SCALAR_LEN};
+use crate::PrecompileError;
+use std::vec::Vec; // TODO: Remove this import, `cargo clippy --no-default-features` needs it to disambiguate
+
```
kevaundray, line 4: I haven't investigated this error, though its happening on another PR too

crates/precompile/src/bn128.rs
```diff
@@ -4,11 +4,11 @@ use crate::{
 };
 use std::vec::Vec;
 
-use substrate::{
+use arkworks::{
     encode_g1_point, g1_point_add, g1_point_mul, pairing_check, read_g1_point, read_g2_point,
     read_scalar,
 };
-mod substrate;
+mod arkworks;
```
kevaundray, line 11: This was mainly a quick hacky way to enable arkworks -- alternative would have been to branch off #2266 , which would have been overkill to just test CI performance

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Panics if the input is not at least 32 bytes long.
+#[inline]
+fn read_fq(input_be: &[u8]) -> Result<Fq, PrecompileError> {
+    // TODO: maybe assert input_be is 32 bytes, then change panic note to _exactly_ 32 bytes?
+
+    let mut input_le = [0u8; FQ_LEN];
+    input_le.copy_from_slice(input_be);
+
+    // Reverse in-place to convert from big-endian to little-endian.
+    input_le.reverse();
```
kevaundray, line 27: arkworks uses little endian while eip1962 uses big endian, so we need to reverse the input

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+        y.serialize_uncompressed(&mut y_bytes[..])
+            .expect("Failed to serialize x coordinate");
+
+        // Convert to big endian by reversing the bytes.
+        x_bytes.reverse();
+        y_bytes.reverse();
+
+        // Place x in the first half, y in the second half.
+        output[0..FQ_LEN].copy_from_slice(&x_bytes);
+        output[FQ_LEN..(FQ_LEN * 2)].copy_from_slice(&y_bytes);
```
kevaundray, line 140: We could encode x into x_bytes, then write to output and then re-use x_bytes -- though this stack allocation is fast, so its not necessary imo

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+/// will be (0,0,1) instead of (0,1,0)
+#[inline]
+fn new_g1_point(px: Fq, py: Fq) -> Result<G1Affine, PrecompileError> {
+    if px.is_zero() && py.is_zero() {
+        Ok(G1Affine::zero())
+    } else {
+        let point = G1Affine::new_unchecked(px, py);
+        if !point.is_on_curve() || !point.is_in_correct_subgroup_assuming_on_curve() {
+            return Err(PrecompileError::Bn128AffineGFailedToCreate);
+        }
```
kevaundray, line 67: For future reference: We cannot use `G1Affine::new` because that triggers an assert if the point is not on the curve.

kevaundray, line 67: Another small note `||` are short-circuit operations.

- `a || b` will not check `b` if `a` is true

- `a` in this case is `!point.is_on_curve()` so if the point is not on the curve, meaning it is invalid, then we don't check if it is in the correct subgroup. 

Perhaps it would be clearer to write this as:

```rust
if !point.is_on_curve() {
   return return Err(PrecompileError::Bn128AffineGFailedToCreate);
}
if !point.is_in_correct_subgroup_assuming_on_curve() {
   return Err(PrecompileError::Bn128AffineGFailedToCreate);
}
```

crates/precompile/src/bn128/arkworks.rs
rakita: This is fine to leave it as is (If you mean `use std::vec::Vec`)

crates/precompile/src/bn128/arkworks.rs
rakita: Would add this as inline comment, just to point why `new_unchecked` is used

rakita: This is fine, short-circuit operations are okay,

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+/// handled specifically because `AffineG2` is not capable of
+/// representing such a point.
+/// In particular, when we convert from `AffineG2` to `G2`, the point
+/// will be (0,0,1) instead of (0,1,0)
+#[inline]
+fn new_g2_point(x: Fq2, y: Fq2) -> Result<G2Affine, PrecompileError> {
+    let point = if x.is_zero() && y.is_zero() {
+        G2Affine::zero()
+    } else {
+        let point = G2Affine::new_unchecked(x, y);
```
rakita, line 88:
```suggestion
        // We cannot use `G1Affine::new` because that triggers an assert if the point is not on the curve.
        let point = G2Affine::new_unchecked(x, y);
```

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Converts a G1 point in Jacobian coordinates to affine coordinates and
+/// serializes the x and y coordinates as big-endian byte arrays.
+///
+/// Note: If the point is the point at infinity, this function returns
+/// all zeroes.
+#[inline]
+pub(super) fn encode_g1_point(point: G1Affine) -> [u8; G1_LEN] {
+    let mut output = [0u8; G1_LEN];
+
+    if let Some((x, y)) = point.xy() {
```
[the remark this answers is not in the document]
rakita: Nit, this would return right away.
```rust
let Some((x, y)) = point.xy() else {
     return output;
}
```

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,213 @@
...
+/// Takes a byte slice and attempts to interpret the first 32 bytes as an
+/// elliptic curve field element. Returns an error if the bytes do not form
+/// a valid field element.
+///
+/// # Panics
+///
+/// Panics if the input is not at least 32 bytes long.
+#[inline]
+fn read_fq(input_be: &[u8]) -> Result<Fq, PrecompileError> {
+    // TODO: maybe assert input_be is 32 bytes, then change panic note to _exactly_ 32 bytes?
```
rakita, line 21: It would be a good hint if we have this.

PR #2286, opened 2025-03-21
fix: ensure transactions succeed when balance check is disabled

Wodann:
I discovered that code for the "disable balance check" feature was removed in the previous refactors, here.

This PR fixes the functionality.

### DIFF
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -37,7 +37,7 @@ pub fn validate_tx_against_state<
     let account = context.journal().load_account_code(tx_caller)?;
     let account = account.data.info.clone();
 
-    validate_tx_against_account(&account, context, U256::ZERO)?;
+    validate_tx_against_account::<_, ERROR>(&account, context, U256::ZERO)?;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


crates/handler/src/validation.rs
```diff
@@ -279,11 +282,20 @@ pub fn validate_tx_against_account<CTX: ContextTr>(
 
     // Check if account has enough balance for `gas_limit * max_fee`` and value transfer.
     // Transfer will be done inside `*_inner` functions.
-    if balance_check > account.balance && !context.cfg().is_balance_check_disabled() {
-        return Err(InvalidTransaction::LackOfFundForMaxFee {
-            fee: Box::new(balance_check),
-            balance: Box::new(account.balance),
-        });
+    if balance_check > account.balance {
+        if context.cfg().is_balance_check_disabled() {
```
rakita, line 286: Previous code did remove balance check, but it would in deduct_caller reduce it to zero, is there a need for the balance to remain unchanged?

Wodann: The current implementation causes an `OutOfFunds` when deducting the balance, in later execution calls.

I can try to find all missed checks there?

rakita: hm it is about value that is sent.

rakita: Lets introduce a logic inside deduc_caller that would leave at least `value` amount of balance if cfg is enabled

Wodann: Done in https://github.com/bluealloy/revm/pull/2286/commits/885b273a309807691e47ad64b9670ef0cd014d7a

PR #2287, opened 2025-03-21
chore: Add MUL benchmark for EIP1962

kevaundray:
This is a quick way to add benchmark for BN_MUL. 

It takes the first point from BN_ADD, and uses the x coordinate of the second point from BN_ADD as the scalar.


rakita (approved):
lgtm

PR #2305, opened 2025-03-25
feat: Add a wrapper for arkworks for EIP196

kevaundray:
This PR supercedes and is based off of #2266 .

It uses the same code that #2266, however it replaces the matter-labs implementaton for EIP9162 with an arkworks implementation. This arkworks implementation was benchmarked in #2285 -- the code in the arkworks.rs file in that PR was essentially copy and pasted.


rakita (approved):
lgtm, amazing work @kevaundray!

crates/precompile/src/bn128/arkworks.rs
```diff
@@ -0,0 +1,214 @@
...
+/// handled specifically because `AffineG1` is not capable of
+/// representing such a point.
+/// In particular, when we convert from `AffineG1` to `G1`, the point
+/// will be (0,0,1) instead of (0,1,0)
+#[inline]
+fn new_g1_point(px: Fq, py: Fq) -> Result<G1Affine, PrecompileError> {
+    if px.is_zero() && py.is_zero() {
+        Ok(G1Affine::zero())
+    } else {
+        let point = G1Affine::new_unchecked(px, py);
```
rakita, line 64:
```suggestion
        // We cannot use `G1Affine::new` because that triggers an assert if the point is not on the curve.
        let point = G1Affine::new_unchecked(px, py);
```

PR #2311, opened 2025-03-25
feat: Add JournalInner

rakita:
Introduce JournalInner that extracts field and logic from journal


crates/context/src/journaled_state/init.rs
```diff
@@ -1,49 +0,0 @@
-use super::Journal;
-use database_interface::EmptyDB;
-
-/// A cloneable version of JournaledState that uses EmptyDB.
-/// Used to clone the journaled state and for initialization of new journaled state.
-pub type JournalInit = Journal<EmptyDB>;
```
zerosnacks, line 6: The removal of this type may be unintended, or is it not the correct way to initialize with an EmptyDB anymore

rakita: There is no need for it as `JournalInner` now contains all journaled statea and has same purpose as `JournalInit`

rakita:
There was no major changes to logic but it degraded by few us. The part that is shown in codspeed as red, I would like to optimize in different PR, to clone precompiles only when spec is different. Both of regresions seems related to HashMap inserts but it does not show what part of code it is.

PR #2312, opened 2025-03-26
test: add EIP-170 contract code size limit tests

jackwener:
* Refactored the `test_eip3860_initcode_size_limit` test by extracting the contract deployment logic into a new helper function `deploy_contract`.
* Added new tests for EIP-3860 to verify both failure and success scenarios when the contract size exceeds or meets the `MAX_INITCODE_SIZE`.
* Added new tests for EIP-170 to verify both failure and success scenarios when the contract size exceeds or meets the limit of 0x6000 bytes.
* Added tests for EIP-170 to verify the behavior of the CREATE opcode when creating sub-contracts from a factory contract, including both failure and success scenarios.


rakita (approved):
In general, those cases are covered by eest and ethereum/tests statetests. But those are good tests let us merge them.

PR #2313, opened 2025-03-26
chore: Move all blst related methods into blst wrapper

kevaundray:
This is mostly just a code move so that only the blst wrapper is importing from the blst crate


crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -39,7 +33,7 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         return Err(PrecompileError::OutOfGas);
     }
 
-    let mut g1_points: Vec<blst_p1_affine> = Vec::with_capacity(k);
```
kevaundray, line 42: This type can be inferred, removed it so we don't need to explicitly import blst

crates/precompile/src/bls12_381/g2.rs
```diff
@@ -1,129 +0,0 @@
...
-        return Err(PrecompileError::Other(format!(
-            "Input should be {PADDED_G2_LENGTH} bytes, was {}",
-            input.len()
-        )));
-    }
-
-    let mut input_fps = [&[0; FP_LENGTH]; 4];
-    for i in 0..4 {
-        input_fps[i] = remove_padding(&input[i * PADDED_FP_LENGTH..(i + 1) * PADDED_FP_LENGTH])?;
-    }
```
kevaundray, line 107: This code likely does not belong in blst since its blst agnostic -- I refrained from making any changes like this, just so that it has a clean diff

crates/precompile/src/bls12_381/utils.rs
```diff
@@ -33,39 +19,8 @@ pub(super) fn remove_padding(input: &[u8]) -> Result<&[u8; FP_LENGTH], Precompil
...
-        blst_scalar_from_bendian(&mut out, input.as_ptr())
-    };
-
-    Ok(out)
-}
-
 /// Checks if the input is a valid big-endian representation of a field element.
-fn is_valid_be(input: &[u8; 48]) -> bool {
+pub(super) fn is_valid_be(input: &[u8; 48]) -> bool {
     for (i, modulo) in input.iter().zip(MODULUS_REPR.iter()) {
```
DaniPopes, line 24: unrelated but arrays/slices can be compared directly: `*input < MODULUS_REPR`

kevaundray: That's a good point, let me change it

crates/precompile/src/bls12_381/blst.rs
```diff
@@ -314,3 +322,283 @@ pub(super) fn pairing_check(pairs: &[(blst_p1_affine, blst_p2_affine)]) -> bool
...
+    // SAFETY: Out comes from fixed length array, input is a blst value.
+    unsafe {
+        fp_to_bytes(&mut out[..PADDED_FP_LENGTH], &(*input).x);
+        fp_to_bytes(&mut out[PADDED_FP_LENGTH..], &(*input).y);
+    }
+    out
+}
+
+/// Encodes a single finite field element into byte slice with padding.
+fn fp_to_bytes(out: &mut [u8], input: *const blst_fp) {
```
DaniPopes, line 338: can these functions take in references instead of raw pointers?

kevaundray: Yep -- I'll remove raw pointers from the public and private API of the blst wrapper

rakita (approved):
lgtm

PR #2314, opened 2025-03-26
chore: clean up blst wrapper doc comment and types

kevaundray:
This is based off of #2313 , mainly cleans up the abstraction and refactors comments.

Most notable changes:

- Padding removal is done in the precompile function versus in the crypto backend
- the crypto backend API now takes in a typed scalar, and you need to call `read_scalar` which guarantees that the slice that is used to construct the scalar is 32 bytes


crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -54,8 +49,11 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         if encoded_g1_element.iter().all(|i| *i == 0) {
             continue;
         }
+
+        let [a_x, a_y] = remove_g1_padding(encoded_g1_element)?;
```
kevaundray, line 53: removing the padding is blst agnostic, so it has been moved into utils.rs from blst.rs

crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -67,8 +65,7 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         }
 
         g1_points.push(p0_aff);
-
-        scalars_bytes.extend_from_slice(&extract_scalar_input(encoded_scalar)?.b);
+        scalars.push(read_scalar(encoded_scalar)?);
```
kevaundray, line 68: This now reads in a scalar and the rest of the API uses this scalar type. This means that you cannot use any of the methods which require scalars without calling read_scalar first, which checks that the byte slice is 32 bytes.

crates/precompile/src/bls12_381/blst.rs
```diff
@@ -314,3 +335,298 @@ pub(super) fn pairing_check(pairs: &[(blst_p1_affine, blst_p2_affine)]) -> bool
...
+/// We do not check that the scalar is a canonical Fr element, because the EIP specifies:
+/// * The corresponding integer is not required to be less than or equal than main subgroup order
+///   `q`.
+pub(super) fn read_scalar(input: &[u8]) -> Result<blst_scalar, PrecompileError> {
+    if input.len() != SCALAR_LENGTH {
+        return Err(PrecompileError::Other(format!(
+            "Input should be {SCALAR_LENGTH} bytes, was {}",
+            input.len()
+        )));
+    }
```
kevaundray, line 619: We could move this check higher up and either change the input to `&[u8;SCALAR_LENGTH]` or add an assert in here

crates/precompile/src/bls12_381/g2_add.rs
```diff
@@ -26,12 +27,15 @@ pub(super) fn g2_add(input: &Bytes, gas_limit: u64) -> PrecompileResult {
         )));
     }
 
+    let [a_x_0, a_x_1, a_y_0, a_y_1] = remove_g2_padding(&input[..PADDED_G2_LENGTH])?;
+    let [b_x_0, b_x_1, b_y_0, b_y_1] = remove_g2_padding(&input[PADDED_G2_LENGTH..])?;
+
```
kevaundray, line 32: Most of the changes is removing the padding at the precompile level versus the crypto wrapper level -- so the backend only knows about "correctly formatted" points instead of the EVM specific padding that gets added

crates/precompile/src/bls12_381/blst.rs
```diff
@@ -115,14 +124,23 @@ fn p1_scalar_mul(p: &blst_p1_affine, scalar: &[u8]) -> blst_p1_affine {
...
     // SAFETY: all inputs are valid blst types
-    unsafe { blst_p2_mult(&mut result, &p_jacobian, scalar.as_ptr(), scalar.len() * 8) };
+    unsafe {
+        blst_p2_mult(
+            &mut result,
+            &p_jacobian,
+            scalar.b.as_ptr(),
+            scalar.b.len() * 8,
+        )
+    };
```
kevaundray, line 143: This is now safer because by definition, we cannot pass in a scalar with more than 32 bytes

rakita: It does remove uncertanty

crates/precompile/src/bls12_381/utils.rs
```diff
@@ -17,8 +19,34 @@ pub(super) fn remove_padding(input: &[u8]) -> Result<&[u8; FP_LENGTH], Precompil
     }
     Ok(unpadded.try_into().unwrap())
 }
+/// remove_g1_padding removes the padding applied to the Fp elements that constitute the
+/// encoded G1 element.
+pub(super) fn remove_g1_padding(input: &[u8]) -> Result<[&[u8; 48]; 2], PrecompileError> {
```
rakita, line 24:
```suggestion
pub(super) fn remove_g1_padding(input: &[u8]) -> Result<[&[u8; FP_LENGTH]; 2], PrecompileError> {
```

crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -34,7 +35,7 @@ pub(super) fn g1_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
     }
 
     let mut g1_points: Vec<_> = Vec::with_capacity(k);
-    let mut scalars_bytes: Vec<u8> = Vec::with_capacity(k * SCALAR_LENGTH);
+    let mut scalars: Vec<_> = Vec::with_capacity(k);
```
rakita, line 38:
```suggestion
    let mut scalars = Vec::with_capacity(k);
```
We can probably remove  `Vec<_>` fully

crates/precompile/src/bls12_381/g1_msm.rs
rakita: Yeah this is a lot nicer!

crates/precompile/src/bls12_381/g2_msm.rs
```diff
@@ -34,23 +35,26 @@ pub(super) fn g2_msm(input: &Bytes, gas_limit: u64) -> PrecompileResult {
     }
 
     let mut g2_points: Vec<_> = Vec::with_capacity(k);
-    let mut scalars_bytes: Vec<u8> = Vec::with_capacity(k * SCALAR_LENGTH);
+    let mut scalars: Vec<_> = Vec::with_capacity(k);
```
rakita, line 38:
```suggestion
    let mut scalars = Vec::with_capacity(k);
```

rakita (approved):
lgtm, left few nits

crates/precompile/src/bls12_381/g2_add.rs
rakita: This makes it simpler

PR #2316, opened 2025-03-26
feat: Add arkworks wrapper for bls12-381

kevaundray:
This is based off of #2314


kevaundray:
cfg flags have not been added yet -- related to #2172

crates/precompile/src/bls12_381/blst.rs
```diff
@@ -314,3 +336,311 @@ pub(super) fn pairing_check(pairs: &[(blst_p1_affine, blst_p2_affine)]) -> bool
...
+        // order `q`.
+        blst_scalar_from_bendian(&mut out, input.as_ptr())
+    };
+
+    Ok(out)
+}
+
+/// Checks if the input is a valid big-endian representation of a field element.
+fn is_valid_be(input: &[u8; 48]) -> bool {
+    for (i, modulo) in input.iter().zip(MODULUS_REPR.iter()) {
```
kevaundray, line 637: can likely move this MODULUS_REPR constant into blst since it is not needed by any other backend

crates/precompile/src/bls12_381_utils.rs
```diff
@@ -17,43 +13,3 @@ pub fn msm_required_gas(k: usize, discount_table: &[u16], multiplication_cost: u
...
-        PrecompileWithAddress(G2_ADD_ADDRESS, |_, _| {
-            Err(PrecompileError::Fatal(
-                "no_std is not supported for BLS12-381 precompiles".into(),
-            ))
-        }),
-        PrecompileWithAddress(G2_MSM_ADDRESS, |_, _| {
-            Err(PrecompileError::Fatal(
-                "no_std is not supported for BLS12-381 precompiles".into(),
-            ))
-        }),
```
kevaundray, line 42: Deleted this because arkworks should support no_std -- I guess one thing to note is that to use no_std, you would need to do `cargo b --no-default-features --features arkworks-bls12-381` , no-default-features to remove `std` and arkworks-bls12-381 to switch to arkworks instead of blst.

There is an asymetry here I guess between bn128 which uses blst by default and bls12-381 which uses arkworks as an optional

crates/revm/Cargo.toml
```diff
@@ -93,7 +93,7 @@ c-kzg = [
 	"precompile/c-kzg",
 ] # `kzg-rs` is not audited but useful for `no_std` environment, use it with causing and default to `c-kzg` if possible.
 kzg-rs = ["precompile/kzg-rs"]
-blst = ["precompile/blst"]
+arkworks-bls12-381 = ["precompile/arkworks-bls12-381"]
```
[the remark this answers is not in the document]
kevaundray: Changed

crates/precompile/src/bls12_381_utils.rs
kevaundray: Have changed it to use blst as optional (happy to change back if needed)

kevaundray:
> ## [CodSpeed Performance Report](https://codspeed.io/bluealloy/revm/branches/kevaundray%3Akw%2Fadd-arkworks-bls12-381)
> ### Merging #2316 will **not alter performance**
> Comparing `kevaundray:kw/add-arkworks-bls12-381` ([0c526ea](https://github.com/bluealloy/revm/commit/0c526ea99271ef00733436c138e0d00e56cb6581)) with `main` ([76e1266](https://github.com/bluealloy/revm/commit/76e12668f93c49d26aa85c24b6d092b6b2acef2c))
> 
> ### Summary
> `✅ 8` untouched benchmarks

This should not be true -- happy to block on this while there are no benchmarks for bls12-381

crates/revm/Cargo.toml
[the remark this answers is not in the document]
kevaundray: unresolving this and will wait for Rakita to say what way this should be done -- seems you prefer for blst to be the main one and arkworks to be optional from this https://github.com/bluealloy/revm/issues/2172#issuecomment-2714215559 ?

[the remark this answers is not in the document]
rakita: This was about security consideration. `zkcrypto/bls12_381` is not audited or maintained that well but it has no_std support. Arkworks is a different story and this is not the case, and i would be fine to make default.

Decision would depend on performance. blst rust binding has some things that are done in threads I am not sure if this is case for arkworks, or how faster arkworks against blst if threading is not needed. Made a task for this https://github.com/bluealloy/revm/issues/2322

Would leave blst as default for now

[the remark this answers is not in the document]
kevaundray: Makes sense to me :)

crates/precompile/src/bls12_381/blst.rs
kevaundray: I can do this in a separate PR since its a breaking change, as MODULUS_REPR is public

rakita: ack

bins/revme/Cargo.toml
```diff
@@ -9,15 +9,14 @@ license.workspace = true
...
-# revm
 revm = { workspace = true, features = [
     "std",
     "hashbrown",
     "c-kzg",
     "blst",
     "serde-json",
     "hashbrown",
-] }
+] } # revm
```
kevaundray, line 19: will revert -- auto cargo fmt

crates/precompile/Cargo.toml
```diff
@@ -64,9 +65,12 @@ c-kzg = { workspace = true, optional = true, features = [
 # Optionally use `kzg-rs` for a pure Rust implementation of KZG point evaluation.
 kzg-rs = { workspace = true, optional = true }
 
-# BLS12-381 precompiles
+# Use the BLS12-381 implementation of blst for EIP2537
 blst = { workspace = true, optional = true }
```
kevaundray, line 69: Making blst optional but default because it doesn't compile with the riscv backend

crates/optimism/src/evm.rs
```diff
@@ -283,7 +283,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "blst")]
     fn test_halted_tx_call_bls12_381_g1_add_out_of_gas() {
         let ctx = Context::op()
             .modify_tx_chained(|tx| {
```
kevaundray, line 288: Since both arkworks and blst are optional, these would need to be added back, since we test with no-default-features and no feature flags

kevaundray: Done

rakita (approved):
cool!

PR #2317, opened 2025-03-26
feat: cache precompile warming

rakita:
Sets spec every time in runtime, changed warmed precompiles only if there is change in precompiles set.

Introduce a bool in `PrecompileProviders::set_spec_id` that would notify us if precompiles got changed.

With caching of precompiles we are removing additional HashMap clone of those addresses


rakita:
Fixes degradation from https://github.com/bluealloy/revm/pulls?q=is%3Apr+journal and improves it even more.

PR #2318, opened 2025-03-26
feat: provide more context to precompiles

bernard-wagner:
To provide greater support for custom precompiles, it would be useful to pass `InputsImpl` and `is_static` to `PrecompileProvider::run` instead of only a subset of the fields passed as inputs to the Interpreter. There are some use cases for custom chains that inspect the caller or should not be allowed in a static context (state-changing precompiles).


rakita (approved):
Makes sense and PR lgtm

Issue #2308, opened 2025-03-25
Add benchmark for Evm build


rakita:
Benching and having a measurement of how long building of Evm takes can be beneficial to track. This is mostly time needed to init instructions.

Blocked by https://github.com/bluealloy/revm/pull/2295

shane-moore:
hey @rakita, looks like https://github.com/bluealloy/revm/pull/2295 is merged. Perhaps this ticket is good to start? If so, I'd be happy to take a look

rakita:
It is good to go

PR #2319, opened 2025-03-26
chore: add criterion benchmark for evm build

shane-moore:
## Changes
- This PR closes https://github.com/bluealloy/revm/issues/2308 by adding a benchmark to measures the how long the EVM build takes to initialization instructions via `evm.replay()` 

## Note:
- the `build_mainnet` step was left out of the benchmark as to focus on time spent to initialize instructions.  Let me know if the intention was to include the `build_mainnet` into the benchmark as well, and I'll happily add it 🤙

### DIFF
--- a/bins/revme/src/cmd/bench.rs
+++ b/bins/revme/src/cmd/bench.rs
@@ -1,5 +1,6 @@
 pub mod analysis;
 pub mod burntpix;
+pub mod evm_build;
 pub mod snailtracer;
 pub mod transfer;
 
@@ -11,6 +12,7 @@ pub enum BenchName {
     Burntpix,
     Snailtracer,
     Transfer,
+    EvmBuild,
 }
 
 impl BenchName {
@@ -19,6 +21,7 @@ impl BenchName {
         BenchName::Burntpix,
         BenchName::Snailtracer,
         BenchName::Transfer,
+        BenchName::EvmBuild,
     ];
 
     pub fn as_str(self) -> &'static str {
@@ -27,6 +30,7 @@ impl BenchName {
             BenchName::Burntpix => "burntpix",
             BenchName::Snailtracer => "snailtracer",
             BenchName::Transfer => "transfer",
+            BenchName::EvmBuild => "evm_build",
         }
     }
 }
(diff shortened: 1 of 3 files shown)


shane-moore:
hi @rakita, this PR is ready for review whenever you get a chance

bins/revme/src/cmd/bench/evm_build.rs
```diff
@@ -0,0 +1,12 @@
+use criterion::Criterion;
+use revm::{Context, ExecuteEvm, MainBuilder, MainContext};
+
+pub fn run(criterion: &mut Criterion) {
+    let mut evm = Context::mainnet().build_mainnet();
+
+    criterion.bench_function("evm_build", |b| {
+        b.iter(|| {
+            let _ = evm.replay().unwrap();
```
rakita, line 9:
```suggestion
           let _ = Context::mainnet().build_mainnet();
```

Buildering of EVM should be here, not a replay.

bins/revme/src/cmd/bench.rs
```diff
@@ -27,6 +30,7 @@ impl BenchName {
             BenchName::Burntpix => "burntpix",
             BenchName::Snailtracer => "snailtracer",
             BenchName::Transfer => "transfer",
+            BenchName::EvmBuild => "evm_build",
```
rakita, line 33:
```suggestion
            BenchName::EvmBuild => "evm-build",
```

For outside names using dash is better.

bins/revme/src/cmd/bench/evm_build.rs
shane-moore: gotcha! updated to https://github.com/bluealloy/revm/blob/e537c96d8c224603eeccb2bfb2a62407bf5de7a2/bins/revme/src/cmd/bench/evm_build.rs#L4-L9

bins/revme/src/cmd/bench.rs
shane-moore: great point @rakita, updated to: https://github.com/bluealloy/revm/blob/e537c96d8c224603eeccb2bfb2a62407bf5de7a2/bins/revme/src/cmd/bench.rs#L33

agreed that `cargo bench --package revme -- evm-build` does look cleaner

rakita (approved):

Issue #2307, opened 2025-03-25
revme statetest should fail on no json's found


rakita:
Return error if there are no json at given path or path is not existing.

`revm statetest path/` can give a false positive if the path is given by mistake.

DioBr4nd0:
Hii @rakita I  want to work  on this issue.

rakita:
@DioBr4nd0 have asigned you

PR #2320, opened 2025-03-26
Chore: add check for path and existence existence

startup-dreamer:
Closes: #2307

Adds the path and json existence checks.


rakita (approved):
cool, thanks

PR #2327, opened 2025-03-27
chore: add bls12-381 benchmarks

### LINKED ISSUE #2322: Benchmarks for bls precompiles
With the introduction of arkworks bls backend we now have two sets of libraries. https://github.com/bluealloy/revm/pull/2316

We need to add benchmarks that would guide us on decision on who to put as a default.

kevaundray:
closes #2322

### DIFF
--- a/crates/precompile/benches/bench.rs
+++ b/crates/precompile/benches/bench.rs
@@ -1,3 +1,5 @@
+pub mod eip2537;
+
 use criterion::{criterion_group, criterion_main, Criterion};
 use primitives::{eip4844::VERSIONED_HASH_VERSION_KZG, hex, keccak256, Bytes, U256};
 use revm_precompile::{
@@ -121,6 +123,14 @@ pub fn benchmark_crypto_precompiles(c: &mut Criterion) {
     let gas = 50000;
     let output = run(&kzg_input, gas).unwrap();
     println!("gas used by kzg precompile: {:?}", output.gas_used);
+
+    eip2537::add_g1_add_benches(&mut group);
+    eip2537::add_g2_add_benches(&mut group);
+    eip2537::add_g1_msm_benches(&mut group);
+    eip2537::add_g2_msm_benches(&mut group);
+    eip2537::add_pairing_benches(&mut group);
+    eip2537::add_map_fp_to_g1_benches(&mut group);
+    eip2537::add_map_fp2_to_g2_benches(&mut group);
 
     group.bench_function(group_name("ecrecover precompile"), |b| {
         b.iter(|| ec_recover_run(&message_and_signature, u64::MAX).unwrap())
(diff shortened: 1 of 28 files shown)


crates/precompile/benches/eip2537.rs
```diff
@@ -0,0 +1,495 @@
...
+const RNG_SEED: u64 = 42;
+const G1_ADD_TEST_VECTORS: usize = 10;
+const G2_ADD_TEST_VECTORS: usize = 10;
+const G1_MSM_TEST_VECTORS: usize = 5;
+const G2_MSM_TEST_VECTORS: usize = 5;
+const MAX_MSM_SIZE: usize = 16;
+const PAIRING_TEST_VECTORS: usize = 5;
+const MAX_PAIRING_PAIRS: usize = 10;
+const MAP_FP_TO_G1_TEST_VECTORS: usize = 10;
+const MAP_FP2_TO_G2_TEST_VECTORS: usize = 10;
```
kevaundray, line 22: This is overkill, will remove -- the rationale was that some of the operations might be too fast, so each benchmark run will cause a lot of variance. To overcome this, we can benchmark a N batches of the operation instead. But checking the CI, I see that bn128_add does not show up often in other PRs for example

kevaundray: has been changed now

crates/precompile/benches/eip2537.rs
```diff
@@ -0,0 +1,332 @@
...
+use criterion::{measurement::Measurement, BenchmarkGroup};
+use primitives::Bytes;
+use rand::{rngs::StdRng, SeedableRng};
+use revm_precompile::bls12_381_const::{
+    FP_LENGTH, FP_PAD_BY, PADDED_FP_LENGTH, PADDED_G1_LENGTH, PADDED_G2_LENGTH,
+};
+
+const RNG_SEED: u64 = 42;
+const MAX_MSM_SIZE: usize = 16;
+const MAX_PAIRING_PAIRS: usize = 10;
```
kevaundray, line 14: These are arbitrary, MAX_MSM_SIZE might be better at 256 for the realistic maximum that will happen in the EVM?

For Pairing, 1/2 is most common, but 8/16 sound reasonable as a max

kevaundray: changed to 256 and 16, and also added cases for MSM and pairing of size 2

crates/precompile/benches/bench.rs
```diff
@@ -122,6 +124,14 @@ pub fn benchmark_crypto_precompiles(c: &mut Criterion) {
     let output = run(&kzg_input, gas).unwrap();
     println!("gas used by kzg precompile: {:?}", output.gas_used);
 
+    eip2537::add_g1_add_benches(&mut group);
+    eip2537::add_g2_add_benches(&mut group);
+    eip2537::add_g1_msm_benches(&mut group);
+    eip2537::add_g2_msm_benches(&mut group);
+    eip2537::add_pairing_benches(&mut group);
+    eip2537::add_map_fp_to_g1_benches(&mut group);
+    eip2537::add_map_fp2_to_g2_benches(&mut group);
```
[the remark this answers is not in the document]
rakita: It is fine, I think we moved benches from another place, and printlns were just left there. They don't do anything crucial.

rakita (approved):
lgtm

PR #2330, opened 2025-03-28
fix(isthmus): Correctly filter refunds for deposit transactions

clabby:
## Overview

Corrects the current condition for operator fee refunds in trunk. Previously the check was using a source hash heuristic that wasn't correct - what we want is to not refund the operator fee for deposit transactions.

Also adds a test for expected refund behavior for both deposit transactions and normal transactions.

note - this bug isn't present in tag v64 (v19.7.0)

### DIFF
--- a/crates/optimism/src/handler.rs
+++ b/crates/optimism/src/handler.rs
@@ -256,7 +256,7 @@ where
         self.mainnet.reimburse_caller(evm, exec_result)?;
 
         let context = evm.ctx();
-        if matches!(context.tx().source_hash(), Some(source_hash) if source_hash.is_zero()) {
+        if context.tx().tx_type() != DEPOSIT_TRANSACTION_TYPE {
             let caller = context.tx().caller();
             let spec = context.cfg().spec();
             let operator_fee_refund = context.chain().operator_fee_refund(exec_result.gas(), spec);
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this would otherwise cause panics because the L1BlockInfo isnt loaded for deposit txs

rakita (approved):

mattsse:
definitely panics:

> [op-el-2-op-reth-op-node-op-kurtosis] thread 'Tree Task' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/op-revm-1.0.0-alpha.6/src/l1block.rs:164:14:
[op-el-2-op-reth-op-node-op-kurtosis] Missing operator fee scalar for isthmus L1 Block
[op-el-2-op-reth-op-node-op-kurtosis] stack backtrace:
[op-el-2-op-reth-op-node-op-kurtosis] note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[op-el-2-op-reth-op-node-op-kurtosis] 2025-03-27T22:10:47.295132Z ERROR Fatal error
[op-el-2-op-reth-op-node-op-kurtosis] 2025-03-27T22:10:47.295159Z ERROR Fatal error in consensus engine
[op-el-2-op-reth-op-node-op-kurtosis] 2025-03-27T22:10:47.295202Z ERROR shutting down due to error
[op-el-2-op-reth-op-node-op-kurtosis] Error: Fatal error in consensus engine
[op-el-2-op-reth-op-node-op-kurtosis] 
[op-el-2-op-reth-op-node-op-kurtosis] Location:
[op-el-2-op-reth-op-node-op-kurtosis]     /home/runner/work/reth/reth/crates/node/builder/src/launch/engine.rs:341:43

PR #2332, opened 2025-03-28
refactor: simplify Eof::data_slice, make it more readable

JackDrogon:
simplify Eof::data_slice, make it more readable

### DIFF
--- a/crates/bytecode/src/eof.rs
+++ b/crates/bytecode/src/eof.rs
@@ -84,11 +84,13 @@ impl Eof {
     /// If offset is greater than the length of the raw bytes, an empty slice is returned.
     /// If len is greater than the length of the raw bytes, the slice is truncated to the length of the raw bytes.
     pub fn data_slice(&self, offset: usize, len: usize) -> &[u8] {
-        self.body
-            .data_section
-            .get(offset..)
-            .and_then(|bytes| bytes.get(..min(len, bytes.len())))
-            .unwrap_or(&[])
+        let data_len = self.body.data_section.len();
+        if len == 0 || offset >= data_len {
+            return &[];
+        }
+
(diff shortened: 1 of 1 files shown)


rakita:
I prefer the chaining version, it is more redable for me

JackDrogon:
> I prefer the chaining version, it is more redable for me

@rakita chaining version is easy to write, but not easy to understand.
I'll give you an example and elaborate.

source code
```rust
/// Returns a slice of the raw bytes.
/// If offset is greater than the length of the raw bytes, an empty slice is returned.
/// If len is greater than the length of the raw bytes, the slice is truncated to the length of the raw bytes.
pub fn data_slice(&self, offset: usize, len: usize) -> &[u8] {
    self.body
        .data_section
        .get(offset..)
        .and_then(|bytes| bytes.get(..min(len, bytes.len())))
        .unwrap_or(&[])
}
```
The two most important things here are the two if logics. one is `If offset is greater than the length of the raw bytes, an empty slice is returned.`, one is `If len is greater than the length of the raw bytes, the slice is truncated to the length of the raw bytes.`. 
How a newcomer has to understand it becomes complicated.
When he wants to understand to the first if, he needs to read these codes
```
      self.body
            .data_section
            .get(offset..)...
            .unwrap_or(&[])
```
This understanding is cut and dried

the second if 
```
            .get(offset..)
            .and_then(|bytes| bytes.get(..min(len, bytes.len())))
            .unwrap_or(&[])
```
  And the logic about getting `&[offset..len]` is also cut into two parts one to get `&[offset..end]` first, and then to determine to get `&[offset..len]`.

This chaining version is a great burden of understanding for those who read the code

rakita:
So a newcomer needs to get familiar with the functional aspect of rust

PR #2334, opened 2025-03-28
refactor: move Eof::new to first place && add doc for some pub functions

JackDrogon:
move Eof::new to first place && add doc for some pub functions


rakita (approved):
thanks

PR #2336, opened 2025-03-28
chore!: Move modulus constant into blst wrapper

kevaundray:
This is a breaking change since MODULUS_REPR was marked as public.

The constant is being used to check if a serialized field element is valid, however it is only needed because blst does not expose a clean way to check this. arkworks for example, allows you to do this.

I don't _think_ anyone should have been using this, but it is a breaking change nonetheless.


crates/precompile/src/bls12_381/blst.rs
```diff
@@ -2,7 +2,7 @@
 
 use crate::{
     bls12_381_const::{
-        FP_LENGTH, FP_PAD_BY, MODULUS_REPR, PADDED_FP_LENGTH, PADDED_G1_LENGTH, PADDED_G2_LENGTH,
+        FP_LENGTH, FP_PAD_BY, PADDED_FP_LENGTH, PADDED_G1_LENGTH, PADDED_G2_LENGTH,
         SCALAR_LENGTH, SCALAR_LENGTH_BITS,
```
rakita, line 6:
```suggestion
        FP_LENGTH, FP_PAD_BY, PADDED_FP_LENGTH, PADDED_G1_LENGTH, PADDED_G2_LENGTH, SCALAR_LENGTH,
        SCALAR_LENGTH_BITS,
```

PR #2346, opened 2025-03-30
chore: Try PUSH2-JMP optimization

kevaundray:
Pushing a PR to see if it has any effect on performance -- not familiar with this part of the codebase, so likely not the way it should be implemented

### DIFF
--- a/crates/interpreter/src/instructions/stack.rs
+++ b/crates/interpreter/src/instructions/stack.rs
@@ -1,6 +1,6 @@
 use crate::{
     gas,
-    instructions::utility::cast_slice_to_u256,
+    instructions::{control, utility::cast_slice_to_u256},
     interpreter::Interpreter,
     interpreter_types::{Immediates, InterpreterTypes, Jumps, LoopControl, RuntimeFlag, StackTr},
     Host,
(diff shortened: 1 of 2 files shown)


kevaundray:
This has zero effect which might imply either:

-  Stack operations are very fast, so this is negligible
-  Maybe there are no benchmarks with a lot of static jumps
- This is implemented incorrectly

crates/interpreter/src/instructions/stack.rs
```diff
@@ -33,6 +33,38 @@ pub fn push<const N: usize, WIRE: InterpreterTypes, H: Host + ?Sized>(
...
+    // Special optimization for PUSH2 followed by JUMP
+    //
+    // PUSH2 will push the jump destination onto the stack
+    if N == 2 {
+        // Try to read the immediate value (jump destination)
+        // TODO: check what happens if there is less than N bytes to read
+        // TODO: The if statement is here assuming it will read as much as it can
+        // TODO: so it is a boundary check
+        let imm = interpreter.bytecode.read_slice(N);
+        if imm.len() == N {
```
rakita, line 46: This check can be removed, all bytecodes are padded with zeroes to ensure that push always has immediate and that STOP is always last opcode

crates/interpreter/src/instructions/stack.rs
```diff
@@ -33,6 +33,38 @@ pub fn push<const N: usize, WIRE: InterpreterTypes, H: Host + ?Sized>(
...
+
+    // Special optimization for PUSH2 followed by JUMP
+    //
+    // PUSH2 will push the jump destination onto the stack
+    if N == 2 {
+        // Try to read the immediate value (jump destination)
+        // TODO: check what happens if there is less than N bytes to read
+        // TODO: The if statement is here assuming it will read as much as it can
+        // TODO: so it is a boundary check
+        let imm = interpreter.bytecode.read_slice(N);
```
rakita, line 45: `.read_u16()` should be there somewhere.

rakita:
> Stack operations are very fast, so this is negligible
They are, but it all depends on the patterns. There are a few patterns (Pairs or triplets of opcodes) that would benefit from replacing it with one opcode. Unfortunatly I couldn't find the tweet where this was analyzed.

> Maybe there are no benchmarks with a lot of static jumps
Hard to tell without checking on bytecode.

> This is implemented incorrectly
Yep code reads u16 but we need u8 for opcode.

crates/interpreter/src/instructions/stack.rs
```diff
@@ -33,6 +33,38 @@ pub fn push<const N: usize, WIRE: InterpreterTypes, H: Host + ?Sized>(
...
+        if imm.len() == N {
+            let jump_dest = ((imm[0] as usize) << 8) | (imm[1] as usize);
+
+            // Try to read the next opcode after PUSH2 and its immediate bytes
+            let next_op = interpreter.bytecode.read_offset_u16(N as isize);
```
rakita, line 50: This would read `u16` (two bytes), but you want to read `u8`

Would structure the code as something like.
```
interpreter.bytecode.relative_jump(N as size);
let next_op = interpreter.bytecode.read_u8().
if next_op == opcode::JUMP {

    ...
    interpreter.bytecode.relative_jump(1)
} else {
    push!(interpreter, U256::from(imm as u64);
}
return;
}

kevaundray:
This is not as performant as I was expecting -- should we close this, since it _does_ introduce added complexity in push?

rakita:
> This is not as performant as I was expecting -- should we close this, since it does introduce added complexity in push?

Sure. The biggest complexity is not about logic inside Revm but with Inspections and tracing, where with it we magically skip one instruction.

PR #2352, opened 2025-04-01
test(op-inspector): Add test for inspecting logs

emhane:

Adds test for verifying correct liking of OP evm to inspector


rakita (approved):
lgtm

Issue #2354, opened 2025-04-01
`Account` helper builder functions


rakita:
Adding few utilities would help with easier usage.

* `with_info` where `AccountInfo` can be set
* `with_storage` where input is Iterator<Item = EvmStorageSlot>`
* All functions for mark's for example `with_selfdestruct_mark` or `with_touched_mark()` etc.

07Vaishnavi-Singh:
@rakita can I take this?

rakita:
Sure, assigned

07Vaishnavi-Singh:
ty!

rakita:
Merged

PR #2355, opened 2025-04-01
feat: Account helper builder functions

07Vaishnavi-Singh:
Closes - #2354

### Changes:
Added methods to Account:
- [x] with_info: Sets account information.
- [x] with_storage: Populates storage from an iterator of storage slots.
- [x] with_selfdestruct_mark: Marks the account as self-destructed.
- [x] with_touched_mark: Marks the account as touched.
- [x] with_created_mark: Marks the account as newly created.
- [x] with_cold_mark: Marks the account as cold.
- [x] with_warm_mark: Marks the account as warm.
- [x] with_warm: Marks the account as warm without returning the previous state.
- [x] Added tests

### DIFF
--- a/crates/state/src/lib.rs
+++ b/crates/state/src/lib.rs
@@ -11,6 +11,7 @@ pub use bytecode::Bytecode;
 pub use primitives;
 pub use types::{EvmState, EvmStorage, TransientStorage};
 
+#[allow(unused_imports)]
 use crate::primitives::KECCAK_EMPTY;
 use bitflags::bitflags;
 use core::hash::Hash;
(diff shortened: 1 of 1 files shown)


crates/state/src/lib.rs
```diff
@@ -11,6 +11,8 @@ pub use bytecode::Bytecode;
 pub use primitives;
 pub use types::{EvmState, EvmStorage, TransientStorage};
 
+#[allow(unused_imports)]
+use crate::primitives::KECCAK_EMPTY;
```
rakita, line 15:
```suggestion
```

rakita (approved):

crates/state/src/lib.rs
07Vaishnavi-Singh: yep, my bad I should run cargo clean and then test, will keep in mind from now ty!

rakita:
Merging after CI

PR #2371, opened 2025-04-05
chore: set gas_priority_fee to None in TxEnv

### LINKED ISSUE #2369: set `gas_priority_fee` to None in TxEnv
Having default value on TxEvn to be legacy type but setting `gas_priority_fee` to Some is problematic

https://github.com/bluealloy/revm/blob/a268f354d0da2248e354b632672a790bcb1165a8/crates/context/src/tx.rs#L93

stevencartavia:
closes #2369


rakita (approved):
lgtm

PR #2374, opened 2025-04-05
add get_account fn to load account once

### LINKED ISSUE #2356: Analize potential merging of gas validation and caller deduction fn
`validate_tx_against_account` fn in validation phase loads caller account and checks nonce/code and if it has enough balance to cover maximum gas usage. 

`deduct_caller` function in pre execution phase of execution deducts caller balance with maximum amount it can spend

Fetching account two times from hashmap is waste of CPU, and checking same things in two places is a potential bug risk (already happened in op-revm). When merging those two we need tobe careful about op-revm as it is used there

nagxsan:
Closes #2356 

# Summary
- Created a `get_account` function to get a single `Account` instance.
- Modify `validate` and `pre_execution` functions to pass relevant account parameters for usage instead of loading the account again.
- Update inspector code.

### DIFF
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -12,6 +12,7 @@ use context_interface::{
     Cfg, Database, JournalTr, Transaction,
 };
 use interpreter::{FrameInput, Gas, InitialAndFloorGas};
+use state::{Account, AccountInfo};
 use std::{vec, vec::Vec};
 
 pub trait EvmTrError<EVM: EvmTr>:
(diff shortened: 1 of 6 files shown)


rakita:
It is not possible to do it this way, as you are cloning the Account and not returning a mut ref, and It exposes the Account to the Handler that is not a great abstraction.

The solution for this, as stated in the issue, is to merge two functions together so logic that is very similar can be found in one place.

Would recommend discarding this branch and starting from scratch

nagxsan:
Got it, will do so thanks

PR #2375, opened 2025-04-06
fix: Effective gas price should check tx type

nemnem7:
* Modified effective_gas_price function in transaction.rs file so it behaves differently depending on the type of transaction. 
* Added tests to tx.rs file to cover multiple cases of different transactions types using effecive_gas_price


crates/context/interface/src/transaction.rs
```diff
@@ -137,11 +137,15 @@ pub trait Transaction {
...
-        min(max_fee, base_fee.saturating_add(max_priority_fee))
+        if self.tx_type() == 0 || self.tx_type() == TransactionType::Eip2930 as u8 {
+            self.gas_price()
+        } else {
+            let max_fee = self.gas_price();
+            let Some(max_priority_fee) = self.max_priority_fee_per_gas() else {
+                return max_fee;
+            };
+            min(max_fee, base_fee.saturating_add(max_priority_fee))
+        }
```
rakita, line 148: More readable code would be: 
```suggestion
        if self.tx_type() == 0 || self.tx_type() == TransactionType::Eip2930 as u8 {
            return self.gas_price();
        }
        
        // for EIP-1559 tx and onwards gas_price represents maximum price.
        let max_price = self.gas_price();
        let Some(max_priority_fee) = self.max_priority_fee_per_gas() else {
            return max_price;
        };
        min(max_fee, base_fee.saturating_add(max_priority_fee))
```

crates/context/src/tx.rs
```diff
@@ -208,3 +208,184 @@ impl Transaction for TxEnv {
         self.gas_priority_fee
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_effective_gas_price_legacy() {
```
rakita, line 217: Lets simplify all these tests and make one with a template.

`fn effective_gas_setup(type: TransactionType, gas_price: u128, priority_fee: Option<u128>) -> u128` 
Base fee can be by default 100

And in test you can do
```
assert_eq!(100, effective_gas_setup(TransactionType::Eip1559, 100, None);
assert_eq!(90, effective_gas_setup(TransactionType::Legacy, 90, None);
```
etc

crates/context/interface/src/transaction.rs
nemnem7: Done

crates/interpreter/src/instructions/system.rs
```diff
@@ -234,165 +234,177 @@ pub fn memory_resize(
     Some(memory_offset)
 }
 
-// TODO : Tests
```
rakita, line 237: Do `git merge` so this changes are removed

crates/context/interface/src/transaction.rs
```diff
@@ -137,11 +137,16 @@ pub trait Transaction {
     ///
     /// While for transactions after Eip1559 it is minimum of max_fee and `base + max_priority_fee`.
     fn effective_gas_price(&self, base_fee: u128) -> u128 {
-        let max_fee = self.gas_price();
+        if self.tx_type() == 0 || self.tx_type() == TransactionType::Eip2930 as u8 {
```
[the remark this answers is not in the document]
nemnem7: Done

rakita (approved):
lgtm! waiting for CI

PR #2381, opened 2025-04-07
fix: transaction type check for `block.transactions`

mdqst:
updated the code to safely check if `block.transactions` is of type `Full` before processing, preventing any panics.
now it handles the transactions more reliably.


mattsse (commented):
this is equivalent, unclear what this changes

rakita:
Unneded change

PR #2385, opened 2025-04-08
chore: make blob params u64

rkrasiuk:
## Description

Make blob params u64.


mattsse (approved):
we need them to be larger than u8 
u64 is def overkill but since these params are u64 usually this is preferred

rakita (approved):
lgtm

PR #2398, opened 2025-04-09
Fix the ordering of eofcreate frame traces

shemnon:
EOF create was placing trace summaries at the beginning of the trace, and not at the end.


rakita (approved):
thanks

PR #2399, opened 2025-04-09
feat: mirror evm api on opevm

mattsse:
mirrors the same functions that evm has.

I noticed that none of OpEvm and Evm have a single line of docs, could we also document these given that those are the main types?


clabby (approved):
Thanks :)

rakita (approved):

rakita:
Gakons suggestion to add lint for all public function to have docs is good. Will take a stab at it.

PR #2409, opened 2025-04-14
Bytecode(chore): add more check in `LegacyAnalyzedBytecode::new()`

chen4903:
According to this [comment](https://github.com/bluealloy/revm/blob/main/crates/bytecode/src/legacy/analyzed.rs#L17-L23):
```rust
/// # Bytecode Padding
///
/// All legacy bytecode is padded with 33 zero bytes at the end. This padding ensures the
/// bytecode always ends with a valid STOP (0x00) opcode. The reason for 33 bytes padding (and not one byte)
/// is handling the edge cases  where a PUSH32 opcode appears at the end of the original
/// bytecode without enough remaining bytes for its immediate data. Original bytecode length
/// is stored in order to be able to copy original bytecode.
```
But at the moment, we [do NOT check the 33 padding](https://github.com/bluealloy/revm/blob/main/crates/bytecode/src/legacy/analyzed.rs#L61-L86) in the `LegacyAnalyzedBytecode::new()`, but only[ add 33 zero bytes padding](https://github.com/bluealloy/revm/blob/main/crates/bytecode/src/legacy/raw.rs#L20) in `LegacyRawBytecode::into_analyzed()`. Therefore, we can still create non-compliant LegacyAnalyzedBytecode through the `LegacyAnalyzedBytecode::new()` function.

I add more check when creating `LegacyAnalyzedBytecode`

### DIFF
--- a/crates/bytecode/src/legacy/analyzed.rs
+++ b/crates/bytecode/src/legacy/analyzed.rs
@@ -78,6 +78,15 @@ impl LegacyAnalyzedBytecode {
             panic!("last bytecode byte should be STOP (0x00)");
         }
 
+        if bytecode.len() <= 33 {
+            panic!("bytecode length must greater than 33");
+        }
+
+        let last_33_bytes = &bytecode[bytecode.len() - 33..];
+        if !last_33_bytes.iter().all(|&byte| byte == 0) {
+            panic!("last 33 bytes must be zero");
+        }
+
         Self {
             bytecode,
             original_len,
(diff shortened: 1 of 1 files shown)


chen4903:
```
failures:
    validation::tests::test_eip170_create_opcode_size_limit_failure
    validation::tests::test_eip170_create_opcode_size_limit_success
    validation::tests::test_eip3860_initcode_size_limit_success
```

rakita:
Would omit this check as it can be misleading. It is true that we add 32 zeroes padding, but this is not necessary, and it depends if last byte is zero or PUSH without immediates.

PR #2410, opened 2025-04-14
feat: add tests for max input size op bls precompiles

### LINKED ISSUE #2408: Test input limit for OP bls precompiles
We found out that we don't have tests for it in https://github.com/bluealloy/revm/pull/2406

varun-doshi:
Closes #2408


rakita (approved):
lgtm

PR #2423, opened 2025-04-16
feat(bytecode): improve legacy bytecode padding

### LINKED ISSUE #2414: Pad legacy bytecode only if needed.
Presently we are alway padding legacy bytecode by 33 bytes. That is a worst edge case where last opcode i PUSH32 so we padd 32 immediates and 33'th byte is STOP.

 There are few changed needed for this.
1. Lets analysis return number of needed padded size: https://github.com/bluealloy/revm/blob/9747df4c2c14b1d568b4bfd4bc7c73ad59392979/crates/bytecode/src/legacy/analysis.rs#L12
2. Do resizing after analysis and only when padded_size is not zero: https://github.com/bluealloy/revm/blob/9747df4c2c14b1d568b4bfd4bc7c73ad59392979/crates/bytecode/src/legacy/raw.rs#L20

Bisht13:
# Improve Legacy Bytecode Analysis

## Description
Refactors bytecode analysis to handle padding internally and improve safety. Adds comprehensive test coverage for various bytecode scenarios.

## Changes
- `analyze_legacy` now handles padding internally and returns both jump table and padded bytecode
- Added early return for empty bytecode
- Improved safety documentation for unsafe pointer operations
- Added extensive test suite covering edge cases and padding scenarios

## Testing
Added tests for:
- Empty bytecode
- JUMPDEST positions
- PUSH operations
- Invalid opcodes
- Padding scenarios

Closes #2414

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -17,7 +17,7 @@ use std::{sync::Arc, vec, vec::Vec};
 ///
 /// Undefined behavior if the bytecode does not end with a valid STOP opcode. Please check
 /// [`crate::LegacyAnalyzedBytecode::new`] for details on how the bytecode is validated.
-pub fn analyze_legacy(mut bytecode: Vec<u8>) -> (JumpTable, Bytes) {
+pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
     let mut jumps: BitVec<u8> = bitvec![u8, Lsb0; 0; bytecode.len()];
 
     if bytecode.is_empty() {
@@ -28,7 +28,7 @@ pub fn analyze_legacy(mut bytecode: Vec<u8>) -> (JumpTable, Bytes) {
     let start = range.start;
     let mut iterator = start;
     let end = range.end;
-    let mut opcode = unsafe { *iterator };
+    let mut opcode = 0;
 
     while iterator < end {
         opcode = unsafe { *iterator };
@@ -52,10 +52,13 @@ pub fn analyze_legacy(mut bytecode: Vec<u8>) -> (JumpTable, Bytes) {
     // If we're at the end and last opcode is not STOP, we need 1 more byte
     let padding_size = (iterator as usize) - (end as usize) + (opcode != opcode::STOP) as usize;
     if padding_size > 0 {
-        bytecode.resize(bytecode.len() + padding_size, 0);
+        let mut padded_bytecode = Vec::with_capacity(bytecode.len() + padding_size);
+        padded_bytecode.extend_from_slice(&bytecode);
+        padded_bytecode.extend(vec![0; padding_size]);
+        (JumpTable(Arc::new(jumps)), Bytes::from(padded_bytecode))
+    } else {
+        (JumpTable(Arc::new(jumps)), bytecode)
     }
-
-    (JumpTable(Arc::new(jumps)), Bytes::from(bytecode))
 }
 
 mod tests {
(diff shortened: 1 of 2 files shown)


crates/bytecode/src/legacy/analysis.rs
```diff
@@ -1,23 +1,37 @@
...
-    let range = bytetecode.as_ptr_range();
+    if bytecode.is_empty() {
+        return (JumpTable(Arc::new(jumps)), Bytes::from(vec![opcode::STOP]));
+    }
+
+    let range = bytecode.as_ptr_range();
     let start = range.start;
     let mut iterator = start;
     let end = range.end;
+    let mut opcode = unsafe { *iterator };
```
rakita, line 31:
```suggestion
    let mut opcode = 0;
```

Nit this can be zero;

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -1,23 +1,37 @@
...
+/// The function uses unsafe pointer arithmetic, but maintains the following invariants:
+/// - The iterator never advances beyond the end of the bytecode
+/// - All pointer offsets are within bounds of the bytecode
+/// - The jump table is never accessed beyond its allocated size
+///
 /// Undefined behavior if the bytecode does not end with a valid STOP opcode. Please check
 /// [`crate::LegacyAnalyzedBytecode::new`] for details on how the bytecode is validated.
-pub fn analyze_legacy(bytetecode: &[u8]) -> JumpTable {
-    let mut jumps: BitVec<u8> = bitvec![u8, Lsb0; 0; bytetecode.len()];
+pub fn analyze_legacy(mut bytecode: Vec<u8>) -> (JumpTable, Bytes) {
```
rakita, line 20:
```suggestion
pub fn analyze_legacy(bytecode: &Bytes) -> (JumpTable, Bytes) {
```

If we use mut Vec<u8> we are cloning full bytecode from original Bytes. so API should be something similar to take `&Bytes` and return `Bytes` that can be same bytes or it can be extended one.

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -34,5 +48,135 @@ pub fn analyze_legacy(bytetecode: &[u8]) -> JumpTable {
         }
     }
 
-    JumpTable(Arc::new(jumps))
+    // Calculate padding needed to ensure bytecode ends with STOP
+    // If we're at the end and last opcode is not STOP, we need 1 more byte
+    let padding_size = (iterator as usize) - (end as usize) + (opcode != opcode::STOP) as usize;
+    if padding_size > 0 {
+        bytecode.resize(bytecode.len() + padding_size, 0);
```
rakita, line 55: When you change sig to take &Bytes.

We can change this part to
```rust
        let len = self.0.len();
        let mut padded_bytecode = Vec::with_capacity(len + padding_size);
        padded_bytecode.extend_from_slice(&self.0);
        padded_bytecode.resize(len + 33, 0);
```

crates/bytecode/src/legacy/analysis.rs
Bisht13: I feel taking ownership of `Bytes` is better because:

1. Zero-copy optimization: In the common case (no padding needed), we just move the original `Bytes` without cloning
2. Inevitable ownership: Callers need to own the potentially padded result anyway, so taking a reference would force an eventual clone
3. Clear API contract: Taking ownership clearly communicates that the function will either return your bytes or new padded bytes

I've pushed the change according to this, let me know if you differ.

rakita (approved):
lgtm, good change

PR #2424, opened 2025-04-16
Replace InterpreterAction with InterpreterTypes::Output

GarmashAlex:
The implementation adds a where clause to ensure that IW::Output implements From<InterpreterAction> for proper type conversion, allowing custom output types to be used with the interpreter while maintaining backward compatibility.


rakita (approved):
lgtm

Issue #2418, opened 2025-04-15
Helper set functions for AccountInfo


rakita:
Function that sets bytecode that would set both the `code` and `code_hash`

fn docs should point out this is expensive as it calculated bytecode hash.

Other function `with_*` should have their duplicate with `set_*` so it is easier to modify values

shane-moore:
hey @rakita, would be happy to take a look at this

PR #2429, opened 2025-04-18
chore: add AccountInfo setters

shane-moore:
Closes https://github.com/bluealloy/revm/issues/2418

### DIFF
--- a/crates/state/src/account_info.rs
+++ b/crates/state/src/account_info.rs
@@ -132,6 +132,11 @@ impl AccountInfo {
     }
 
     /// Sets the [`AccountInfo`] `code_hash` and clears any cached bytecode.
+    ///
+    /// # Note
+    ///
+    /// Calling this after `set_code(...)` will remove the bytecode you just set.
+    /// If you intend to mutate the code, use only `set_code`.
     #[inline]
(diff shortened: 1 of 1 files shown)


shane-moore:
@rakita, this PR is ready for review 🤙

crates/state/src/account_info.rs
```diff
@@ -117,6 +117,40 @@ impl AccountInfo {
...
+    }
+
+    /// Sets the [`AccountInfo`] `nonce`.
+    #[inline]
+    pub fn set_nonce(&mut self, nonce: u64) -> &mut Self {
+        self.nonce = nonce;
+        self
+    }
+
+    /// Sets the [`AccountInfo`] `code_hash` and clears any cached bytecode.
```
mattsse, line 134: we should also add a note here, because this could be called incorrectly with `set_code().set_code_hash()`

mattsse (dismissed):
these lgtm.

setting the code to none when we set the codehash seems fine

crates/state/src/account_info.rs
shane-moore: I feel you on that, @mattsse. updated the comment in latest commit https://github.com/bluealloy/revm/pull/2429/commits/604869effb264627defdf3e21e120fc30b4962f5

Issue #2436, opened 2025-04-22
perf: avoid allocating in `LegacyAnalyzedBytecode::default`


DaniPopes:
We can micro-optimize `LegacyAnalyzedBytecode::default` (used in `Bytecode::default`) by creating a singleton static instance with `OnceLock` similar to precompiles, and cloning that instead of creating a new bitmap allocation on every call.

PR #2439, opened 2025-04-22
perf: cache and use JumpTable::default

DaniPopes:
The `Bytecode::default` jump table doesn't need to contain any data. Even if it had to, we can still cache it with a singleton as it's an Arc, and then just clone the singleton every time we need a new default jump table.

Closes https://github.com/bluealloy/revm/issues/2436.


rakita (approved):
lgtm. CI pending

PR #2440, opened 2025-04-22
perf(journal): flatten journal

rakita:
Flatten journal entries to one Vec


crates/context/src/journal/inner.rs
```diff
@@ -74,7 +73,7 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
             state: HashMap::default(),
             transient_storage: TransientStorage::default(),
             logs: Vec::new(),
-            journal: vec![vec![]],
+            journal: Vec::with_capacity(4 * 1024),
```
DaniPopes, line 76: this is `4KiB * size_of::<ENTRY>()=88 = 352KiB` which seems a bit excessive?

rakita: Good find! And apology for missing this comment

PR #2443, opened 2025-04-23
feat(database): re-export all public items from alloydb when feature …

ethever:
feat(database): re-export all public items from alloydb when feature is enabled

### DIFF
--- a/crates/database/src/lib.rs
+++ b/crates/database/src/lib.rs
@@ -13,7 +13,7 @@ pub mod in_memory_db;
 pub mod states;
 
 #[cfg(feature = "alloydb")]
-pub use alloydb::{AlloyDB, BlockId};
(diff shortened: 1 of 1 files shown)


crates/database/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub mod in_memory_db;
 pub mod states;
 
 #[cfg(feature = "alloydb")]
-pub use alloydb::{AlloyDB, BlockId};
+pub use alloydb::*;
```
rakita, line 16: Supportive, but lets export them one by one. It is nicer to see what is made public.

rakita (approved):
nit

ethever:
updated

rakita (approved):
Nice

PR #2453, opened 2025-04-26
feat: replace input Bytes and refactored code where required

### LINKED ISSUE #2361: Use range from SharedMemory for CALL input
Replace input (`Bytes`) with enum that contains `Range` or `Bytes` in CalInput.

https://github.com/bluealloy/revm/blob/6b73a3c6875f5875e55e80d9b6edb83c99b4bcb9/crates/interpreter/src/interpreter_action/call_inputs.rs#L9

Idea is to don't clone bytes from shared memory when CALL is called but to give a range of memory where we can find that input.

This new enum probably needs to be propagated in few places and `Memory` trait should probably add `fn global_slice(range)` function that would allow accessing input that is in section of parent call. See what else need to be changed when new enum is changed in `CallInput`

CreateInputs can't use range as Bytes are needed to create Bytecode. We could force it, but it would require unsafe code and could be a footgun.

EOFCreateInputs has a input that can be changed to new enum:
https://github.com/bluealloy/revm/blob/6b73a3c6875f5875e55e80d9b6edb83c99b4bcb9/crates/interpreter/src/interpreter_action/eof_create_inputs.rs#L22

First lets do https://github.com/bluealloy/revm/issues/2360 than this PR

PoulavBhowmick03:
Potentially fixes #2361

Created an enum `CallInput` and changed the input of Bytes to CallInput, and changed code wherever was needed


PoulavBhowmick03:
@rakita is this the intended solution?

rakita:
Have skimmed over it and it looks great. Will see to detail review it and merge it, and in other PR address those TODO's that you left.

PoulavBhowmick03:
> Have skimmed over it and it looks great. Will see to detail review it and merge it, and in other PR address those TODO's that you left.

alright! thanks

crates/handler/src/execution.rs
```diff
@@ -31,7 +31,9 @@ pub fn create_init_frame(tx: &impl Transaction, spec: SpecId, gas_limit: u64) ->
                     tx.caller(),
                     tx.value(),
                     gas_limit,
-                    EOFCreateKind::Tx { initdata: input },
+                    EOFCreateKind::Tx {
+                        initdata: input.clone(),
```
rakita, line 35:
```suggestion
                        initdata: input,
```

It seems unnecessary to clone it

crates/handler/src/precompile_provider.rs
```diff
@@ -99,14 +99,21 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
         let Some(precompile) = self.precompiles.get(address) else {
             return Ok(None);
         };
+        let input_bytes = match &inputs.input {
+            CallInput::Range(_range) => {
+                // Get from parent memory (need access to memory)
+                todo!("Implement memory range access")
+            }
+            CallInput::Bytes(bytes) => bytes.clone(),
```
rakita, line 107: Could probably go without cloning and using a reference.

crates/interpreter/src/interpreter.rs
```diff
@@ -86,7 +86,7 @@ impl Default for Interpreter<EthInterpreter> {
             InputsImpl {
                 target_address: Address::ZERO,
                 caller_address: Address::ZERO,
-                input: Bytes::default(),
+                input: CallInput::Bytes(Bytes::default()),
```
rakita, line 89: We could hgave `CallInput::default()` that would create `CallInput::Bytes(Bytes::default())`

crates/interpreter/src/interpreter/input.rs
```diff
@@ -22,7 +22,13 @@ impl InputsTr for InputsImpl {
     }
 
     fn input(&self) -> &[u8] {
-        &self.input
+        match &self.input {
+            CallInput::Range(_range) => {
```
rakita, line 26: Will need to move this to have access to shared memory, but for this PR this lgtm

crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -1,12 +1,26 @@
 use core::ops::Range;
 use primitives::{Address, Bytes, U256};
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum CallInput {
```
rakita, line 6: Docs are missing

rakita (approved):
PR lgtm, left few nits to be fixed

PoulavBhowmick03:
> PR lgtm, left few nits to be fixed

made the changes, can you ptal and whether i need to change the docs for the enum

crates/handler/src/precompile_provider.rs
rakita: Will check this later

rakita (approved):

PR #2458, opened 2025-04-28
refactor: simplify opcode iter

DaniPopes:
See also `OpcodesIter` in revmc which also returns the immediate byte slice: https://github.com/paradigmxyz/revmc/blob/c3e2a3c8331032cc3875bdfce27d4247e4c7facb/crates/revmc/src/bytecode/opcode.rs#L44

### DIFF
--- a/crates/bytecode/src/iter.rs
+++ b/crates/bytecode/src/iter.rs
@@ -6,45 +6,50 @@ use crate::{opcode, Bytecode, OpCode};
 /// without dealing with the immediate values that follow instructions.
 #[derive(Debug, Clone)]
 pub struct BytecodeIterator<'a> {
-    /// Reference to the underlying bytecode bytes
-    bytes: &'a [u8],
-    /// Current position in the bytecode
-    position: usize,
-    /// End position in the bytecode (to handle original length for legacy bytecode)
-    end: usize,
+    /// Start pointer of the bytecode. Only used to calculate [`position`](Self::position).
+    start: *const u8,
+    /// Iterator over the bytecode bytes.
+    bytes: core::slice::Iter<'a, u8>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/bytecode/src/iter.rs
```diff
@@ -6,66 +6,81 @@ use crate::{opcode, Bytecode, OpCode};
...
+    #[inline]
     pub fn new(bytecode: &'a Bytecode) -> Self {
-        let bytes = bytecode.bytecode();
-        let end = match bytecode {
-            Bytecode::LegacyAnalyzed(analyzed) => analyzed.original_len(),
-            Bytecode::Eip7702(_) => 0,
-            _ => bytes.len(),
-        };
-
+        let bytes = bytecode.original_byte_slice();
```
rakita, line 19: `original_byte_slice` bytes for Eip7702 would iterate over address, and for EOF it would iterate over header.

You should use `.bytecode()` for Lagacy/EOF and for Eip7702 a empty vec

crates/bytecode/src/iter.rs
```diff
@@ -74,37 +89,29 @@ impl<'a> BytecodeIterator<'a> {
...
-        }
-
-        // Get the opcode first with bounds check
-        let opcode = *self.bytes.get(self.position)?;
-        self.skip_to_next_opcode();
-        Some(opcode)
+        self.bytes
+            .next()
+            .copied()
+            .inspect(|&current| self.skip_immediate(current))
```
rakita, line 97: Like for the inspect pattern

rakita (changes requested):
Left a comment for bytes that we use to iterate on, but in general lgtm

rakita (approved):

PR #2459, opened 2025-04-28
chore: make OPCODE_INFO a static

DaniPopes:
`const`s get duplicated at all callsites so it's possible that multiple callers (in multiple crates) end up duplicating the array.


rakita (approved):
good find

PR #2461, opened 2025-04-29
Storage Types Alias

ssolit:
This PR introduces type aliases `StorageKey` and `StorageValue`, following the pattern established in alloy-core. Both are currently aliased to `U256`, as used in revm, but the goal is to improve code clarity and maintainability across the codebase.

By explicitly naming these types:
* We make their roles in the storage system more obvious (key vs value).
* We gain flexibility to adjust the underlying types in the future without widespread refactoring.

For context, `StorageKey` is aliased to `B256` in alloy-core, while revm uses `U256`. At some point it may be useful/necessary to bring these into alignment.


rakita (commented):
This is cool, will recheck changes later but this lgtm

rakita (approved):
Awesome change! Merging it

PR #2464, opened 2025-04-30
Add a custom address to the CreateScheme.

MathieuDutSik:
In [Linera](https://github.com/linera-io/linera-protocol), we use REVM to run EVM apps.

In our usage, it would be very nice if we could add a specific Custom address besides the `Create/Create2` classic possibilities. This would simplify the code a lot.

If needed, I can feature gate this possibility.


rakita (approved):
lgtm, although this enum is public, it is tied to internal call flow so it is fine to change.

PR #2468, opened 2025-05-02
feat(alloydb): Added condition in AlloyDB to check for empty accounts post EIP 161

### LINKED ISSUE #2454: [Bug] AlloyDB should not always return a Some(AccountInfo)
Hello community

Image

I found a bug with alloydb that always returns a non None account, which affects the logical judgment of EIP 161 [here](https://github.com/bluealloy/revm/blob/4e2dcf4b25a2abf94424957a95dca31e82e0c1dc/crates/state/src/lib.rs#L47). I discovered the issue while replaying 0x1ae5f96a8b446ddd51e96019472f7a31fcad3d2e774373ecfbb2ba94ff72fc04 using the example:  https://github.com/bluealloy/revm/blob/main/examples/block_traces/src/main.rs

# Reference

https://github.com/bluealloy/revm/blob/main/crates/database/src/alloydb.rs#L63
https://github.com/bluealloy/revm/blob/4e2dcf4b25a2abf94424957a95dca31e82e0c1dc/crates/state/src/lib.rs#L43

PoulavBhowmick03:
Potentially fixes #2454

### DIFF
--- a/crates/database/src/alloydb.rs
+++ b/crates/database/src/alloydb.rs
@@ -6,7 +6,7 @@ use alloy_provider::{
 use alloy_transport::TransportError;
 use core::error::Error;
 use database_interface::{async_db::DatabaseAsyncRef, DBErrorMarker};
-use primitives::{Address, B256, U256};
+use primitives::{hardfork::SpecId, Address, B256, U256};
 use state::{AccountInfo, Bytecode};
 use std::fmt::Display;
 
@@ -39,15 +39,17 @@ pub struct AlloyDB<N: Network, P: Provider<N>> {
     /// The block number on which the queries will be based on.
     block_number: BlockId,
     _marker: core::marker::PhantomData<fn() -> N>,
+    spec_id: SpecId,
 }
 
 impl<N: Network, P: Provider<N>> AlloyDB<N, P> {
     /// Creates a new AlloyDB instance, with a [Provider] and a block.
-    pub fn new(provider: P, block_number: BlockId) -> Self {
+    pub fn new(provider: P, block_number: BlockId, spec_id: SpecId) -> Self {
         Self {
             provider,
             block_number,
             _marker: core::marker::PhantomData,
+            spec_id,
         }
     }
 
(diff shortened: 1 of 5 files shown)


crates/database/src/alloydb.rs
```diff
@@ -78,10 +80,15 @@ impl<N: Network, P: Provider<N>> DatabaseAsyncRef for AlloyDB<N, P> {
 
         let balance = balance?;
         let code = Bytecode::new_raw(code?.0.into());
-        let code_hash = code.hash_slow();
         let nonce = nonce?;
+        let code_hash = code.hash_slow();
 
-        Ok(Some(AccountInfo::new(balance, nonce, code_hash, code)))
+        let empty_acc = SpecId::is_enabled_in(self.spec_id, SpecId::SPURIOUS_DRAGON);
+        if empty_acc && nonce == 0 && balance.is_zero() && code.is_empty() {
```
rakita, line 87: This is not a solution for it.

I don't think there is right fix for this. You would need to fetch a additional info that is not available and that is if account exist or not. As of now with just `nonce == 0 && balance.is_zero() && code.is_empty()` we don't know if account is empty (pre state clear empty) or not.

rakita:
I was looking at JSONRPC https://ethereum.github.io/execution-apis/api-documentation/ and I couldn't find a rpc that would return if account is empty or not pre state clear EIP. So I don't think this can be fixed.

Would be good to add a doc comment that would explain that using `AlloyDB` before state clear could give incorrect gas bcs of inability of jsonrpc to distinct empty and empty not existing accounts.

PoulavBhowmick03:
> I was looking at JSONRPC https://ethereum.github.io/execution-apis/api-documentation/ and I couldn't find a rpc that would return if account is empty or not pre state clear EIP. So I don't think this can be fixed.
> 
> Would be good to add a doc comment that would explain that using `AlloyDB` before state clear could give incorrect gas bcs of inability of jsonrpc to distinct empty and empty not existing accounts.

I see. Makes sense, alright will add the doc comment in that manner

rakita:
Closing this PR as it is not possible to fix it

PR #2495, opened 2025-05-09
fix(op): mark caller account as touched

PinelliaC:
Currently, `op-revm` does not invoke `mark_touch` on the `validate_against_state_and_deduct_caller`. 
Adding this call will bring the behavior in line with [bluealloy/revm v71 (see handler/src/pre_execution.rs line 174).](https://github.com/bluealloy/revm/blob/v71/crates/handler/src/pre_execution.rs#L174)


rakita (approved):
good catch

PR #2496, opened 2025-05-09
fix(examples/block_traces): ensure traces/ directory is created before writing

ethever:
Should we ensure traces/ directory is created before writing?

When running the examples/block_traces example, it fails with the following error if the traces/ directory does not exist:

``` bash
thread 'main' panicked at examples/block_traces/src/main.rs:98:15:
Failed to open file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```


rakita (approved):
lgtm

PR #2507, opened 2025-05-10
feat: added CallInput::bytes<CTX>(ctx: &CTX) -> Bytes {} function

### LINKED ISSUE #2502: Add `CallInput::bytes<CTX>(ctx: &CTX) -> Bytes {}` function
`fn bytes` should use ctx to close input from shared memory similar to:
https://github.com/bluealloy/revm/blob/e6009b3a4bb9f19c66043015dfac1aab308d2cbd/crates/handler/src/precompile_provider.rs#L109-L118

PoulavBhowmick03:
Fixes #2502

### DIFF
--- a/crates/interpreter/src/interpreter_action/call_inputs.rs
+++ b/crates/interpreter/src/interpreter_action/call_inputs.rs
@@ -1,6 +1,6 @@
+use context_interface::{ContextTr, LocalContextTr};
 use core::ops::Range;
 use primitives::{Address, Bytes, U256};
-
 /// Input enum for a call.
 ///
 /// As CallInput uses shared memory buffer it can get overridden if not used directly when call happens.
@@ -31,6 +31,27 @@ impl CallInput {
     /// Returns `true` if the call input is empty.
     pub fn is_empty(&self) -> bool {
         self.len() == 0
+    }
+
+    /// Returns the bytes of the call input.
+    /// # Allocation
+    /// If this `CallInput` is a `SharedBuffer`, the slice will be copied
+    /// into a fresh `Bytes` buffer, i.e., this always allocates when reading
+    /// from shared memory.
+    pub fn bytes<CTX>(&self, ctx: &mut CTX) -> Bytes
+    where
+        CTX: ContextTr,
+    {
+        match self {
+            CallInput::Bytes(bytes) => bytes.clone(),
+            CallInput::SharedBuffer(range) => {
+                if let Some(slice) = ctx.local().shared_memory_buffer_slice(range.clone()) {
+                    Bytes::from(slice.to_vec())
+                } else {
+                    Bytes::new()
+                }
+            }
+        }
     }
 }
 
--- a/crates/interpreter/src/interpreter_action/call_inputs.rs
+++ b/crates/interpreter/src/interpreter_action/call_inputs.rs
@@ -34,23 +34,25 @@ impl CallInput {
     }
 
     /// Returns the bytes of the call input.
+    ///
+    /// SharedMemory buffer can be shrunked or overwritten if the child call returns the
+    /// shared memory context to its parent, the range in `CallInput::SharedBuffer` can show unexpected data.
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -32,6 +32,23 @@ impl CallInput {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    /// Returns the bytes of the call input.
+    pub fn bytes<CTX>(&self, ctx: &mut CTX) -> Bytes
```
mattsse, line 37: this makes sense, 

I assume there is now easy way to return something like `Cow` here because we can't unify `Bytes` and `Ref<[u8]>` 

we should at least add a note here that this will allocate if the data is shared

PoulavBhowmick03: > this makes sense,
> 
> I assume there is now easy way to return something like `Cow` here because we can't unify `Bytes` and `Ref<[u8]>`
> 
> we should at least add a note here that this will allocate if the data is shared

that makes sense. but for now then should we stick to Bytes and add a note, or do we change it to return Cow?

crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -32,6 +32,27 @@ impl CallInput {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    /// Returns the bytes of the call input.
+    /// # Allocation
+    /// If this `CallInput` is a `SharedBuffer`, the slice will be copied
+    /// into a fresh `Bytes` buffer, i.e., this always allocates when reading
+    /// from shared memory.
+    pub fn bytes<CTX>(&self, ctx: &mut CTX) -> Bytes
```
rakita, line 41:
```suggestion
    ///
    /// SharedMemory buffer can be shrunked or overwritten if the child call returns the
    /// shared memory context to its parent, the range in `CallInput::SharedBuffer` can show unexpected data.
    ///
    /// # Allocation
    ///
    /// If this `CallInput` is a `SharedBuffer`, the slice will be copied
    /// into a fresh `Bytes` buffer, which can pose a performance penalty.
    pub fn bytes<CTX>(&self, ctx: &mut CTX) -> Bytes
```

crates/interpreter/src/interpreter_action/call_inputs.rs
```diff
@@ -32,6 +32,27 @@ impl CallInput {
...
+        CTX: ContextTr,
+    {
+        match self {
+            CallInput::Bytes(bytes) => bytes.clone(),
+            CallInput::SharedBuffer(range) => {
+                if let Some(slice) = ctx.local().shared_memory_buffer_slice(range.clone()) {
+                    Bytes::from(slice.to_vec())
+                } else {
+                    Bytes::new()
+                }
```
rakita, line 52:
```suggestion
                ctx.local().shared_memory_buffer_slice(range.clone()).map(|b| Bytes::from(b.to_vec).unwrap_or_default()
```

Is slightly better

rakita (approved):
lgtm, left few nits

crates/interpreter/src/interpreter_action/call_inputs.rs
rakita: Didn't find a way to use `Bytes` and `Ref`, `Ref` sucks.

@PoulavBhowmick03 note would suffice for now

crates/interpreter/src/interpreter_action/call_inputs.rs
rakita: probably need a fmt.

PoulavBhowmick03:
@rakita ptal now

rakita (approved):

zerosnacks:
Thanks! Was just looking for this as well

PR #2511, opened 2025-05-12
feat: make Journal::set_code to be EIP-7702 zero address bytecode aware

### LINKED ISSUE #2501: Journal::set_code to be EIP-7702 zero address bytecode aware
make set_code EIP-7702 aware, if address of bytecode is zero remove the bytecode from account.

More info in 8 point in Behaviour section: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md#behavior
```
If address is 0x0000000000000000000000000000000000000000, do not write the delegation indicator. Clear the account's code by reseting the account's code hash to the empty code hash
```

louisbrown0212:
Resolves #2501

### DIFF
--- a/crates/context/src/journal/inner.rs
+++ b/crates/context/src/journal/inner.rs
@@ -178,6 +178,13 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
     /// Assume account is warm.
     #[inline]
     pub fn set_code(&mut self, address: Address, code: Bytecode) {
+        if let Bytecode::Eip7702(eip7702_bytecode) = &code {
+            if eip7702_bytecode.address().is_zero() {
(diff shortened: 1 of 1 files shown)


crates/context/src/journal/inner.rs
```diff
@@ -176,8 +176,17 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
     /// Use it only if you know that acc is warm.
     ///
     /// Assume account is warm.
+    ///
+    /// In case of EIP-7702 code with zero address, the bytecode will be erased.
```
rakita, line 180:
```suggestion
    /// In case of EIP-7702 code with zero address, the bytecode will be erased.
```

PR #2517, opened 2025-05-14
feat: transact multi tx

### LINKED ISSUE #2362: Multi transaction execution
Execution multiple transactions can help with performance as we don't need to re-load account inside Journal, this means fewer hashmap lookups on both inner state and other state. Biggest example is on the beneficiary account that is always inserted inside the journal, and this overhead should've been removed.

The idea is to have an incarnation inside accounts of Journal state. This incarnation would allow us to mark accounts that previous transactions touched/loaded and look at them as cold loaded, even if they are inside inner state. This would aggregating all block state into one Hashmap and skip aggregation/commit that is needed on single transaction execution.

Additionally as the journal would stay unchanged, it would allow transaction reverts.

rakita:
closes #2362


gakonst:
What's the impact on perf?

bins/revme/Cargo.toml
```diff
@@ -12,7 +12,6 @@ repository.workspace = true
 # revm
 revm = { workspace = true, features = [
     "std",
-    "hashbrown",
```
rakita, line 15: duplicated feature

PR #2520, opened 2025-05-15
chore: enable devcontainer

zouguangxian:
This PR enables devcontainer, which provides an isolated and consistent development environment, furthermore, it is helpful to use codespaces for development.


rakita:
Do not personally use it, but it doesn't hurt to have it

PR #2521, opened 2025-05-15
feat: support riscv32-unknown-none-elf

zouguangxian:
This PR enhances the codebase's compatibility with no_std and platforms lacking atomic support by:

- Conditionally importing and using Arc or Rc based on the presence of atomics and the std feature.
- Updating trait definitions to use #[cfg_attr] for auto_impl, ensuring correct trait implementations for both atomic and non-atomic targets.
- Adding extern crate alloc where necessary for no_std environments.
- Applying these changes across multiple crates, including bytecode, context/interface, database/interface, handler, and interpreter.

These changes improve portability and make the codebase more robust for embedded and constrained environments. No functional logic is changed; only trait bounds and imports are updated for broader platform support.

### DIFF
--- a/crates/bytecode/src/bytecode.rs
+++ b/crates/bytecode/src/bytecode.rs
@@ -4,6 +4,8 @@
 //! - Legacy bytecode with jump table analysis. Found in [`LegacyAnalyzedBytecode`]
 //! - EOF ( EMV Object Format) bytecode introduced in Osaka that.
 //! - EIP-7702 bytecode, introduces in Prague and contains address to delegated account.
+#[cfg(not(feature = "std"))]
+extern crate alloc;
 
 use crate::{
     eip7702::{Eip7702Bytecode, EIP7702_MAGIC_BYTES},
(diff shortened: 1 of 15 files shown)


rakita:
Is there a concrete need for this? Until now, I was not asked or saw that this is a blocker for anyone. We are building project in `riscv32imac-unknown-none-elf`. This PR would drop an atomic requirement.

zouguangxian:
To enable integration with modern zero-knowledge virtual machines (zkVMs) like Jolt and SP1, revm needs to support compilation for the riscv32im-unknown-none-elf target. This environment—based on a 32-bit RISC-V architecture without an operating system and using the ELF binary format—is the standard runtime for many zkVMs. Supporting this target allows revm to execute transactions within zkVMs, enabling verifiable EVM execution.

rakita:
Will see to do this slightly differently as a feature flag in revm_primitives, making it less intrusive.

kevaundray:
> To enable integration with modern zero-knowledge virtual machines (zkVMs) like Jolt and SP1, revm needs to support compilation for the riscv32im-unknown-none-elf target. This environment—based on a 32-bit RISC-V architecture without an operating system and using the ELF binary format—is the standard runtime for many zkVMs. Supporting this target allows revm to execute transactions within zkVMs, enabling verifiable EVM execution.

I think most zkVMs should view the atomic operations as no-ops -- If you try to compile reth using Sp1, then it should work even though reth uses `Arc` in quite a few places for example.

I think we also want to move towards zkVMs supporting imac + floating points which is what we call `gc` where:

- The atomic operations are no-ops
- The floating point operations are software emulated which is expensive but okay because the EL state transition function should hopefully not be using floating points. Maybe somewhere in the stdlib.

`gc` ie general compute, is advantageous to support because most other higher level languages like go/c# will support it

rakita:
Thanks for the insights @kevaundray. I didn't notice any usage of floats.

Closing this.

Issue #2477, opened 2025-05-06
Detach blob params from SpecId


mattsse:
with https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7892.md

blob params are scheduled by timestamp, so this will no longer work or needs some changes so that it supports multiple SpecId -> params pairs

https://github.com/bluealloy/revm/blob/6b057f1fe5a21039bae0cf290f136bdc9ee6177d/crates/context/src/cfg.rs#L127-L137

due to how this works, we could either just promote blobparams as setting in the CfgEnv?

https://github.com/bluealloy/revm/blob/6b057f1fe5a21039bae0cf290f136bdc9ee6177d/crates/context/src/cfg.rs#L152-L152

from reth pov that would be the easiest solution because we do

`params_at_timestamp()` -> BlobParams and could set maxblobcount to cfgenv like this

ref:

https://github.com/paradigmxyz/reth/blob/b5c9984fd4351b063bf14357dd67fc56bd79a05d/crates/ethereum/evm/src/lib.rs#L154-L160

cc @rkrasiuk

rakita:
Have made max blobs param optional. If param is not set max limit will not be checked

PR #2532, opened 2025-05-21
feat: make blob max number optional

rakita:
Closing https://github.com/bluealloy/revm/issues/2477

Have made this parameter optional as it only checks for the transaction blob len is not more than limi. It does not affect the gas calculations.

### DIFF
--- a/crates/context/src/cfg.rs
+++ b/crates/context/src/cfg.rs
@@ -24,11 +24,10 @@ pub struct CfgEnv<SPEC = SpecId> {
     pub limit_contract_code_size: Option<usize>,
     /// Skips the nonce validation against the account's nonce
     pub disable_nonce_check: bool,
-    /// Blob target count. EIP-7840 Add blob schedule to EL config files.
+    /// Blob max count. EIP-7840 Add blob schedule to EL config files.
     ///
-    /// If this config is not set, the default value will be used.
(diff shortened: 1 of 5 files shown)


crates/context/src/cfg.rs
```diff
@@ -125,15 +125,19 @@ impl<SPEC> CfgEnv<SPEC> {
...
-        blob_params.sort_by_key(|(id, _, _)| *id);
-        self.blob_target_and_max_count = blob_params;
+    pub fn set_blob_max_and_target_count(&mut self, blob_target: u64, blob_max_count: u64) {
+        self.blob_target_and_max_count = Some((blob_target, blob_max_count));
+    }
+
+    /// Clears the blob target and max count over hardforks.
+    pub fn clear_blob_max_and_target_count(&mut self) {
+        self.blob_target_and_max_count = None;
     }
```
mattsse, line 141: these makes sense, since this is breaking this should highlight all locations where we need to update this in reth

mattsse (approved):
makes sense,

ptal @rkrasiuk

crates/context/src/cfg.rs
rakita: Evm uses only blob_max_count field, and that is to only check if tx has less than max, so making this check optional is fine. And if num of blobs are checked outside of revm there is no need to use `set_blob_max_count` fn

PR #2545, opened 2025-05-26
chore: use sha3 by default

rakita:


lispc:
hi @rakita , actually this PR is a breaking change.   

In alloy-primitives, if "native-keccak" is enabled, and none of "tiny-keccak" / "sha3-keccak" is enabled, then it will choose "native-keccak".

After this PR, since here we always use "sha3-keccak", it is no longer possible to use "native-keccak".

So maybe we'd better expose "revm/sha3-keccak" as a feature flag, who wants to use it can enable it.

Thanks!

rakita:
@lispc I see, lets do exactly what you propose.

PR #2550, opened 2025-05-28
WIP: feat(Osaka): eip7907 Meter Contract Code Size And Increase Limit

jgresham:
EIP: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7907.md

- [X] Add new max contract size and max initcode size variables
- [X] Add additional account code warm status
- [x] (cold access) Implement additional gas fees for size > previous max sizes
- [x] set code warm after `create`/`create2`
- [x] Test new limit of max initcode & contract size
- [ ] Test code cold/warm statuses. Test with txn containing multiple calls to same account code.
- [ ] TBD on [eip7702](https://ethereum-magicians.org/t/eip-7907-meter-contract-code-size-and-increase-limit/23156/36?u=jgresham) and `EXTCODESIZE` related changes

execution specs reference implementation https://github.com/ethereum/execution-specs/pull/1231/files

I'll try to get more of this ready and tested in advance of fusaka devnet-2 testing

### DIFF
--- a/crates/context/interface/src/journaled_state.rs
+++ b/crates/context/interface/src/journaled_state.rs
@@ -171,7 +171,11 @@ pub trait JournalTr {
     ) -> Result<StateCodeLoad<B256>, <Self::Database as Database>::Error> {
         let acc = self.load_account_code(address)?;
         if acc.is_empty() {
-            return Ok(StateCodeLoad::new(B256::ZERO, acc.is_cold, acc.is_code_cold));
+            return Ok(StateCodeLoad::new(
+                B256::ZERO,
+                acc.is_cold,
+                acc.is_code_cold,
+            ));
         }
         // SAFETY: Safe to unwrap as load_code will insert code if it is empty.
         let code = acc.info.code.as_ref().unwrap();
@@ -303,17 +307,17 @@ pub struct StateCodeLoad<T> {
 }
 
 impl<T> Deref for StateCodeLoad<T> {
-  type Target = T;
+    type Target = T;
 
-  fn deref(&self) -> &Self::Target {
-      &self.data
-  }
+    fn deref(&self) -> &Self::Target {
+        &self.data
+    }
 }
 
 impl<T> DerefMut for StateCodeLoad<T> {
-  fn deref_mut(&mut self) -> &mut Self::Target {
-      &mut self.data
-  }
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.data
+    }
 }
 
 impl<T> StateCodeLoad<T> {
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -307,10 +307,13 @@ mod tests {
     use crate::{ExecuteCommitEvm, MainBuilder, MainContext};
     use bytecode::opcode;
     use context::{
-        result::{EVMError, ExecutionResult, HaltReason, InvalidTransaction, Output}, CfgEnv, Context, TxEnv
+        result::{EVMError, ExecutionResult, HaltReason, InvalidTransaction, Output},
+        CfgEnv, Context, TxEnv,
     };
     use database::{CacheDB, EmptyDB};
-    use primitives::{address, Address, Bytes, TxKind, eip3860, eip7825, eip7907, hardfork::SpecId};
+    use primitives::{
+        address, eip3860, eip7825, eip7907, hardfork::SpecId, Address, Bytes, TxKind,
+    };
 
     fn deploy_contract(
         bytecode: Bytes,
@@ -324,7 +327,9 @@ mod tests {
         let block_and_txn_gas_limit_cap = gas_limit.unwrap_or(eip7825::TX_GAS_LIMIT_CAP);
         cfg.tx_gas_limit_cap = Some(block_and_txn_gas_limit_cap);
 
-        let ctx = Context::mainnet().with_db(CacheDB::<EmptyDB>::default()).with_cfg(cfg);
+        let ctx = Context::mainnet()
+            .with_db(CacheDB::<EmptyDB>::default())
+            .with_cfg(cfg);
 
         let mut evm = ctx.build_mainnet();
         evm.transact_commit(TxEnv {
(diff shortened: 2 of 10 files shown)


rakita:
Appreciate the effort! Lets do this more holistically.

Change this field https://github.com/bluealloy/revm/blob/f8a9630ddd71e6a388d9706e64b0a2ca4c81b136/crates/context/src/cfg.rs#L24 and add `fn max_code_size(&self) -> usize` to `CfgEnv` that would give us a runtime limit depending on SpecId, or if concrete value is set

jgresham:
> Appreciate the effort! Lets do this more holistically.
> 
> Change this field
> 
> https://github.com/bluealloy/revm/blob/f8a9630ddd71e6a388d9706e64b0a2ca4c81b136/crates/context/src/cfg.rs#L24
> 
> and add `fn max_code_size(&self) -> usize` to `CfgEnv` that would give us a runtime limit depending on SpecId, or if concrete value is set

will do! any and all feedback and suggestions is much appreciated

jgresham:
> Appreciate the effort! Lets do this more holistically.
> 
> Change this field
> 
> https://github.com/bluealloy/revm/blob/f8a9630ddd71e6a388d9706e64b0a2ca4c81b136/crates/context/src/cfg.rs#L24
> 
> and add `fn max_code_size(&self) -> usize` to `CfgEnv` that would give us a runtime limit depending on SpecId, or if concrete value is set

Added specId conditional to `max_code_size`, but wasn't sure what the change to `pub limit_contract_code_size: Option<usize>` should be.

Also, I'm planning to work on the gas metering and additional code warmed (account code loaded) state next. I can tag you when I think this PR (and the EIP is more finalized) if you'd like, otherwise any reviews in the meantime are appreciated.

crates/handler/src/frame.rs
```diff
@@ -751,8 +751,8 @@ pub fn return_create<JOURNAL: JournalTr>(
         return;
     }
 
-    // EIP-170: Contract code size limit
-    // By default limit is 0x6000 (~25kb)
+    // EIP-170 & EIP-7907: Contract code size limit
+    // By default limit is 0x40000 (~262kb)
```
rakita: Would add additional info about this and not override the previous comment.
```suggestion
    // EIP-170: Contract code size limit to 0x6000 (~25kb)
    // EIP-7907 increased this limit to 0x40000 (~262kb).
```

crates/primitives/src/constants.rs
```diff
@@ -17,8 +17,14 @@ pub const STACK_LIMIT: usize = 1024;
 
 /// EIP-3860: Limit and meter initcode
 ///
-/// Limit of maximum initcode size is `2 * MAX_CODE_SIZE`.
-pub const MAX_INITCODE_SIZE: usize = 2 * eip170::MAX_CODE_SIZE;
+/// Limit of maximum initcode size is `2 * MAX_CODE_SIZE_170` (EIP-170).
+/// Used for testing as chains haven't yet upgraded to EIP-7907.
+pub const MAX_INITCODE_SIZE_3860: usize = 2 * eip170::MAX_CODE_SIZE_170;
```
rakita: Need to cleanup consts and how they are used.

`MAX_INITCODE_SIZE` should be put in `eip3860.rs` and references as `eip3860::MAX_INITCODE_SIZE` additional constant should be placed in `eip7907.rs`.

`MAX_CODE_SIZE_170` const name should stay as `MAX_CODE_SIZE` but on every place it gets used `use primitives::eip170` should be included and constant used as `eip170::MAX_CODE_SIZE`

crates/context/interface/src/journaled_state.rs
```diff
@@ -290,6 +290,53 @@ impl<T> StateLoad<T> {
     }
 }
 
+/// Duplicates StateLoad, adding code cold status.
```
jgresham, line 297: @rakita I needed to return `is_code_cold` from account load calls to use it for gas calculations. Is there a better way to extend `StateLoad` without duplicating the rest of the code here? `StateLoad` is widely used and was hesitant to change it

crates/handler/src/validation.rs
```diff
@@ -307,30 +307,62 @@ mod tests {
...
     ) -> Result<ExecutionResult, EVMError<core::convert::Infallible>> {
-        let ctx = Context::mainnet().with_db(CacheDB::<EmptyDB>::default());
+        let mut cfg = CfgEnv::default();
+        if let Some(spec_id) = spec_id {
+            cfg.spec = spec_id;
+        }
+        let block_and_txn_gas_limit_cap = gas_limit.unwrap_or(eip7825::TX_GAS_LIMIT_CAP);
+        cfg.tx_gas_limit_cap = Some(block_and_txn_gas_limit_cap);
+
+        let ctx = Context::mainnet().with_db(CacheDB::<EmptyDB>::default()).with_cfg(cfg);
```
jgresham: @rakita I was thinking to add more tests to check the journal/db if code is warmed properly and txns with multiple `CALL` and `CREATE`. Should this be done in this repo and if so are there any good examples of this? or should these scenarios be covered in the `EEST` tests?

crates/interpreter/src/gas/calc.rs
```diff
@@ -113,14 +113,18 @@ pub const fn copy_cost_verylow(len: usize) -> Option<u64> {
...
+    let mut base_gas = if spec_id.is_enabled_in(SpecId::BERLIN) {
         warm_cold_cost(is_cold)
     } else if spec_id.is_enabled_in(SpecId::TANGERINE) {
         700
     } else {
         20
     };
+    if spec_id.is_enabled_in(SpecId::OSAKA) && is_code_cold {
+        let large_contract_cost = large_contract_code_size_cost(len);
+        base_gas += large_contract_cost;
```
jgresham: Saw code around this doing safe adds, but was getting an error

rakita:
@jgresham we had ian nternal discussion and are probably changing it to just increase of contract limit increase. Will take this over to push it over the line.

crates/context/interface/src/journaled_state.rs
rakita: Looks reasonable,

rakita:
Closing in favour of https://github.com/bluealloy/revm/pull/2605 . Awesome work @jgresham!

jgresham:
> Closing in favour of #2605 . Awesome work @jgresham!

awesome, thanks!

PR #2561, opened 2025-05-31
refactor: unify calling of journal account loading

rakita:
Unify `initial_account_load` into `load_account_optional`


rakita:
hm this changes does not effect the Interpreter and codspeed shows the degradation is there. It is not related

PR #2565, opened 2025-06-02
chore(docs): context crate lints

rakita:
Add docs for crate


rakita:
bench not touched, interesting observation is that difference in performance is related to malloc on `EthInterpreter::new` (Probably 1024 stack alloc)

PR #2568, opened 2025-06-02
chore(docs): add lints to database-interface and op-revm crates

rakita:
Added docs to database-interface and op-revm

Simplify `L1Block::data_gas` impl (change covered by tests)


rakita:
Change is docs related.

PR #2569, opened 2025-06-02
chore(op-revm): type alias for default OptimismEvm 

0xDmtri:
Changes:

1. Introduced type alias OptimismEvm for the default OpEvm configuration. 
2. Renamed for consistency with Mainnet Revm conventions and improved clarity. 

If you prefer the original name, let me know, and I'll discard this PR and include only the type alias :)

### DIFF
--- a/crates/op-revm/src/api.rs
+++ b/crates/op-revm/src/api.rs
@@ -5,5 +5,5 @@ pub mod default_ctx;
 pub mod exec;
 
 pub use builder::OpBuilder;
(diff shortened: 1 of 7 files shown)


rakita:
It is better to use `Op` prefix as it is a unified name used everywhere.

0xDmtri:
> It is better to use `Op` prefix as it is a unified name used everywhere.

Alright, ill revert back to what it was not :)

PR #2570, opened 2025-06-02
feat(database): Implement DatabaseRef for State

### LINKED ISSUE #2566: Implement DatabaseRef for State
I would suggest to have both traits (`Database` and `DatabaseRef`) implemented for `State`. 

Currently `State` implements the `Database` trait, but not `DatabaseRef`. In case where e.g. an account is only read it seems more reasonable to just use the reference. This allows to reuse functions for the `State` that only accept `DatabaseRef`.

For State here:
https://github.com/bluealloy/revm/blob/9f94a2a4cd03f3cbad68ee4966a13c7a32ecda6a/crates/database/src/states/state.rs#L28-L29

Implement the equivalent of `Database` for `DatabaseRef`:
https://github.com/bluealloy/revm/blob/9f94a2a4cd03f3cbad68ee4966a13c7a32ecda6a/crates/database/src/states/state.rs#L228

cakevm:
Fixes #2566, implementing `DatabaseRef` for `State`. 

The ref db implementation does not write to the cache, it tries to read from the cache, then optional from bundle state, and if both fails from the database. For the `storage_ref` I am not 100% sure if I got the account status check right and if bundle state is correctly used here.

### DIFF
--- a/crates/database/src/states/state.rs
+++ b/crates/database/src/states/state.rs
@@ -2,7 +2,6 @@ use super::{
     bundle_state::BundleRetention, cache::CacheState, plain_account::PlainStorage, BundleState,
     CacheAccount, StateBuilder, TransitionAccount, TransitionState,
 };
-use crate::AccountStatus;
 use bytecode::Bytecode;
 use database_interface::{Database, DatabaseCommit, DatabaseRef, EmptyDB};
 use primitives::{hash_map, Address, HashMap, StorageKey, StorageValue, B256, BLOCK_HASH_HISTORY};
@@ -319,18 +318,14 @@ impl<DB: DatabaseRef> DatabaseRef for State<DB> {
     type Error = DB::Error;
 
     fn basic_ref(&self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
-        match self.cache.accounts.get(&address) {
-            Some(account) => {
-                // Account is already in cache
+        // Account is already in cache
+        if let Some(account) = self.cache.accounts.get(&address) {
+            return Ok(account.account_info());
+        }
+        // If bundle state is used, check if account is in bundle state
+        if self.use_preloaded_bundle {
+            if let Some(account) = self.bundle_state.account(&address) {
                 return Ok(account.account_info());
-            }
-            None => {
-                if self.use_preloaded_bundle {
-                    // Check if account is in bundle state
-                    if let Some(account) = self.bundle_state.account(&address) {
-                        return Ok(CacheAccount::from(account.clone()).account_info());
-                    }
-                }
             }
         }
         // If not found, load it from database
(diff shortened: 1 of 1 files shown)


crates/database/src/states/state.rs
```diff
@@ -314,6 +315,102 @@ impl<DB: Database> DatabaseCommit for State<DB> {
...
+            }
+            None => {
+                if self.use_preloaded_bundle {
+                    // Check if account is in bundle state
+                    if let Some(account) = self.bundle_state.account(&address) {
+                        return Ok(CacheAccount::from(account.clone()).account_info());
+                    }
+                }
+            }
+        }
```
rakita: Lets simplify this path
```rust
if let Some(account) = self.cache.accounts.get(... {
      return Ok(..
}

if self.use _preloaded . {

}

self.database.basic_ref(address)
```

crates/database/src/states/state.rs
```diff
@@ -314,6 +315,102 @@ impl<DB: Database> DatabaseCommit for State<DB> {
...
+        match self.cache.accounts.get(&address) {
+            Some(account) => {
+                // Account is already in cache
+                return Ok(account.account_info());
+            }
+            None => {
+                if self.use_preloaded_bundle {
+                    // Check if account is in bundle state
+                    if let Some(account) = self.bundle_state.account(&address) {
+                        return Ok(CacheAccount::from(account.clone()).account_info());
```
rakita: Try not to clone full account, but to just extract account info from it. Cloning is expensive as it probably contain storage.

crates/database/src/states/state.rs
rakita: Do this for rest of calls

crates/database/src/states/state.rs
```diff
@@ -314,6 +315,102 @@ impl<DB: Database> DatabaseCommit for State<DB> {
...
+            Some(acc_entry) => {
+                // Check if account is destroyed or not existing
+                if matches!(
+                    acc_entry.status,
+                    AccountStatus::Destroyed
+                        | AccountStatus::DestroyedAgain
+                        | AccountStatus::LoadedNotExisting
+                ) {
+                    return Ok(StorageValue::ZERO);
+                }
```
rakita: Please look how this is done in `Database` impl https://github.com/bluealloy/revm/pull/2570/files#diff-9a82735f9034f0152663798e390258a8e5b56aee5afe6236c6a28c9214c54297L263

crates/database/src/states/state.rs
cakevm: Thanks, look much better.

crates/database/src/states/state.rs
cakevm: In my understanding it can happen that the account is maybe not found in the cache here. For that reason the last bastion is always the database lookup. I adapted the logic as in `Database`, if there is a cache and the `PlainAccount` is populated, then check if the storage exist. If storage not exists, check if the account is destroyed, or fallback to database lookup.

rakita (approved):
lgtm

PR #2571, opened 2025-06-03
refactor: use a vec-based map to improve the performance of small key/value set access

dyxushuai:
# WIP

draft now for benching with codspeed!


dyxushuai:
Use `vec map` instead of `hash map` directly is not work, I am working on another approach, using `vec+hash` map with index exported

PR #2574, opened 2025-06-03
chore: Avoid clone before converting ref BundleAccount to CacheAccount

cakevm:
I just learned in #2570, that the clone here could be avoided. Since the conversion is not used somewhere else I adapted it.

### DIFF
--- a/crates/database/src/states/cache_account.rs
+++ b/crates/database/src/states/cache_account.rs
@@ -13,6 +13,12 @@ pub struct CacheAccount {
     pub status: AccountStatus,
 }
 
+impl From<BundleAccount> for CacheAccount {
+    fn from(account: BundleAccount) -> Self {
+        CacheAccount::from(&account)
+    }
+}
+
 impl From<&BundleAccount> for CacheAccount {
     fn from(account: &BundleAccount) -> Self {
         let storage = account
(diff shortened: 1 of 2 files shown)


crates/database/src/states/cache_account.rs
```diff
@@ -13,8 +13,8 @@ pub struct CacheAccount {
     pub status: AccountStatus,
 }
 
-impl From<BundleAccount> for CacheAccount {
-    fn from(account: BundleAccount) -> Self {
+impl From<&BundleAccount> for CacheAccount {
```
mattsse, line 22: this is a breaking change, but idk if used in reth

but we can also just keep
`impl From<BundleAccount> for CacheAccount {`

mattsse (dismissed):

crates/database/src/states/cache_account.rs
cakevm: Sure, did not like to break anything. We can have both :)

crates/database/src/states/state.rs
```diff
@@ -188,9 +188,7 @@ impl<DB: Database> State<DB> {
             hash_map::Entry::Vacant(entry) => {
                 if self.use_preloaded_bundle {
                     // Load account from bundle state
-                    if let Some(account) =
-                        self.bundle_state.account(&address).cloned().map(Into::into)
-                    {
+                    if let Some(account) = self.bundle_state.account(&address).map(Into::into) {
```
rakita, line 191: We should skip this cloning as it is unnecessary to clone the storage. Lets do this in he other PR.

PR #2580, opened 2025-06-04
refactor!: remove `TxEnv::tx_type` field

klkvr:
This field is very easy to forget about, and have caused weird bugs multiple times already. It can also always be easily inferred from the specified `TxEnv` values.

This PR removes the `tx_env` and adjusts logic to be invoked based on present fields instead


klkvr:
It seems that there are some statetests failing that don't make much sense in scope of our usage patterns — e.g there are tests asserting that eip7702 tx with empty auth list will fail validation

however, the way `TxEnv` is handled with this PR is that transaction will not be considered to be 7702 tx unless it has a non-empty authorization list

should we just ignore those kinds of tests that don't fit into our tx representation?

rakita:
Not relevant anymore

PR #2584, opened 2025-06-05
chore: use iter_batched for transfer bench

Rjected:
I noticed the clone ending up in the profiles, instead we can probably use iter_batched here


bins/revme/src/cmd/bench/analysis.rs
```diff
@@ -23,8 +23,15 @@ pub fn run(criterion: &mut Criterion) {
     };
     let mut evm = context.build_mainnet();
     criterion.bench_function("analysis", |b| {
-        b.iter(|| {
-            let _ = evm.transact(tx.clone());
-        });
+        b.iter_batched(
+            || {
+                // create a transaction input
+                tx.clone()
```
mattsse, line 29: ah yeah this is better

rakita (approved):
cool!

PR #2586, opened 2025-06-06
refactor: use index map to accelerate the speed of obtaining value by index

dyxushuai:


dyxushuai:
Sorry for the delay due to some personal issues, Now I am free from the matters. Over the past few days, I've been trying to add identity information to the `Address` type to take advantage of the `O(1)` speed of `IndexMap`'s index query.
But the `IndexMap` of non-index queries is 20% slower than `HashMap` because of the `Vec::growth`, and we need to make drastic changes to the architecture, so it may not be worth it. What do you think?  @rakita 

Work progess:
- Add `AddressTr` for patching identity for `Address`
```rust
/// Trait for converting between `Address` and `NamedAddress`.
pub trait AddressTr: Copy {
    /// Converts `self` into an `Address`.
    fn into_address(self) -> Address;

    /// Converts `self` into a `NamedAddress`.
    fn into_named_address(self) -> NamedAddress;

    /// Converts `self` into a `NamedAddress::Caller`.
    fn to_caller(self) -> NamedAddress {
        NamedAddress::Caller(self.into_address())
    }

    /// Converts `self` into a `NamedAddress::Target`.
    fn to_target(self) -> NamedAddress {
        NamedAddress::Target(self.into_address())
    }

    /// Converts `self` into a `NamedAddress::Beneficiary`.
    fn to_beneficiary(self) -> NamedAddress {
        NamedAddress::Beneficiary(self.into_address())
    }
}
```
- Use `IndexMap` for index queries when the identity of the `Address` is known.
- Warms some accounts access, be like the: beneficiary
- ...

Do you think I am on the right way? 😢 

I feel it might be more reasonable to pass loaded accounts in the context(need to pay attention to handling the update operations). 🤔

rakita:
We could potentially do a custom allocation that is done in pages (Something like vec of vecs) as we always just add account to the inner state.

rakita:
Have taken a different approach for this, by making Journal `AddressOrId` aware, caller and target can be saved inside Interpreter, and we can load both by index or by address.

Have added additional structs as `AddressAndId` or `AddressAndOptionalId` to facilitate all variants.

Paged vec can be worked on later so we dont have penalty of reallocationg on vec expansion.

PR is still wip but the branch is here: https://github.com/bluealloy/revm/compare/rakita/account_id?expand=1

Will close this PR.
@dyxushuai thank you for bringing this to the project, will see to take it over the line.

Issue #2585, opened 2025-06-05
Add config setting for disable EIP1559 max fee< prio fee rule


mattsse:
Apparently this is allowed in arbitrum

https://github.com/bluealloy/revm/blob/215f4e08e844995b60567a72325fbabab84bceff/crates/handler/src/validation.rs#L31-L34

causing issues with replaying arbitrum transactions:

https://github.com/foundry-rs/foundry/issues/10689

example tx: https://arbiscan.io/tx/0xb5e8b1cd23b8c07dd422b90002cb237afd4140e5d931a419ffbbb966bb31832e

since this is currently hard enforced.

it would be nice to have a setting like:

https://github.com/bluealloy/revm/blob/215f4e08e844995b60567a72325fbabab84bceff/crates/context/src/cfg.rs#L64-L65

that we can use to disable this behaviour in foundry

## TODO
* introduce a new bool setting and function like `disable_priority_fee_check:bool`

https://github.com/bluealloy/revm/blob/215f4e08e844995b60567a72325fbabab84bceff/crates/context/src/cfg.rs#L271-L271
the overhead here is negligible because a single additional bool.
* skip fee check if enabled
https://github.com/bluealloy/revm/blob/215f4e08e844995b60567a72325fbabab84bceff/crates/handler/src/validation.rs#L161-L165

frankudoags:
May I work on this?

PR #2588, opened 2025-06-06
feat: add optional priority fee check configuration

frankudoags:
closes https://github.com/bluealloy/revm/issues/2585

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -52,6 +52,7 @@ pub trait Cfg {
     /// Returns whether the base fee check is disabled.
     fn is_base_fee_check_disabled(&self) -> bool;
 
+    /// Returns whether the priority fee check is disabled.
     fn is_priority_fee_check_disabled(&self) -> bool;
 }
 
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -51,6 +51,8 @@ pub trait Cfg {
 
     /// Returns whether the base fee check is disabled.
     fn is_base_fee_check_disabled(&self) -> bool;
+
+    fn is_priority_fee_check_disabled(&self) -> bool;
 }
 
 /// What bytecode analysis to perform
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -123,6 +123,8 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
         });
     }
 
+    let disable_max_priority_check = context.cfg().is_priority_fee_check_disabled();
+
     match tx_type {
         TransactionType::Legacy => {
             // Gas price must be at least the basefee.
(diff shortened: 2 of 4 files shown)


crates/handler/src/validation.rs
```diff
@@ -147,22 +147,34 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
...
             }
 
-            validate_priority_fee_tx(
-                tx.max_fee_per_gas(),
-                tx.max_priority_fee_per_gas().unwrap_or_default(),
-                base_fee,
-            )?;
+            if Some(context.cfg().chain_id()) != tx.chain_id() {
+                return Err(InvalidTransaction::InvalidChainId);
+            }
```
mattsse: this looks new, this shouldn't be here?

crates/handler/src/validation.rs
```diff
@@ -177,11 +189,17 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
...
```
mattsse: same

crates/handler/src/validation.rs
```diff
@@ -147,22 +147,34 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
...
```
mattsse: same

crates/context/interface/src/cfg.rs
```diff
@@ -51,6 +51,8 @@ pub trait Cfg {
 
     /// Returns whether the base fee check is disabled.
     fn is_base_fee_check_disabled(&self) -> bool;
+
```
mattsse, line 54: let's also add a one liner doc here

mattsse (changes requested):
smol nits

frankudoags:
fixed all, didn't sync fork, all good now

crates/handler/src/validation.rs
```diff
@@ -147,22 +147,26 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
...
                 return Err(InvalidTransaction::Eip4844NotSupported);
             }
 
-            validate_priority_fee_tx(
-                tx.max_fee_per_gas(),
-                tx.max_priority_fee_per_gas().unwrap_or_default(),
-                base_fee,
-            )?;
+            if !context.cfg().is_priority_fee_check_disabled() {
+                validate_priority_fee_tx(
```
rakita: There are two checks here, one for 

`max_priority_fee` > `max_fee` that we want to skip, and least from this tx: https://arbiscan.io/tx/0xb5e8b1cd23b8c07dd422b90002cb237afd4140e5d931a419ffbbb966bb31832e

And the second one is to check for enough fee to cover basefee, this one needs to stay.

You can expand `validate_priority_fee_tx` and add an additional field that would skip first check.

frankudoags: thanks, good catch. fixed

frankudoags:
kindly take a look now @rakita

crates/handler/src/validation.rs
```diff
@@ -120,6 +121,8 @@ pub fn validate_tx_env<CTX: ContextTr, Error>(
         });
     }
 
+    let disable_max_priority_check = context.cfg().is_priority_fee_check_disabled();
```
rakita: nit: `disable_priority_fee_check` just to unify names

crates/handler/src/validation.rs
```diff
@@ -27,8 +27,9 @@ pub fn validate_priority_fee_tx(
     max_fee: u128,
     max_priority_fee: u128,
     base_fee: Option<u128>,
+    ignore_priority_fee: bool,
```
rakita: nit: unify namings `disable_priority_fee_check`

PR #2595, opened 2025-06-07
feat: added instruction_result fn in LoopControl trait 

### LINKED ISSUE #2592: Make it easier to access InstructionResult from LoopControl trait
Add a `instruction_result` function to loop control trait that would return `Option<InstructionResult>`. InstructionResult can be found inside `InterpreterAction` (if action is set)

Soubhik-10:
closes #2592

### DIFF
--- a/crates/interpreter/src/interpreter/ext_bytecode.rs
+++ b/crates/interpreter/src/interpreter/ext_bytecode.rs
@@ -93,7 +93,9 @@ impl LoopControl for ExtBytecode {
 
     #[inline]
     fn instruction_result(&self) -> Option<InstructionResult> {
-        self.action.as_ref().unwrap().instruction_result()
+        self.action
+            .as_ref()
+            .and_then(|action| action.instruction_result())
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -90,6 +90,11 @@ impl LoopControl for ExtBytecode {
     fn action(&mut self) -> &mut Option<InterpreterAction> {
         &mut self.action
     }
+
+    #[inline]
+    fn instruction_result(&self) -> Option<InstructionResult> {
+        self.action.as_ref().unwrap().instruction_result()
```
mattsse: this should use `.and_then` instead of the unwrap

mattsse (changes requested):

mattsse (approved):

crates/interpreter/src/interpreter/ext_bytecode.rs
```diff
@@ -90,6 +90,13 @@ impl LoopControl for ExtBytecode {
     fn action(&mut self) -> &mut Option<InterpreterAction> {
         &mut self.action
     }
+
+    #[inline]
+    fn instruction_result(&self) -> Option<InstructionResult> {
```
rakita: This can be part of the default implementation of the trait function as we already have a `fn action`

PR #2605, opened 2025-06-09
feat(devnet2): eip7907 Meter Contract Code Size And Increase Limit

rakita:
Same commit as `fusaka/devnet2` branch

### DIFF
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -65,22 +65,28 @@ pub fn extcodesize<WIRE: InterpreterTypes, H: Host + ?Sized>(
 ) {
     popn_top!([], top, context.interpreter);
     let address = top.into_address();
-    let Some(code) = context.host.load_account_code(address) else {
+    let Some(code_size) = context.host.load_account_code_size(address) else {
         context
             .interpreter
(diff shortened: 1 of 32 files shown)


crates/interpreter/src/instructions/host.rs
```diff
@@ -65,22 +65,26 @@ pub fn extcodesize<WIRE: InterpreterTypes, H: Host + ?Sized>(
...
         context
             .interpreter
             .halt(InstructionResult::FatalExternalError);
         return;
     };
     let spec_id = context.interpreter.runtime_flag.spec_id();
-    if spec_id.is_enabled_in(BERLIN) {
-        gas!(context.interpreter, warm_cold_cost(code.is_cold));
+    let gas = if spec_id.is_enabled_in(OSAKA) {
+        5000
```
rakita: Will revert this for now, would assume this will be added to devnet3. 5k was added for code size index, but it is a bug as warm/cold was not accounted.

rakita: For devnet2 behaviour for extcodesize stays the same and it will not make code warm and not additional cost is added.

rakita:
Removed from fork

PR #2609, opened 2025-06-10
perf: selfbalance cache

DaniPopes:
Prob not worth but opening because I had implemented it anyway.


DaniPopes:
This is also just wrong because it doesnt get invalidated

PR #2598, opened 2025-06-09
feat: add clz opcode

Rjected:
Fixes https://github.com/bluealloy/revm/issues/2600

PR #2610, opened 2025-06-10
feat: add clz

Rjected:
This is a copy of https://github.com/bluealloy/revm/pull/2598 but rebased on top of fusaka/devnet2


rakita (approved):
lgtm

PR #2611, opened 2025-06-10
feat(devnet2): double contract size limit

rakita:
Doubling contract size limit


crates/bytecode/src/eof/verification.rs
```diff
@@ -24,7 +21,7 @@ pub fn validate_raw_eof_inner(
     raw: Bytes,
     first_code_type: Option<CodeType>,
 ) -> Result<Eof, EofError> {
-    if raw.len() > MAX_INITCODE_SIZE {
+    if raw.len() > eip7907::MAX_INITCODE_SIZE {
```
rakita, line 24: eof related, not important

PR #2615, opened 2025-06-10
perf: pool Stack buffer allocations

DaniPopes:
Complete implementation, WIP code structure: pls advise @rakita

Ideally it would be `Pool<Stack>` but importing Stack would be a dependency cycle, so we have to transmute the reference from Vec to Stack

Even more ideally Pool doesn-t exist and the Frame stack is re-used by clearing the frame and passing a &mut to initialize it, but this is very invasive...

edit: done here https://github.com/bluealloy/revm/pull/2617


DaniPopes:
Subsumed by https://github.com/bluealloy/revm/pull/2617

PR #2616, opened 2025-06-10
chore: rename `transact` methods

klkvr:
transact -> transact_one
transact_finalize -> transact
transact_multi -> transact_many


mattsse (approved):
this is better and would unify naming with what we already have in reth

PR #2617, opened 2025-06-10
perf: re-use frame allocation

DaniPopes:
Generalizes https://github.com/bluealloy/revm/pull/2615 to frames.


crates/handler/src/frame.rs
```diff
@@ -268,33 +327,33 @@ where
...
-                is_static,
-                false,
-                context.cfg().spec().into(),
-                gas_limit,
-            ),
+            memory,
+            ExtBytecode::new_with_hash(bytecode, code_hash),
+            interpreter_input,
+            is_static,
+            is_ext_delegate_call,
```
DaniPopes:
```suggestion
            false,
```

DaniPopes:
This will obviously use more RAM across transactions; this breaks 32bit state tests as they now run OOM, what should we do @rakita ? we can special case 32bit to clear the map across blocks, or just ignore the test?

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -16,7 +16,7 @@ use std::{rc::Rc, vec::Vec};
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SharedMemory {
     /// The underlying buffer.
-    buffer: Rc<RefCell<Vec<u8>>>,
+    buffer: Option<Rc<RefCell<Vec<u8>>>>,
```
rakita, line 19: We can go without option, and just initialise with default empty vec.

rakita:
> This will obviously use more RAM across transactions; this breaks 32bit state tests as they now run OOM, what should we do @rakita ? we can special case 32bit to clear the map across blocks, or just ignore the test?

We need to check this assumption

crates/handler/src/handler.rs
```diff
@@ -527,3 +531,9 @@ pub trait Handler {
         Err(error)
     }
 }
+
+#[inline]
+fn frame_stack<'a, F: Frame<Evm: EvmTr>>(evm: &mut F::Evm) -> &'a mut FrameStack<F> {
+    let f = evm.ctx_mut().local_mut().frame_stack();
```
rakita: Honestly, this is not that bad, we could take a `local_mut().frame_stack()` inside `run_exec_loop` and return it back after fn finishes.

Would make `run_exec_loop` have a `&mut FrameStack` field so it is easier to handle error handles of multiple `?`.

crates/interpreter/src/interpreter/shared_memory.rs
DaniPopes: Rc allocates on its own

crates/handler/src/handler.rs
```diff
@@ -414,12 +416,13 @@ pub trait Handler {
...
         frame_stack(evm).end_init(token);
         loop {
-            let (frame, new_frame) = frame_stack(evm).get();
+            let frame = frame_stack(evm).get();
             let call_or_result = self.frame_call(frame, evm)?;
 
             let result = match call_or_result {
                 ItemOrResult::Item(init) => {
-                    match self.frame_init(frame, new_frame, evm, init)? {
+                    let new_frame = frame_stack(evm).get_next();
```
DaniPopes: the problem with this was that frame and new_frame are in the same vec so we got borrow errors, this is a good cleanup

rakita:
Have investigated why memory oom occurs with https://docs.rs/dhat/latest/dhat/ and have noticed that when frame_stack drops u128 values, the previously set heap allocations will not be deallocated and would leak (including boxed frame input and `Bytes`).

It feels the fault is with the frames in general and how they are burrowed inside Handler (hard to access), have tried a few ways to handle this better. Having `frame_stack` inside `Evm` looked the best way. Hope to have a PR for it today

DaniPopes:
oh yeah that makes sense, i overlooked the drop

rakita:
superseded by https://github.com/bluealloy/revm/pull/2636

Issue #2437, opened 2025-04-22
perf: remove an indirection in `JumpTable`


DaniPopes:
`JumpTable` is currently defined as `Arc<BitVec>`: https://github.com/bluealloy/revm/blob/2e09f1e69f7fce568a9ff4aea06ea40ed1693c19/crates/bytecode/src/legacy/jump_map.rs#L8

However this setup occurs 2 indirections per lookup, once for the `Arc`, once for the `BitVec` (pointer to pointer to bits)

Investigate if we can change this to `Arc<BitSlice>` or something similar, where the `Arc` pointer itself is the pointer to the bits

This would require making the field private and exposing more methods for creating the jumpmap and accessing properties

PR #2618, opened 2025-06-11
perf: faster JumpTable bits lookup

borngraced:
Cache `Arc<BitVec<u8>>` raw pointer and use bit operations for 7-8x performance improvement in `JumpTable` access/lookups via `is_valid`. Fixes https://github.com/bluealloy/revm/issues/2437

## Benchmark

Implementation    | Time per op | Ops/sec | Speedup
------------------|-------------|---------|--------
Cached Pointer    | 13.40 ns    | 74.6M   | 7.9x faster
Arc Deref         | 106.26 ns   | 9.4M    | baseline

Pattern     | Cached    | Arc       | Speedup
------------|-----------|-----------|--------
Sequential  | 16.44 ns  | 100.77 ns | 6.1x
Random      | 12.57 ns  | 98.48 ns  | 7.8x

Both sequential and random access patterns show 6-8x speedup. All tests pass with identical results

PS: I included the benchmark test so anyone can run see the result <3

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -2,7 +2,7 @@ use super::JumpTable;
 use crate::opcode;
 use bitvec::{bitvec, order::Lsb0, vec::BitVec};
 use primitives::Bytes;
-use std::{sync::Arc, vec, vec::Vec};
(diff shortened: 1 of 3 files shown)


rakita:
Awesome!

PR #2629, opened 2025-06-12
chore: build less benchmark binaries

DaniPopes:
cargo makes a binary for every file in benches/*, tests/* etc. but we dont actually want that


DaniPopes:
just renames benchmarks so that cargo doesnt build unnecessary binaries

Issue #2626, opened 2025-06-12
Mark account as None if NotExisting on cache.insert_accout_info


rakita:
After this line, add a change from `NotExisting` to `None` account status from `DbAccount`.
https://github.com/bluealloy/revm/blob/c9a017aa6a4c378b494ff41d2d360a106014b6ab/crates/database/src/in_memory_db.rs#L136

LevBeta:
This should be solved by #2630

LevBeta:
This issue can be closed

rakita:
done, thanks @LevBeta

PR #2630, opened 2025-06-12
fix: change account state to None if NotExisting on insert_account_info

LevBeta:
This change fixes the error reported [here](https://t.me/c/1543757393/5408) and resolves issue #2626.

By setting the AccountState to None, this ensures correct behavior when the previous state before calling set_existing_account was None.

### DIFF
--- a/crates/database/src/in_memory_db.rs
+++ b/crates/database/src/in_memory_db.rs
@@ -133,7 +133,11 @@ impl<ExtDB> CacheDB<ExtDB> {
     /// Inserts account info but not override storage
     pub fn insert_account_info(&mut self, address: Address, mut info: AccountInfo) {
         self.insert_contract(&mut info);
-        self.cache.accounts.entry(address).or_default().info = info;
+        let account_entry = self.cache.accounts.entry(address).or_default();
+        account_entry.info = info;
+        if account_entry.account_state == AccountState::NotExisting {
+            account_entry.account_state = AccountState::None;
+        }
     }
 
     /// Wraps the cache in a [CacheDB], creating a nested cache.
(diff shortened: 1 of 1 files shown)


crates/database/src/in_memory_db.rs
```diff
@@ -133,7 +133,11 @@ impl<ExtDB> CacheDB<ExtDB> {
     /// Inserts account info but not override storage
     pub fn insert_account_info(&mut self, address: Address, mut info: AccountInfo) {
         self.insert_contract(&mut info);
-        self.cache.accounts.entry(address).or_default().info = info;
+        let account_entry = self.cache.accounts.entry(address).or_default();
+        account_entry.info = info;
+        if account_entry.account_state == AccountState::NotExisting {
+            account_entry.account_state = AccountState::None;
+        }
```
borngraced: nit: this fn could look cleaner if you could update [DbAccount](https://github.com/borngraced/revm/blob/main/crates/database/src/in_memory_db.rs#L373-L373) with
```rust
#[inline(always)]
pub fn update_info(&mut self, info: AccountInfo) {
      self.info = info
 }

 #[inline(always)]
 pub fn update_account_state(&mut self, state: AccountState) {
      self.account_state = state
 }
```

then we can just call them straightaway here..like
```rust
     account.update_info(info);
     account.update_account_state(AccountState::None);
```

LevBeta: Yeah, I like that approach.

Just wondering if you are also suggesting to not check the previous state of the account before setting it to None

borngraced: You should do the checking ofc..Sorry I didn't include it in my suggestion but you should for sure.

LevBeta: Perfect, was just double checking

PR #2634, opened 2025-06-13
perf: store coinbase address separately to avoid cloning warm addresses in the common case

### LINKED ISSUE #2627: perf: special case warming coinbase
with #2625 we can avoid cloning the preloaded warm address set if it's unchanged;
it currently only gets appended with coinbase, and tx accesslist
if we special case coinbase into a separate field, we avoid cloning almost always
it might be a small perf gain

borngraced:
separate coinbase address warming from preloaded addresses to avoid unnecessary cloning on every transaction(if coinbase address was warmed ofc). Fixes #2627

### DIFF
--- a/crates/context/interface/src/journaled_state.rs
+++ b/crates/context/interface/src/journaled_state.rs
@@ -72,6 +72,9 @@ pub trait JournalTr {
     /// Warms the account.
     fn warm_account(&mut self, address: Address);
 
+    /// Warms the coinbase account.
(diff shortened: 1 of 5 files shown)


borngraced:
@DaniPopes I thought this PR was merged already without checking so I updated my branch(which caused a hard reset and lost my changes which also marked the PR closed automatically) since I wanted to start working on something else..so sorry about this! 🙏🏾

DaniPopes (approved):

PR #2649, opened 2025-06-18
feat: add fallible conversion from OpHaltReason to HaltReason

Wodann:
Adds a fallible conversion from an OP halt reason to an L1 halt reason.

We need this to convert an OP halt reason back to the `InstructionResult`.

Since `OpHaltReason::FailedDeposit` does not originate from an `InstructionResult`, we cannot map this to an `InstructionResult`. Instead I opted to add a fallible conversion for the halt reason types.


rakita (approved):
lgtm

PR #2654, opened 2025-06-20
fix: implement `PartialEq` for `JumpTable` correctly

klkvr:
Implements `PartialEq` and `Hash` manually for `JumpTable` to make sure we're not ending up comparing pointers

Also made `table` field private to ensure that pointer is always valid


rakita (approved):
lgtm

PR #2661, opened 2025-06-23
feat: optional_eip3541

bernard-wagner:


rakita (approved):
At some point, I will rework the configuration per EIP, but now, this lgtm!

PR #2662, opened 2025-06-24
feat: implement `Transaction` for `Either`

klkvr:


mattsse (approved):

crates/context/interface/src/transaction/either.rs
```diff
@@ -0,0 +1,127 @@
+use super::Transaction;
+use either::Either;
+use primitives::{Address, Bytes, TxKind, B256, U256};
+
+impl<L, R> Transaction for Either<L, R>
```
rakita, line 5: What is a use case for Transaction Either?

rakita (approved):
lgtm

crates/context/interface/src/transaction/either.rs
klkvr: e.g if you're building an OP rollup with a custom transaction type you'd want to do `Either<OpTxEnv, MyTx>`

PR #2663, opened 2025-06-24
docs: fix copy-pasted inner doc comments

quentin:
Hi, the inner doc comment for a few `lib.rs` were probably copy-pasted from https://github.com/bluealloy/revm/blob/435153d1b3e708bf48100fd2bf73a07aa02e1f94/crates/op-revm/src/lib.rs#L1

I quickly inferred better descriptions, better suggestions welcomed.


rakita (approved):
Cool, thanks!

Issue #2667, opened 2025-06-26
blake2b avx2 impl


rakita:
Have looked around and found avx2 code impl that we can reuse: https://github.com/oconnor663/blake2_simd/blob/82b3e2aee4d2384aabbeb146058301ff0dbd453f/blake2b/src/avx2.rs#L146-L424

And it should replace this fn https://github.com/bluealloy/revm/blob/803261837a134d0ca6cf5d376830c781c827a030/crates/precompile/src/blake2.rs#L104

There are few modification that needs to be done. Eth precompile has N round so we need to check if round counter has expired or not more often. Inputs that I can deduce are`block` is `m`, `words` is `h`,  `count` is `t`.`last_block` looks like a mask for `f` but not sure what `last_node` is.

This algorithm aligns with what Nethermind has https://github.com/NethermindEth/nethermind/blob/cc953a1e1375d1d8133b92655211597336709412/src/Nethermind/Nethermind.Crypto/Blake2/Blake2Avx2.cs#L48-L697

rakita:
Done

PR #2670, opened 2025-06-26
feat: blake2 avx2

rakita:
Have run few simple tests and it seems that it works.
Tested with: `cross test --target i686-unknown-linux-gnu --package revm-precompile --lib -- blake2::tests::perfblake2 --exact --show-output`

Need to test performance on linux/win machine with avx2 enabled.

Code is adapted from: https://github.com/oconnor663/blake2_simd/blob/82b3e2aee4d2384aabbeb146058301ff0dbd453f/blake2b/src/avx2.rs

It allows us around 1.8x faster execution. needs to be compiled with `RUSTFLAGS='-C target-feature=+avx2' cargo ` to be enabled and in runtime check is done if avx2 is present (if it is not present defaul to previous implementation).

closing: #2667

### DIFF
--- a/crates/precompile/src/blake2.rs
+++ b/crates/precompile/src/blake2.rs
@@ -132,8 +132,7 @@ pub mod algo {
         {
             // only if it is compiled with avx2 flag and it is std, we can use avx2.
             if std::is_x86_feature_detected!("avx2") {
-                println!("avx2 is run");
-                // avx2 gives 40% performance boost over portable implementation
+                // avx2 is 1.8x more performant than portable implementation.
                 unsafe {
                     super::avx2::compress_block(
                         rounds,
@@ -147,7 +146,7 @@ pub mod algo {
                 return;
             }
         }
-        println!("avx2 is not available");
+
         // if avx2 is not available, use the fallback portable implementation
 
         // Read m values
@@ -203,7 +202,7 @@ macro_rules! _MM_SHUFFLE {
 }
 
 /// Code adapted from https://github.com/oconnor663/blake2_simd/blob/82b3e2aee4d2384aabbeb146058301ff0dbd453f/blake2b/src/avx2.rs
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[cfg(all(target_feature = "avx2", feature = "std"))]
 mod avx2 {
     #[cfg(target_arch = "x86")]
     use core::arch::x86::*;


crates/precompile/src/blake2.rs
```diff
@@ -29,20 +30,30 @@ pub fn run(input: &[u8], gas_limit: u64) -> PrecompileResult {
...
-        m[i] = u64::from_le_bytes(input[pos..pos + 8].try_into().unwrap());
-    }
-    let t = [
-        u64::from_le_bytes(input[196..196 + 8].try_into().unwrap()),
-        u64::from_le_bytes(input[204..204 + 8].try_into().unwrap()),
-    ];
+    let t;
+    // Optimized parsing using ptr::read_unaligned for potentially better performance
+
+    let m: [u8; 16 * size_of::<u64>()];
```
DaniPopes:
```suggestion
    let m: &[u8; 16 * size_of::<u64>()];
```

crates/precompile/src/blake2.rs
```diff
@@ -113,21 +162,540 @@ pub mod algo {
...
+
+// Adapted from https://github.com/rust-lang-nursery/stdsimd/pull/479.
+macro_rules! _MM_SHUFFLE {
+    ($z:expr, $y:expr, $x:expr, $w:expr) => {
+        ($z << 6) | ($y << 4) | ($x << 2) | $w
+    };
+}
+
+/// Code adapted from https://github.com/oconnor663/blake2_simd/blob/82b3e2aee4d2384aabbeb146058301ff0dbd453f/blake2b/src/avx2.rs
+#[cfg(target_feature = "avx2")]
```
DaniPopes: you can use std::is_x86_feature_detected for runtime detection instead of only compile time, see e.g https://github.com/DaniPopes/const-hex/blob/3c08e80a20e72cf6133c59a2007224d5f22d9b3e/src/arch/x86.rs#L15

rakita: Didn't know about it, but i dont see a usage of runtime check for this. If binary is compiled with avx2 we should use faster version, right?

DaniPopes: the macro resolves true at compile time if -Ctarget-cpu / -Ctarget-feature is passed, otherwise it resolves at run time

you want to change `#[cfg(target_feature = "avx2")]` to `#[target_feature(enable = "avx2")]` and do something like:
```rust
#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
{
#[cfg(feature = "std")
let enabled = std::is_x86_feature_detected!("avx2");
#[cfg(not(feature = "std")]
let enabled = cfg!(target_feature = "avx2");
if enabled { avx2::compress_block(...) }
}
```

rakita: This is cool! Including it will mean we don't need any additional input on compilation, and the switch is done internally at runtime.

rakita: Tried to build it without `-Ctarget-feature` would make avx2 compile and run, but it is slow. avx code needs to be included with `-Ctarget-feature` flag, and skipped if `is_x86_feature_detected` is false.

DaniPopes: you're missing `#[target_feature(enable = "avx2")]` on the function. if you do both cfg() and is_detected then the is_detected has no impact

PR #2673, opened 2025-06-28
feat: implement Database traits for either::Either

mattsse:
Adds Database, DatabaseCommit, and DatabaseRef trait implementations for the either::Either type.

This allows using `Either<L, R>` as a database where:
- For `Database` and `DatabaseRef`: Both L and R must implement the respective trait with the same error type
- For `DatabaseCommit`: Both L and R must implement the trait

The implementation delegates all calls to either the left or right variant based on the enum value.


rakita (approved):
lgtm

Issue #2597, opened 2025-06-09
origin opcode casts Address to U256 on every iteration


rakita:
Cache Address to U256, it speeds up the ORIGIN opcode

gitToki:
Hi @rakita, could I take this ?

DaniPopes:
Not worth it. https://github.com/bluealloy/revm/pull/2674#issuecomment-3107756558

PR #2674, opened 2025-06-29
feat(interpreter): cache address to U256 conversion for ORIGIN

gitToki:
closes [#2597](https://github.com/bluealloy/revm/issues/2597)

### DIFF
--- a/crates/interpreter/src/host.rs
+++ b/crates/interpreter/src/host.rs
@@ -40,6 +40,8 @@ pub trait Host {
     fn effective_gas_price(&self) -> U256;
     /// Transaction caller, calls `ContextTr::tx().caller()`
     fn caller(&self) -> Address;
+    /// Transaction caller as U256, with caching optimization
+    fn caller_u256(&self) -> U256;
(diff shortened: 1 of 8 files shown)


gitToki:
2 failed test, but seems like it's unrelated to the changes

crates/interpreter/src/host.rs
```diff
@@ -40,6 +40,8 @@ pub trait Host {
     fn effective_gas_price(&self) -> U256;
     /// Transaction caller, calls `ContextTr::tx().caller()`
     fn caller(&self) -> Address;
+    /// Transaction caller as U256, with caching optimization
+    fn caller_u256(&self) -> U256;
```
rakita, line 44: This is not a best abstracton, let us expand LocalContext with a `caller_u256() -> Option<U256>`, so we can check if it is set or not.

DaniPopes:
This is not worth the effort. The only conversion being done is bswap for a couple of registers (bswap/movbe instructions on x86), which has no or negligible impact.

PR #2675, opened 2025-06-29
chore(DONOTMERGE): Example of allowing downstream users to override certain precompile methods using extern C

kevaundray:


kevaundray:
With something like this, we would also remove `kzg-rs` and its implementation details would be done downstream

jsign:
This will be awesome :)

kevaundray:
Also experimented with using the CryptoProvider abstraction that is in an open PR in alloy. This rough commit shows that it could look like: https://github.com/kevaundray/revm/commit/ff038cc18c174a8f2344d4ec2be7b41762b195f4

idea being that we could just remove zkVM related code and just have that trait. The trait can be installed by doing something like:

```rust
  /// let evm = Context::mainnet()
    ///     .with_crypto_provider(Box::new(CustomCryptoProvider))?
    ///     .build_mainnet();
 ```

kevaundray:
Superceded by #2786

PR #2678, opened 2025-06-30
docs: document external state transitions for EIP-4788 and EIP-2935

### LINKED ISSUE #1382: Document external state transitions for EIPs
Some EIPs require state transitions outside of revm to function properly. Currently these are:

- [EIP-4788](https://eips.ethereum.org/EIPS/eip-4788)
- [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935)

This should probably be documented in the book, as well as somewhat documented in the rustdoc itself (i.e. by referring to the book) so it's clear.

Examples for implementations of both EIPs in Reth:

- [EIP-4788](https://github.com/paradigmxyz/reth/pull/4457)
- [EIP-2935](https://github.com/paradigmxyz/reth/pull/7818)

VolodymyrBg:
Add documentation for external state transitions required by EIP-4788 and EIP-2935.
- Introduce a new section in the book explaining how these EIPs require client-side state updates outside of the EVM, with references to the relevant EIPs and example implementations in Reth.
- Update the book summary to include the new section.
- Add a module-level rustdoc comment to system_call.rs clarifying that REVM provides the system call mechanism, but the actual state transitions must be performed by the client, not by the EVM itself.
This improves clarity for users and developers regarding the responsibilities for supporting these EIPs.

Fixes #1382

### DIFF
--- a/book/src/external_state_transitions.md
+++ b/book/src/external_state_transitions.md
@@ -1,17 +1,17 @@
 # External State Transitions (EIP-4788 & EIP-2935)
 
-Some Ethereum Improvement Proposals (EIPs) require state transitions that occur outside of the EVM execution engine, and thus outside of REVM itself. Notably, these are:
+Some Ethereum Improvement Proposals (EIPs) require state transitions that are not triggered by regular user transactions, but are instead performed by the client using special system calls (such as `transact_system_call`). These transitions are part of the EVM state changes, but are initiated by the client at specific block boundaries (pre- or post-block hooks), as required by the EIP.
 
 - [EIP-4788: Beacon block root in the EVM](https://eips.ethereum.org/EIPS/eip-4788)
 - [EIP-2935: Add `blockHash` and `blockNumber` to the EVM](https://eips.ethereum.org/EIPS/eip-2935)
 
 ## What are external state transitions?
 
-External state transitions refer to updates to the Ethereum state that are not performed by the EVM during transaction execution, but are instead performed by the client (or consensus layer) as part of block processing. These are typically required for EIPs that introduce new system contracts or require special state updates at block boundaries.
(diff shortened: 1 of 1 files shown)


rakita:
Would be good to add that there is `.transact_system_call` call that is used to execute pre and post block hooks.

VolodymyrBg:
> Would be good to add that there is `.transact_system_call` call that is used to execute pre and post block hooks.

Yes boss, done

book/src/external_state_transitions.md
```diff
@@ -0,0 +1,39 @@
+# External State Transitions (EIP-4788 & EIP-2935)
+
+Some Ethereum Improvement Proposals (EIPs) require state transitions that occur outside of the EVM execution engine, and thus outside of REVM itself. Notably, these are:
```
rakita: Would formulate this differently, it is not outside of EVM there as there is a `transact_system_call` function that does this change. But this change to state does not have a tx and it is state transition.

book/src/external_state_transitions.md
```diff
@@ -0,0 +1,39 @@
...
+
+EIP-2935 introduces a system contract that stores recent block hashes, allowing contracts to query them. The client is responsible for updating this contract at each block, by calling a system contract at a fixed address (`0x0000F90827F1C53a10cb7A02335B175320002935`) with the new block hash.
+
+- The system call is performed by the client, not by EVM transaction execution.
+- See [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) for full details.
+- Example implementation in Reth: [reth#7818](https://github.com/paradigmxyz/reth/pull/7818)
+
+## How does this affect REVM users?
+
+- REVM itself does not perform these external state transitions. It expects the state to be updated by the client or test harness before EVM execution.
```
rakita: To perform this block state transition system call should be used that Revm supports.

VolodymyrBg:
@rakita Done

PR #2679, opened 2025-06-30
[DO NOT MERGE] measure uni swap exec

rakita:


rakita:
Used as example, closing it

NikitaZ:
We added a v3 example (calling their QuoterV2):
v2 22.164µs
v3 72.626µs

we tried Taiko simulations and got  15.17575µs, so it is 4-5 times faster.

NikitaZ:
add imports

use revm::primitives::alloy_primitives::aliases::U24;
use revm::primitives::alloy_primitives::U160;

NikitaZ:
add to end of main()

    swap_v3(account, account, amount_in, true, &mut cache_db)?;

NikitaZ:
And here is the swap_v3()


pub const QUOTER_V2: Address = address!("61fFE014bA17989E743c5F6cB21bF9697530B21e");
pub const USDC: Address = address!("A0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48");
pub const WETH: Address = address!("C02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");

fn swap_v3(
    from: Address,
    target: Address,
    amount_in: U256,
    is_token0: bool,
    cache_db: &mut AlloyCacheDB,
) -> Result<()> {
    sol! {
        struct QuoteExactInputSingleParams {
            address tokenIn;
            address tokenOut;
            uint256 amountIn;
            uint24 fee;
            uint160 sqrtPriceLimitX96;
        }

        /// @notice Returns the amount out received for a given exact input but for a swap of a single pool
        /// @param params The params for the quote, encoded as `QuoteExactInputSingleParams`
        /// tokenIn The token being swapped in
        /// tokenOut The token being swapped out
        /// fee The fee of the token pool to consider for the pair
        /// amountIn The desired input amount
        /// sqrtPriceLimitX96 The price limit of the pool that cannot be exceeded by the swap
        /// @return amountOut The amount of `tokenOut` that would be received
        /// @return sqrtPriceX96After The sqrt price of the pool after the swap
        /// @return initializedTicksCrossed The number of initialized ticks that the swap crossed
        /// @return gasEstimate The estimate of the gas that the swap consumes
        function quoteExactInputSingle(QuoteExactInputSingleParams memory params)
            external
            returns (
                uint256 amountOut,
                uint160 sqrtPriceX96After,
                uint32 initializedTicksCrossed,
                uint256 gasEstimate
            );
    }

    let params : QuoteExactInputSingleParams = QuoteExactInputSingleParams{
        tokenIn: WETH,
        tokenOut: USDC,
        amountIn: amount_in,
        fee: U24::from(500),
        sqrtPriceLimitX96: U160::ZERO,
    };

    let encoded = quoteExactInputSingleCall { params }.abi_encode();

    let mut evm = Context::mainnet().with_db(cache_db).build_mainnet();

    let tx = TxEnv::builder()
        .caller(from)
        .kind(TxKind::Call(QUOTER_V2))
        .data(encoded.into())
        .value(U256::from(0))
        .nonce(2)
        .build()
        .unwrap();

    // fetch slots from rpc
    let time = std::time::Instant::now();
    let _ = evm.transact(tx.clone()).unwrap();
    println!("Fetch slots: {:?}", time.elapsed());

    let time = std::time::Instant::now();
    for _ in 0..1000 {
        let _ = evm.transact(tx.clone()).unwrap();
    }
    println!("Swap time taken: {:?}", time.elapsed()/1000);
    // run it with `cargo run --package example-uniswap-v2-usdc-swap --profile profiling --bin example-uniswap-v2-usdc-swap`

    let ref_tx = evm.transact_commit(tx).unwrap();

    match ref_tx {
        ExecutionResult::Success { .. } => {
            let output = ref_tx.output().unwrap();
            let (amount_out, sqrt_price_x96_after, initialized_ticks_crossed, gas_estimate) =
            <(U256, U160, u32, U256)>::abi_decode_validate(output)?;
            println!("Amount out: {amount_out}");
            println!("Sqrt price after: {sqrt_price_x96_after}");
            println!("Initialized ticks crossed: {initialized_ticks_crossed}");
            println!("Gas estimate: {gas_estimate}");
        }
        result => return Err(anyhow!("'swap' execution failed: {result:?}")),
    };

    Ok(())
}

NikitaZ:
@rakita what do you think?

rakita:
ooo this week but will take a look afterwards, thanks @NikitaZ

gakonst:
```
let time = std::time::Instant::now();
for _ in 0..1000 {
    let _ = evm.transact(tx.clone()).unwrap();
}
```

you're cloning in the loop, this seems like it pollutes the benchmark right?

NikitaZ:
I'm seeing something like 2.28ms per transaction, tx.clone() should not have any effect at this scale, but I tried
```
    for _ in 0..1000 {
        let _ = evm.replay().unwrap();
    }
```
And it is still 2.28 ms per transaction. Sometimes it shows 2ms per transaction, but in both cases.

NikitaZ:
@gakonst @rakita what do you think?

NikitaZ:
Also swap() (v2 version) uses exactly the same technique with tx.clone()

rakita:
@NikitaZ how did you get `2.28ms` when example above takes 70us?

NikitaZ:
@rakita this puzzled me too, I just ran with default settings (`cargo run --bin example-uniswap-v2-usdc-swap`). My aim was just to show that tx.clone() doesn't have any dramatic effect.

rakita:
@NikitaZ you are missing `--release`: `cargo run --release --bin example-uniswap-v2-usdc-swap`.

NikitaZ:
Ok, `cargo run --package example-uniswap-v2-usdc-swap --profile profiling --bin example-uniswap-v2-usdc-swap` is 10 times faster,

originally it gives
```
Swap time taken: 24.185µs -- v2 swaps
Swap time taken: 243.914µs -- v3 swaps
```
With evm.replay() I see
```
Swap time taken: 246.384µs -- v3 swaps
```

There's about 3% drift from run to run.

It is still strange - v2 time is close to what I reported above on July 3rd, but v3 time became 3x worse.

NikitaZ:
@rakita `cargo run --release --package example-uniswap-v2-usdc-swap --bin example-uniswap-v2-usdc-swap` gives similar numbers to the comment above.

rakita:
It is a crude benchmark without something like `Criterion` to stabilize the time, but I am getting around ~75us (similar to what you got previously). This is the branch https://github.com/bluealloy/revm/tree/rakita/uni-example
```
➜  uniswap_v2_usdc_swap git:(temp) cargo run --release --bin example-uniswap-v2-usdc-swap
    Finished `release` profile [optimized + debuginfo] target(s) in 0.11s
     Running `/Users/draganrakita/workspace/revm/target/release/example-uniswap-v2-usdc-swap`
WETH balance before swap: 1000000000000000000
USDC balance before swap: 0
Fetch slots: 872.232584ms
Swap time taken: 27.856µs
Fetch v3 slots: 1.203663292s
Swap v3 time taken: 74.553µs
Amount out: 427160529
Sqrt price after: 1211927286965354452201799490358665
Initialized ticks crossed: 0
Gas estimate: 84819
OK
➜  uniswap_v2_usdc_swap git:(temp) cargo run --release --bin example-uniswap-v2-usdc-swap
    Finished `release` profile [optimized + debuginfo] target(s) in 0.11s
     Running `/Users/draganrakita/workspace/revm/target/release/example-uniswap-v2-usdc-swap`
WETH balance before swap: 1000000000000000000
USDC balance before swap: 0
Fetch slots: 931.779ms
Swap time taken: 28.284µs
Fetch v3 slots: 1.198997416s
Swap v3 time taken: 74.691µs
Amount out: 427373943
Sqrt price after: 1211624655487018247138973735460057
Initialized ticks crossed: 0
Gas estimate: 84821
OK
➜  uniswap_v2_usdc_swap git:(temp) cargo run --release --bin example-uniswap-v2-usdc-swap
    Finished `release` profile [optimized + debuginfo] target(s) in 0.11s
     Running `/Users/draganrakita/workspace/revm/target/release/example-uniswap-v2-usdc-swap`
WETH balance before swap: 1000000000000000000
USDC balance before swap: 0
Fetch slots: 928.875417ms
Swap time taken: 28.243µs
Fetch v3 slots: 1.196536791s
Swap v3 time taken: 70.828µs
Amount out: 427427095
Sqrt price after: 1211549318315591936532339113403328
Initialized ticks crossed: 0
Gas estimate: 84819
OK
```

NikitaZ:
@rakita I'm getting same numbers as you on your branch, moreover, looking at the diff I found where the 3x difference comes from, my latest code had:
```
    let amount_in = one_ether.mul(U256::from(200));
    swap_v3(account, account, amount_in, true, &mut cache_db)?;
```
So swapping 200ETH takes 3x longer than 0.1 ETH. 

This leads to a question - what amount size should be used for v3 swaps benchmark, for v2 amount doesn't affect the execution time as the formulas are straight, for v3 the bigger the amount the more ticks are crossed and the more iterations of the loop are taken. I think, something around 10 to 100 ETH is a good choice.

rakita:
@NikitaZ amount should not matter at all. But did you increase the amount that is stored for this account, if the balance stayed the same, then logic could have changed. 

https://github.com/bluealloy/revm/blob/70619e4a3c119716d74bd92e0bd1d62817570e6c/examples/uniswap_v2_usdc_swap/src/main.rs#L48-L53

NikitaZ:
@rakita for v3 or v4 the amount matters because it affects the number of iterations of their main loop and the amount of state read from storage. I used QUOTER_V2 not the pool contract and so the balance at the start doesn't matter, quoter exits before the last step of settling amounts. (Quoter reverts the swap transaction at settle amount step and then catches the revert data to get the output amount and returns it to the user). The other option is to use Universal Router, to call a v3 or a v4 pool one needs a contract which implements v3/v4 callback. So QUOTER_V2 does all the calculations but avoids token transfers and storage changes.

rakita:
> @rakita for v3 or v4 the amount matters because it affects the number of iterations of their main loop and the amount of state read from storage

I see, I am not familiar with internals of the quoter contract.
I am getting around `Swap v3 time taken: 183.81µs` and this takes close to 500_000gas

PR #2697, opened 2025-07-07
fix(Inspector): call_end not called on fast return

rakita:
fixing call_end not being called bug. Adding docs for inspector and tests that cover this bug.

### DIFF
--- a/crates/inspector/src/traits.rs
+++ b/crates/inspector/src/traits.rs
@@ -62,8 +62,12 @@ pub trait InspectorEvmTr:
             frame_end(ctx, inspector, &frame_init.frame_input, &mut output);
             return Ok(ItemOrResult::Result(output));
         }
-        if let ItemOrResult::Result(frame) = self.frame_init(frame_init)? {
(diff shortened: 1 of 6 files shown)


crates/inspector/src/traits.rs
```diff
@@ -62,8 +62,12 @@ pub trait InspectorEvmTr:
             frame_end(ctx, inspector, &frame_init.frame_input, &mut output);
             return Ok(ItemOrResult::Result(output));
         }
-        if let ItemOrResult::Result(frame) = self.frame_init(frame_init)? {
-            return Ok(ItemOrResult::Result(frame));
+
+        let frame_input = frame_init.frame_input.clone();
+        if let ItemOrResult::Result(mut output) = self.frame_init(frame_init)? {
+            let (ctx, inspector) = self.ctx_inspector();
+            frame_end(ctx, inspector, &frame_input, &mut output);
```
rakita, line 69: bug fix is here, after `frame_init` return a `Result` ( for example precompile call returns a result right away), we will not call call_end previously.

PR #2699, opened 2025-07-08
fix: gas deduction with `disable_balance_check`

frangio:
The merge of `validate_tx_against_state` and `deduct_caller` done in https://github.com/bluealloy/revm/pull/2460 seems to have broken gas deduction when `disable_balance_check` is enabled. This PR adds a test for this error and proposes a possible fix.

The result is that when balance checks are disabled the effective cost is first deducted with saturation from the caller balance, and then increased if necessary to cover the transaction value.

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -163,7 +163,10 @@ pub fn validate_against_state_and_deduct_caller<
     // subtracting max balance spending with value that is going to be deducted later in the call.
     let gas_balance_spending = effective_balance_spending - tx.value();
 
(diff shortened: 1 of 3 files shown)


rakita (approved):
lgtm, good catch

frangio:
I've applied the same fix to op-revm. I tried to implement a similar test but I wasn't able to get it to show the failure, I'm not sure why.

rakita:
> I've applied the same fix to op-revm. I tried to implement a similar test but I wasn't able to get it to show the failure, I'm not sure why.

I will need to check this, but at first look PR lgtm.

rakita:
All good, have added test for op-revm in local but couldn't push it to repo branch. Will do it in separate branch

PR #2701, opened 2025-07-10
feat(statetest-types): add no_std support for bare metal compatibility

zouguangxian:
This PR adds no_std support to the `revm-statetest-types` crate to enable usage in bare metal and embedded environments. It allows the crate to be used in environments without standard library support.


rakita:
It seems CI is failing,

PR #2703, opened 2025-07-11
refactor: simplify gas calculations by introducing a used() method

jackwener:


crates/interpreter/src/gas.rs
```diff
@@ -73,6 +73,12 @@ impl Gas {
         self.limit - self.remaining
     }
 
+    /// Returns the total amount of gas used.
+    #[inline]
+    pub const fn used(&self) -> u64 {
+        self.spent() - self.refunded() as u64
```
rakita:
```suggestion
        self.spent().saturating_sub(self.refunded().try_into().unwrap_or_default())
```

If used somewhere else before `Gas::set_final_refund` is called, this can underflow.

and casting refund can be a panic as refund can be negative between calls.

crates/interpreter/src/gas.rs
```diff
@@ -73,6 +73,12 @@ impl Gas {
         self.limit - self.remaining
     }
 
+    /// Returns the total amount of gas used.
```
rakita:
```suggestion
    /// Returns the final amount of gas used by subtracting the refund from spent gas. 
```

rakita (commented):
Cool change. added few nits on safer usage of `fn used`

jackwener:
Thanks your review!! @rakita

crates/interpreter/src/gas.rs
```diff
@@ -73,6 +73,12 @@ impl Gas {
         self.limit - self.remaining
     }
 
+    /// Returns the final amount of gas used by subtracting the refund from spent gas.
+    #[inline]
+    pub const fn used(&self) -> u64 {
+        self.spent().saturating_sub(self.refunded() as u64)
```
jackwener, line 79: self.refunded() is `i64`, so `as u64` is safe

rakita: You are right

PR #2705, opened 2025-07-14
feat: Use bytes API for underlying precompile library APIs

kevaundray:
## Motivation

Using revm inside of a zkVM involves patching certain cryptographic libraries using patch-crates.io -- this is quite brittle because one needs to ensure that the versions match up and long term we would like these to be patched using extern C.

One could implement extern C methods under a feature flag, however the more general strategy would be to have a pluggable architecture like https://github.com/alloy-rs/alloy/pull/2634 -- This PR is a precursor to adding the pluggable architecture since the functions that are being plugged in shouldn't rely on the types of a particular library.

### DIFF
--- a/crates/precompile/src/bls12_381/arkworks.rs
+++ b/crates/precompile/src/bls12_381/arkworks.rs
@@ -368,3 +368,68 @@ pub(super) fn pairing_check(pairs: &[(G1Affine, G2Affine)]) -> bool {
     let pairing_result = Bls12_381::multi_pairing(&g1_points, &g2_points);
     pairing_result.0.is_one()
 }
+
+// Byte-oriented versions of the functions for external API compatibility
+
+/// Performs point addition on two G1 points taking byte coordinates and returning encoded result.
+#[inline]
+pub(super) fn p1_add_affine_bytes(
+    a_x: &[u8; FP_LENGTH],
+    a_y: &[u8; FP_LENGTH],
+    b_x: &[u8; FP_LENGTH],
+    b_y: &[u8; FP_LENGTH],
+) -> Result<[u8; PADDED_G1_LENGTH], PrecompileError> {
+    // Parse first point
+    let p1 = read_g1_no_subgroup_check(a_x, a_y)?;
+
+    // Parse second point
+    let p2 = read_g1_no_subgroup_check(b_x, b_y)?;
+
+    // Perform addition
+    let result = p1_add_affine(&p1, &p2);
+
+    // Encode result
+    Ok(encode_g1_point(&result))
+}
+
+/// Performs point addition on two G2 points taking byte coordinates and returning encoded result.
+#[inline]
+#[allow(clippy::too_many_arguments)]
+pub(super) fn p2_add_affine_bytes(
+    a_x_0: &[u8; FP_LENGTH],
+    a_x_1: &[u8; FP_LENGTH],
+    a_y_0: &[u8; FP_LENGTH],
+    a_y_1: &[u8; FP_LENGTH],
+    b_x_0: &[u8; FP_LENGTH],
+    b_x_1: &[u8; FP_LENGTH],
+    b_y_0: &[u8; FP_LENGTH],
+    b_y_1: &[u8; FP_LENGTH],
+) -> Result<[u8; PADDED_G2_LENGTH], PrecompileError> {
+    // Parse first point
+    let p1 = read_g2_no_subgroup_check(a_x_0, a_x_1, a_y_0, a_y_1)?;
+
+    // Parse second point
+    let p2 = read_g2_no_subgroup_check(b_x_0, b_x_1, b_y_0, b_y_1)?;
+
+    // Perform addition
+    let result = p2_add_affine(&p1, &p2);
+
+    // Encode result
+    Ok(encode_g2_point(&result))
+}
+
+/// Maps a field element to a G1 point from bytes
+#[inline]
+pub(super) fn map_fp_to_g1_bytes(fp_bytes: &[u8; FP_LENGTH]) -> Result<[u8; PADDED_G1_LENGTH], PrecompileError> {
+    let fp = read_fp(fp_bytes)?;
+    let result = map_fp_to_g1(&fp);
+    Ok(encode_g1_point(&result))
+}
+
+/// Maps field elements to a G2 point from bytes
+#[inline]
+pub(super) fn map_fp2_to_g2_bytes(fp2_x: &[u8; FP_LENGTH], fp2_y: &[u8; FP_LENGTH]) -> Result<[u8; PADDED_G2_LENGTH], PrecompileError> {
+    let fp2 = read_fp2(fp2_x, fp2_y)?;
+    let result = map_fp2_to_g2(&fp2);
+    Ok(encode_g2_point(&result))
+}
--- a/crates/precompile/src/bls12_381/arkworks.rs
+++ b/crates/precompile/src/bls12_381/arkworks.rs
@@ -4,6 +4,21 @@ use crate::{
     },
     PrecompileError,
 };
+
+// Type aliases for MSM byte-oriented API
+pub(super) type G1PointScalarPairRef<'a> = (
+    (&'a [u8; FP_LENGTH], &'a [u8; FP_LENGTH]),
+    &'a [u8; SCALAR_LENGTH],
+);
+pub(super) type G2PointScalarPairRef<'a> = (
+    (
+        &'a [u8; FP_LENGTH],
+        &'a [u8; FP_LENGTH], 
+        &'a [u8; FP_LENGTH],
+        &'a [u8; FP_LENGTH],
+    ),
+    &'a [u8; SCALAR_LENGTH],
+);
 use ark_bls12_381::{Bls12_381, Fq, Fq2, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ec::{
     hashing::{curve_maps::wb::WBMap, map_to_curve_hasher::MapToCurve},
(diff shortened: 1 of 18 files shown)


crates/precompile/src/bls12_381/arkworks.rs
```diff
@@ -368,3 +369,196 @@ pub(super) fn pairing_check(pairs: &[(G1Affine, G2Affine)]) -> bool {
...
+
+    // Encode result
+    Ok(encode_g1_point(&result))
+}
+
+/// Performs multi-scalar multiplication (MSM) for G2 points taking byte inputs and returning encoded result.
+#[inline]
+pub(super) fn p2_msm_bytes(
+    point_scalar_pairs: &[(&G2Point, &[u8; SCALAR_LENGTH])],
+) -> Result<[u8; PADDED_G2_LENGTH], PrecompileError> {
```
kevaundray: These should not return padded elements

kevaundray: ie encode_g2_point should happen outside of this module

kevaundray:
Pulled out of #2675

crates/precompile/src/bls12_381/arkworks.rs
kevaundray: This has now been changed

crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -35,60 +36,39 @@ pub fn g1_msm(input: &[u8], gas_limit: u64) -> PrecompileResult {
...
-        }
-
         let [a_x, a_y] = remove_g1_padding(encoded_g1_element)?;
 
-        // NB: Scalar multiplications, MSMs and pairings MUST perform a subgroup check.
-        let p0_aff = read_g1(a_x, a_y)?;
+        // Convert scalar to fixed-size array
+        let scalar_array: [u8; SCALAR_LENGTH] = encoded_scalar
+            .try_into()
+            .map_err(|_| PrecompileError::Other("Invalid scalar length".to_string()))?;
```
rakita:
```suggestion
            .unwrap();
```
We have the exact size, right?

crates/precompile/src/bls12_381/g2_msm.rs
```diff
@@ -34,53 +36,29 @@ pub fn g2_msm(input: &[u8], gas_limit: u64) -> PrecompileResult {
...
-
         let [a_x_0, a_x_1, a_y_0, a_y_1] = remove_g2_padding(encoded_g2_element)?;
 
-        // NB: Scalar multiplications, MSMs and pairings MUST perform a subgroup check.
-        //
-        // So we set the subgroup_check flag to `true`
-        let p0_aff = read_g2(a_x_0, a_x_1, a_y_0, a_y_1)?;
+        // Convert scalar to fixed-size array
+        let scalar_array: [u8; SCALAR_LENGTH] = encoded_scalar
+            .try_into()
```
rakita: `.unwrap()` we encoded_scalar is always exact length

crates/precompile/src/bls12_381/utils.rs
```diff
@@ -51,3 +51,98 @@ pub(super) fn remove_g2_padding(input: &[u8]) -> Result<[&[u8; FP_LENGTH]; 4], P
...
+    padded[FP_PAD_BY..PADDED_FP_LENGTH].copy_from_slice(&unpadded[0..FP_LENGTH]);
+    // x.c1
+    padded[PADDED_FP_LENGTH + FP_PAD_BY..2 * PADDED_FP_LENGTH]
+        .copy_from_slice(&unpadded[FP_LENGTH..2 * FP_LENGTH]);
+    // y.c0
+    padded[2 * PADDED_FP_LENGTH + FP_PAD_BY..3 * PADDED_FP_LENGTH]
+        .copy_from_slice(&unpadded[2 * FP_LENGTH..3 * FP_LENGTH]);
+    // y.c1
+    padded[3 * PADDED_FP_LENGTH + FP_PAD_BY..4 * PADDED_FP_LENGTH]
+        .copy_from_slice(&unpadded[3 * FP_LENGTH..4 * FP_LENGTH]);
```
rakita:
```suggestion
    for i in 0..4 {
        padded[i*PADDED_FP_LENGTH + FP_PAD_BY..(i+1) * PADDED_FP_LENGTH]
        .copy_from_slice(&unpadded[i*FP_LENGTH..(i+1) * FP_LENGTH]);
    }
```

slightly less code

crates/precompile/src/bls12_381/utils.rs
```diff
@@ -51,3 +51,98 @@ pub(super) fn remove_g2_padding(input: &[u8]) -> Result<[&[u8; FP_LENGTH]; 4], P
...
+        "Invalid unpadded G1 point length"
+    );
+
+    let mut padded = [0u8; PADDED_G1_LENGTH];
+
+    // x
+    padded[FP_PAD_BY..PADDED_FP_LENGTH].copy_from_slice(&unpadded[0..FP_LENGTH]);
+    // y
+    padded[PADDED_FP_LENGTH + FP_PAD_BY..2 * PADDED_FP_LENGTH]
+        .copy_from_slice(&unpadded[FP_LENGTH..G1_LENGTH]);
```
rakita:
```suggestion
    for i in 0..2 {
            padded[i*PADDED_FP_LENGTH + FP_PAD_BY..(i+1) * PADDED_FP_LENGTH]
        .copy_from_slice(&unpadded[i*FP_LENGTH..(i+1)*FP_LENGTH]);
    }
```
Slighly less code

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -55,10 +55,18 @@ pub fn run(input: &[u8], gas_limit: u64) -> PrecompileResult {
...
+        .expect("expected `commitment` to be 48 bytes");
+    let z = input[32..64]
+        .try_into()
+        .expect("expected `z` to be 32 bytes");
+    let y = input[64..96]
+        .try_into()
+        .expect("expected `y` to be 32 bytes");
+    let proof = input[144..192]
+        .try_into()
+        .expect("expected `proof` to be 48 bytes");
```
rakita: Would just `.unwrap()` it, it will probably make it in one line, and it is slightly more readable in that way

crates/precompile/src/secp256k1.rs
```diff
@@ -44,16 +44,25 @@ pub fn ec_recover_run(input: &[u8], gas_limit: u64) -> PrecompileResult {
         return Ok(PrecompileOutput::new(ECRECOVER_BASE, Bytes::new()));
     }
 
-    let msg = <&B256>::try_from(&input[0..32]).unwrap();
+    let msg: [u8; 32] = input[0..32].try_into().expect("expected 32 bytes");
```
rakita: Should we use `<&B256>::try_from` it is probably more performant? Would leave it as is rn

crates/precompile/src/bls12_381/g1_msm.rs
```diff
@@ -35,60 +36,39 @@ pub fn g1_msm(input: &[u8], gas_limit: u64) -> PrecompileResult {
         return Err(PrecompileError::OutOfGas);
     }
 
-    let mut g1_points: Vec<_> = Vec::with_capacity(k);
-    let mut scalars = Vec::with_capacity(k);
+    let mut point_scalar_pairs: Vec<(G1Point, [u8; SCALAR_LENGTH])> = Vec::with_capacity(k);
+
     for i in 0..k {
```
rakita: Would remove this for loop, the only thing that we do here is removal of the padding and adding creating new vec. And inside crypto provider you would again create an additional vec to transform the fields.


We can transmute it and send it padded and leave crypto provider to do it in the loop:

```rust
    let points = unsafe {
        mem::transmute::<&[u8], &[([u8; PADDED_G1_LENGTH], [u8; SCALAR_LENGTH])]>(&input)
    };
```

kevaundray: I had initially done this to have a create a clean abstraction since the padding is an EVM specific operation and the cryptography modules don't have any context about the EVM -- If we can't avoid the extra allocation, I can  use a transmute and remove padding in the cryptography module. Perhaps we could use an iterator?

kevaundray: Removed the initial allocation here 7d3e16982a2a4a68576822081dda4f55313419c7 -- what do you think?

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -77,38 +77,43 @@ pub fn kzg_to_versioned_hash(commitment: &[u8]) -> [u8; 32] {
 
 /// Verify KZG proof.
 #[inline]
-pub fn verify_kzg_proof(commitment: &Bytes48, z: &Bytes32, y: &Bytes32, proof: &Bytes48) -> bool {
+pub fn verify_kzg_proof(
```
kevaundray, line 80: Removes the library specific types from the public API

crates/precompile/src/bls12_381/g1_msm.rs
kevaundray: Merging this into the the dynamic CryptoProvider PR and impl Iterator removes object safety, so the padding abstraction leak might be the better approach

kevaundray: Have also rebased this ontop of https://github.com/bluealloy/revm/pull/2706 to show what it looks like with CryptoProvider

rakita: Iterators are lazy so it makes sense that allocation is skipped here, it looks great. Would need to check just to be sure, but can do that later.

rakita (approved):
lgtm

PR #2706, opened 2025-07-14
feat: Add dynamic crypto provider

kevaundray:
Based off of #2705

This PR includes all of the changes including code restructuring now, to show what the end goal would be. If this looks okay, then I can split this into smaller PRs.

Changes:

- All raw cryptography functions now have byte API
- CryptoProvider trait now defines all of the raw cryptography functions that we need
- DefaultCryptoProvider implements all of the methods on the trait and behaves similar to the previous code. The difference is that now we call `get_crypto_provider` to return `DefaultCryptoProvider` 
- cryptography functions have been moved to crypto_provider directory. It should now be possible to pull out this directory.


crates/precompile/src/crypto_provider.rs
```diff
@@ -0,0 +1,111 @@
...
+///     // ... implementation
+/// }
+///
+/// if !install_provider(MyProvider) {
+///     println!("Provider already installed");
+/// }
+/// ```
+pub fn install_provider<P: CryptoProvider>(provider: P) -> bool {
+    PROVIDER.set(Box::new(Box::new(provider))).is_ok()
+}
```
kevaundray: We can remove this global installation of the crypto backend by having the `run` method for precompiles take a CryptoProvider parameter and it can be installed like:


```rust
new_evm().with_crypto_provider(Box::new(CustomCryptoProvider))?
```

And if `with_crypto_provider` is missing then it uses the default

kevaundray: Did this way initially so it doesn't change the API

crates/precompile/src/bls12_381.rs
```diff
@@ -2,15 +2,8 @@
...
-    if #[cfg(feature = "blst")]{
-        mod blst;
-        use blst as crypto_backend;
-    } else {
-        mod arkworks;
-        use arkworks as crypto_backend;
-    }
-}
+// Re-export type aliases from crypto_provider for backward compatibility
+pub use crate::crypto_provider::bls12_381::{G1Point, G2Point, PairingPair};
```
kevaundray: remove

crates/precompile/src/crypto_provider/blake2/avx2.rs
kevaundray, line 1: This file was only moved

crates/precompile/src/crypto_provider/mod.rs
```diff
@@ -0,0 +1,567 @@
...
+    /// `true` if the proof is valid, `false` otherwise.
+    #[cfg(any(feature = "c-kzg", feature = "kzg-rs"))]
+    fn kzg_verify_proof(
+        &self,
+        commitment: &[u8; 48],
+        z: &[u8; 32],
+        y: &[u8; 32],
+        proof: &[u8; 48],
+    ) -> bool;
+
```
kevaundray, line 197: Preserving the old behaviour, essentially if c-kzg or kzg-rs are not enabled, then this is not available. It should be possible to reuse one of the bls12_381 crates to implement this if needed

crates/precompile/src/bn128.rs
```diff
@@ -1,26 +1,14 @@
 //! BN128 precompiles added in [`EIP-1962`](https://eips.ethereum.org/EIPS/eip-1962)
+
+use crate::crypto_provider::bn128::{
+    G1_LENGTH as G1_LEN, G2_LENGTH as G2_LEN, SCALAR_LENGTH as SCALAR_LEN,
+};
```
kevaundray, line 5: Doing this type alias to reduce the diff

crates/precompile/src/bls12_381.rs
```diff
@@ -2,16 +2,6 @@
 //! For more details check modules for each precompile.
 use crate::PrecompileWithAddress;
 
-cfg_if::cfg_if! {
```
kevaundray, line 5: precompile modules no longer have these conditional compilations -- most/all of them were being used for switching between different cryptography libraries

crates/precompile/src/crypto_provider/mod.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// * `pk` - The uncompressed public key (secp256r1::PUBKEY_LENGTH bytes: 0x04 || x || y)
+    ///
+    /// # Returns
+    /// `true` if the signature is valid, `false` otherwise.
+    fn secp256r1_verify(
+        &self,
+        msg: &[u8; secp256r1::MESSAGE_HASH_LENGTH],
+        sig: &[u8; secp256r1::SIGNATURE_LENGTH],
+        pk: &[u8; secp256r1::PUBKEY_LENGTH],
+    ) -> bool;
```
kevaundray, line 232: This would change with #2736

crates/precompile/src/crypto_provider/mod.rs
```diff
@@ -0,0 +1,618 @@
...
+    /// # Arguments
+    /// * `fp2` - Fp2 element as two bls12_381::FP_LENGTH-byte field elements
+    ///
+    /// # Returns
+    /// The mapped G2 point as bls12_381::G2_LENGTH bytes (unpadded).
+    fn bls12_381_map_fp2_to_g2(
+        &self,
+        fp2_x: &[u8; bls12_381::FP_LENGTH],
+        fp2_y: &[u8; bls12_381::FP_LENGTH],
+    ) -> Result<[u8; bls12_381::G2_LENGTH], PrecompileError>;
```
kevaundray, line 175: Technically speaking, putting PrecompileError here and in the APIs for the cryptography code is leaking an abstraction, but I think it can be cleaned up after

kevaundray:
Superceded by #2786

PR #2709, opened 2025-07-14
feat(devnet3): eip7907 Meter Contract Code Size And Increase Limit

rakita:
devnet-3 changes for EIP-7907


rakita:
Removed from fork

PR #2711, opened 2025-07-15
feat: add a way for precompiles to revert

klkvr:
Adds `reverted` flag to `PrecompileOutput` that allows precompiles to revert

### DIFF
--- a/crates/handler/src/precompile_provider.rs
+++ b/crates/handler/src/precompile_provider.rs
@@ -124,7 +124,11 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
             Ok(output) => {
                 let underflow = result.gas.record_cost(output.gas_used);
                 assert!(underflow, "Gas underflow is not possible");
-                result.result = InstructionResult::Return;
+                result.result = if output.reverted {
+                    InstructionResult::Revert
+                } else {
+                    InstructionResult::Return
+                };
                 result.output = output.bytes;
             }
             Err(PrecompileError::Fatal(e)) => return Err(e),
--- a/crates/precompile/src/interface.rs
+++ b/crates/precompile/src/interface.rs
@@ -16,12 +16,33 @@ pub struct PrecompileOutput {
     pub gas_used: u64,
     /// Output bytes
     pub bytes: Bytes,
+    /// Whether the precompile reverted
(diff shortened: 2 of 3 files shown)


crates/precompile/src/interface.rs
```diff
@@ -16,12 +16,33 @@ pub struct PrecompileOutput {
     pub gas_used: u64,
     /// Output bytes
     pub bytes: Bytes,
+    /// Whether the precompile reverted
+    pub reverted: bool,
```
mattsse, line 20: this seems like the least invasive option to me,

the use case here is valid, since there exist precompiles that act as revert

mattsse (approved):

rakita (approved):
lgtm

crates/handler/src/precompile_provider.rs
```diff
@@ -124,7 +124,11 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
             Ok(output) => {
                 let underflow = result.gas.record_cost(output.gas_used);
                 assert!(underflow, "Gas underflow is not possible");
-                result.result = InstructionResult::Return;
+                result.result = if output.reverted {
+                    InstructionResult::Revert
+                } else {
+                    InstructionResult::Return
+                };
```
will-2012, line 131: https://github.com/alloy-rs/evm/blob/666cb3bf0a6511e210e93b1d2e88d80fcf85708c/crates/evm/src/precompiles.rs#L370-L392    
   ```
        let precompile_result = precompile.call(PrecompileInput {
            data: input_bytes,
            gas: gas_limit,
            caller: inputs.caller_address,
            value: inputs.call_value,
            internals: EvmInternals::new(journal, &context.block),
        });


        match precompile_result {
            Ok(output) => {
                let underflow = result.gas.record_cost(output.gas_used);
                assert!(underflow, "Gas underflow is not possible");
                result.result = InstructionResult::Return;
                result.output = output.bytes;
            }
            Err(PrecompileError::Fatal(e)) => return Err(e),
            Err(e) => {
                result.result = if e.is_oog() {
                    InstructionResult::PrecompileOOG
                } else {
                    InstructionResult::PrecompileError
                };
            }
```

Does alloy-rs/evm need to handle the revert error?

PR #2712, opened 2025-07-15
perf: add asm-sha2 feature for sha2 precompile

Rjected:
Turns out `sha2` has an `asm` feature, added a feature for enabling this


rakita (approved):
Nice!

PR #2715, opened 2025-07-15
chore: remove State bound from JournalTr in Handler::Evm

sam-iamm:
Removes the unused State bound from JournalTr in Handler::Evm. This removal allows journals to use the Handler trait even if they don't return EvmState from JournalTr::finalize. A concrete example is a JournalTr implementor that maintains some custom state in addition to EvmState.


rakita (approved):
nice change

PR #2731, opened 2025-07-18
feat: simplify inspector interpreter loop

rakita:
Removes additional check for loop end in inspector interpreter loop. Comparing it with older this looked even worse 

Will merge inspector on main first https://github.com/bluealloy/revm/pull/2730 so it can be compared more easily.


rakita:
Superseeded by https://github.com/bluealloy/revm/pull/2797

PR #2734, opened 2025-07-18
chore(experiment): refactor blake2 input parsing

kevaundray:


kevaundray:
running perfblake locally implied that this was faster, however when comparing to master using codspeed, its slightly slower: https://github.com/bluealloy/revm/pull/2735#issuecomment-3090980258

kevaundray:
Adding 100K and 200K rounds to have larger inputs to confirm

crates/precompile/src/blake2.rs
```diff
@@ -238,6 +225,7 @@ mod avx2 {
         let flags = set4(count_low(count), count_high(count), last_block, last_node);
         let mut d = xor(loadu(iv_high), flags);
 
+        let block: &[u8; BLOCKBYTES] = std::mem::transmute(block);
```
rakita, line 228: This is probably fine as endiness is the same for all targets that have avx

PR #2735, opened 2025-07-18
chore: Add blake2 benchmarks

kevaundray:


crates/precompile/bench/blake2.rs
```diff
@@ -0,0 +1,116 @@
...
+        ];
+        let m = [0u8; 128];
+        let t = [0u64, 0u64];
+        b.iter(|| {
+            let mut h_copy = h;
+            blake2::algo::compress(
+                black_box(12),
+                &mut h_copy,
+                black_box(&m),
+                black_box(t),
```
kevaundray, line 111: can remove -- so we are only benchmarking the precompile in the way that it will be called

rakita: What was a reason for measuring 12 rounds. Is it to compare with other blake implementations?

kevaundray: It was mainly because blake2b uses 12 rounds

PR #2736, opened 2025-07-19
chore: use `EncodedPoint` to decode uncompressed public key

kevaundray:
Use `EncodedPoint` to decode the uncompressed public key. EncodedPoint allows us to bypass sec1 encoding and use the 64 bytes public key.


crates/precompile/src/secp256r1.rs
```diff
@@ -69,8 +72,10 @@ pub fn verify_impl(input: &[u8]) -> Option<()> {
 
     // Can fail only if the input is not exact length.
     let signature = Signature::from_slice(sig).ok()?;
-    // Can fail if the input is not valid, so we have to propagate the error.
-    let public_key = VerifyingKey::from_sec1_bytes(&uncompressed_pk).ok()?;
+    // Decode the uncompressed public key using EncodedPoint
+    let encoded_point = EncodedPoint::from_bytes(&uncompressed_pk).ok()?;
+    // Create VerifyingKey from the encoded point
+    let public_key = VerifyingKey::from_encoded_point(&encoded_point).ok()?;
```
kevaundray: Currently this is a no-op -- will change to use untagged_bytes

kevaundray:
PR is failing due to codspeed (unrelated) -- did this initial commit to show its a noop. The next commit applies the relevant changes

crates/precompile/src/secp256r1.rs
kevaundray: changed in  a849d68

crates/precompile/src/secp256r1.rs
```diff
@@ -62,15 +65,12 @@ pub fn verify_impl(input: &[u8]) -> Option<()> {
     // x, y: public key
     let pk = &input[96..160];
 
-    // Prepend 0x04 to the public key: uncompressed form
-    let mut uncompressed_pk = [0u8; 65];
-    uncompressed_pk[0] = 0x04;
-    uncompressed_pk[1..].copy_from_slice(pk);
```
kevaundray, line 92: The 0x04 was because of `from_sec1_bytes` -- using `untagged_bytes` means we don't need to do this and can avoid this allocation plus the validation of 0x04 tag

rakita (approved):
cool

PR #2737, opened 2025-07-20
feat(experiment): use arkworks for kzg_point_eval_precompile

kevaundray:
This is a rough draft to see what the code would look like if we reused the bls12-381 implementations to implement kzg_point_eval_precompile instead of pulling in c-kzg


crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -86,7 +96,122 @@ pub fn verify_kzg_proof(commitment: &Bytes48, z: &Bytes32, y: &Bytes32, proof: &
...
+        let Ok(z_fr) = read_scalar_canonical(&z.0) else {
+            return false;
+        };
+        let Ok(y_fr) = read_scalar_canonical(&y.0) else {
+            return false;
+        };
+
+        // Get the trusted setup G2 point [τ]₂
+        // TODO: This only needs to be done once and cached
+        let tau_g2 = get_trusted_setup_g2();
```
kevaundray, line 153: This should be cached -- though its not affecting performance, so probably okay:

I did some rough benchmarks of this and it was taking about 60 microseconds, so its not taking up a lot of time.

The whole function takes 1ms for arkworks and about 0.8ms for c-kzg --the 20% is likely coming from the pairing being more efficient, this could be checked by using blst instead of arkworks since c-kzg using blst

kevaundray: (Noting that 1ms is what I would have expected this to take and is not slow)

crates/precompile/src/kzg_point_evaluation.rs
```diff
@@ -86,7 +96,122 @@ pub fn verify_kzg_proof(commitment: &Bytes48, z: &Bytes32, y: &Bytes32, proof: &
             let env = kzg_rs::EnvKzgSettings::default();
             let kzg_settings = env.get();
             KzgProof::verify_kzg_proof(commitment, z, y, proof, kzg_settings).unwrap_or(false)
+        } else {
+            bls12_381_backend::verify_kzg_proof(commitment, z, y, proof)
```
kevaundray: Avoids a KzgSettings file given that revm never needs to create proofs and the setup information needed to verify proofs is fixed and small

kevaundray:
This has only been ran against the basic_test in the module which means for completeness it's probably okay whp, but would need to be ran against the spec tests to check for soundness

kevaundray:
If this PR is desired, then I think waiting until the changes in #2675 are merged would be good to reduce the number of conflicts (this PR is a lot simpler)

kevaundray:
Closing due to 2809

PR #2738, opened 2025-07-20
refactor: remove generic impl for Host

### LINKED ISSUE #2732: [feat] Remove generic implementation for `Host` trait
https://github.com/bluealloy/revm/blob/a494e389df59a245d296acbe9fcbdaf75265658e/crates/interpreter/src/host.rs#L91

Currently, `revm::interpreter::Host` is automatically implemented for any struct that implements `revm::context::ContextTr`, which prevents any downstream specialization of `[Host](revm::interpreter::Host)` trait implementation of customized context. Implementing `revm::interpreter::Host` for a customized context will result in "comflict implementation" compilation error. 

Currently, I think there is only one context struct implemented in the revm crates, i.e., `Context`. Is it possible to explicitly implement `revm::interpreter::Host` only for `revm::Context`?

Troublor:
This PR resolves #2732. 

The generic implementation for `revm::interpreter::Host` is replaced with an explicit implementation for `revm::context::Context` struct type. This decoupling enables the specialization of host behavior for customized context types. 

PS: I added one dependency in the `revm-interpreter`, making it additionally depend on `revm-context`. I am not sure whether this dependency is appropriate or not, or whether there are other ways to do this.


Troublor:
@rakita Then there will be cycle dependency I think. The crate `revm-interpreter`, which defines `Host`, already depends on `context_interface`, which defines `ContextTr`.

rakita:
> @rakita Then there will be cycle dependency I think. The crate revm-interpreter, which defines Host, already depends on context_interface, which defines ContextTr.

@Troublor, we could move `Host` to the `context-interface` to have interface traits in one place

Troublor:
@rakita Updated. Please see if it is appropriate.

rakita (approved):
Looks great

PR #2741, opened 2025-07-21
feat(precompile): update p256 verify osaka gas cost

rkrasiuk:
## Description

Update the gas cost for `P256VERIFY` precompile.


rakita (approved):
lgtm

PR #2750, opened 2025-07-21
chore: move code into `crypto` folder

kevaundray:


crates/precompile/src/crypto/blake2/mod.rs
```diff
@@ -0,0 +1,567 @@
+//! Blake2 cryptographic implementations
+
+// Re-export the main compress function for external use
+pub use algo::compress;
+
+/// Blake2 algorithm
+pub mod algo {
```
kevaundray: copied from previous blake2.rs file

kevaundray:
Pulled out of #2706

crates/precompile/bench/blake2.rs
```diff
@@ -87,30 +87,4 @@ pub fn add_benches(group: &mut BenchmarkGroup<'_, criterion::measurement::WallTi
             black_box(blake2::run(black_box(input), u64::MAX).unwrap());
         });
     });
-
-    // Benchmark just the compression function with different round counts
-    group.bench_function("blake2/compress_12_rounds", |b| {
```
kevaundray, line 92: I think it might make sense to benchmark precompiles and the underlying cryptography separately -- we do still benchmark 12 rounds via the `run` method so we don't lose much here

crates/precompile/src/bls12_381.rs
```diff
@@ -2,26 +2,8 @@
 //! For more details check modules for each precompile.
 use crate::PrecompileWithAddress;
 
-cfg_if::cfg_if! {
```
kevaundray, line 5: precompile/evm specific code no longer has the cfg_if code -- these were mainly being used to switch between different cryptography libraries. This now lives inside of `crypto`  folder

crates/precompile/bench/blake2.rs
rakita: hm precompiles is a crypto with a gas calculation.

kevaundray:
We departed from the implementation of #2705 , so this was superseded by #2786 -- we have another issue for restructuring the new code which is #2782

PR #2751, opened 2025-07-22
feat(bytecode): add version getter + make version dynamic

KumaCrypto:
### Extract version from bytecode and add getter

* Replaced hardcoded `EIP7702_VERSION` with a value from raw to allow dynamic version parsing.
  * This doesn't change current behavior - only one version is still allowed, and it's validated in `new_raw`.
* Added `version()` getter.


rakita (approved):
lgtm

PR #2753, opened 2025-07-22
feat(osaka): update EIP-7825 constant

rkrasiuk:
## Description

Update EIP-7825 constant. Ref https://github.com/ethereum/EIPs/pull/9986


rakita:
For failed tests, remove files from op-revm/tests/testdata and rum test again. Thos would regenerate fixtures that were changed

PR #2755, opened 2025-07-22
fix: system call should have 30M gas limit

rakita:
System call tx should have fixed 30_000_000 gas. Currently it does not effect anything


crates/handler/src/system_call.rs
```diff
@@ -62,7 +62,7 @@ impl SystemCallTx for TxEnv {
             .caller(caller)
             .data(data)
             .kind(TxKind::Call(system_contract_address))
-            .gas_limit(eip7825::TX_GAS_LIMIT_CAP)
+            .gas_limit(30_000_000)
```
kevaundray, line 65: Mainly curious; where does this 30M come from?

PR #2758, opened 2025-07-22
chore: Add CryptoProvider trait + impl

kevaundray:
Based off of #2750 superecedes #2706 so the diff includes that:

- Adds the CryptoProvider trait
- Refactors cryptography calls to use the get_provider() method for cryptography code


kevaundray:
Superceded by #2786

PR #2760, opened 2025-07-22
perf: interpreter improvements

DaniPopes:
See individual commits


mattsse:
das feuer bruder

rakita:
Awesome finds! It looks that the biggest thing is marking `fn halt` as cold.

it is unfortunate that degradation didn't happened on PR when this fn was introduced but in EOF removal PR where code was removed, I guess with code removed the compiler switched its strategy for this fn from not inlining to inlining. 

https://codspeed.io/bluealloy/revm/benchmarks/bins/revme/benches/evm.rs::benches::evm::snailtracer?runnerMode=Instrumentation

rakita (approved):

PR #2762, opened 2025-07-23
chore: evaluate instruction table initializer at compile time

DaniPopes:


gakonst:
Nice -- this is the kind of change that I wonder why doesn't rust do by itself?

DaniPopes:
Because it does what you tell it to, generally it won't evaluate something at compile time just because it's `const fn` unless you force it through `const X: ...` or `const {}` because it technically has different semantics (initializing the array on the stack and moving elements into it, vs copying a constant)

mattsse (approved):

rakita (approved):

PR #2772, opened 2025-07-23
chore: Add dyn Crypto trait to PrecompileFn

kevaundray:


crates/handler/src/precompile_provider.rs
```diff
@@ -98,9 +98,6 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
         _is_static: bool,
         gas_limit: u64,
     ) -> Result<Option<InterpreterResult>, String> {
-        let Some(precompile) = self.precompiles.get(address) else {
-            return Ok(None);
-        };
```
kevaundray: We now check for this in `self.precompiles.execute` -- will revert this and just deref the fn pointer, then explicitly pass in the dyn trait

kevaundray: https://github.com/bluealloy/revm/pull/2772/commits/666a47ca68ebdfad889f14fb0bd0e82089ae0157

mattsse (commented):
this is quite an invasive change just to make the sha256 fn configurable

perhaps a better approach would be to have dedicated precompile fns that use different impls and then simply replace the entire precompile

kevaundray:
> this is quite an invasive change just to make the sha256 fn configurable
> 
> perhaps a better approach would be to have dedicated precompile fns that use different impls and then simply replace the entire precompile

This would be for all of the precompiles fwiw -- I just did sha256 first to show it working.

I was hoping to not need to replace the entire precompile because the main thing we care about switching out is the pure cryptography part and not the EVM related parts

kevaundray:
@mattsse we would essentially be adding all of these methods: https://github.com/bluealloy/revm/blob/3d00601df95ae699ecf8f344abb59a841d562e59/crates/precompile/src/crypto.rs#L42

rakita:
This is contained only within `Precompile`, so it does not change PrecompileProvider and how it is handled. And intention is to have an exchangeable crypto module for every precompile, so it is easy to use.

rakita (approved):
Good as a skeleton.

rakita (approved):

PR #2776, opened 2025-07-23
perf: improve inspector loop

DaniPopes:


crates/inspector/src/handler.rs
```diff
@@ -182,21 +183,39 @@ where
     CTX: ContextTr<Journal: JournalExt> + Host,
     IT: InterpreterTypes,
 {
-    let mut log_num = context.journal_mut().logs().len();
-    // Main loop
-    while interpreter.bytecode.is_not_end() {
-        // Get current opcode.
-        let opcode = interpreter.bytecode.opcode();
+    if interpreter.bytecode.is_not_end() {
```
rakita: check for `is_not_end` can be removed as bytecode will always point to correct opcode.

rakita (approved):
one nit, otherwise lgtm!

PR #2777, opened 2025-07-23
feat: Add bn128 to CryptoTrait

kevaundray:


kevaundray:
Checking the diff to implement all of them in one PR

kevaundray:
Removing the dynamic dispatch in [f567bec](https://github.com/bluealloy/revm/pull/2777/commits/f567bec08551db9494db5b0cd48a185edd779cee) removed the 3% regression -- so it seems that calling a round of blake2f is too cheap such that its performance is being affected by whether it is being statically called or via dynamic dispatch with non-inlining

kevaundray:
Superceded by #2786

PR #2780, opened 2025-07-23
chore: collapse debug info for interpreter macros

DaniPopes:
Without this attribute, debug info will point into the macro, but this is quite annoying when debugging/profiling for simple macros. e.g. `gas!` will show up in a separate stack frame in `samply`.

Reference: https://doc.rust-lang.org/nightly/reference/attributes/debugger.html#r-attributes.debugger.collapse_debuginfo


mattsse (approved):
TIL

rakita (approved):

PR #2786, opened 2025-07-24
chore: reuse global crypto provide idea

rakita:
Reusing idea from: https://github.com/bluealloy/revm/pull/2706
Build on top of: https://github.com/bluealloy/revm/pull/2777

### DIFF
--- a/crates/precompile/src/interface.rs
+++ b/crates/precompile/src/interface.rs
@@ -132,13 +132,8 @@ pub trait Crypto: Send + Sync + Debug {
 
     /// secp256r1 (P-256) signature verification.
     #[inline]
-    fn secp256r1_verify_signature(
-        &self,
-        msg: &[u8; 32],
-        sig: &[u8; 64],
-        pk: &[u8; 64],
-    ) -> Result<bool, PrecompileError> {
-        Ok(crate::secp256r1::verify_signature(*msg, *sig, *pk).is_some())
+    fn secp256r1_verify_signature(&self, msg: &[u8; 32], sig: &[u8; 64], pk: &[u8; 64]) -> bool {
+        crate::secp256r1::verify_signature(*msg, *sig, *pk).is_some()
     }
 
     /// KZG point evaluation.
--- a/crates/precompile/src/secp256r1.rs
+++ b/crates/precompile/src/secp256r1.rs
@@ -68,7 +68,7 @@ fn p256_verify_inner(input: &[u8], gas_limit: u64, gas_cost: u64) -> PrecompileR
     if gas_cost > gas_limit {
         return Err(PrecompileError::OutOfGas);
     }
-    let result = if verify_impl(input).is_some() {
+    let result = if verify_impl(input) {
         B256::with_last_byte(1).into()
     } else {
         Bytes::new()
@@ -78,9 +78,9 @@ fn p256_verify_inner(input: &[u8], gas_limit: u64, gas_cost: u64) -> PrecompileR
 
 /// Returns `Some(())` if the signature included in the input byte slice is
 /// valid, `None` otherwise.
-pub fn verify_impl(input: &[u8]) -> Option<()> {
+pub fn verify_impl(input: &[u8]) -> bool {
     if input.len() != 160 {
-        return None;
+        return false;
     }
 
     // msg signed (msg is already the hash of the original message)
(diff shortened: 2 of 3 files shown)


crates/precompile/src/interface.rs
```diff
@@ -1,8 +1,28 @@
 //! Interface for the precompiles. It contains the precompile result type,
 //! the precompile output type, and the precompile error type.
 use core::fmt::{self, Debug};
+use once_cell::race::OnceBox;
 use primitives::Bytes;
-use std::{boxed::Box, string::String};
+
+extern crate alloc;
```
DaniPopes: this is not necessary

DaniPopes (approved):

crates/precompile/src/interface.rs
```diff
@@ -48,18 +66,143 @@ impl PrecompileOutput {
...
+
+    /// BLS12-381 G2 addition (returns 192-byte unpadded G2 point)
+    fn bls12_381_g2_add(&self, a: G2Point, b: G2Point) -> Result<[u8; 192], PrecompileError> {
+        crate::bls12_381::crypto_backend::p2_add_affine_bytes(a, b)
+    }
+
+    /// BLS12-381 G2 multi-scalar multiplication (returns 192-byte unpadded G2 point)
+    fn bls12_381_g2_msm<'a>(
+        &self,
+        pairs: Box<dyn Iterator<Item = Result<G2PointScalar, PrecompileError>> + 'a>,
```
DaniPopes: i think these `Box<dyn Iterator>` can be `&mut dyn`

kevaundray: yeah I think switching to &mut dyn Iterator would make sense here

crates/precompile/src/interface.rs
```diff
@@ -48,18 +66,143 @@ impl PrecompileOutput {
...
+
+    /// BN128 elliptic curve scalar multiplication.
+    #[inline]
+    fn bn128_g1_mul(&self, point: &[u8], scalar: &[u8]) -> Result<[u8; 64], PrecompileError> {
+        crate::bn128::crypto_backend::g1_point_mul(point, scalar)
+    }
+
+    /// BN128 pairing check.
+    #[inline]
+    fn bn128_pairing_check(&self, pairs: &[(&[u8], &[u8])]) -> Result<bool, PrecompileError> {
```
kevaundray, line 105: I can change this to use Iterator in a followup PR

crates/precompile/src/interface.rs
```diff
@@ -48,18 +66,143 @@ impl PrecompileOutput {
...
+    }
+
+    /// secp256r1 (P-256) signature verification.
+    #[inline]
+    fn secp256r1_verify_signature(
+        &self,
+        msg: &[u8; 32],
+        sig: &[u8; 64],
+        pk: &[u8; 64],
+    ) -> Result<bool, PrecompileError> {
```
kevaundray: Can also change this in a followup PR to be like verify_kzg_proof; `Result<bool, Error>` Since it always returns Ok

crates/precompile/src/lib.rs
```diff
@@ -67,32 +67,19 @@ pub fn calc_linear_cost_u32(len: usize, base: u64, word: u64) -> u64 {
 }
 
 /// Precompiles contain map of precompile addresses to functions and HashSet of precompile addresses.
-#[derive(Debug)]
+#[derive(Default, Debug)]
```
kevaundray: I think we can Derive clone again

crates/precompile/src/lib.rs
```diff
@@ -112,25 +99,14 @@ impl Precompiles {
     }
 
     /// Creates a new Precompiles instance with a custom crypto implementation.
-    pub fn with_crypto(spec: PrecompileSpecId, crypto: Box<dyn Crypto>) -> Self {
+    pub fn with_crypto(spec: PrecompileSpecId) -> Self {
```
kevaundray: I think we can remove this `with_crypto` method, though thinking out loud, I wonder if we can make the `install_crypto` method crate-visible and use this method to call `install_crypto`

crates/precompile/src/lib.rs
rakita: Added it

crates/precompile/src/lib.rs
rakita: Yeah, it does not make sense to have this fn. Have reexported `install_crypto` inside `revm` so it is slightly easier to use

PR #2787, opened 2025-07-24
chore: add OnceLock re-export with no_std support

DaniPopes:
Same as `alloy_primitives::map` that uses `hashbrown`, here we use `once_cell::race::OnceBox` for `no_std` `OnceLock`, while preferring `std::sync::OnceLock`.


DaniPopes:
Small regressions look like allocator/inlining noise

rakita (approved):

PR #2789, opened 2025-07-24
chore: add rust-version and note about MSRV

DaniPopes:
Having rust-version set helps Cargo produce better error messages if the installed version is lower than the minimum specified.


kevaundray (approved):
Minor note: reth is targetting [1.86](https://github.com/paradigmxyz/reth/blob/876e964cbcbd85ade6fdec40bb35f08690332854/Cargo.toml#L4)

rakita (approved):

PR #2797, opened 2025-07-24
chore: fix inspector, cleanup loop

DaniPopes:
Applies https://github.com/bluealloy/revm/pull/2731

### DIFF
--- a/crates/inspector/src/handler.rs
+++ b/crates/inspector/src/handler.rs
@@ -4,8 +4,8 @@ use handler::{evm::FrameTr, EvmTr, FrameResult, Handler, ItemOrResult};
 use interpreter::{
     instructions::InstructionTable,
     interpreter_types::{Jumps, LoopControl},
-    FrameInput, Host, InitialAndFloorGas, InstructionContext, InstructionResult, Interpreter,
-    InterpreterAction, InterpreterTypes,
+    FrameInput, Host, InitialAndFloorGas, InstructionResult, Interpreter, InterpreterAction,
+    InterpreterTypes,
 };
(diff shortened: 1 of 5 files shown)


crates/inspector/src/handler.rs
```diff
@@ -185,32 +185,20 @@ where
...
 
-        let done = interpreter.bytecode.is_end();
-        if done {
-            interpreter.bytecode.revert_to_previous_pointer();
-        }
-
         inspector.step_end(interpreter, context);
 
-        if done {
+        if interpreter.bytecode.is_end() {
```
DaniPopes, line 201: idt it matters if `is_end` calculated before or after step_end, but technically this is now done after

rakita: It should definitely be after `step_end`. If `step_end` sets action, it should break the interpreter loop

crates/interpreter/src/interpreter.rs
```diff
@@ -252,22 +249,8 @@ impl<IW: InterpreterTypes> Interpreter<IW> {
...
-                interpreter: self,
-                host,
-            };
-            // Execute instruction.
-            instruction_table[opcode as usize](context);
+            self.step(instruction_table, host);
         }
-        self.bytecode.revert_to_previous_pointer();
-
         self.take_next_action()
```
rakita, line 254:
```suggestion
        self.bytecode.reset_action()
        self.take_next_action()
```

rakita: Okay, this is already done in take_next_action

rakita (approved):
lgtm

PR #2799, opened 2025-07-25
refactor checkpoint

jackwener:


rakita:
Funny enough, I have read somewhere that this was a way how initial EVMs had implemented reverts. Db journaling is optimisation that every evm does now.

This is not good path forward

enlightened88:
Hello friend, pick up your drop in succinct, deadline 2 days.

PR #2801, opened 2025-07-25
chore: fix run-tests.sh

DaniPopes:


scripts/run-tests.sh
```diff
@@ -1,7 +1,7 @@
-# ./run-tests --help
+#!/usr/bin/env bash
+set -eo pipefail
```
rakita, line 2: Did you encounter a problem without `pipefail`?

rakita (approved):

scripts/run-tests.sh
DaniPopes: set -e doesnt work in pipes so this adds some extra robustness

the error was shebang not being in the first line

rakita: I see

PR #2807, opened 2025-07-25
fix(op-revm): system tx not enveloped

emhane:
Bug: system tx not enveloped, triggering panic `"all not deposit tx have enveloped tx"` in handler.
Fix: sets enveloped tx equal to `Some(Bytes::default())`

Adds test covering `OpEvm::transact_system_call_with_caller`, increasing coverage of _op-revm/src/api/exec.rs_ from 59% to 79%.

### DIFF
--- a/crates/op-revm/src/transaction/abstraction.rs
+++ b/crates/op-revm/src/transaction/abstraction.rs
@@ -89,11 +89,15 @@ impl<TX: Transaction + SystemCallTx> SystemCallTx for OpTransaction<TX> {
         system_contract_address: Address,
         data: Bytes,
     ) -> Self {
-        OpTransaction::new(TX::new_system_tx_with_caller(
+        let mut tx = OpTransaction::new(TX::new_system_tx_with_caller(
             caller,
             system_contract_address,
             data,
-        ))
+        ));
+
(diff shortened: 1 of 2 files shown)


crates/op-revm/src/transaction/abstraction.rs
```diff
@@ -89,11 +89,15 @@ impl<TX: Transaction + SystemCallTx> SystemCallTx for OpTransaction<TX> {
...
+        let mut tx = OpTransaction::new(TX::new_system_tx_with_caller(
             caller,
             system_contract_address,
             data,
-        ))
+        ));
+
+        tx.enveloped_tx = Some(Bytes::default());
+
+        tx
```
emhane, line 100: I'm not sure if we need to be checking if this is pre-regolith here actually. I didn't find out how revm performs the actual eip2718 encoding since it seems alloy's impl can't be used here as revm has its own transaction trait ? @rakita

rakita: Revm does not perform eip2718 encoding, it is a field.

Calling `replay` on system "tx" is an undefined behaviour (as fields for this tx are undefined), so having empty bytes for envelop seems appropriate.

PR #2808, opened 2025-07-25
feat: add system transaction inspection support

### LINKED ISSUE #2798: feat request: Add Inspecting support for System Transactons
At the moment, normal transactions have inspecting variants, but system transactions don't have a way to enable inspecting. The system_transaction transaction abstraction is used in bera-reth to add protocol enforced, gas-free, nonce-free transactions at top of block, but we also want it to look like a normal transaction in block explorers and via RPC methods (which is not the case for other system calls like eip4788). 
Inspecting doesn't seem to be possible for such transactions. (most frequently used in `debug_traceTransaction` and `debug_traceBlock` apis.

## Additional Context

In Reth, there exists the `Trace` trait which is the entry point for transaction traction tracing (see https://github.com/paradigmxyz/reth/blob/818e01773accb34ef807136fec343179678b40b5/crates/rpc/rpc-eth-api/src/helpers/trace.rs#L30-L48)

This eventually calls into `alloy-evm` as such:
```rs
    fn transact_raw(
        &mut self,
        tx: Self::Tx,
    ) -> Result<ResultAndState<Self::HaltReason>, Self::Error> {
        if self.inspect {
            self.inner.inspect_tx(tx)
        } else {
            self.inner.transact(tx)
        }
    }
```

However, transact system call reaches here, which has we can see has no check for `inspect` as there is no `inspect` equivalent for system txs:
```rs
    fn transact_system_call(
        &mut self,
        caller: Address,
        contract: Address,
        data: Bytes,
    ) -> Result<ResultAndState<Self::HaltReason>, Self::Error> {
        self.inner.transact_system_call_with_caller_finalize(caller, contract, data)
    }
```

Looking deeper into the implementations in revm, I don't see an obvious reason why `inspect` couldn't exist for system txs unless I'm mistaken. Thanks!

rezbera:
Adds inspection support for system transactions to enable tracing with debug_traceTransaction and debug_traceBlock APIs.

## Changes

- Added InspectSystemCallEvm trait extending InspectEvm and SystemCallEvm
- Added InspectorHandler::inspect_run_system_call method for system call inspection
- Implemented inspection support for mainnet EVM with comprehensive test coverage
- Re-exported new traits in main revm crate

Closes #2798

### DIFF
--- a/crates/inspector/src/inspect.rs
+++ b/crates/inspector/src/inspect.rs
@@ -1,5 +1,7 @@
 use context::result::ExecResultAndState;
-use handler::{system_call::SYSTEM_ADDRESS, ExecuteCommitEvm, ExecuteEvm, SystemCallEvm, SystemCallCommitEvm};
+use handler::{
+    system_call::SYSTEM_ADDRESS, ExecuteCommitEvm, ExecuteEvm, SystemCallCommitEvm, SystemCallEvm,
+};
 use primitives::{Address, Bytes};
 
 /// InspectEvm is a API that allows inspecting the EVM.
@@ -93,11 +95,7 @@ pub trait InspectSystemCallEvm: InspectEvm + SystemCallEvm {
         system_contract_address: Address,
         data: Bytes,
     ) -> Result<Self::ExecutionResult, Self::Error> {
-        self.inspect_system_call_with_caller_one(
-            SYSTEM_ADDRESS,
-            system_contract_address,
-            data,
-        )
+        self.inspect_system_call_with_caller_one(SYSTEM_ADDRESS, system_contract_address, data)
     }
 
     /// Inspect a system call with the current inspector and a custom caller.
@@ -160,11 +158,8 @@ pub trait InspectSystemCallEvm: InspectEvm + SystemCallEvm {
         data: Bytes,
         inspector: Self::Inspector,
     ) -> Result<ExecResultAndState<Self::ExecutionResult, Self::State>, Self::Error> {
-        let output = self.inspect_system_call_with_inspector_one(
-            system_contract_address,
-            data,
-            inspector,
-        )?;
+        let output =
+            self.inspect_system_call_with_inspector_one(system_contract_address, data, inspector)?;
         let state = self.finalize();
         Ok(ExecResultAndState::new(output, state))
     }
@@ -207,11 +202,8 @@ pub trait InspectSystemCallCommitEvm:
         data: Bytes,
         inspector: Self::Inspector,
     ) -> Result<Self::ExecutionResult, Self::Error> {
-        let output = self.inspect_system_call_with_inspector_one(
-            system_contract_address,
-            data,
-            inspector,
-        )?;
+        let output =
+            self.inspect_system_call_with_inspector_one(system_contract_address, data, inspector)?;
         self.commit_inner();
         Ok(output)
     }
--- a/crates/inspector/src/inspect.rs
+++ b/crates/inspector/src/inspect.rs
@@ -1,5 +1,6 @@
 use context::result::ExecResultAndState;
-use handler::{ExecuteCommitEvm, ExecuteEvm};
+use handler::{system_call::SYSTEM_ADDRESS, ExecuteCommitEvm, ExecuteEvm, SystemCallEvm, SystemCallCommitEvm};
+use primitives::{Address, Bytes};
 
 /// InspectEvm is a API that allows inspecting the EVM.
 ///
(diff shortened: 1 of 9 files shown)


rezzmah:
tested with a system call in bera-reth and its working as intended (after updating alloy-evm + bera-reth ofc). The following tx is executed as a system tx, and is now traceable. 
```sh
curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"debug_traceTransaction","params":["0xf13ff9cde2232ffece74532fc34110eaa38bb01d8cba415c9eba8390319030b6",{"tracer":"callTracer"}],"id":1}' \
      http://localhost:8545/ | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1097  100   935  100   162  11553   2001 --:--:-- --:--:-- --:--:-- 13712
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "from": "0xfffffffffffffffffffffffffffffffffffffffe",
    "gas": "0x1000000",
    "gasUsed": "0x42d9",
    "to": "0x4200000000000000000000000000000000000042",
    "input": "0x60644a6b000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000309070f3f3d1ef0bd7e3186f7945261291d9363005eb126c64660daec91d51514bbcc50e8fb1d0f632bad82f2901e43d7600000000000000000000000000000000",
    "calls": [
      {
        "from": "0x4200000000000000000000000000000000000042",
        "gas": "0x1c27be0",
        "gasUsed": "0x19f0",
        "to": "0x4200000000000000000000000000000000000043",
        "input": "0x999da65b000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000309070f3f3d1ef0bd7e3186f7945261291d9363005eb126c64660daec91d51514bbcc50e8fb1d0f632bad82f2901e43d7600000000000000000000000000000000",
        "value": "0x0",
        "type": "CALL"
      }
    ],
    "value": "0x0",
    "type": "CALL"
  }
}
```

rakita (commented):
Needs to be added to op-revm

crates/inspector/src/inspect.rs
```diff
@@ -76,3 +77,88 @@ pub trait InspectCommitEvm: InspectEvm + ExecuteCommitEvm {
...
+///
+/// It extends [`InspectEvm`] and [`SystemCallEvm`] traits to provide inspection
+/// capabilities for system transactions, enabling tracing and debugging of
+/// system calls similar to regular transactions.
+pub trait InspectSystemCallEvm: InspectEvm + SystemCallEvm {
+    /// Inspect a system call with the current inspector.
+    ///
+    /// Similar to [`InspectEvm::inspect_one_tx`] but for system calls.
+    /// Uses [`SYSTEM_ADDRESS`] as the caller.
+    fn inspect_system_call_one(
```
rakita: Inspection naming is `inspect_one_tx` so it would be nicer for this to be `inspect_one_system_call`

crates/inspector/src/inspect.rs
```diff
@@ -76,3 +77,88 @@ pub trait InspectCommitEvm: InspectEvm + ExecuteCommitEvm {
...
+        system_contract_address: Address,
+        data: Bytes,
+    ) -> Result<Self::ExecutionResult, Self::Error> {
+        self.inspect_system_call_with_caller_one(SYSTEM_ADDRESS, system_contract_address, data)
+    }
+
+    /// Inspect a system call with the current inspector and a custom caller.
+    ///
+    /// Similar to [`InspectEvm::inspect_one_tx`] but for system calls with a custom caller.
+    fn inspect_system_call_with_caller_one(
```
rakita: `inspect_one_system_call_with_caller`

crates/inspector/src/inspect.rs
```diff
@@ -76,3 +77,88 @@ pub trait InspectCommitEvm: InspectEvm + ExecuteCommitEvm {
...
+        let output =
+            self.inspect_system_call_with_caller_one(caller, system_contract_address, data)?;
+        let state = self.finalize();
+        Ok(ExecResultAndState::new(output, state))
+    }
+
+    /// Inspect a system call with a given inspector.
+    ///
+    /// Similar to [`InspectEvm::inspect_one`] but for system calls.
+    fn inspect_system_call_with_inspector_one(
```
rakita:
```suggestion
    fn inspect_one_system_call_with_inspector(
```

rakita (approved):

PR #2815, opened 2025-07-28
test(op-revm): Adds caller nonce assertion to op-revm intergation tests

emhane:
Asserts caller nonce post-execution for existing op-revm integration tests


crates/op-revm/tests/integration.rs
```diff
@@ -1087,5 +1184,10 @@ fn test_system_call() {
     // Run evm.
     let output = evm.replay().unwrap();
 
+    assert_eq!(
+        output.state.get(&Address::ZERO).map(|a| a.info.nonce),
+        Some(0)
+    );
```
emhane: double checking that this the expected behaviour for system call?

rakita: Yes, system call does not touch caller to beneficiary account

rakita:
No need to check the nonce every time with assert as it is checked/saved inside json fixture.

This PR should order it so it will be easier to diff in github: https://github.com/bluealloy/revm/pull/2813

crates/op-revm/tests/integration.rs
emhane: if caller account wouldn't be touched, then it should be `None` ~~like for deposit~~, not `Some(0)`. adding an assertion here to verifying that caller has not been marked as touched, the test fails.
```rust
assert!(!output.state.get(&Address::ZERO).unwrap().is_touched());
```

rakita: Ah this is touched after `.replay`. If you check account after `system_call`, it will not be touched.

Although I see expectation for `replay` to behave the same as system call, but this would make things complex and not performant. The intention for `replay` is to be removed somewhere in the future and for users to rely on `.transact(tx)`.

emhane: true! that replay statement isn't necessary there. thanks.

rakita (approved):

Issue #2564, opened 2025-06-01
Implement manual deserialize for AccountInfo


mattsse:
```rust
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub struct AccountInfo {
```

matches the non standard `eth_getAccountInfo` endpoint response:

https://github.com/alloy-rs/alloy/blob/c04ef82f3a63c97cb3332a78b3c18dca47966ad3/crates/rpc-types-eth/src/account.rs#L9-L21

this means for an rpc db this could be replaced with single rpc call if the node supports that endpoint and directly deserialize into revm AccountInfo if deserialize is manually impl so that the codehash is computed during deserde

```
   let (nonce, balance, code) = tokio::join!(nonce, balance, code,);
```

## TODO
* add manual deserde that in addition to the current format also supports the eth_getAccountInfo response

0xMushow:
Can I take this one please?

rakita:
Go for it @0xMushow

0xkenj1:
Hi! Is this issue still open/available? I’d be happy to take it on :)
@0xMushow @rakita @mattsse

Pratham1812:
Hey @rakita @mattsse  was also looking to try out this out . Needed some more context if possible. I am little new to Rust dev and still improving.
Any pointers will be appreciated.

Nanle-code:
Is this issue still open @rakita @mattsse 
I'd love to give it a go

PR #2816, opened 2025-07-28
feat: removed padding in case last opcode is terminating or unknown

### LINKED ISSUE #2796: Avoid padding if last opcode is terminating
In bytecode analysis we always pad with one STOP opcode (0 byte) at the end, if the last opcode is not already a STOP:
https://github.com/bluealloy/revm/blob/e42a93a86580da9c861e568f24d86482532f3560/crates/bytecode/src/legacy/analysis.rs#L42

However we can avoid doing this if the last opcode is known to be terminating, e.g. is unknown or has the terminating flag set

wdyt @rakita

Pratham1812:
### Removes padding with `STOP` opcode in case last opcode is terminating or unknown

- Creates a wrapper function around already existing `is_terminating` function to check if the last opcode is unknown or terminating
- Introduces a new test to check the implementation
- Removes failing tests.

Fixes #2796

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -4,6 +4,17 @@ use bitvec::{bitvec, order::Lsb0, vec::BitVec};
 use primitives::Bytes;
 use std::vec::Vec;
 
+/// Check if an opcode is terminating (will end execution)
+fn is_terminating_opcode(opcode: u8) -> bool {
+    // First check if it's a known opcode
+    if let Some(opcode_info) = opcode::OpCode::info_by_op(opcode) {
+        opcode_info.is_terminating()
+    } else {
+        // Unknown opcodes are also terminating (cause INVALID instruction behavior)
+        true
+    }
+}
+
 /// Analyzes the bytecode for use in [`LegacyAnalyzedBytecode`](crate::LegacyAnalyzedBytecode).
 ///
 /// See [`LegacyAnalyzedBytecode`](crate::LegacyAnalyzedBytecode) for more details.
@@ -39,7 +50,12 @@ pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
         }
     }
 
-    let padding = (iterator as usize) - (end as usize) + (opcode != opcode::STOP) as usize;
+    let overflow_padding = (iterator as usize) - (end as usize);
+
+    let stop_padding = if !is_terminating_opcode(opcode) { 1 } else { 0 };
+
+    let padding = overflow_padding + stop_padding;
+
     let bytecode = if padding > 0 {
         let mut padded = Vec::with_capacity(bytecode.len() + padding);
         padded.extend_from_slice(&bytecode);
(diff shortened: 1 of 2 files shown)


crates/bytecode/src/legacy/analysis.rs
```diff
@@ -4,6 +4,17 @@ use bitvec::{bitvec, order::Lsb0, vec::BitVec};
 use primitives::Bytes;
 use std::vec::Vec;
 
+/// Check if an opcode is terminating (will end execution)
+fn is_terminating_opcode(opcode: u8) -> bool {
+    // First check if it's a known opcode
+    if let Some(opcode_info) = opcode::OpCode::info_by_op(opcode) {
```
rakita: We should only consider known opcodes. Comment from https://github.com/bluealloy/revm/issues/2796#issuecomment-3117233884

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -39,7 +50,12 @@ pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
         }
     }
 
-    let padding = (iterator as usize) - (end as usize) + (opcode != opcode::STOP) as usize;
+    let overflow_padding = (iterator as usize) - (end as usize);
+
+    let stop_padding = if !is_terminating_opcode(opcode) { 1 } else { 0 };
```
rakita:
```suggestion
    let stop_padding = OpCode::info_by_op(opcode).map(|o| !o.is_terminating() as usize).unwrap_or(1);
```

can replace the function.

crates/bytecode/src/legacy/analyzed.rs
```diff
@@ -77,10 +76,6 @@ impl LegacyAnalyzedBytecode {
             "jump table length is less than original length"
         );
         assert!(!bytecode.is_empty(), "bytecode cannot be empty");
-        assert!(
-            bytecode.last() == Some(&opcode::STOP),
```
rakita, line 81: would be good to check if opcode is terminating

Pratham1812: On it.

crates/bytecode/src/legacy/analyzed.rs
```diff
@@ -150,14 +156,6 @@ mod tests {
         let _ = LegacyAnalyzedBytecode::new(bytecode.bytecode, bytecode.original_len, jump_table);
     }
 
-    #[test]
-    #[should_panic(expected = "last bytecode byte should be STOP (0x00)")]
-    fn test_panic_on_non_stop_bytecode() {
-        let bytecode = Bytes::from_static(&[opcode::PUSH1, 0x01]);
-        let jump_table = JumpTable::new(bitvec![u8, Lsb0; 0; 2]);
-        let _ = LegacyAnalyzedBytecode::new(bytecode, 2, jump_table);
-    }
```
rakita, line 159: Lets revert this as it is needed to test `last bytecode byte should be terminating`

rakita (approved):
lgtm, pending test revert

Pratham1812:
Hey @rakita was also looking to try out this one  #2564 . Needed some more context if possible. I am little new to Rust dev and still improving.
Any pointers will be appreciated.

rakita:
> Hey @rakita was also looking to try out this one https://github.com/bluealloy/revm/issues/2564 . Needed some more context if possible. I am little new to Rust dev and still improving.
Any pointers will be appreciated.

It is a small task, check `eth_getAccountInfo ` rpc and how it returns account and align AccountInfo to be serialized to it. Very similar how alloy does it: https://github.com/alloy-rs/alloy/blob/c04ef82f3a63c97cb3332a78b3c18dca47966ad3/crates/rpc-types-eth/src/account.rs#L9-L21

PR #2819, opened 2025-07-28
feat: precompile id

rkrasiuk:
## Description

Add unique identifier for precompiles.

### DIFF
--- a/crates/precompile/src/lib.rs
+++ b/crates/precompile/src/lib.rs
@@ -71,8 +71,8 @@ pub fn calc_linear_cost_u32(len: usize, base: u64, word: u64) -> u64 {
 #[derive(Clone, Default, Debug)]
 pub struct Precompiles {
     /// Precompiles
-    inner: HashMap<Address, PrecompileFn>,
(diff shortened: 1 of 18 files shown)


crates/precompile/src/lib.rs
```diff
@@ -300,33 +300,94 @@ impl Precompiles {
...
+    /// Pairing operations between a set of pairs of (G1, G2) points.
+    Bls12Pairing,
+    /// Base field element mapping into the G1 point.
+    Bls12MapFpToGp1,
+    /// Extension field element mapping into the G2 point.
+    Bls12MapFp2ToGp2,
+    /// ECDSA signature verification over the secp256r1 elliptic curve (also known as P-256 or prime256v1).
+    P256Verify,
+    /// Custom precompile identifier.
+    Custom(String),
```
mattsse: this should be `Other(Cow<'static, str>)`

and a helper fn to create a static one

mattsse (changes requested):

rakita:
hm, why is this needed?

PR #2823, opened 2025-07-29
feat: small perf rm b.lo instr

rakita:
wip.

we dont care about gas if it is confirmed it is oog

or as eth tests are failing, we can set remaining to zero if action is oog


rakita:
There is no benefit to this, as probably checks for b.lo are already inlined

PR #2846, opened 2025-08-04
feat: optimize access to precompile short addreesses

rakita:
All mainnet precompiles are under the first 100 addresses, we can use this knowledge and optimise access to those functions so we dont need to access the hashmap

PR #2849, opened 2025-08-05
feat: short address for journal cold/warm check

rakita:
Added optimisation for precompile short address access. This is an extension of https://github.com/bluealloy/revm/pull/2846

Previously access list account was not loaded from database but inserted into the warm_accounts hashset. This made sense if the account from the access list is not accessed in execution. This PR switches this behaviour and always loads all access lists at the start to mark the warm.


crates/context/src/journal/inner.rs
```diff
@@ -166,9 +145,9 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
         *depth = 0;
         logs.clear();
         *transaction_id += 1;
+
         // Clear coinbase address warming for next tx
-        *warm_coinbase_address = None;
-        reset_preloaded_addresses(warm_preloaded_addresses, precompiles);
+        warm_addresses.clear_coinbase();
```
DaniPopes, line 150: this doesn't clear precompiles anymore?

DaniPopes: i guess it's always reset in set_precompile_addresses

rakita: `set_precompile_addresses` will reset it, and it is only called on spec_id change or if precompile address is empty.

DaniPopes:
lgtm

PR #2851, opened 2025-08-05
fix: correct typos across codebase

radik878:
Fix various typos in comments and documentation, such as:

- “Can be slightly faster that” → “Can be slightly faster than”
- “RJUMV” → “RJUMPV”
- “a address” → “an address”
- “this fields” → “these fields”
- “has a *_mut variant” → “have a *_mut variant”
- “frm instruction” → “from instruction”
- “Returns authorization the chain id” → “Returns the chain id from the authorization”
- “all Ethereum transaction” → “all Ethereum transactions”

…and similar typo fixes throughout the codebase.

### DIFF
--- a/crates/bytecode/src/opcode.rs
+++ b/crates/bytecode/src/opcode.rs
@@ -160,7 +160,7 @@ impl OpCode {
 
     /// Returns the number of both input and output stack elements.
     ///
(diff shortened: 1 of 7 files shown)


rakita:
Thanks for finding those, for the sake of not attracting typo bot, will make a new PR

PR #2862, opened 2025-08-07
perf: compute changed_storage lazily in CacheState::apply_account_state

nkysg:
self-destructed accounts and empty accounts under EIP-161 don't need to compute changed_storage


rakita:
No need for this, as this is only called on the path of empty touched account and those no longer exist.

PR #2866, opened 2025-08-08
fix: correct internal function reference in calc_excess_blob_gas comment

Fibonacci747:
Fix incorrect comment in calc_excess_blob_gas function

The comment incorrectly stated that the function uses calc_excess_blob_gas 
internally, when it actually uses calc_excess_blob_gas_osaka.


rakita:

PR #2867, opened 2025-08-08
Remove debug macro from test code

aso20455:
Remove accidentally left dbg! macro from test_step_halt test function. The debug output is not needed as the test already has proper assertions to verify the expected behavior


rakita:

PR #2868, opened 2025-08-08
perf: small performance and safety improvements

rakita:
- Remove unwrap() in primitives::short_address by using direct array indexing
- Use lazy allocation with or_insert_with to avoid unnecessary Vec allocations
- Pre-allocate Vec capacity in State::increment_balances based on iterator hint
- Simplify iterator pattern in bundle_state to use destructuring
- Use mem::take instead of mem::replace for bool values
- Add #[inline] attributes to small hot-path functions in state module
- Use unwrap_or_default() instead of unwrap_or(0) for cleaner code

These are small incremental improvements that enhance performance and code safety without changing functionality.

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -56,7 +56,7 @@ pub const SHORT_ADDRESS_CAP: usize = 300;
 #[inline]
 pub fn short_address(address: &Address) -> Option<usize> {
     if address.0[..18].iter().all(|b| *b == 0) {
-        let short_address = u16::from_be_bytes(address.0[18..].try_into().unwrap()) as usize;
+        let short_address = u16::from_be_bytes([address.0[18], address.0[19]]) as usize;
         if short_address < SHORT_ADDRESS_CAP {
             return Some(short_address);
         }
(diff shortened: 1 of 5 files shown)


crates/primitives/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub const SHORT_ADDRESS_CAP: usize = 300;
 #[inline]
 pub fn short_address(address: &Address) -> Option<usize> {
     if address.0[..18].iter().all(|b| *b == 0) {
-        let short_address = u16::from_be_bytes(address.0[18..].try_into().unwrap()) as usize;
+        let short_address = u16::from_be_bytes([address.0[18], address.0[19]]) as usize;
```
kevaundray, line 59: Noting: an address is 20 bytes, so this is equivalent. 

Hopefully the compiler can figure it out and avoid the bounds checks

kevaundray (approved):
LGTM

crates/primitives/src/lib.rs
rakita: It is a constant index to an array that is of fixed size, compiler has all the information needed to optimise it

PR #2871, opened 2025-08-08
feat: impl inspector for tuple

klkvr:
Implements `Inspector` for tuple of 2 inspectors that will be invoked one after another.

It seems that there's no other implementation allowing for combining 2 inspectors behaviors in that way


sorpaas:
> It seems that there's no other implementation allowing for combining 2 inspectors behaviors in that way

[Rust EVM](https://github.com/rust-ethereum/evm) already supported this a long time ago. We can do both internal inspection as well as single stepping. We allow extending the `Etable`, so an arbitrary number of inspectors can be chained.

Anyway, always happy to learn the interesting design choices made by other impls!

PR #2874, opened 2025-08-10
refactor: use HashMap::is_empty() instead of len() == 0

strmfos:
Replace `self.inner.len() == 0` with `self.inner.is_empty()` in Precompiles::is_empty() method. This follows Rust idioms and best practices, potentially improving readability and performance


rakita:

PR #2877, opened 2025-08-11
refactor: add gas refund tracking to journal and host interfaces

### LINKED ISSUE #2145: Refactor: migrate from call local refund to global journal refund

jackwener:
close #2145

- Introduced `record_refund`, `refund`, and `reset_refund` methods in the `JournalTr` trait to manage gas refunds at the journal level.
- Updated the `Host` trait to include a `record_refund` method for recording gas refunds.
- Implemented the corresponding methods in `DummyHost` and `JournalInner`.
- Adjusted gas handling in various components to ensure refunds are recorded correctly, adhering to EIP-3529 limits.
- Refactored related functions in the handler and interpreter to utilize the new refund tracking system.

### DIFF
--- a/crates/context/src/journal/inner.rs
+++ b/crates/context/src/journal/inner.rs
@@ -147,11 +147,8 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
             warm_addresses,
             refund,
         } = self;
-        let is_spurious_dragon_enabled = spec.is_enabled_in(SPURIOUS_DRAGON);
         // iterate over all journals entries and revert our global state
-        journal.drain(..).rev().for_each(|entry| {
-            entry.revert(state, None, is_spurious_dragon_enabled);
-        });
+        Self::revert_journal_entries_impl(journal.drain(..), refund, state, None, *spec);
         transient_storage.clear();
         *depth = 0;
         logs.clear();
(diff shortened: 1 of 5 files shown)


crates/context/src/journal/inner.rs
```diff
@@ -855,3 +875,28 @@ pub fn sload_with_account<DB: Database, ENTRY: JournalEntryTr>(
...
+    pub fn refund(&self) -> i64 {
+        self.refund
+    }
+
+    /// Resets the gas refund to zero.
+    ///
+    /// This is typically called at the start of a new transaction.
+    #[inline]
+    pub fn reset_refund(&mut self) {
+        self.refund = 0;
```
rakita: This function is not something that we should call

crates/context/src/journal/inner.rs
```diff
@@ -855,3 +875,28 @@ pub fn sload_with_account<DB: Database, ENTRY: JournalEntryTr>(
...
 }
+
+impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
+    /// Records a gas refund for the current transaction.
+    ///
+    /// This accumulates refunds at the global journal level, avoiding negative
+    /// refund values being visible during individual call frame execution.
+    #[inline]
+    pub fn record_refund(&mut self, refund: i64) {
+        self.refund += refund;
```
rakita, line 880: Journal entry for refund should be made so refund can be reverted in case of revert/oog/stackoverflow etc.

jackwener: > case of revert/oog/stackoverflow etc.

Look like these aren't covered by test.

rakita: Yeah will ping ef testing team

rakita:
This looks okay now, will see to merge it a little bit later so it does not affect downstream inspectors in the next release.

PR #2878, opened 2025-08-11
feat(Jurnal): Accounts saved in index map.

rakita:
Journal function now contain `AddressOrId` field that allow fetching account by or or by address

Output state is now a index map, that contains pages of accounts and hashmap that maps address to the index. 

Analysis:
codspeed times are very slower than running on machine but even with them we can see order of magnitude difference between transfer and CALLs. 1000tx transfers with commit take 2.8ms that means one transfer is 2.8us and degradation is 0.2us per transaction, while 50 CALLs are 97us and it is improved to 91 µs
And when i run this on machine, `revme bench transfer-multi` I get `transact_commit_1000txs` time:  `[387.81 µs 393.68 µs 398.68 µs]` which means on the real machine it is 0.4us per transaction and diff is 0.05us or 50ns .

For context 50ns degradation for 3 billion transactions (number of mainnet transactions) are few (two/three) minutes of difference.

My conclusion is even if transact_commit_1000tx is worse by 7%, an improvement of 7% for CALL or %5 SSTORE is worth a lot more.


crates/context/interface/src/journaled_state.rs
```diff
@@ -31,53 +32,53 @@ pub trait JournalTr {
     /// Loads the storage from database if not found in Journal state.
     fn sload(
         &mut self,
-        address: Address,
+        address_or_id: AddressOrId,
```
rakita, line 35: Interface got changed so it now can process both account id or address.

Account id is obtained on loading of the account and it can be used to fetch the account faster from journal.

crates/context/interface/src/journaled_state.rs
```diff
@@ -31,53 +32,53 @@ pub trait JournalTr {
...
     fn warm_account(
         &mut self,
         address: Address,
-    ) -> Result<(), <Self::Database as Database>::Error> {
+    ) -> Result<AddressAndId, <Self::Database as Database>::Error> {
         self.warm_account_and_storage(address, [])
     }
 
-    /// Warms the coinbase account.
-    fn warm_coinbase_account(&mut self, address: Address);
```
rakita, line 78: We can load coinbase from the start and make it warm (if needed), as fetching it in post execution state is going to be fast with account id.

crates/context/interface/src/journaled_state.rs
```diff
@@ -86,94 +87,112 @@ pub trait JournalTr {
...
+    ) -> Result<StateLoad<(&mut Account, AddressAndId)>, <Self::Database as Database>::Error>;
+
+    /// Returns the account by id.
+    fn get_account_mut(&mut self, account_id: AccountId) -> (&mut Account, AddressAndId);
+
+    /// Fast fetch the account by id.
+    fn get_account(&mut self, account_id: AccountId) -> (&Account, AddressAndId);
+
+    /// Sets the caller id.
+    fn set_caller_address_id(&mut self, id: AddressAndId);
```
rakita, line 136: Caller/coinbase/txtarget addresses and ids are saved in Journal so they can be fetched on demand.

crates/database/src/states/cache.rs
```diff
@@ -89,9 +89,12 @@ impl CacheState {
...
 
     /// Applies output of revm execution and create account transitions that are used to build BundleState.
-    pub fn apply_evm_state(&mut self, evm_state: EvmState) -> Vec<(Address, TransitionAccount)> {
+    pub fn apply_evm_state(
+        &mut self,
+        mut evm_state: EvmState,
+    ) -> Vec<(Address, TransitionAccount)> {
         let mut transitions = Vec::with_capacity(evm_state.len());
-        for (address, account) in evm_state {
+        for (account, address) in evm_state.take_accounts().into_iter().flatten() {
```
rakita, line 97: Iterating over accounts

crates/handler/src/mainnet_builder.rs
```diff
@@ -119,11 +119,13 @@ mod test {
             .unwrap()
             .state;
 
-        let auth_acc = state.get(&signer.address()).unwrap();
-        assert_eq!(auth_acc.info.code, Some(Bytecode::new_eip7702(FFADDRESS)));
-        assert_eq!(auth_acc.info.nonce, 1);
+        println!("state: {:#?}", state);
```
onbjerg, line 122:
```suggestion
```

rakita:
Obsolete

PR #2879, opened 2025-08-11
chore(bytecode): remove unused Debug import

Snezhkko:
Removed use core::fmt::Debug; from crates/bytecode/src/bytecode.rs since #[derive(Debug)] does not require it. No functional changes; cleans up an unused import warning.


rakita (approved):
thanks

PR #2891, opened 2025-08-16
opt: optimize Stack methods for performance and safety

jackwener:
This pull request refactors several methods in the `Stack` implementation to use unsafe memory operations for improved performance.


jackwener:
BTW, Why we don't use `[U256;STACK_LIMIT]`? It has higher performance.

PR #2892, opened 2025-08-17
fix(handler): correct transaction ID decrement logic

jackwener:


crates/op-revm/src/handler.rs
```diff
@@ -436,7 +436,7 @@ where
             let old_balance = acc.info.balance;
 
             // decrement transaction id as it was incremented when we discarded the tx.
-            acc.transaction_id -= acc.transaction_id;
+            acc.transaction_id -= 1;
```
jackwener, line 439: `acc.transaction_id` always is 0

rakita (approved):

PR #2900, opened 2025-08-19
chore: remove tx type check

mattsse:
I believe this is reasonable because a non 4844 tx shouldnt have blobhashes, so there's no need to check txtype, which is inconvenient for forge testing for example

ref https://github.com/foundry-rs/foundry/pull/11355


rakita:
This breaks the notion of tx type check

PR #2902, opened 2025-08-20
feat: Add conversions to alloy `CallKind`

moricho:
`CallKind` has been moved from revm-inspector to alloy in https://github.com/alloy-rs/alloy/pull/2779.
To complete the migration, `CallKind` should have conversions to `CreateScheme` and `CallScheme` in revm that revm-inspector originally has in its repository: https://github.com/paradigmxyz/revm-inspectors/blob/74349ffba7654b770f325a8fa3167106a20a13bf/src/tracing/types.rs#L536-L555

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,9 +68,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.0.12"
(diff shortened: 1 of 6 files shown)


rakita:
This creates a strange dependency on alloy

PR #2905, opened 2025-08-20
Feature: Support bubbling up Precompile error messages 

### LINKED ISSUE #2898: Support bubbling up Precompile error messages
currently we do:

https://github.com/bluealloy/revm/blob/6f9a57c203f4c44031733ba93abb0e862e290a1a/crates/handler/src/precompile_provider.rs#L135-L142

which strips away the error message for the 

>  InstructionResult::PrecompileError

this makes debugging custom precompiles and user feedback more difficult

unsure what the implications of adding a `Box<dyn Error>` value to precompilerror variant are, but this would be nice to have if a precompile could convey what exactly failed

we could also make this part of the InterpreterResult if that is cheaper somehow

mablr:
## Motivation
This PR enhances precompile error reporting and closes #2898.

## Solution
Attempt to implement a mechanism that records potential precompile error in context if depth == 0, in order to bubble it up in output as Bytes. Idea proposed by Rakita https://github.com/bluealloy/revm/pull/2905#pullrequestreview-3155479883.

## TODO
- [x] make tests
- [x] doc

### DIFF
--- a/crates/context/interface/src/local.rs
+++ b/crates/context/interface/src/local.rs
@@ -3,7 +3,7 @@ use core::{
     cell::{Ref, RefCell},
     ops::Range,
 };
-use std::{rc::Rc, vec::Vec};
+use std::{rc::Rc, string::String, vec::Vec};
 
 /// Non-empty, item-pooling Vec.
 #[derive(Debug, Clone)]
@@ -202,6 +202,19 @@ pub trait LocalContextTr {
 
     /// Clear the local context.
     fn clear(&mut self);
+
+    /// Set the error message for a precompile error, if any.
+    ///
+    /// This is used to bubble up precompile error messages when the
+    /// transaction directly targets a precompile (depth == 1).
+    fn set_precompile_error_context(&mut self, _output: String) {}
+
+    /// Take and clear the precompile error context, if present.
+    ///
+    /// Returns `Some(String)` if a precompile error message was recorded.
+    fn take_precompile_error_context(&mut self) -> Option<String> {
+        None
+    }
 }
 
 #[cfg(test)]
--- a/crates/context/interface/src/local.rs
+++ b/crates/context/interface/src/local.rs
@@ -207,14 +207,12 @@ pub trait LocalContextTr {
     ///
     /// This is used to bubble up precompile error messages when the
     /// transaction directly targets a precompile (depth == 1).
-    fn set_precompile_error_context(&mut self, _output: String) {}
+    fn set_precompile_error_context(&mut self, _output: String);
 
     /// Take and clear the precompile error context, if present.
     ///
     /// Returns `Some(String)` if a precompile error message was recorded.
-    fn take_precompile_error_context(&mut self) -> Option<String> {
-        None
-    }
+    fn take_precompile_error_context(&mut self) -> Option<String>;
 }
 
 #[cfg(test)]
--- a/crates/context/interface/src/result.rs
+++ b/crates/context/interface/src/result.rs
@@ -568,7 +568,7 @@ pub enum SuccessReason {
 /// Indicates that the EVM has experienced an exceptional halt.
 ///
 /// This causes execution to immediately end with all gas being consumed.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum HaltReason {
     /// Out of gas error.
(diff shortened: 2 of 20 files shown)


crates/context/interface/Cargo.toml
```diff
@@ -22,6 +22,7 @@ workspace = true
 primitives.workspace = true
 database-interface.workspace = true
 state.workspace = true
+precompile.workspace = true
```
rakita: This introduces weird dependency

rakita (changes requested):
We should have a different approach for this. Only place where we can return the error outside of evm is if transaction target points to the precompile.

We should introduce the mechanism that if depth == 0 and precompile returns the error for this error to be saved in Context and returned in output

crates/context/interface/src/result.rs
```diff
@@ -568,7 +568,7 @@ pub enum SuccessReason {
 /// Indicates that the EVM has experienced an exceptional halt.
 ///
 /// This causes execution to immediately end with all gas being consumed.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
mablr, line 571: I hope removing `Copy` here doesn't affect anything outside revm.

crates/context/interface/src/result.rs
```diff
@@ -591,6 +591,8 @@ pub enum HaltReason {
     CreateCollision,
     /// Precompile error.
     PrecompileError,
+    /// Precompile error with message from context.
+    PrecompileErrorWithContext(String),
```
mablr, line 595: We can also forward directly `PrecompileError` type...

crates/handler/src/precompile_provider.rs
```diff
@@ -102,20 +102,23 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
...
                 }
-            }
-            CallInput::Bytes(bytes) => bytes.0.iter().as_slice(),
+                CallInput::Bytes(bytes) => bytes.0.iter().as_slice(),
+            };
+            precompile.execute(input_bytes, inputs.gas_limit)
         };
 
-        match precompile.execute(input_bytes, inputs.gas_limit) {
+        match exec_result {
```
mablr, line 121: This was necessary to have read-only context dropped before eventually setting later precompile error message.

crates/context/interface/src/result.rs
rakita: Should be fine

crates/context/interface/src/result.rs
rakita: New item is more clear, we can remove `PrecompileError` later

crates/inspector/src/eip3155.rs
```diff
@@ -283,7 +283,7 @@ where
     fn call_end(&mut self, context: &mut CTX, _: &CallInputs, outcome: &mut CallOutcome) {
         self.gas_inspector.call_end(outcome);
 
-        if context.journal_mut().depth() == 0 {
+        if context.journal_mut().depth() == 1 {
```
rakita:
```suggestion
        if context.journal_mut().depth() == 0 {
```

This probably should be 0 , as call_end has decremented depth

crates/inspector/src/eip3155.rs
```diff
@@ -294,7 +294,7 @@ where
     fn create_end(&mut self, context: &mut CTX, _: &CreateInputs, outcome: &mut CreateOutcome) {
         self.gas_inspector.create_end(outcome);
 
-        if context.journal_mut().depth() == 0 {
+        if context.journal_mut().depth() == 1 {
```
rakita:
```suggestion
        if context.journal_mut().depth() == 0 {
```

crates/context/interface/src/local.rs
```diff
@@ -202,6 +202,19 @@ pub trait LocalContextTr {
 
     /// Clear the local context.
     fn clear(&mut self);
+
+    /// Set the error message for a precompile error, if any.
+    ///
+    /// This is used to bubble up precompile error messages when the
+    /// transaction directly targets a precompile (depth == 1).
+    fn set_precompile_error_context(&mut self, _output: String) {}
```
rakita:
```suggestion
    fn set_precompile_error_context(&mut self, _output: String);
```

Lets make it necesarry

crates/context/interface/src/local.rs
```diff
@@ -202,6 +202,19 @@ pub trait LocalContextTr {
...
+
+    /// Take and clear the precompile error context, if present.
+    ///
+    /// Returns `Some(String)` if a precompile error message was recorded.
+    fn take_precompile_error_context(&mut self) -> Option<String> {
+        None
+    }
```
rakita:
```suggestion
    /// Returns `Some(String)` if a precompile error message was recorded.
    fn take_precompile_error_context(&mut self) -> Option<String>;
```

PR #2907, opened 2025-08-21
refactor(precompile): add new specific `PrecompileError` variants

### LINKED ISSUE #2906: Make `PrecompileError` compatible with `Copy` trait
While working on #2898. I realized that bubbling up `PrecompileError` requires the `Copy` to not impact performance with nasty `.clone()`'s.

This will involve the removal of variants with `String` value.

https://github.com/bluealloy/revm/blob/ae1655da953356fa6707e529bec09f80dd88abde/crates/precompile/src/interface.rs#L233-L236

- `Fatal` actually doesn't seem to be used
- `Other` is mostly a cheat to not create dedicated variants (this is remediable)

Once we'll have that, #2898 impl will be straightforward.

mablr:
*Initially the goal was to make `PrecompileError` compatible with `Copy`, but this actually not possible. We can close #2906. (as not planned)*

So I just propose to extend `PrecompileError` enum as in cases where it is feasible, more specific variants should be used in place of the Other() error variant.

### DIFF
--- a/crates/precompile/src/interface.rs
+++ b/crates/precompile/src/interface.rs
@@ -115,7 +115,7 @@ pub trait Crypto: Send + Sync + Debug {
         msg: &[u8; 32],
     ) -> Result<[u8; 32], PrecompileError> {
         crate::secp256k1::ecrecover_bytes(*sig, recid, *msg)
-            .ok_or_else(|| PrecompileError::other("ecrecover failed"))
+            .ok_or(PrecompileError::Secp256k1RecoverFailed)
     }
 
     /// Modular exponentiation.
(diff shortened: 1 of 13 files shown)


rakita (changes requested):
`Fatal` and `Other` PrecompileErrors are needed for users of lib, so they should be brought back.

I do like that this PR adds errors of new precompiles, so we should merge that part in

crates/handler/src/precompile_provider.rs
```diff
@@ -132,7 +131,6 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
                 };
                 result.output = output.bytes;
             }
-            Err(PrecompileError::Fatal(e)) => return Err(e),
```
rakita: Revert this

crates/op-revm/src/precompiles.rs
```diff
@@ -189,29 +186,23 @@ pub mod bls12_381 {
     /// Run the g1 msm precompile with Optimism input limit.
     pub fn run_g1_msm(input: &[u8], gas_limit: u64) -> PrecompileResult {
         if input.len() > ISTHMUS_G1_MSM_MAX_INPUT_SIZE {
-            return Err(PrecompileError::Other(
```
rakita: This file should be reverted back, op-revm errors shouldn't be leaked inside revm.

crates/precompile/src/interface.rs
```diff
@@ -230,18 +230,67 @@ pub enum PrecompileError {
...
+    /// KZG input length error
+    KzgInvalidInputLength,
+    /// secp256k1 ecrecover failed
+    Secp256k1RecoverFailed,
+    /// Isthmus G1 msm input length error
+    IsthmusG1MsmInputLength,
+    /// Isthmus G2 msm input length error
+    IsthmusG2MsmInputLength,
+    /// Isthmus pairing input length error
+    IsthmusPairingInputLength,
```
rakita: Errors leaked from op-revm

examples/custom_precompile_journal/src/precompile_provider.rs
```diff
@@ -113,7 +113,7 @@ fn run_custom_precompile<CTX: ContextTr>(
         // Write storage operation
         handle_write_storage(context, &input_bytes, gas_limit)
     } else {
-        Err(PrecompileError::Other("Invalid input length".to_string()))
+        Err(PrecompileError::InputLength)
```
rakita: Errors from examples shouldn't leak inside revm

crates/precompile/src/interface.rs
```diff
@@ -230,18 +230,67 @@ pub enum PrecompileError {
...
+    /// Custom precompile input length error
+    InputLength,
+    /// Custom precompile storage operation failed
+    StorageOperationFailed,
+    /// Custom precompile balance operation failed
+    BalanceOperationFailed,
+    /// Custom precompile transfer operation failed
+    TransferFailed,
```
rakita: Errors leaked from examples

mablr:
@rakita Thanks for your insightful feedback.

I'll remove the errors which are not supposed to be inside revm (custom/op/examples related). And restore the error variants i deleted.

Given that, making `PrecompileError` `Copy`-compatible isn't possible, so i'll rename this PR as well.

crates/handler/src/precompile_provider.rs
mablr: Just out of curiosity, is this `Fatal` variant for custom precompiles ? I was actually unable to find any use of it in the revm codebase or example.

rakita: It is not used inside revm, but to allow functionality where variant of evm would like to abort the currect transaction.

For example if precompile fetched storage from database and database is rpc that fails, you would like to throw fatal error in this case

mablr: Thanks for the clarification!

PR #2909, opened 2025-08-22
feat!: Remove kzg-rs

### LINKED ISSUE #2831: Removing `kzg-rs` as a dependency
Rationale:

Since we now can use arkworks as a no_std alternative and the kzg-rs usecase for zkEVMs can now be satisfied with https://github.com/bluealloy/revm/pull/2786 -- I believe it is possible to now remove kzg-rs as a dependency.

-------

cc @jtguibas and @0xWOLAND to see if there is anything I'm missing here

kevaundray:
Resolves #2831

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


rakita:
lgtm, will wait for response from sp1 before merging

Issue #2917, opened 2025-08-25
Testing rv64imac-unknown-unknown-elf


kevaundray:
## Motivation

Currently most zkVMs target a variant of rv32 -- we are planning to move to rv64 because this captures a larger amount of languages.

## Ask

Currently the CI only tests for rv32imac, however long term it would be great to test for rv64. In theory, if it compiles for rv32imac, it should compile for rv64imac -- the places where this might not hold are cases where the code is explicitly matching on targets

rakita:
Doesn't hurt to use add, to at least show outsiders that revm compiles to it. There is not a lot of benefit as if revm compiles to riscv32, it will compile riscv64.

PR is merged

PR #2918, opened 2025-08-25
Add rv64imac-unknown-unknown-elf to ci

kevaundray:
#2917

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,22 +32,23 @@ jobs:
       - run: cargo test --workspace ${{ matrix.flags }}
 
   check-no-std:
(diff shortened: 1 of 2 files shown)


kevaundray:
This doesn't add much to the CI so might make sense to pre-emptively add and watch for regressions -- or we can wait for when most zkVMs have migrated

kevaundray:
> It does not hurt to add it as is a quick task, but it does not test anything that riscv32imc does not cover

It would mainly be testing for dependencies that have code like this: https://github.com/ferrilab/radium/blob/cb3ef4b5fd2dd9a1109b86ab54926df8348f695c/build.rs#L92

ie code that matches on specific targets

PR #2934, opened 2025-08-30
chore: added amsterdam in spec id

### LINKED ISSUE #2932: [Feature]: Add Amsterdam variant in SpecId
We have this already confirmed here [Hardfork Meta - Glamsterdam](https://eips.ethereum.org/EIPS/eip-7773)
So as prep for [BAL](https://eips.ethereum.org/EIPS/eip-7928) integration , we should add Amsterdam in the SpecId here 
https://github.com/bluealloy/revm/blob/a8916288952ca65ead1b0fd7aae20341e396b1c6/crates/primitives/src/hardfork.rs#L16-L17

Soubhik-10:
closes #2932

### DIFF
--- a/crates/precompile/src/lib.rs
+++ b/crates/precompile/src/lib.rs
@@ -427,7 +427,7 @@ impl PrecompileSpecId {
             BERLIN | LONDON | ARROW_GLACIER | GRAY_GLACIER | MERGE | SHANGHAI => Self::BERLIN,
             CANCUN => Self::CANCUN,
             PRAGUE => Self::PRAGUE,
-            OSAKA => Self::OSAKA,
+            OSAKA | AMSTERDAM => Self::OSAKA,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/precompile/src/lib.rs
```diff
@@ -427,7 +427,7 @@ impl PrecompileSpecId {
             BERLIN | LONDON | ARROW_GLACIER | GRAY_GLACIER | MERGE | SHANGHAI => Self::BERLIN,
             CANCUN => Self::CANCUN,
             PRAGUE => Self::PRAGUE,
-            OSAKA => Self::OSAKA,
+            OSAKA | AMSTERDAM => Self::OSAKA,
```
mattsse, line 430: should we add a todo here?
but for now this seems okay

rakita: Should be fine without todo and we can add Amsterdam in PrecompileSpecId when EIPs for precompile gets included

rakita (approved):

PR #2935, opened 2025-09-01
feat(revme): ef blockchain tests cli

rakita:
Currently running:
legacytests/Cancun/BlockchainTests/GeneralStateTests and
legacytests/Cancun/BlockchainTests/ValidBlocks/

And checking only the post state, nothing more.

It is run with: `revme blockchaintest path`


rakita:
This PR is mostly a command to revme and some types in revm-statetest-types, it does not touch the main crates in any meaningful way.

Issue #2938, opened 2025-09-02
failed block verification caused by the gas refunding for empty accounts in eip-7702 transaction


silathdiir:
Hi, I found a block verification issue when replays some blocks by reth executor. Seems it's caused by the gas refunding for empty accounts in an eip-7702 transaction.

There's an example code - [reth-executor-example](https://github.com/eason1981/reth-executor-example), this issue could be reproduced as:

- replace [this constant](https://github.com/eason1981/reth-executor-example/blob/main/src/main.rs#L36) with a real RPC URL of mainnet
- build and run
- the [validate_block_post_execution](https://github.com/eason1981/reth-executor-example/blob/main/src/main.rs#L126) call will be failed with an error:
```
called `Result::unwrap()` on an `Err` value: BlockGasUsed { gas: GotExpected { got: 7133691, expected: 7146191 } ...
```

I tried to check the transactions of the failed block 23043259, the gas used of [this transaction](https://etherscan.io/tx/0xd8bf27e11adde9b5eea6799a3e74a9fc6e4429126898243b6b27575da06c5109) is not right, the difference is 12500.

It could be fixed by replacing with the patched revm (tag v82) in [Cargo.toml](https://github.com/eason1981/reth-executor-example/blob/main/Cargo.toml#L57), and fix [this line](https://github.com/bluealloy/revm/blob/main/crates/handler/src/pre_execution.rs#L233) of revm-handler:
```
- if !(authority_acc.is_empty() && authority_acc.is_loaded_as_not_existing_not_touched()) {
+ if !(authority_acc.is_empty() || authority_acc.is_loaded_as_not_existing_not_touched()) {
```

I try to send a straightforward PR to fix it: https://github.com/bluealloy/revm/pull/2939

rakita:
This part looks correct, and reth does sync this block/tx correctly, it would be a big issue if this is not the case.

Second part of this check was added for optimism where they still had some **empty existing** accounts, so checking if it is both empty and loading as not existing and not touched makes sense.

rakita:
Maybe the problem is somewhere else

silathdiir:
I see. Perhaps that empty account is supposed to satisfy `authority_acc.is_loaded_as_not_existing_not_touched()`, but it doesn’t in my example.
I will debug more, thanks.

sandakersmann:
Reth 1.6.0: 

[...]

PR #2939, opened 2025-09-02
fix the eip-7702 refunding condition for empty accounts

silathdiir:
Close https://github.com/bluealloy/revm/issues/2938

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -230,7 +230,7 @@ pub fn apply_eip7702_auth_list<
         }
 
         // 7. Add `PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST` gas to the global refund counter if `authority` exists in the trie.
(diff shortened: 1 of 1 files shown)


rakita (changes requested):
Last condition was added for empty not existing accounts, it does not make sense to change && to ||

sandakersmann:
Reth 1.6.0:

`EVM reported invalid transaction (0xe05a49ce326297d27d83ad15efa4aa773e1a7adc7a242b0c142d165dd3eac610): nonce 0 too low, expected 2`

Probably a bug in Revm.

rakita:
> Reth 1.6.0:
> 
> `EVM reported invalid transaction (0xe05a49ce326297d27d83ad15efa4aa773e1a7adc7a242b0c142d165dd3eac610): nonce 0 too low, expected 2`
> 
> Probably a bug in Revm.

It is not, as this is gas diff it would fail differently, not on nonce check, and tx in question is EIP-1559 not EIP-7702, so not related to this at all. Current bug in reth is merkle hashing related.

rakita:
Will close this PR so that people dont get confused

PR #2940, opened 2025-09-03
Respect SpecId for code and initcode size limits

Forostovec:
Make CfgEnv spec-aware for code and initcode size limits. Use EIP-7907 limits from Osaka onward while preserving explicit overrides and the doubling rule for initcode when code limit is overridden. This aligns implementation with comments, tests, and EIP expectations.


rakita:
That EIP from osaka got dropped, so present code has correct behaviour

PR #2941, opened 2025-09-05
fix: interpreter_result_mut should return mutable reference

Troublor:
I found the `interpreter_result_mut` function of `FrameResult` isdoes not return a mutable reference. Is that an expected implementation? This PR modifies the method to return a mutable reference. 
https://github.com/bluealloy/revm/blob/0ca6564f02004976f533cacf8821fed09d801e0a/crates/handler/src/frame_data.rs#L94


rakita (approved):
lgtm

rakita:
Test is failing for fetching some docker stuff, not related to the PR.

PR #2943, opened 2025-09-06
chore(op-revm): rm redundant phantom

nkysg:
https://github.com/bluealloy/revm/blob/4d247890e00ad62edded3b2d0f08845bd9523063/crates/handler/src/mainnet_handler.rs#L8-L12

MainnetHandler already uses _phantom


rakita (approved):
cool

PR #2946, opened 2025-09-08
chore: remove unused generic from validate_tx_env and fix call site

Forostovec:
The validate_tx_env function declared a second generic parameter (Error) that was never used and the function always returned Result<(), InvalidTransaction>. This created the false impression that the error type was configurable, while the implementation was fixed. The unused generic caused confusion and added no value. This commit removes the extraneous generic and updates the sole call in validate_env to use only ::<CTX>, aligning with the module’s error-handling design where inner validators return InvalidTransaction and the outer layer maps to the handler’s error type.


rakita (approved):
lgtm

PR #2947, opened 2025-09-08
chore: update `EthFrame::invalid` visibility

0xKitsune:
This PR updates `EthFrame::invalid` visibility to `pub` to enable consumers of the lib to utilize this function.

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -68,7 +68,7 @@ impl Default for EthFrame<EthInterpreter> {
 }
 
 impl EthFrame<EthInterpreter> {
-    fn invalid() -> Self {
+    pub fn invalid() -> Self {
         Self::do_default(Interpreter::invalid())
     }
 


crates/handler/src/frame.rs
```diff
@@ -68,7 +68,7 @@ impl Default for EthFrame<EthInterpreter> {
 }
 
 impl EthFrame<EthInterpreter> {
-    fn invalid() -> Self {
+    pub fn invalid() -> Self {
```
mattsse, line 72:
```suggestion
    /// Creates an new invalid [`EthFrame`].
    pub const fn invalid() -> Self {
```

mattsse:
ah sigh ...
maybe clippy is confused

crates/handler/src/frame.rs
```diff
@@ -68,7 +68,8 @@ impl Default for EthFrame<EthInterpreter> {
 }
 
 impl EthFrame<EthInterpreter> {
-    fn invalid() -> Self {
+    /// Creates an new invalid [`EthFrame`].
+    pub const fn invalid() -> Self {
```
rakita:
```suggestion
    pub fn invalid() -> Self {
```

PR #2948, opened 2025-09-08
Avoid panic on non-UTF filenames in statetest runner

prestoalvarez:
### Description
- **Summary**: Replace unsafe `file_name().unwrap().to_str().unwrap()` with a safe check in `skip_test` to prevent panics on non-UTF or missing filenames.
- **Change**: In `bins/revme/src/cmd/statetest/runner.rs`, use `and_then(|s| s.to_str())` with early return (`false`) if the name is invalid. Added a brief comment.
- **Rationale**: CLI should not crash on valid Unix paths with non-UTF names; default behavior is to not skip when the filename is unknown.

### DIFF
--- a/bins/revme/src/cmd/statetest/runner.rs
+++ b/bins/revme/src/cmd/statetest/runner.rs
@@ -86,11 +86,10 @@ pub fn find_all_json_tests(path: &Path) -> Vec<PathBuf> {
 /// Check if a test should be skipped based on its filename
 /// Some tests are known to be problematic or take too long
 fn skip_test(path: &Path) -> bool {
-    let Some(_name) = path.file_name().and_then(|s| s.to_str()) else {
+    let Some(name) = path.file_name().and_then(|s| s.to_str()) else {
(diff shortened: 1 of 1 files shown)


bins/revme/src/cmd/statetest/runner.rs
```diff
@@ -86,6 +86,10 @@ pub fn find_all_json_tests(path: &Path) -> Vec<PathBuf> {
 /// Check if a test should be skipped based on its filename
 /// Some tests are known to be problematic or take too long
 fn skip_test(path: &Path) -> bool {
+    let Some(_name) = path.file_name().and_then(|s| s.to_str()) else {
+        // Non-UTF file names or missing file name: do not skip by default.
+        return false;
+    };
     let name = path.file_name().unwrap().to_str().unwrap();
```
rakita:
```suggestion
    let Some(name) = path.file_name().and_then(|s| s.to_str()) else {
        // Non-UTF file names or missing file name: do not skip by default.
        return false;
    };
```
This should probably work

PR #2950, opened 2025-09-09
chore: remove duplicate comment for TOTAL_COST_FLOOR_PER_TOKEN constant

aso20455:
Remove redundant inline comment that duplicates the doc comment above TOTAL_COST_FLOOR_PER_TOKEN constant. The doc comment already provides the same information about EIP-2028, making the inline comment unnecessary and improving code readability.


rakita (approved):
thanks

PR #2951, opened 2025-09-09
chore(inspector): remove dead skip flag

MozirDmitriy:
The skip flag was never set anywhere and only read in step_end, making it dead code. Removing it simplifies the tracer, avoids confusion about unimplemented behavior, and aligns with EIP-3155 which expects an entry per operation (there’s no need to conditionally skip output).


rakita (approved):
nice find

rakita:
wainting for CI before merging

PR #2953, opened 2025-09-09
Prevent panic in operator fee calculation when unset; default to zero

sashaodessa:
### Summary
Safely handle missing `operator_fee_scalar` and `operator_fee_constant` by returning zero instead of panicking.

### Rationale
Before Isthmus, operator fees should be zero. Using `expect(...)` could cause unexpected panics in valid configurations where those fields are unset.

### Changes
- `crates/op-revm/src/l1block.rs`: Replace `expect(...)` with safe `Option` handling in `operator_fee_charge_inner`, returning `U256::ZERO` if values are missing.


rakita:
It is fine to panic here, better to panic than to return the wrong value

PR #2956, opened 2025-09-10
chore: add `SECURITY.md` 

zerosnacks:
This makes it easier to find for users as it shows up as a `Security` tab

Screenshot from 2025-09-10 12-02-17

If you would prefer to include / change it to `security@ithaca.xyz` this is of course also a possibility


rakita:
This is nice!

Revm does not belong to Ithaca, so it does not make sense to use ithaca email.

PR #2957, opened 2025-09-10
fix(op-revm): clear enveloped_tx for deposit txs in build_fill and align docs

GarmashAlex:
Ensure deposit transactions never carry enveloped bytes when using build_fill:
- If tx_type == DEPOSIT_TRANSACTION_TYPE: enforce source_hash default and set enveloped_tx = None
- If source_hash != B256::ZERO: set tx_type to deposit and set enveloped_tx = None

This matches repository assumptions and tests where non-deposit txs must have enveloped bytes (used for L1 cost/operator fee) while deposit txs should not. Previously the comment stated this, but the code didn’t clear enveloped_tx, leading to possible confusion and inconsistent normalization in build_fill.


rakita (approved):
lgtm

PR #2960, opened 2025-09-11
fix: CTX bound in InspectorEvmTr

mdqst:
adjusted `CTX` so the associated `Journal` type properly implements `JournalExt`.


rakita:
Does not matter

PR #2962, opened 2025-09-11
Fix infinite recursion in EthPrecompiles PrecompileProvider methods

sashaodessa:
### Description
- **What**: Replace self-recursive implementations of `warm_addresses` and `contains` with proper delegation to `precompiles`.
- **Where**: `crates/handler/src/precompile_provider.rs`
- **Why**: Self-recursion caused a potential stack overflow at runtime and made the methods non-functional.

### Changes
- `warm_addresses` now returns `Box::new(self.precompiles.addresses().cloned())`.
- `contains` now calls `self.precompiles.contains(address)`.

### Impact
- Eliminates stack overflow risk and restores correct precompile lookups.
- No behavior change beyond fixing the bug; public API remains the same.


sashaodessa:
@mattsse merge?

PR #2963, opened 2025-09-11
feat: send bytecode with call input

rakita:
Loading of bytecode account can be done before we make a frame so we can skip additional load that was done there


rakita:
~Strange that it degrades by that much~

PR #2964, opened 2025-09-11
fix(bytecode): saturating immediate skip to preserve slice provenance

MozirDmitriy:
Replace unwrap_or_default() in skip_immediate with a saturating re-slice from the same buffer: compute remaining = self.bytes.as_slice(), skip = min(immediate_size, remaining.len()), then set self.bytes = remaining[skip..].iter().
Add a debug_assert to flag broken invariants if immediate_size ever exceeds the remaining length.
Clarify the safety comment in position() to spell out the same-allocation invariant for pointer offset calculation.
These changes are necessary to avoid potential undefined behavior if PUSH immediates were ever truncated: unwrap_or_default() could switch to a different &[] allocation, making position()’s pointer difference cross-allocation. While current analysis pads bytecode to prevent truncation, this hardening makes the iterator safe under future changes and preserves provenance guarantees.

### DIFF
--- a/crates/bytecode/src/iter.rs
+++ b/crates/bytecode/src/iter.rs
@@ -40,7 +40,10 @@ impl<'a> BytecodeIterator<'a> {
     /// Returns the current position in the bytecode.
     #[inline]
     pub fn position(&self) -> usize {
-        // SAFETY: `start` always points to the start of the bytecode.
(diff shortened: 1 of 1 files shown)


rakita:
Current code looks okay

PR #2965, opened 2025-09-12
chore: prealloc few frames

rakita:


crates/context/interface/src/local.rs
```diff
@@ -18,9 +18,22 @@ impl<T> Default for FrameStack<T> {
...
 }
 
+impl<T: Default> FrameStack<T> {
+    /// Creates a new stack with preallocated items by calling `T::default()` `len` times.
+    /// Index will still be `None` until `end_init` is called.
+    pub fn new_prealloc(len: usize) -> Self {
+        let mut s = Self::new();
+        for _ in 0..len {
+            s.stack.push(T::default());
+        }
```
mattsse: isn't this space already allocated via new?

should this instead do `with_capacity(len)`?

rakita: Should add the `with_capacity(len)`, but the main idea here is the call for `T::default()`, which would, for the  Interpreter, allocate 1024 items for the stack.

crates/context/interface/src/local.rs
```diff
@@ -18,9 +18,22 @@ impl<T> Default for FrameStack<T> {
     }
 }
 
+impl<T: Default> FrameStack<T> {
+    /// Creates a new stack with preallocated items by calling `T::default()` `len` times.
+    /// Index will still be `None` until `end_init` is called.
+    pub fn new_prealloc(len: usize) -> Self {
+        let mut stack = Vec::with_capacity(core::cmp::min(len, 4));
```
mattsse: hmm this min doesnt make much sense if we always push len items

rakita: True, it was added as there was assumption that capacity is at least 2. I fixed this by allocating vec in `get_next` and it looks more correct to do.

PR #2966, opened 2025-09-12
chore(database): remove redundant ToString import from EmptyDB

Snezhkko:
The ToString trait is available via the standard prelude under the crate’s default std feature, so importing std::string::ToString is unnecessary. Other modules already call .to_string() without explicit imports. Removing this import avoids redundancy and improves consistency, while also preventing an unconditional std path that could hinder potential no_std portability.


Snezhkko:
shame on me

PR #2971, opened 2025-09-15
Read ETH_RPC_URL in block_traces example; remove hardcoded Infura URL

vtjl10:
### Description
Replaces the hardcoded Infura RPC URL in `examples/block_traces/src/main.rs` with the `ETH_RPC_URL` environment variable and provides a clear error if it’s not set.

- **Why**: improves security (no leaked key), reproducibility, and configurability.
- **Change scope**: example-only; no behavior change beyond config source.
- **Run**
```bash
export ETH_RPC_URL="https://mainnet.infura.io/v3/<PROJECT_ID>"
# or e.g. http://localhost:8545
cargo run -p example-block-traces
```


rakita:
This is example, it is fine to hardcode

PR #2972, opened 2025-09-15
Interpreter: make debug_unreachable safe by default; gate unreachable_unchecked behind feature

vtjl10:
- Replaces UB-prone unreachable_unchecked in release builds with safe unreachable!() by default.
- Introduces optional feature flag unchecked_unreachable to re-enable unreachable_unchecked for peak performance.

- Avoids undefined behavior in production if assumptions are violated, which is critical for an interpreter executing untrusted bytecode.


rakita:
Not necessary

PR #2973, opened 2025-09-15
op-revm: treat empty input as zero operator fee in operator_fee_charge

zeevick10:
### Description
- **Summary**: Align `operator_fee_charge` with its doc comment by returning zero for empty inputs.
- **Rationale**: The function’s comment states that deposit or empty inputs should yield zero fee; previously, only the deposit prefix (0x7E) was handled.
- **Change**: Added `input.is_empty()` to the early-return condition in `l1block.rs`.
- **Impact**: Prevents overcharging operator fees for empty inputs; behavior now matches documentation.


rakita (approved):

rakita:
Although this can't be triggered as all enveloped tx are not empty, it is good to align it with comment

PR #2974, opened 2025-09-15
refactor: remove outdated TODO comment in FrameStack test

forkfury:
Simplify FrameStack test by removing intermediate variable and outdated TODO comment.


rakita:
It is already part of different PR

PR #2975, opened 2025-09-15
feat: add transaction index tracking in transact_many for better debugging

forkfury:
Add transaction index tracking in transact_many method to improve debugging capabilities when multiple transactions fail.

- Use enumerate() to track transaction index
- Add comment indicating index availability for debugging


rakita:
Thank you, will remove TODO in saparate PR

PR #2977, opened 2025-09-16
fix(handler): write CALL output to memory only on success

radik878:
Prevent copying CALL return data into the caller’s output buffer on REVERT. EVM semantics require that the output buffer is written only on successful calls; on failure (including REVERT) the buffer must remain unchanged and return data should be accessed via RETURNDATACOPY. We still erase remaining gas for both success and revert and keep refunds only on success. This aligns with EIP-140/EIP-211 and avoids incorrect memory mutations on reverted subcalls.


rakita:
This is not correct

PR #2980, opened 2025-09-16
Add an option to disable "fee-charge" on `op-revm`

ngotchac:
An option to disable fee charges, only meaningful on `op-revm` is added, which doesn't charge additional fees, in addition from the default fee from the gas limit and gas price.

This is useful for `eth_call`s which should disable all fees.

See https://github.com/paradigmxyz/reth/issues/18470

### DIFF
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -115,7 +115,7 @@ where
         let mut additional_cost = U256::ZERO;
 
         // The L1-cost fee is only computed for Optimism non-deposit transactions.
-        if !is_deposit {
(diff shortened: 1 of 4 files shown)


crates/op-revm/src/handler.rs
```diff
@@ -115,7 +115,7 @@ where
         let mut additional_cost = U256::ZERO;
 
         // The L1-cost fee is only computed for Optimism non-deposit transactions.
-        if !is_deposit {
+        if !is_deposit && !ctx.cfg().is_fee_charge_disabled() {
```
mattsse, line 118: checks out

rakita (approved):
Would introduce configurable flags in the future, but for the sake of moving forward/unblocking this lgtm

PR #2982, opened 2025-09-16
docs: Clarify that legacy bytecode padding is variable, not fixed

gap-editor:
The comment for the `bytecode()` method was misleading, claiming the padding was always 32 bytes. The actual padding is variable (up to 33 bytes).

This change corrects the comment to be more accurate.


rakita:
Thanks for pointing this out

PR #2985, opened 2025-09-17
feat: allow EIP-7623 to be disabled

### LINKED ISSUE #2936: Configurability of EIP-7623 for L2 rollups
Hi,

We are currently using REVM in the context of an EVM compatible L2 rollup. As we prepare to activate the Prague upgrade, we noticed that EIP-7623 doesn’t seem to make much sense in our L2 context.

Is there a way to deactivate this EIP, or could it be made configurable in REVM?

Thanks.

EDIT: Would be happy to contribute if that’s feasible — just need a couple of pointers on the most straightforward way to implement it.

rodibozman:
Closes #2936.

Let me know if I missed something or if you want me to add a line in a CHANGELOG.

(cc @rakita)

### DIFF
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -224,9 +224,8 @@ pub trait Handler {
         // Calculate final refund and add EIP-7702 refund to gas.
         self.refund(evm, exec_result, eip7702_gas_refund);
         // Ensure gas floor is met and minimum floor gas is spent.
-        if !evm.ctx().cfg().is_eip7623_disabled() {
-            self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
-        }
+        // if `cfg.is_eip7623_disabled` is true, floor gas will be set to zero
+        self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
         // Return unused gas to caller
         self.reimburse_caller(evm, exec_result)?;
         // Pay transaction fees to beneficiary
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -224,7 +224,9 @@ pub trait Handler {
         // Calculate final refund and add EIP-7702 refund to gas.
         self.refund(evm, exec_result, eip7702_gas_refund);
         // Ensure gas floor is met and minimum floor gas is spent.
-        self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
+        if !evm.ctx().cfg().is_eip7623_disabled() {
+            self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
+        }
         // Return unused gas to caller
         self.reimburse_caller(evm, exec_result)?;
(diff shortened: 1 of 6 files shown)


crates/handler/src/validation.rs
```diff
@@ -237,7 +238,8 @@ pub fn validate_initial_tx_gas(
 
     // EIP-7623: Increase calldata cost
     // floor gas should be less than gas limit.
-    if spec.is_enabled_in(SpecId::PRAGUE) && gas.floor_gas > tx.gas_limit() {
+    if !is_eip7623_disabled && spec.is_enabled_in(SpecId::PRAGUE) && gas.floor_gas > tx.gas_limit()
```
rodibozman: This might not be strictly necessary since the `floor_gas` will be 0 if EIP-7623 is disabled. I just found it more clear to still add it in the condition.

Let me know if you want me to remove it.

rodibozman:
Hey @mattsse, @rakita, can one of you guys please have a look at this when you have time? Thanks.

crates/interpreter/src/gas/calc.rs
```diff
@@ -395,6 +395,7 @@ pub fn calculate_initial_tx_gas(
     access_list_accounts: u64,
     access_list_storages: u64,
     authorization_list_num: u64,
+    is_eip7623_disabled: bool,
```
rakita: Lets do it differently, and make a change only in the place where this function is called. So we dont need to propagate this flag this deep

rakita: In `validate_initial_tx_gas` function

rodibozman: I reduced the scope of the code.

I had to at least propagate the flag to `validate_initial_tx_gas` from `validate` otherwise if `GasFloorMoreThanGasLimit` is raised I'm loosing the information about the `initial_gas` in `validate_initial_tx_gas` from `handler`.

Let me know if you had something else in mind.

rakita (commented):
lgtm

crates/handler/src/handler.rs
```diff
@@ -224,7 +224,9 @@ pub trait Handler {
         // Calculate final refund and add EIP-7702 refund to gas.
         self.refund(evm, exec_result, eip7702_gas_refund);
         // Ensure gas floor is met and minimum floor gas is spent.
-        self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
+        if !evm.ctx().cfg().is_eip7623_disabled() {
+            self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
+        }
```
rakita:
```suggestion
        // if `cfg.is_eip7623_disabled` is true, floor gas will be set to zero
        self.eip7623_check_gas_floor(evm, exec_result, init_and_floor_gas);
```

Can revert as floor gas is zero

rakita (approved):
waiting for CI, rest lgtm

PR #2986, opened 2025-09-17
fix race in global crypto provider installation using OnceLock::set

vtjl10:
- **Summary**: Replace racy install_crypto logic with atomic OnceLock::set to ensure accurate success reporting under contention.
- **Problem**: install_crypto could return true even if another thread initialized CRYPTO first.
- **Solution**: Use OnceLock::set(...).is_ok(); add set to the no_std OnceLock wrapper for consistent API.
- **Changes**:
  - primitives: add OnceLock::set in no_std wrapper; std path delegates to std::sync::OnceLock::set
  - precompile/interface: rewrite install_crypto to return CRYPTO.set(...).is_ok()

### DIFF
--- a/crates/primitives/src/once_lock.rs
+++ b/crates/primitives/src/once_lock.rs
@@ -41,7 +41,7 @@ mod no_std_impl {
         pub fn get(&self) -> Option<&T> {
             self.inner.get()
         }
-    
+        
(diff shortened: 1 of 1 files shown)


rakita:
Superseded by #2997 thank you for finding it

PR #2990, opened 2025-09-17
fix: remove dummy authorization when building eip-7702 txs

### LINKED ISSUE #2989: Inconsistency between validation and transaction building
Hey,

By looking at [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702):

> Non-empty authorization list required
Set code transactions are required to have at least one authorization to be considered valid. This is to disincentivize senders from using type 4 transactions as a generic transaction format, because this transaction has different implications on the transaction pool than, say, [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) transactions.

in `crates/handler/src/validation.rs`:

```rust
            let auth_list_len = tx.authorization_list_len();
            // The transaction is considered invalid if the length of authorization_list is zero.
            if auth_list_len == 0 {
                return Err(InvalidTransaction::EmptyAuthorizationList);
            }
```

whereas in `crates/context/src/tx.rs`:

```rust
                    // authorization list can be empty
                    if self.authorization_list.is_empty() {
                        // add dummy authorization
                        self.authorization_list =
                            vec![Either::Right(RecoveredAuthorization::new_unchecked(
                                Authorization {
                                    chain_id: U256::from(self.chain_id.unwrap_or(1)),
                                    address: self.caller,
                                    nonce: self.nonce,
                                },
                                RecoveredAuthority::Invalid,
                            ))];
                    }
```

It seems the inconsistency comes from the transaction building. I can propose a quickfix to remove the code in `tx.rs` and let the validation raise the issue if the authorization list is empty as intended.

rodibozman:
Closes #2989.

### DIFF
--- a/crates/context/src/tx.rs
+++ b/crates/context/src/tx.rs
@@ -3,8 +3,7 @@ use crate::TransactionType;
 use context_interface::{
     either::Either,
     transaction::{
-        AccessList, AccessListItem, Authorization, RecoveredAuthority, RecoveredAuthorization,
-        SignedAuthorization, Transaction,
+        AccessList, AccessListItem, RecoveredAuthorization, SignedAuthorization, Transaction,
     },
 };
 use core::fmt::Debug;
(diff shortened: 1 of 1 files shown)


mattsse (commented):
imo most of these defaults are a bit problematic

rodibozman:
Closing. See https://github.com/bluealloy/revm/issues/2989#issuecomment-3303606420.

PR #2991, opened 2025-09-17
chore: refactor validate_against_state_and_deduct_caller

rakita:


crates/handler/src/pre_execution.rs
```diff
@@ -106,73 +125,84 @@ pub fn validate_account_nonce_and_code(
...
+
+    let new_balance =
+        deduct_caller_balance_with_components(caller_account.info.balance, tx, block, cfg)?;
+
+    // make changes to the account
+    let old_balance = caller_touch_and_change(caller_account, new_balance, tx.kind().is_call());
+
+    // journal the change
     journal.caller_accounting_journal_entry(tx.caller(), old_balance, tx.kind().is_call());
+
```
rakita, line 205: @klkvr @mattsse can you take a look? This function is a lot simpler than before as few things got abstracted with functions

examples/erc20_gas/src/handler.rs
```diff
@@ -49,77 +49,33 @@ where
...
+        // validate account nonce and code
+        validate_account_nonce_and_code_with_components(&mut caller_account.info, tx, cfg)?;
 
-        if tx.kind().is_call() {
-            caller_account.info.nonce = caller_account.info.nonce.saturating_add(1);
-        }
-
-        // Touch account so we know it is changed.
-        caller_account.mark_touch();
+        let new_balance = deduct_caller_balance_with_components(account_balance, tx, block, cfg)?;
```
rakita, line 68: All max and effective balance deduction and checks are wrapped inside this function

examples/erc20_gas/src/handler.rs
```diff
@@ -49,77 +49,33 @@ where
...
-            tx.nonce(),
-            is_eip3607_disabled,
-            is_nonce_check_disabled,
-        )?;
+        // validate account nonce and code
+        validate_account_nonce_and_code_with_components(&mut caller_account.info, tx, cfg)?;
 
-        if tx.kind().is_call() {
-            caller_account.info.nonce = caller_account.info.nonce.saturating_add(1);
-        }
```
rakita, line 75: Touching of account and nonce bumps are now inside `caller_touch_and_change`

examples/erc20_gas/src/handler.rs
```diff
@@ -49,77 +49,33 @@ where
...
     fn validate_against_state_and_deduct_caller(&self, evm: &mut Self::Evm) -> Result<(), ERROR> {
-        let context = evm.ctx();
-        let basefee = context.block().basefee() as u128;
-        let blob_price = context.block().blob_gasprice().unwrap_or_default();
-        let is_balance_check_disabled = context.cfg().is_balance_check_disabled();
-        let is_eip3607_disabled = context.cfg().is_eip3607_disabled();
-        let is_nonce_check_disabled = context.cfg().is_nonce_check_disabled();
-        let caller = context.tx().caller();
-        let value = context.tx().value();
+        let (tx, block, cfg, journal) = evm.ctx_mut().tx_block_cfg_journal_mut();
```
rakita, line 52: No need for temp fields, and we can use `tx_block_cfg_journal_mut` to get all needed context objects.

crates/op-revm/src/l1block.rs
```diff
@@ -260,6 +260,33 @@ impl L1BlockInfo {
         tx_l1_cost
     }
 
+    /// Calculate additional transaction cost with [OpTxTr].
+    #[inline]
+    pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> U256 {
```
mattsse, line 266: can we use track_caller here, because this panics and mention that

crates/op-revm/src/handler.rs
```diff
@@ -95,121 +96,74 @@ where
...
 
-        // subtracting max balance spending with value that is going to be deducted later in the call.
-        let gas_balance_spending = effective_balance_spending - tx.value();
+        let caller_account = journal.load_account_code(tx.caller())?.data;
 
-        // If the transaction is not a deposit transaction, subtract the L1 data fee from the
-        // caller's balance directly after minting the requested amount of ETH.
-        // Additionally deduct the operator fee from the caller's account.
-        //
-        // In case of deposit additional cost will be zero.
```
mattsse, line 194: I feel like this is useful context, that we now no longer have?

crates/context/interface/src/transaction.rs
```diff
@@ -179,6 +180,21 @@ pub trait Transaction {
         Ok(max_balance_spending)
     }
 
+    /// Checks if the caller has enough balance to cover the maximum balance spending of this transaction.
+    ///
+    /// Internally calls [`Self::max_balance_spending`] and checks if the balance is enough.
+    #[inline]
+    fn has_enough_balance(&self, balance: U256) -> Result<(), InvalidTransaction> {
```
mattsse: has_ usually indicates a bool response, this is more like `ensure_`

crates/op-revm/src/handler.rs
rakita: This comment became obsolete, as flows for deposit and ordinary transactions are now split for clarity.

And it is not fully correct, mint is only available for deposit transactions. And a deposit transaction does not have additional fee (l1 cost and operator fee)

Would assume the intention for this is to mint first, then to deduct the effective gas.

crates/context/interface/src/transaction.rs
rakita: `ensure_` does sounds nicer

rakita:
last PR to chunk all changes is here https://github.com/bluealloy/revm/pull/3069

All of rest are merged

PR #2992, opened 2025-09-18
feat:  Introduced `all_mut` and `all` functions to ContextTr

### LINKED ISSUE #2988: Idea for ContextTr, create a `all_mut` and `all` functions
We have a lot of specialisations for getters, as for some use cases, you would need both a mutable journal and a reference to tx.

So having a `fn all(&self) ->  (&Tx, &Block, &Journal)` and `fn all_mut(&mut self) -> (&Tx, &Block, &Cfg, &mut Journal, &mut Local, &mut Chain...)  would allow for only those two functions to be required to be implemented and all of rest to have default implementation.

This allows a smaller footprint when implementing ContextTr on struct. And in case that some function pair is not present `tx_cfg_journal_mut`, `all_mut` can be used in its place.

developeruche:
closes #2988

### DIFF
--- a/crates/context/interface/src/context.rs
+++ b/crates/context/interface/src/context.rs
@@ -30,23 +30,47 @@ pub trait ContextTr: Host {
     type Chain;
     /// Local context type
     type Local: LocalContextTr;
-
+    
+    /// Get all contexts
+    fn all(&self) -> (&Self::Block, &Self::Tx, &Self::Cfg, &Self::Db, &Self::Journal, &Self::Chain, &Self::Local);
+    
+    /// Get all contexts mutably
+    fn all_mut(&mut self) -> (&mut Self::Block, &mut Self::Tx, &mut Self::Cfg, &mut Self::Journal, &mut Self::Chain, &mut Self::Local);
+    
     /// Get the transaction
-    fn tx(&self) -> &Self::Tx;
+    fn tx(&self) -> &Self::Tx {
+        let (_, tx, _, _, _, _, _) = self.all();
+        tx
+    }
     /// Get the block
-    fn block(&self) -> &Self::Block;
+    fn block(&self) -> &Self::Block {
+        let (block, _, _, _, _, _, _) = self.all();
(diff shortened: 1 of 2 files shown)


developeruche:
Couldn't add `db_mut` to `all_mut` because of the shared mutable ref to `journaled_state`, settle for a default impl

rakita:
> Couldn't add db_mut to all_mut because of the shared mutable ref to journaled_state, settle for a default impl

Makes sense, db is part of journal

crates/context/interface/src/context.rs
```diff
@@ -31,40 +32,98 @@ pub trait ContextTr: Host {
...
+        &Self::Local,
+    );
+
+    /// Get all contexts mutably
+    fn all_mut(
+        &mut self,
+    ) -> (
+        &mut Self::Block,
+        &mut Self::Tx,
+        &mut Self::Cfg,
```
rakita: Those three should never be changed in execution. So we should return
```suggestion
        &Self::Block,
        &Self::Tx,
        &Self::Cfg,
```

rakita (changes requested):
one nit

crates/context/interface/src/context.rs
rakita: @developeruche, not to remove them, but to make a ref in place of mut.

And make `all` and `all_mut` have `#[inline]

developeruche: Oh, alright... it felt redundant since the reference version was already available using `all`.

would fix it now!

rakita: There is a need to read something from ref tx and set it inside mut journal, so for this usecase it needs to be returned.

crates/context/src/context.rs
```diff
@@ -56,63 +56,47 @@ impl<
     type Local = LOCAL;
 
     #[inline]
```
developeruche, line 58: #[inline] have been applied

PR #2993, opened 2025-09-18
refactor: revamp tx build

rodibozman:
I'm currently running some state transitions tests on my implementation of REVM's interface to run transactions.

One test is expecting a transaction to be rejected if the authorization list is empty in an EIP-7702 transaction. While reading the code of `tx.rs`, in particular `build`. I was thinking that reorganising the transaction building and the condition could make sense as it allows to catch error earlier at building phase instead of validation **IF** the tx type is missing.
In my implementation, I don’t have direct access to the transaction type, so if I first derive it from the fields and then apply the checks, it will allow me to reject invalid transactions earlier.
This is just a proposition I wanted to have your thoughts on that.

It also allowed me to spot a missing field in a sanity test for EIP-7702 transactions in an early stage at tx building, see 0ffad112cab0233584c3f684577e25ae87636db8.

A bit more context: the state transition tests I'm running are from [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests) which don't explicit the tx type in the generated fixtures. So I'd need to derive the tx type prior to the tx build to benefit from that early exit.

### DIFF
--- a/crates/context/src/tx.rs
+++ b/crates/context/src/tx.rs
@@ -396,68 +396,7 @@ impl TxEnvBuilder {
     }
 
     /// Build the final [`TxEnv`] with default values for missing fields.
-    pub fn build_fill(mut self) -> TxEnv {
-        if let Some(tx_type) = self.tx_type {
-            match TransactionType::from(tx_type) {
-                TransactionType::Legacy => {
-                    // do nothing
-                }
-                TransactionType::Eip2930 => {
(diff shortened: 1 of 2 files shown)


rodibozman:
I think we could also simplify `derive_tx_type` as some checks should be redundant with what's done inside `build` if tx type wasn't set.

Waiting for your thoughts first. Let me know if I missed something.

rakita:
I think this is not worth doing,

PR #2994, opened 2025-09-18
Set l2_block in try_fetch for pre-Isthmus forks; add reload tests

Snezhkko:
Set l2_block in L1BlockInfo::try_fetch for Bedrock/Regolith and Ecotone (pre-Isthmus) branches to align caching/reload semantics with Isthmus.
Reason: OpHandler checks ctx.chain().l2_block != block_number to decide whether to reload. When l2_block was left at U256::ZERO pre-Isthmus, the handler reloaded on every non-deposit tx within the same block, defeating the intended cache and causing redundant DB fetches.
This change ensures consistent behavior across forks, reduces unnecessary DB reads, and matches the documented intent (“reloaded if not current block”).
Add tests mirroring the Isthmus case to verify reload behavior for Regolith and Ecotone pre-Isthmus: 
- test_reload_l1_block_info_regolith
- test_reload_l1_block_info_ecotone_pre_isthmus

### DIFF
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -787,6 +787,124 @@ mod tests {
     }
 
     #[test]
+    fn test_reload_l1_block_info_regolith() {
+        const BLOCK_NUM: U256 = uint!(200_U256);
+        const L1_BASE_FEE: U256 = uint!(7_U256);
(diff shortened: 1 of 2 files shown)


rakita (approved):
This makes sense, good find.

PR #2998, opened 2025-09-19
Feat: Add Str(Cow<'static, str>) to InvalidTransaction error enum

### LINKED ISSUE #2995: Add Str(&'str) to InvalidTransaction error enum
Having a `'str` error is helpful for custom errors without having dealocation/allocation of string inside enum.

developeruche:
Closes #2995


rakita (approved):
regresion is not related to the change

PR #3000, opened 2025-09-19
feat: add transaction index to batch execution error handling

crStiv:
Add `TransactionIndexedError` type to include transaction index when batch transaction execution fails.

**Changes:**
- Add `TransactionIndexedError<Error>` struct in `context/interface/src/result.rs`
- Update `transact_many`, `transact_many_finalize`, and `transact_many_commit` methods to return indexed errors
- Add comprehensive tests for the new functionality
- Remove TODO comment about adding tx index to error


rakita (approved):
Makes sense and change lgtm. Will include it after next release

rakita (approved):
lgtm

PR #3007, opened 2025-09-23
fix: add missing is_fee_charge_disabled check

mattsse:
https://github.com/bluealloy/revm/pull/2980 followup that also adds the check to the reimbursment.

and also only applies this after loading the l1info for good measure, just in case

cc @ngotchac 

## TODO

* this needs some sanity tests


ngotchac:
Arf, sorry I missed that! Thx for the fix

PR #3008, opened 2025-09-23
Use short-circuit logical operators in U256 conversion macros

viktorking7:
## Description
- **What**: Replace bitwise boolean ops with logical `&&`/`||` in `as_u64_saturated!` and `as_usize_or_fail_ret!` within `crates/interpreter/src/instructions/macros.rs`. Added brief comments for clarity.
- **Why**: Improves readability and uses short-circuit semantics; no behavioral changes.
- **Impact**: No public API changes; negligible micro-optimization; existing tests remain valid.


rakita:
The current code is better, reasons against impact are not good enough

PR #3009, opened 2025-09-24
feat: Flush out usage of custom frame, example in MyEvm

### LINKED ISSUE #2984: Flush out usage of custom frame, make an example
Make a example of CustomFrame and flush out this use case in revm

rakita:
Closes #2984 

`InspectorEvmTr` now depends on `InspectorFrame` and not on EthFrame

### DIFF
--- a/crates/context/interface/src/result.rs
+++ b/crates/context/interface/src/result.rs
@@ -646,6 +646,54 @@ pub enum OutOfGasError {
     ReentrancySentry,
 }
 
+/// Error that includes transaction index for batch transaction processing.
+#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


crates/inspector/src/traits.rs
```diff
@@ -18,28 +18,62 @@ use crate::{
 /// It is used inside [`crate::InspectorHandler`] to extend evm with support for inspection.
 pub trait InspectorEvmTr:
     EvmTr<
-    Frame = EthFrame<EthInterpreter>,
+    Frame: InspectorFrameTr<IT = EthInterpreter>,
```
rakita, line 21: Removes dependency on `EthFrame` and introduces `InspectorFrameTr` that has `fn eth_frame() -> Option<&EthFrame>`. cc @klkvr @mattsse

rakita:
Superseeded by linked PRs

PR #3011, opened 2025-09-25
Safe handling of legacytests repo in run-tests fixtures setup

vtjl10:
### Summary
Ensure reliable cloning/updating of the `legacytests` repo during fixtures setup in `scripts/run-tests.sh`.

### Motivation
Previously, `git clone` targeted a pre-created directory, which could fail due to Git’s requirement for a non-existing destination, breaking the test run.

### Changes
- Do not pre-create `LEGACY_DIR`.
- If `LEGACY_DIR` is a Git repo: fetch with depth 1 and hard reset to `FETCH_HEAD`.
- If `LEGACY_DIR` is a non-Git directory: remove and re-clone.
- If `LEGACY_DIR` does not exist: clone with `--depth 1`.
- Added clear English comments.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -101,7 +101,7 @@ download_and_extract() {
 # Download all fixtures
 download_fixtures() {
     echo "Creating fixtures directory structure..."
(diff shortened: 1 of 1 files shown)


rakita:
Not accepting AI commits

PR #3015, opened 2025-09-26
Fix division by zero handling in arithmetic operations

maradini77:
## Description

Fixed incorrect behavior when dividing by zero in `DIV` and `MOD` instructions. Previously, when the divisor was zero, the operations would leave the original dividend value on the stack instead of returning zero as specified by the EVM specification.

### Changes
- **`div` function**: Now returns `U256::ZERO` when dividing by zero
- **`rem` function**: Now returns `U256::ZERO` when taking modulo by zero

**Files modified:**
- `crates/interpreter/src/instructions/arithmetic.rs`


rakita:
No bug there

PR #3021, opened 2025-09-28
Gate OP-specific payouts in reward_beneficiary when fee charging disabled

sashass1315:
- This change adds a cfg.is_fee_charge_disabled() check inside reward_beneficiary to skip Optimism-specific payouts (L1 fee vault, base fee vault, operator fee) when fee charging is disabled, e.g., during eth_call simulations on OP chains.
- The base (mainnet) beneficiary reward remains intact, preserving standard EVM tip behavior.
- This aligns with existing gating in pre-execution deduction and operator fee refund, preventing unintended minting to OP fee recipients when no OP fees are being charged.

### DIFF
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -341,6 +341,10 @@ where
         }
 
         self.mainnet.reward_beneficiary(evm, frame_result)?;
+        // If fee charging is disabled (e.g., eth_call simulations), skip OP-specific payouts.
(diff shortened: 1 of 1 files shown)


rakita:
Although it does make sense, the biggest usage of `is_fee_charge_disabled` is to not fail the transaction when the account does not have enough balance. For users they dont care for benficiary rewards. Same logic is aplied for balance checks so i will leave it as it is, at least for now

PR #3028, opened 2025-09-30
fix(interpreter): remove redundant stack underflow check in LOG instruction

Galoretka:
This change removes a duplicate stack length check before calling popn::<N>() in the LOG instruction and uses halt_underflow() in the None arm for consistency. The explicit len() < N guard is unnecessary because popn::<N>() already performs an underflow-safe check and returns None when insufficient items are present. Keeping both checks adds duplication without improving safety and can confuse future maintenance.

### DIFF
--- a/crates/interpreter/src/instructions/host.rs
+++ b/crates/interpreter/src/instructions/host.rs
@@ -367,10 +367,6 @@ pub fn log<const N: usize, H: Host + ?Sized>(
         resize_memory!(context.interpreter, offset, len);
         Bytes::copy_from_slice(context.interpreter.memory.slice_len(offset, len).as_ref())
     };
-    if context.interpreter.stack.len() < N {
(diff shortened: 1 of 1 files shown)


crates/interpreter/src/instructions/host.rs
```diff
@@ -372,7 +372,7 @@ pub fn log<const N: usize, H: Host + ?Sized>(
         return;
     }
     let Some(topics) = context.interpreter.stack.popn::<N>() else {
-        context.interpreter.halt(InstructionResult::StackUnderflow);
+        context.interpreter.halt_underflow();
```
rakita, line 371: Redundant check from line 370 was not removed.

Galoretka: > Redundant check from line 370 was not removed.

Yes, sorry , corrected

PR #3030, opened 2025-09-30
chore: helper function effective_balance_spending_without_value

rakita:
Helper function as we have three places where value needs to be deducted. 

Exctracted from https://github.com/bluealloy/revm/pull/2991

### DIFF
--- a/crates/context/interface/src/transaction.rs
+++ b/crates/context/interface/src/transaction.rs
@@ -18,6 +18,7 @@ use auto_impl::auto_impl;
 use core::cmp::min;
 use core::fmt::Debug;
 use primitives::{eip4844::GAS_PER_BLOB, Address, Bytes, TxKind, B256, U256};
+use std::boxed::Box;
 
 /// Transaction validity error types.
 pub trait TransactionError: Debug + core::error::Error {}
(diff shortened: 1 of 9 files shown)


crates/context/interface/src/transaction.rs
```diff
@@ -210,4 +211,16 @@ pub trait Transaction {
 
         Ok(effective_balance_spending)
     }
+
+    /// Returns the effective balance calculated with [`Self::effective_balance_spending`] but without the value.
```
mattsse, line 234: could we clarify a bit more that this deducts the tx value and is

```
gas_limit * max_fee + additional_gas_cost
```

crates/context/interface/src/transaction.rs
```diff
@@ -210,4 +211,16 @@ pub trait Transaction {
...
+    #[inline]
+    fn effective_balance_spending_without_value(
+        &self,
+        base_fee: u128,
+        blob_price: u128,
+    ) -> Result<U256, InvalidTransaction> {
+        Ok(self
+            .effective_balance_spending(base_fee, blob_price)?
+            .saturating_sub(self.value()))
```
mattsse: this now uses saturating sub, which seems okay because this is expected to be called after the max_balance_spending check but we could highlight this perhaps?

mattsse (commented):
makes sense to me, some doc nits

crates/context/interface/src/transaction.rs
rakita: saturating_sub is not needed here, as `effective_balance_spending` adds the value, and we are deducting it here. It is not related to `max_balance_spending`

crates/context/interface/src/transaction.rs
```diff
@@ -210,4 +214,18 @@ pub trait Transaction {
...
         Ok(effective_balance_spending)
     }
+
+    /// Returns the effective balance calculated with [`Self::effective_balance_spending`] but without the value.
+    ///
+    /// Effective balance is always strictly less than [`Self::max_balance_spending`].
+    ///
+    /// This functions returns `tx.effective_gas_price * tx.gas_limit + blob_price * tx.total_blob_gas`.
+    #[inline]
+    fn effective_balance_spending_without_value(
```
klkvr: should this be called `gas_balance_spending`?

rakita: It feels better

rakita:
`I'm personally fine with keeping this duplicated` what did you mean by duplicated?

klkvr:
> I'm personally fine with keeping this duplicated what did you mean by duplicated?

I meant that this is just a single line of code deduplicating which might not be worth the indirection and a new trait method

rakita:
I see, as it is a trait method with a default implementation, it will not require reimplementation it. For this case it made sense to add it.

It does replace fetching two values from the same place `tx.effective_*() - tx.value` with one fetch `tx.gas_balance_*()`

PR #3032, opened 2025-09-30
chore: helper caller_touch_and_change added

rakita:
Refactor account changed from `validate_against_state_and_deduct_caller` into one function. Split from PR https://github.com/bluealloy/revm/pull/2991

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -152,7 +152,7 @@ pub fn validate_against_state_and_deduct_caller<
         new_balance = new_balance.max(tx.value());
     }
 
-    let old_balance = caller_account.caller_touch_and_change(new_balance, tx.kind().is_call());
+    let old_balance = caller_account.caller_initial_modification(new_balance, tx.kind().is_call());
 
     journal.caller_accounting_journal_entry(tx.caller(), old_balance, tx.kind().is_call());
     Ok(())
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -199,7 +199,8 @@ where
             new_balance = new_balance.max(tx.value());
         }
 
-        let old_balance = caller_account.caller_touch_and_change(new_balance, tx.kind().is_call());
+        let old_balance =
+            caller_account.caller_initial_modification(new_balance, tx.kind().is_call());
 
         // NOTE: all changes to the caller account should journaled so in case of error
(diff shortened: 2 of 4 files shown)


crates/handler/src/pre_execution.rs
```diff
@@ -106,6 +106,27 @@ pub fn validate_account_nonce_and_code(
     Ok(())
 }
 
+/// Make changes to the caller account.
+///
+/// Returns the old balance.
+#[inline]
+pub fn caller_touch_and_change(
+    caller_account: &mut Account,
```
mattsse: should this be moved into `impl Account`?

crates/handler/src/pre_execution.rs
```diff
@@ -106,6 +106,27 @@ pub fn validate_account_nonce_and_code(
     Ok(())
 }
 
+/// Make changes to the caller account.
+///
+/// Returns the old balance.
+#[inline]
+pub fn caller_touch_and_change(
```
mattsse: this should also mention that this bumps the nonce if it is a call

mattsse (commented):
makes sense, one idea and doc nit

crates/handler/src/pre_execution.rs
rakita: Yes

crates/handler/src/pre_execution.rs
```diff
@@ -161,16 +161,7 @@ pub fn validate_against_state_and_deduct_caller<
...
-    let old_balance = caller_account.info.balance;
-    // Touch account so we know it is changed.
-    caller_account.mark_touch();
-    caller_account.info.balance = new_balance;
-
-    // Bump the nonce for calls. Nonce for CREATE will be bumped in `make_create_frame`.
-    if tx.kind().is_call() {
-        // Nonce is already checked
-        caller_account.info.nonce = caller_account.info.nonce.saturating_add(1);
-    }
```
klkvr, line 164: wondering if we could bump `call` nonces in `make_call_frame` too? right now the flow is not very intuitive and moving it into `caller_touch_and_change` just makes it harder to reason about

rakita: Caller nonce is not bumped for CALL/STATICCALL/etc, while the caller of CREATE/CREATE opcode is bumped. That is why nonce bump is here.

In future i would even like to add the value deduction here, but this requires changing more senstive code, so I am delaying it for later.

PR #3033, opened 2025-09-30
chore: add ensure_enough_balance helper

rakita:
Helper function for `ensure_enough_balance` for Transaction that compares balance with max amount that tx could spend.

Extracted from https://github.com/bluealloy/revm/pull/2991

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -16,7 +16,6 @@ use core::cmp::Ordering;
 use primitives::StorageKey;
 use primitives::{eip7702, hardfork::SpecId, KECCAK_EMPTY, U256};
 use state::AccountInfo;
-use std::boxed::Box;
 
 /// Loads and warms accounts for execution, including precompiles and access list.
 pub fn load_accounts<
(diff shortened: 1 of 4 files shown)


crates/handler/src/pre_execution.rs
```diff
@@ -132,16 +131,8 @@ pub fn validate_against_state_and_deduct_caller<
         is_nonce_check_disabled,
     )?;
 
-    let max_balance_spending = tx.max_balance_spending()?;
-
-    // Check if account has enough balance for `gas_limit * max_fee`` and value transfer.
```
mattsse, line 137: can we keep this doc?

crates/op-revm/src/handler.rs
```diff
@@ -170,14 +168,16 @@ where
...
-            // skip max balance check for deposit transactions.
-            // this check for deposit was skipped previously in `validate_tx_against_state` function
-            return Err(InvalidTransaction::LackOfFundForMaxFee {
-                fee: Box::new(max_balance_spending),
-                balance: Box::new(new_balance),
-            }
-            .into());
+        if !is_deposit && !is_balance_check_disabled {
+            // check additional cost and deduct it from the caller's balances
+            let Some(balance) = new_balance.checked_sub(additional_cost) else {
```
mattsse, line 173: ah so this was moved inside here replacing `max_balance_spending > new_balance` ?

crates/op-revm/src/handler.rs
```diff
@@ -158,8 +158,6 @@ where
             )?;
         }
 
-        let max_balance_spending = tx.max_balance_spending()?.saturating_add(additional_cost);
```
mattsse, line 161: we no longer chheck that this doesn't overflow for all txs, is this right?

mattsse (commented):
makes sense but unsure about the op changes

crates/op-revm/src/handler.rs
rakita: Yes, it was never checked. Deposit tx skips this check.

crates/op-revm/src/handler.rs
rakita: Yes, the deposit and the rest of the transaction overlapped a lot, and it wasn't very clear to look.

`max_balance_spending > new_balance` is now done in `ensure_enough_balance` and additional_cost is checked first before `ensure_` is called

rakita: And `additional_cost` is not applicable for deposits, it is zero for it

PR #3034, opened 2025-10-01
fix: сorrect Homestead DAO spec mapping

Bilogweb3:
Map SpecName::HomesteadToDaoAt5 to SpecId::DAO_FORK so DAO fork statetests run with the intended hardfork rules.


Bilogweb3:
Updated to use HOMESTEAD as suggested, thanks for the review

rakita:
PR not needed

PR #3040, opened 2025-10-02
chore: helper calculate_caller_fee

rakita:
A lot of cleanup for mainet and erc20 example. op-revm requires spliting deposit and other txs flows


klkvr (approved):
nice

PR #3041, opened 2025-10-02
chore(op): split paths for deposit tx in caller deduction

rakita:
Change is a lot smaller than it is shown in the diff. In split diff it is a little bit better.

Flow for deposit is extracted and put behind one `if is_deposit` rest of code is not touched, and the only change is the removal of `is_deposit` from condition checks

This is a precondition to add other helper functions to ordinary tx checks.

### DIFF
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -152,11 +152,12 @@ where
 
         if !ctx.cfg().is_fee_charge_disabled() {
             // account for additional cost of l1 fee and operator fee
-            let enveloped_tx = ctx
-                .tx()
-                .enveloped_tx()
-                .expect("all not deposit tx have enveloped tx")
-                .clone();
+            // account for additional cost of l1 fee and operator fee
+            let Some(enveloped_tx) = ctx.tx().enveloped_tx().cloned() else {
+                return Err(ERROR::from_string(
+                    "[OPTIMISM] Failed to load enveloped transaction.".into(),
+                ));
+            };
 
             // compute L1 cost
             additional_cost = ctx.chain_mut().calculate_tx_l1_cost(&enveloped_tx, spec);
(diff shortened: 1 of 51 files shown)


crates/op-revm/src/handler.rs
```diff
@@ -106,66 +106,88 @@ where
...
-                        ctx.chain()
-                            .operator_fee_charge(&enveloped_tx, gas_limit, spec);
-                    additional_cost = additional_cost.saturating_add(operator_fee_charge);
-                }
+            let old_balance =
+                caller_account.caller_initial_modification(new_balance, tx.kind().is_call());
+
+            // NOTE: all changes to the caller account should journaled so in case of error
+            // we can revert the changes.
+            journal.caller_accounting_journal_entry(tx.caller(), old_balance, tx.kind().is_call());
```
klkvr, line 140: can `caller_initial_modification` ever be called without `caller_accounting_journal_entry`?

the naming here is very confusing now and it's hard to understand what methods are doing without reading through them

klkvr (approved):
this seems equivalent, and we now need to maintain 2 codepaths which should be ok

crates/op-revm/src/handler.rs
rakita: I am unsure what to do here, to make it more intuitive.

```
journal.set_balance(caller, balance)
journal.bump(caller)
journal.touch(caller)
```
Would be nicer. But we already have a `caller_account` loaded so we need to directly use it,  but than again, we need to journal this change inside journal.

The reason why it is behind `caller_*` functions is for reusability.

Maybe we could go in a general direction and add `account.set_balance(journal, new_balance)`, `account.bump_nonce(journal)`.

rakita: I will make an additional PR for this to unblock this PR, and it will be easier to discuss

PR #3042, opened 2025-10-02
fix: silent walkdir error handling in test discovery

viktorking7:
## Description

Replace silent `.ok()` error swallowing with explicit error logging in `find_all_json_tests()` to improve debugging visibility when filesystem traversal fails.

**Changes:**
- Log walkdir errors to stderr instead of silently ignoring them
- Maintain existing behavior for test file collection

**Impact:**
- Better debugging experience when directory access issues occur

Fixes potential silent failures that could make test runs appear to "lose" test files without explanation.


rakita:
It is okay for it to silently skip errors on reading files

PR #3044, opened 2025-10-02
chore: remove redundant Vec import in handler API

reject-i:
Removed the redundant Vec import from crates/handler/src/api.rs. Inlined the fully qualified std::vec::Vec path so both std and no_std builds succeed.


rakita:
Lol, no

PR #3045, opened 2025-10-03
refactor(handler): extract duplicate gas price validation

RyanBGG:
Eliminates duplicate basefee validation logic in Legacy and EIP-2930 transaction branches by extracting it into a `validate_legacy_gas_price` helper function.


crates/handler/src/validation.rs
```diff
@@ -23,6 +23,20 @@ pub fn validate_env<CTX: ContextTr, ERROR: From<InvalidHeader> + From<InvalidTra
     validate_tx_env::<CTX>(context, spec).map_err(Into::into)
 }
 
+/// Validate legacy transaction gas price against basefee.
+pub fn validate_legacy_gas_price(
```
rakita, line 28:
```suggestion
#[inline]
pub fn validate_legacy_gas_price(
```

PR #3051, opened 2025-10-05
perf: use touch_account() in caller_accounting_journal_entry()

chen4903:
Optimize `caller_accounting_journal_entry()`, we use the same structure: uniformly set the account to touched state via `touch_account()`, without manually setting it in `caller_initial_modification()`.


rakita:
Doesn't make sense

PR #3052, opened 2025-10-05
chore: check whether the new created account is precompile contract

chen4903:


rakita:
Doesn't make sense

chen4903:
> Doesn't make sense

I am sorry. But could I know where do we check "[Account is not precompile](https://github.com/bluealloy/revm/pull/3052/files#diff-4a3cb6e0f849d37593327051f92df315c1d5bdf7644f7c2679955d535f0a1bdcR459)"?

rakita:
That comment can be improved, it is correct but after a Parity/Geth bug of state clear of precompile, all precompile accounts have none zero nonce, so we don't need to do this check.

And the way this is done in PR makes no sense, and it will introduce bugs for custom precompiles

PR #3053, opened 2025-10-06
fix: remove redundant U256::from on Host getters in instructions

Forostovec:
- Replace U256::from(context.host.block_number()) with context.host.block_number() in block_info.rs.
- Replace U256::from(context.host.effective_gas_price()) with context.host.effective_gas_price() in tx_info.rs.
- Aligns with Host methods already returning U256 and matches style of adjacent instructions.


rakita (approved):
lgtm

PR #3059, opened 2025-10-07
Use inspector only when needed in blockchaintest to reduce overhead

prestoalvarez:
### PR Description
- Why
  - Avoid unnecessary tracing overhead and stdout noise when not required.
  - Align behavior with intent: enable tracing only for structured output or diagnostics.
- What
  - In `bins/revme/src/cmd/blockchaintest.rs`, build EVM with an inspector only when `json_output` or `print_env_on_error` is enabled, using `Either<NoOpInspector, TracerEip3155>` to keep a stable EVM type.


rakita:
Previous is fine, Inspector is not touched unless `inspect_tx` is called

PR #3062, opened 2025-10-07
 fix: clamp gas spent against negative refunds

futreall:
Fixed net gas accounting so negative refunds no longer zero out the actual gas spent

### DIFF
--- a/crates/interpreter/src/gas.rs
+++ b/crates/interpreter/src/gas.rs
@@ -35,27 +35,39 @@ impl Gas {
 
     /// Returns the gas limit.
     #[inline]
-    pub const fn limit(&self) -> u64 { self.limit }
+    pub const fn limit(&self) -> u64 {
+        self.limit
+    }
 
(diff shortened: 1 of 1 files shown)


rakita:
Not a solution

futreall:
> Not a solution

@rakita hit a real bug here. Example:

let mut gas = Gas::new(1_000);
gas.record_cost(1_000);
gas.record_refund(-24_000);

assert_eq!(gas.spent_sub_refunded(), 0); // today

Negative refund is cast to u64, so we underflow and the floor / reimburse logic thinks we spent nothing. This happens for EIP‑3529 penalties (SELFDESTRUCT, SSTORE). Fix keeps the math signed and clamps to u64, so eip7623_check_gas_floor and reimburse_caller stay correct. Added unit test `gas::tests::spent_sub_refunded_handles_negative_refund` that fails on main and passes with this change. Thanks.

PR #3063, opened 2025-10-07
refactor(database): remove unnecessary Send+Sync bounds from TryDatabaseCommit for Arc

radik878:
Relax trait bounds on the Arc<Db> implementation of TryDatabaseCommit by removing Send + Sync. Arc::get_mut does not require these bounds, and dropping them broadens applicability in single-threaded contexts without changing behavior.


rakita:
That is true

PR #3064, opened 2025-10-08
fix: remove unused Jumps import from CountInspector

CreeptoGengar:
The trait was imported but never used in the implementation, as CountInspector only accesses opcode information through interp.bytecode.opcode() which doesn't require the Jumps trait directly.


CreeptoGengar:
my bad, opcode belongs to jumps not interpreter

PR #3066, opened 2025-10-08
fix(inspector): saturate negative refunded gas to zero in EIP-3155 tracer

viktorking7:
### Summary
- Fix incorrect JSON refund value caused by sign-unsafe cast in `crates/inspector/src/eip3155.rs`.
- Negative refunded gas no longer wraps to a huge `u64`.

### Problem
- `i64` to `u64` cast (`as`) in `step_end` caused wrap-around for negative `self.refunded`, corrupting tracer output.

### Change
- Replace `refund: self.refunded as u64` with `refund: self.refunded.max(0) as u64` and add a short explanatory comment.

### Rationale
- Ensures data correctness and avoids misleading analytics for tools consuming EIP-3155 traces.

### DIFF
--- a/crates/inspector/src/eip3155.rs
+++ b/crates/inspector/src/eip3155.rs
@@ -253,7 +253,7 @@ where
             stack: &self.stack,
             depth: context.journal_mut().depth() as u64,
             return_data: "0x",
-            refund: self.refunded as u64,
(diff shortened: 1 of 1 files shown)


rakita:
This is not a way. This PR would be an end goal https://github.com/bluealloy/revm/pull/2877

PR #3067, opened 2025-10-08
fix(kzg/blst): use uncompress for compressed G1/G2 inputs

MozirDmitriy:
Problem: The BLST backend used uncompressed deserialization functions (blst_p1_deserialize, blst_p2_deserialize) for compressed inputs (48-byte G1 and 96-byte G2). This mismatched the documented input formats, risked incorrect parsing and potential FFI UB, and contradicted inline comments and the arkworks implementation.

Fix: Replaced with the correct compressed APIs:
- parse_g1_compressed: blst_p1_uncompress for 48-byte compressed G1.
- get_trusted_setup_g2: blst_p2_uncompress for 96-byte compressed G2; updated the comment to reflect uncompression.

### DIFF
--- a/crates/precompile/src/kzg_point_evaluation/blst.rs
+++ b/crates/precompile/src/kzg_point_evaluation/blst.rs
@@ -68,8 +68,8 @@ fn get_trusted_setup_g2() -> blst_p2_affine {
     let mut g2_affine = blst_p2_affine::default();
     unsafe {
         // The compressed format has x coordinate and a flag bit for y
-        // We use deserialize_compressed which handles this automatically
-        let result = blst::blst_p2_deserialize(&mut g2_affine, TRUSTED_SETUP_TAU_G2_BYTES.as_ptr());
+        // We use uncompress which handles this automatically
+        let result = blst::blst_p2_uncompress(&mut g2_affine, TRUSTED_SETUP_TAU_G2_BYTES.as_ptr());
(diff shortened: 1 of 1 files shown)


kevaundray:
Great PR!

I believe the previous code still worked because it checked whether the serialization was encoded, see [here]( https://github.com/supranational/blst/blob/374b7cde7f3272ecde053e5c3f20ab65baa6cec1/src/e2.c#L395) which is called  [by revm from here](https://github.com/supranational/blst/blob/374b7cde7f3272ecde053e5c3f20ab65baa6cec1/src/e2.c#L408)

This was what the previous comment was noting when it said that it was automatically deserialized

rakita:
> This was what the previous comment was noting when it said that it was automatically deserialized

I know that I checked it and it worked, this makes sense, thanks for pointing it out. 

Maybe crafted input for blst_p1_deserialize could potentially read out of scope data and panic

kevaundray:
> > This was what the previous comment was noting when it said that it was automatically deserialized
> 
> I know that I checked it and it worked, this makes sense, thanks for pointing it out.
> 
> Maybe crafted input for blst_p1_deserialize could potentially read out of scope data and panic

I don't think its possible because the function reads the encoding from the first byte and then uses that to determine what constraints to place on the length -- though I think the change in the PR makes things clearer so overall a good change

PR #3070, opened 2025-10-09
feat: BAL EIP-7928

rakita:
Implementation of EIP-7928

This breaks the `EvmState` serde as it introduces an additional field for `Account` an `original_info`. Original info is used to calculate BAL. Discussion pending, maybe it is fine, but adding custom serde serialisation could work here.

### DIFF
--- a/bins/revme/src/cmd/blockchaintest.rs
+++ b/bins/revme/src/cmd/blockchaintest.rs
@@ -870,7 +870,7 @@ fn execute_blockchain_test(
                             println!("{}", serde_json::to_string(&output).unwrap());
                         } else {
                             eprintln!(
-                                "⚠️  Skipping block {block_idx} due to expected failure: {exception}"
+                                "⚠️  Skipping block {block_idx}: transaction unexpectedly succeeded (expected failure: {exception})"
                             );
                         }
                         break; // Skip to next block
--- a/crates/database/src/states/state.rs
+++ b/crates/database/src/states/state.rs
@@ -145,16 +145,39 @@ impl<DB: Database> State<DB> {
     /// If the account is not found in the cache, it will be loaded from the
     /// database and inserted into the cache.
     pub fn load_cache_account(&mut self, address: Address) -> Result<&mut CacheAccount, DB::Error> {
-        match self.cache.accounts.entry(address) {
+        Self::load_cache_account_with(
+            &mut self.cache,
+            self.use_preloaded_bundle,
+            &self.bundle_state,
+            &mut self.database,
+            address,
+        )
+    }
+
+    /// Get a mutable reference to the [`CacheAccount`] for the given address.
+    ///
+    /// If the account is not found in the cache, it will be loaded from the
+    /// database and inserted into the cache.
+    ///
+    /// This function accepts destructed fields of [`Self`] as arguments and
+    /// returns a cached account with the lifetime of the provided cache reference.
+    fn load_cache_account_with<'a>(
+        cache: &'a mut CacheState,
+        use_preloaded_bundle: bool,
+        bundle_state: &BundleState,
+        database: &mut DB,
+        address: Address,
+    ) -> Result<&'a mut CacheAccount, DB::Error> {
+        Ok(match cache.accounts.entry(address) {
             hash_map::Entry::Vacant(entry) => {
-                if self.use_preloaded_bundle {
+                if use_preloaded_bundle {
                     // Load account from bundle state
-                    if let Some(account) = self.bundle_state.account(&address).map(Into::into) {
+                    if let Some(account) = bundle_state.account(&address).map(Into::into) {
                         return Ok(entry.insert(account));
                     }
                 }
                 // If not found in bundle, load it from database
-                let info = self.database.basic(address)?;
+                let info = database.basic(address)?;
                 let account = match info {
                     None => CacheAccount::new_loaded_not_existing(),
                     Some(acc) if acc.is_empty() => {
@@ -162,10 +185,10 @@ impl<DB: Database> State<DB> {
                     }
                     Some(acc) => CacheAccount::new_loaded(acc, HashMap::default()),
                 };
-                Ok(entry.insert(account))
+                entry.insert(account)
             }
-            hash_map::Entry::Occupied(entry) => Ok(entry.into_mut()),
-        }
+            hash_map::Entry::Occupied(entry) => entry.into_mut(),
+        })
     }
 
     // TODO : Make cache aware of transitions dropping by having global transition counter.
@@ -223,13 +246,13 @@ impl<DB: Database> State<DB> {
     #[inline]
     fn storage(&mut self, address: Address, index: StorageKey) -> Result<StorageValue, DB::Error> {
         // If account is not found in cache, it will be loaded from database.
-        let account = if let Some(account) = self.cache.accounts.get_mut(&address) {
-            account
-        } else {
-            self.load_cache_account(address)?;
-            // safe to unwrap as account is loaded a line above.
-            self.cache.accounts.get_mut(&address).unwrap()
-        };
+        let account = Self::load_cache_account_with(
+            &mut self.cache,
+            self.use_preloaded_bundle,
+            &self.bundle_state,
+            &mut self.database,
+            address,
+        )?;
 
         // Account will always be some, but if it is not, StorageValue::ZERO will be returned.
         let is_storage_known = account.status.is_storage_known();
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -265,7 +265,10 @@ pub trait Handler {
     ///
     /// Verifies the initial cost does not exceed the transaction gas limit.
     #[inline]
-    fn validate_initial_tx_gas(&self, evm: &Self::Evm) -> Result<InitialAndFloorGas, Self::Error> {
+    fn validate_initial_tx_gas(
+        &self,
+        evm: &mut Self::Evm,
+    ) -> Result<InitialAndFloorGas, Self::Error> {
         let ctx = evm.ctx_ref();
         validation::validate_initial_tx_gas(
             ctx.tx(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,9 +171,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-json-abi"
-version = "1.4.1"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5513d5e6bd1cba6bdcf5373470f559f320c05c8c59493b6e98912fbe6733943f"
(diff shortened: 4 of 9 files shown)


bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -719,6 +721,17 @@ fn execute_blockchain_test(
...
+        let bal_test = block
+            .block_access_list
+            .as_ref()
+            .map(|bal| Bal::try_from(bal.clone()).ok())
+            .flatten()
+            .map(|bal| Arc::new(bal));
+
+        state.bal = bal_test;
+        state.bal_index = 0;
+        state.bal_builder = Some(Bal::new());
```
rakita: Initialize bal building by putting Some with empty Bal inside Db.

bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -719,6 +721,17 @@ fn execute_blockchain_test(
...
         }
 
+        let bal_test = block
+            .block_access_list
+            .as_ref()
+            .map(|bal| Bal::try_from(bal.clone()).ok())
+            .flatten()
+            .map(|bal| Arc::new(bal));
+
+        state.bal = bal_test;
```
rakita: This will come into play with parallel execution. It can be set to None for integration as it is not needed.

Used here to test that fetching from BAL values is okay

bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -804,6 +817,9 @@ fn execute_blockchain_test(
                 }
             };
 
+            // bump bal index
+            evm.db_mut().bal_index += 1;
```
rakita: Hand increment bal_index, should probably add trait for it, or do it automatically on every commit. The reason why it is not automatic is that there is multiple systems call that commit to state so an automatic bump would not work for them

crates/state/src/bal.rs
```diff
@@ -0,0 +1,263 @@
...
+
+///Block access index (0 for pre-execution, 1..n for transactions, n+1 for post-execution)
+pub type BalIndex = u64;
+
+/// BAL structure.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Bal {
+    /// Accounts bal.
+    pub accounts: IndexMap<Address, AccountBal>,
```
klkvr, line 34: would `BTreeMap` not work here? iiuc we need accounts to be sorted lexicographically? I'd also be open to changing alloy types to the ones we use here

crates/state/src/bal/writes.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Get value from BAL.
+    pub fn get(&self, bal_index: BalIndex) -> Option<T> {
+        let mut last_item = None;
+        for (index, item) in self.writes.iter() {
+            // if index is greater than bal_index we return the last item.
+            if index >= &bal_index {
+                return last_item;
+            }
+            last_item = Some(item.clone());
+        }
```
klkvr, line 33: probably worth doing a binary search here

crates/state/src/bal/writes.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// If BalIndex is same as last it will override the value.
+    pub fn update(&mut self, index: BalIndex, original_value: &T, value: T) {
+        self.update_with_key(index, original_value, value, |i| i);
+    }
+
+    /// Insert a value into the builder.
+    ///
+    /// If BalIndex is same as last it will override the value.
+    #[inline]
+    pub fn update_with_key<K: PartialEq, F>(
```
klkvr, line 95: let's document that this assumes that index is either 0 or >= last

crates/context/src/journal/inner.rs
```diff
@@ -40,6 +43,13 @@ pub struct JournalInner<ENTRY> {
     ///
     /// This ID is used in `Self::state` to determine if account/storage is touched/warm/cold.
     pub transaction_id: usize,
+    /// BAL for the state, if None BAL values are not used and it will not be applied on the top
+    /// of loaded account and storages.
+    pub bal: Option<BalWithIndex>,
+    /// If Account or storage is not found in the BAL, the error is stored here.
+    ///
+    /// Execution will continue even if the error is present and it will be taken at the end of the execution.
+    pub bal_error: Option<BalError>,
```
klkvr: do you think it's possible to handle this fully at db level? i.e custom `Database` handles all of the bal state overlay

crates/state/src/bal/account.rs
```diff
@@ -0,0 +1,310 @@
...
+
+        self.account_info
+            .update(bal_index, &account.original_info, &account.info);
+
+        self.storage.update(bal_index, &account.storage);
+    }
+
+    /// Create account from alloy account changes.
+    #[inline]
+    pub fn try_from_alloy(
```
klkvr, line 69: not super happy about us having to do this

crates/handler/src/handler.rs
```diff
@@ -145,6 +145,7 @@ pub trait Handler {
         &mut self,
         evm: &mut Self::Evm,
     ) -> Result<ExecutionResult<Self::HaltReason>, Self::Error> {
+        self.configure(evm)?;
```
klkvr: this is my least favourite kind of sdk changes because now every project overriding this fn needs to apply same diff which is super easy to miss, any chance we could avoid this?

crates/state/src/bal.rs
rakita: It is slighlty better to have IndexMap so you could index your account to fetch storage. Even HashMap would work here if you sort it in the end.

crates/context/src/journal/inner.rs
rakita: Have made a change in this PR: https://github.com/bluealloy/revm/pull/3119 . It is equivalent to this, bu better abstracted.

crates/state/src/bal/account.rs
rakita: It is a transition between revm inner type and rpc type from alloy

crates/handler/src/handler.rs
rakita: This will be removed as it is not needed.

An idea for configuration is something that is probably going to be needed for InstructionProvider that can set a different gas table. 

Good thing about placement of this call is that it is at the top of the function call chain, so it is very unlikely that it is overwritten by someone.

crates/state/src/bal/account.rs
klkvr: yep ideally we unify them

rakita: It is never a good idea to force one into another

crates/handler/src/handler.rs
rakita: Removed from PR

crates/database/src/states/state.rs
```diff
@@ -225,13 +236,84 @@ impl<DB: Database> State<DB> {
...
+        } else {
+            unreachable!("For accessing any storage account is guaranteed to be loaded beforehand")
+        };
+        Ok(storage)
+    }
 }
 
 impl<DB: Database> Database for State<DB> {
-    type Error = DB::Error;
+    type Error = BalDatabaseError<DB::Error>;
```
klkvr: hmm I think it's undesired that we break API like that

crates/database/src/states/state.rs
```diff
@@ -225,13 +236,84 @@ impl<DB: Database> State<DB> {
...
-        self.load_cache_account(address).map(|a| a.account_info())
+        let basic = self
+            .load_cache_account(address)
+            .map(|a| a.account_info())
+            .map_err(BalDatabaseError::Database)?;
+        // will populate account code if there was a bal change to it. If there is no change
+        // it will be fetched in code_by_hash.
+        self.bal_state
+            .basic(address, basic)
+            .map_err(Self::Error::from)
```
klkvr: not sure I understand what we're doing here

does this overlay BAL on top of the state? why don't we query the BAL first then? what do errors represent?

crates/state/src/account_info.rs
```diff
@@ -17,6 +17,8 @@ pub struct AccountInfo {
     pub nonce: u64,
     /// Hash of the raw bytes in `code`, or [`KECCAK_EMPTY`].
     pub code_hash: B256,
+    /// Storage id.
+    pub storage_id: Option<usize>,
```
klkvr: can we document what this is and what it's used for?

crates/database/interface/src/lib.rs
```diff
@@ -63,6 +64,20 @@ pub trait Database {
...
+    #[inline]
+    fn storage_by_account_id(
+        &mut self,
+        address: Address,
+        account_id: usize,
+        storage_key: StorageKey,
+    ) -> Result<StorageValue, Self::Error> {
+        let _ = account_id;
+        self.storage(address, storage_key)
+    }
```
klkvr: also worth documenting what is `account_id`

crates/database/src/states/state.rs
rakita: We are going back and forth on this. It should be fine as this error is again wrapped in main EVMError as result of transact.

cc @mattsse

crates/database/interface/src/bal.rs
```diff
@@ -0,0 +1,341 @@
...
+    #[inline]
+    pub fn reset_bal_index(&mut self) {
+        self.bal_index = 0;
+    }
+
+    /// Bump BAL index.
+    #[inline]
+    pub fn bump_bal_index(&mut self) {
+        self.bal_index += 1;
+    }
```
mattsse, line 47: we need a function here that explicitly sets the index

mattsse (changes requested):

crates/database/interface/src/lib.rs
```diff
@@ -62,6 +63,20 @@ pub trait Database {
...
         -> Result<StorageValue, Self::Error>;
 
+    /// Gets storage value of account by its id.
+    ///
+    /// Default implementation is to call [`Database::storage`] method.
+    #[inline]
+    fn storage_by_account_id(
+        &mut self,
+        address: Address,
+        account_id: usize,
```
mattsse, line 75: yeah unsure what account id is and why this is leaking into database trait here

crates/state/src/account_info.rs
rakita: Yes, it is a hint for the database that allows us to access the storage faster as we know the index of account.

It is set by database, and it is called by Journal with `Database::storage_by_account_id`

crates/database/src/states/state.rs
rakita: Have changes from here. We first check if accounts exist in BAL (if BAL is present). If they do, we load the full account from this database and then apply changes from BAL to it.

rakita:
CI fail on test stable is CI problem, it passed for nightly variant.

PR #3071, opened 2025-10-09
perf: use lazy defaults for env vars in burntpix benchmark

sashaodessa:
## Summary

Replace eager `unwrap_or()` with lazy `unwrap_or_else()` when reading environment variables to avoid unnecessary string allocations.


rakita:
I should probably by default close issue generated by bot accounts (Accounts without visible history or history of typo changes) previous code is okay

sashaodessa:
> I should probably by default close issue generated by bot accounts (Accounts without visible history or history of typo changes) previous code is okay

If a user has a private profile, this does not mean that he is a bot.

PR #3072, opened 2025-10-09
refactor: clarify gas accounting check in custom precompile

viktorking7:
## Description
Renamed misleading variable `underflow` to `cost_recorded` and add a clarifying comment about `record_cost` semantics. Improves readability and reduces risk of future logic mistakes around OOG handling.


rakita:
Marked as spelling

PR #3080, opened 2025-10-10
Fix: potential panic in State::storage fallback

prestoalvarez:
### Description

**Problem**: The `State::storage` method uses `unreachable!()` when an account is unexpectedly not found in cache, which could cause a panic in edge cases.

**Solution**: Replace the panic with a safe fallback to direct database access.


rakita:
This breaks a notion that account will always be loaded first, than storage will be loaded second.

If there is a need to remove this, we should load the account if it is none (with a `.entry()` api)

Will close this PR, be free to open another one if you are up to make a change

PR #3084, opened 2025-10-11
chore(handler): return NonceOverflow on CREATE nonce overflow

Forostovec:
Previously, CREATE returned InstructionResult::Return when the caller nonce increment overflowed, which incorrectly signaled success. This change returns InstructionResult::NonceOverflow instead, aligning with internal semantics and downstream mappings to HaltReason::NonceOverflow. This ensures consumers correctly detect and handle nonce overflow as a halt, not a successful return.


rakita:
Return is by spec

PR #3085, opened 2025-10-12
chore: journal transfer fn cleanup

rakita:
They have the same logic so we can cleanup the old function.

Have added back caller touch, in common path it is not needed as caller is always tocuhed, but just for consistency have reverted it back

### DIFF
--- a/crates/context/src/journal/inner.rs
+++ b/crates/context/src/journal/inner.rs
@@ -343,7 +343,6 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
 
         // sub balance from
         let from_account = self.state.get_mut(&from).unwrap();
-        // no need to touch caller account.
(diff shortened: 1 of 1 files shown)


crates/context/src/journal/inner.rs
klkvr, line 328: can we rename this to `amount`?

crates/context/src/journal/inner.rs
```diff
@@ -344,7 +344,7 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
         // sub balance from
         let from_account = self.state.get_mut(&from).unwrap();
         // no need to touch caller account.
```
klkvr:
```suggestion
```

crates/context/src/journal/inner.rs
rakita: Yes, will do it in additional PR

PR #3086, opened 2025-10-12
feat: JournaledAccount, a nice way to update and track changes

rakita:
JournalTr `load_account` and `load_account_code` now return & and not &mut of Account.

Additional functions that return a journaled account with both Account and Journal (`load_account_mut` and `load_account_code_mut`)

This allows `account.set_balance` `account.touch()` without thinking if journal entry needs to be made

### DIFF
--- a/crates/context/interface/src/journaled_state.rs
+++ b/crates/context/interface/src/journaled_state.rs
@@ -152,6 +152,16 @@ pub trait JournalTr {
         address: Address,
     ) -> Result<StateLoad<&Account>, <Self::Database as Database>::Error>;
 
+    /// Loads the account code, use `load_account_with_code` instead.
+    #[inline]
+    #[deprecated(note = "Use `load_account_with_code` instead")]
+    fn load_account_code(
+        &mut self,
+        address: Address,
+    ) -> Result<StateLoad<&Account>, <Self::Database as Database>::Error> {
+        self.load_account_with_code(address)
+    }
+
     /// Loads the account with code.
     fn load_account_with_code(
         &mut self,
@@ -178,7 +188,7 @@ pub trait JournalTr {
 
     /// Loads the journaled account.
     #[inline]
-    fn load_account_code_mut(
+    fn load_account_with_code_mut(
         &mut self,
         address: Address,
     ) -> Result<
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -167,7 +167,7 @@ pub fn validate_against_state_and_deduct_caller<
     let (block, tx, cfg, journal, _, _) = context.all_mut();
 
     // Load caller's account.
-    let mut caller = journal.load_account_code_mut(tx.caller())?.data;
+    let mut caller = journal.load_account_with_code_mut(tx.caller())?.data;
 
     validate_account_nonce_and_code_with_components(&caller.info, tx, cfg)?;
 
@@ -218,7 +218,7 @@ pub fn apply_eip7702_auth_list<
 
         // warm authority account and check nonce.
         // 4. Add `authority` to `accessed_addresses` (as defined in [EIP-2929](./eip-2929.md).)
-        let mut authority_acc = journal.load_account_code_mut(authority)?;
+        let mut authority_acc = journal.load_account_with_code_mut(authority)?;
 
         // 5. Verify the code of `authority` is either empty or already delegated.
         if let Some(bytecode) = &authority_acc.info.code {
--- a/crates/op-revm/src/handler.rs
+++ b/crates/op-revm/src/handler.rs
@@ -109,7 +109,7 @@ where
             let blob_price = block.blob_gasprice().unwrap_or_default();
             // deposit skips max fee check and just deducts the effective balance spending.
 
-            let mut caller = journal.load_account_code_mut(tx.caller())?.data;
+            let mut caller = journal.load_account_with_code_mut(tx.caller())?.data;
 
             let effective_balance_spending = tx
                 .effective_balance_spending(basefee, blob_price)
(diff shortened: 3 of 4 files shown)


crates/context/interface/src/journaled_state.rs
```diff
@@ -155,20 +162,54 @@ pub trait JournalTr {
     fn load_account(
         &mut self,
         address: Address,
-    ) -> Result<StateLoad<&mut Account>, <Self::Database as Database>::Error>;
+    ) -> Result<StateLoad<&Account>, <Self::Database as Database>::Error>;
```
rakita, line 153: This is biggest (and breaking) change in PR, where `&mut` is becomming `&`. Alternative is `load_account_mut`

crates/context/interface/src/journaled_state.rs
```diff
@@ -155,20 +162,54 @@ pub trait JournalTr {
...
+        address: Address,
+    ) -> Result<
+        StateLoad<JournaledAccount<'_, Self::JournalEntry>>,
+        <Self::Database as Database>::Error,
+    > {
+        self.load_account_mut_optional(address, true)
+    }
+
+    /// Loads the journaled account.
+    fn load_account_mut_optional(
```
rakita: This allows us to fetch the account that can get changed

crates/context/src/journal/inner.rs
```diff
@@ -439,14 +430,6 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
         // Enter subroutine
         let checkpoint = self.checkpoint();
 
-        // Fetch balance of caller.
-        let caller_balance = self.state.get(&caller).unwrap().info.balance;
-        // Check if caller has enough balance to send to the created contract.
-        if caller_balance < balance {
-            self.checkpoint_revert(checkpoint);
-            return Err(TransferError::OutOfFunds);
-        }
```
rakita, line 416: This check was redundunt and was done in `make_create_frame`

crates/context/interface/src/host.rs
```diff
@@ -133,7 +133,7 @@ pub trait Host {
     /// Balance, calls `ContextTr::journal_mut().load_account(address)`
     #[inline]
     fn balance(&mut self, address: Address) -> Option<StateLoad<U256>> {
-        self.load_account_info_skip_cold_load(address, false, false)
+        self.load_account_info_skip_cold_load(address, false, true)
             .ok()
             .map(|load| load.into_state_load(|i| i.balance))
     }
```
klkvr, line 139: why do we need to change this?

crates/context/interface/src/journaled_state.rs
```diff
@@ -143,20 +150,54 @@ pub trait JournalTr {
...
+        &mut self,
+        address: Address,
+        load_code: bool,
+    ) -> Result<
+        StateLoad<JournaledAccount<'_, Self::JournalEntry>>,
+        <Self::Database as Database>::Error,
+    >;
```
klkvr, line 209: should this be called `load_account_mut_optional_code`?

examples/cheatcode_inspector/src/main.rs
```diff
@@ -155,14 +156,11 @@ impl JournalTr for Backend {
...
-    fn load_account(&mut self, address: Address) -> Result<StateLoad<&mut Account>, Infallible> {
+    fn load_account(&mut self, address: Address) -> Result<StateLoad<&Account>, Infallible> {
         self.journaled_state.load_account(address)
     }
 
-    fn load_account_code(
-        &mut self,
-        address: Address,
-    ) -> Result<StateLoad<&mut Account>, Infallible> {
+    fn load_account_code(&mut self, address: Address) -> Result<StateLoad<&Account>, Infallible> {
```
klkvr: it's super confusing that `load_account_code` returns `StateLoad<&Account>` while there's also this

https://github.com/bluealloy/revm/blob/33154d8cbbbcb457987956c46bd1302b44926cac/crates/context/interface/src/host.rs#L169-L182

imo this should be `load_account_with_code`

crates/handler/src/pre_execution.rs
```diff
@@ -248,12 +249,11 @@ pub fn apply_eip7702_auth_list<
...
-        authority_acc.info.code_hash = hash;
-        authority_acc.info.code = Some(bytecode);
+        authority_acc.touch();
+        authority_acc.set_code(hash, bytecode);
 
-        // 9. Increase the nonce of `authority` by one.
-        authority_acc.info.nonce = authority_acc.info.nonce.saturating_add(1);
-        authority_acc.mark_touch();
+        // 9. Increase the nonce of `authority` by one.s
+        authority_acc.bump_nonce();
```
klkvr: should we bundle this into a separate `fn delegate`?

crates/op-revm/src/handler.rs
```diff
@@ -167,10 +165,12 @@ where
 
         let balance = calculate_caller_fee(balance, tx, block, cfg)?;
 
-        let old_balance = caller_account.caller_initial_modification(balance, tx.kind().is_call());
-        // NOTE: all changes to the caller account should journaled so in case of error
-        // we can revert the changes.
-        journal.caller_accounting_journal_entry(tx.caller(), old_balance, tx.kind().is_call());
+        // make changes to the account
+        caller_account.touch();
+        caller_account.set_balance(balance);
```
klkvr: `set_balance` would already touch, right?

klkvr (approved):
some naming nits but nothing critical so approving

crates/context/interface/src/host.rs
rakita: Lets revert it

crates/handler/src/pre_execution.rs
rakita: Nice one

crates/op-revm/src/handler.rs
rakita: Yes, can be removed

PR #3088, opened 2025-10-12
chore: consolidate revme imports

rakita:
Addded `parse` feature set so bytecode does not need to be included. Remove all single imports with one main import of revm inside revme.

No logical change


rakita:
Performance is worse as hashmap flag is removed from revme. This is going to be a follow up PR where statetest-types gets included inside revm so they can share primitive features.

PR #3097, opened 2025-10-14
fix: EIP-7702 refund for non-empty accounts

Wodann:
Since the Spurious Dragon hardfork, accounts are considered empty if the code is empty, balance is zero, and nonce is zero.
Before the Spurious Dragon hardfork, there was a distinction between "existing" and "empty" states.

REVM's codebase models this behaviour [here](https://github.com/bluealloy/revm/blob/main/crates/state/src/lib.rs#L63).

Remote providers, like Alchemy & Infura, return empty account information when querying for non-existing accounts. This breaks the `Journal` logic that would previously set the `status` flags to mark an account as "not existing" [here](https://github.com/bluealloy/revm/blob/main/crates/context/src/journal/inner.rs#L732).

As a result, [this line](https://github.com/bluealloy/revm/blob/main/crates/handler/src/pre_execution.rs#L225) caused the EIP-7702 gas cost refunds to be calculated incorrectly. I added a test that demonstrates this failure.

Since EIP-7702 was introduced after the Spurious Dragon hardfork, the existing logic to check for pre-Spurious dragon non-existing accounts was not necessary at all. Thus I was able to fix the issue by not checking for both conditions.

Since our team is still using REVM v83, I would also like to fix that release v83 and publish a new release. I don't seem to be able to create a PR for that tag. Could you please help me with that?

This is the comparison of our v83 branch with the tag: https://github.com/bluealloy/revm/compare/v83...NomicFoundation:revm:fix/v83-eip7702-refund

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -222,7 +222,7 @@ pub fn apply_eip7702_auth_list<
         }
 
         // 7. Add `PER_EMPTY_ACCOUNT_COST - PER_AUTH_BASE_COST` gas to the global refund counter if `authority` exists in the trie.
-        if !(authority_acc.is_empty() && authority_acc.is_loaded_as_not_existing_not_touched()) {
+        if !authority_acc.is_empty() {
             refunded_accounts += 1;
         }
 


rakita:
That part was a fix for the optimism chain, they had some existing empty accounts (an artifact from state transition), so it was a bug for them.

Just to add, this is by the spec.

rakita:
Would complain to Infura to be honest :)

Wodann:
The fixed line itself is broken, if I'm not mistaken. The correct way to solve it would be using [this function](https://github.com/bluealloy/revm/blob/main/crates/state/src/lib.rs#L63).

The logic should be specific to the hardfork used. The current logic triggers a refund if either of these are not true:
1. account is empty
2. account is loaded as not existing and not touched

For post-Spurious Dragon, the 2nd requirement will always be false, hence why we're seeing the bug.

Since EIP-7702 is always post-Spurious Dragon I decided not to use [the above function](https://github.com/bluealloy/revm/blob/main/crates/state/src/lib.rs#L63) but only check requirement 1.

> Would complain to Infura to be honest :)

Since we're no longer on the Spurious Dragon hardfork, providers, like Infura and Alchemy, are within their right to return empty accounts.

Wodann:
My original description was wrong, so I updated it:

> Remote providers, like Alchemy & Infura, recently switched to returning empty account information when querying for non-existing accounts.

Remote providers didn't make any changes recently. After checking, they've been returning empty account information for a long time (at least three years), potentially always.

Wodann:
> That part was a fix for the optimism chain, they had some existing empty accounts (an artifact from state transition), so it was a bug for them.

I didn't fully understand this comment, but after discussing internally I think it's clear to me:

Something went wrong with OP Mainnet that broke the expectation `empty_account === non_existent_account` post-Spurious Dragon hardfork. As such, it's possible that an empty account can still be an existing account. This requires alternative logic for OP mainnet to support EIP-7702 refunds.

Is that understanding correct?

> Just to add, this is by the spec.

The spec requires checking that the `authority` account is non-existent. Since `empty-account === non-existent-account`, that check by itself should suffice to guarantee it doesn't exist. Alternatively, if we have a marker that's supposed to indicate non-existent accounts, that marker should suffice by itself.

If alternative logic is required for OP, shouldn't that logic be chain-specific for OP? With the current implementation, all chain implementations are affected by the OP-specific behaviour.

rakita:
As this is an artefact of rpc provider, you could check if `account.is_empty()` and return `None` from `Database` trait

Wodann:
> As this is an artefact of rpc provider, you could check if `account.is_empty()` and return `None` from `Database` trait

For completeness, despite discovering the problem through a use case that depends on RPC providers (i.e. replaying live blocks), it isn't unique to RPC providers. Any `Database` implementation that returns an empty account would fail the same way.

E.g., in Hardhat we expose JSON-RPC methods to manually set the nonce, balance, and code of an account. This would allow a user to set an account's values to appear empty.

For the time being, we'll introduce a check in all of our `Database` implementations to see whether `account.is_empty()`. This however will result in the check being done twice; once in the `Database` implementation and once in REVM execution logic.

As such, it still feels like [the logic](https://github.com/bluealloy/revm/pull/3097/files#diff-e00f103b07b32916490ce8d8560f8a1e3a3345c33ad786af0485e2efdcc9dbffR225) in REVM could be improved to only use one concept of "not existing" and document this clearly.

rakita:
> As such, it still feels like [the logic](https://github.com/bluealloy/revm/pull/3097/files#diff-e00f103b07b32916490ce8d8560f8a1e3a3345c33ad786af0485e2efdcc9dbffR225) in REVM could be improved to only use one concept of "not existing" and document this clearly.

Got that doc directly from EIP. There is still a difference between not existing and existing empty, which we need as support for all hardforks is still there.

@Wodann initially i did it just with empty but as I said, it triggered a consensus on op in regards to other clients, so it is not something I can just erase bcs it is inconvenient for rpc (I would if I could).

rakita:
Will close this PR as it is not going to be merged

PR #3100, opened 2025-10-15
fix(handler): use spent gas as gas_used

wetkeyboard:
fixes https://github.com/paradigmxyz/reth/issues/18871


rakita:
This is not correct

PR #3101, opened 2025-10-15
revert: JumpTable old serde format

rakita:
Revert to old serde format so bincode can be supported

### DIFF
--- a/crates/bytecode/src/legacy/jump_map.rs
+++ b/crates/bytecode/src/legacy/jump_map.rs
@@ -69,8 +69,7 @@ impl serde::Serialize for JumpTable {
     where
         S: serde::Serializer,
     {
-        use bitvec::order::Lsb0;
-        let mut bitvec = BitVec::<_, Lsb0>::from_vec(self.table.to_vec());
+        let mut bitvec = BitVec::<u8>::from_vec(self.table.to_vec());
         bitvec.resize(self.len, false);
         bitvec.serialize(serializer)
     }


crates/bytecode/src/legacy/jump_map.rs
```diff
@@ -65,35 +63,27 @@ impl Default for JumpTable {
     }
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for JumpTable {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use bitvec::order::Lsb0;
```
klkvr: why do we need an explicit generic here? isn't it the default one

crates/bytecode/src/legacy/jump_map.rs
```diff
@@ -65,35 +63,27 @@ impl Default for JumpTable {
...
+        let bitvec = BitVec::<_, Lsb0>::from_vec(self.table.to_vec());
```
klkvr: this does not retain any context about the bit length, is it OK? can we add some roundtrip tests?

rakita: There is roundtrip that got triggered, have fixed it in later commit

crates/bytecode/src/legacy/jump_map.rs
rakita: For serialization we do, as `BitVec::from_vec` requires it.

crates/bytecode/src/legacy/jump_map.rs
```diff
@@ -65,35 +63,27 @@ impl Default for JumpTable {
...
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for JumpTable {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use bitvec::order::Lsb0;
+        let mut bitvec = BitVec::<_, Lsb0>::from_vec(self.table.to_vec());
```
klkvr:
```suggestion
        let mut bitvec = BitVec::<u8>::from_vec(self.table.to_vec());
```

PR #3105, opened 2025-10-15
chore: remove redundant alloy-eip7702 from handler dev-dependencies

RyanBGG:
`alloy-eip7702` is already available as a transitive dependency through `context-interface`, making the dev-dependency declaration redundant.

Additionally, referencing features of dev-dependencies in the `std` and `serde` features is not a recommended practice.

### DIFF
--- a/crates/context/Cargo.toml
+++ b/crates/context/Cargo.toml
@@ -31,7 +31,6 @@ cfg-if.workspace = true
 bitvec.workspace = true
 
 # Optional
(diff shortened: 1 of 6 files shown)


rakita:
@RyanBGG this needs some cleanup

RyanBGG:
@rakita all good now.

RyanBGG:
@rakita sorry for the confusing. 

Second commit incorrectly removed `database/std` and `database/serde`  revert now

PR #3110, opened 2025-10-17
feat: Optional Bytecode in CallInput

rakita:
Make bytecode optional in CallInput so bytecode_address can be used if needed.


zerosnacks (approved):
Makes sense :+1: 

Thanks!

PR #3115, opened 2025-10-19
Rename map_frame to map_item for clearer API symmetry

sashaodessa:
### Description
- Rename ItemOrResult::map_frame to map_item to accurately reflect behavior and match map_result.
- Update the single call site in handler/evm.rs.


rakita:
Previous is fine, will stop merging PRs from AI

PR #3116, opened 2025-10-20
feat: dont load access list immediatly

rakita:
Don't load access list immediately at the start. Push the access list in the journal and on load mark accounts/storages warm.

### DIFF
--- a/crates/handler/src/pre_execution.rs
+++ b/crates/handler/src/pre_execution.rs
@@ -55,11 +55,9 @@ pub fn load_accounts<
         if let Some(access_list) = tx.access_list() {
             let mut map: HashMap<Address, HashSet<StorageKey>> = HashMap::default();
             for item in access_list {
-                map.entry(*item.address()).or_default().extend(
-                    item.storage_slots()
-                        .cloned()
-                        .map(|key| U256::from_be_bytes(key.0)),
-                );
+                map.entry(*item.address())
+                    .or_default()
+                    .extend(item.storage_slots().map(|key| U256::from_be_bytes(key.0)));
             }
             journal.warm_access_list(map);
         }


crates/handler/src/pre_execution.rs
```diff
@@ -53,12 +53,15 @@ pub fn load_accounts<
...
         if let Some(access_list) = tx.access_list() {
+            let mut map: HashMap<Address, HashSet<StorageKey>> = HashMap::default();
             for item in access_list {
-                journal.warm_account_and_storage(
-                    *item.address(),
-                    item.storage_slots().map(|i| StorageKey::from_be_bytes(i.0)),
-                )?;
+                map.entry(*item.address()).or_default().extend(
+                    item.storage_slots()
+                        .cloned()
```
borngraced: do you really mean/need to call `cloned` here? looks like it can be omitted

crates/context/src/journal/inner.rs
```diff
@@ -725,18 +724,6 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
...
-            sload_with_account(
-                load.data,
-                db,
-                &mut self.journal,
-                self.transaction_id,
-                address,
-                storage_key,
-                false,
-            )?;
-        }
```
mattsse, line 739: I see, this previously always loaded the accesslist entirely?

crates/context/src/journal/inner.rs
```diff
@@ -755,16 +742,44 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
...
-            db,
-            &mut self.journal,
-            self.transaction_id,
-            address,
-            key,
-            skip_cold_load,
-        )
+
+        let is_newly_created = account.is_created();
+        let (value, is_cold) = match account.storage.entry(key) {
```
mattsse, line 747: is all of this the same as sload_with_account?

mattsse (commented):
i think this all checks out but dont feel equipped to properly review this

crates/context/src/journal/inner.rs
rakita: Yes, it has not changed other than variable names

crates/context/src/journal/inner.rs
rakita: Yeah, it was a small optimisation that would use access list at the start, and allows us to not have access list inside journal

crates/context/interface/src/journaled_state.rs
```diff
@@ -87,20 +87,8 @@ pub trait JournalTr {
...
 
-    /// Warms the account and storage.
-    fn warm_account_and_storage(
-        &mut self,
-        address: Address,
-        storage_keys: impl IntoIterator<Item = StorageKey>,
-    ) -> Result<(), <Self::Database as Database>::Error>;
-
-    /// Warms the account. Internally calls [`JournalTr::warm_account_and_storage`] with empty storage keys.
-    fn warm_account(
```
rakita, line 98: Have quicky searched foundry repo and these two functions are not used.

https://github.com/search?q=repo%3Afoundry-rs%2Ffoundry+warm_account_and_storage&type=code
https://github.com/search?q=repo%3Afoundry-rs%2Ffoundry+warm_account&type=code

rakita:
There are a lot of eest tests that cover access list, so I am confident that this is correct

crates/context/interface/src/journaled_state.rs
```diff
@@ -87,20 +87,8 @@ pub trait JournalTr {
...
-        &mut self,
-        address: Address,
-    ) -> Result<(), <Self::Database as Database>::Error> {
-        self.warm_account_and_storage(address, [])
-    }
+    /// Sets access list inside journal.
+    fn warm_access_list(&mut self, access_list: HashMap<Address, HashSet<StorageKey>>);
```
mattsse, line 91: this seems fine, given that this more or less just internal api

PR #3117, opened 2025-10-20
chore: merge context_interface import

RyanBGG:
Is there a reason to import `context_interface` separately? Should we merge them?


RyanBGG:
found one more such case in [precompile_provider.rs](https://github.com/bluealloy/revm/pull/3117/files#diff-67a0a3d75e9fc8a8cc68e5a0bb2491ce63a1650aa75fab7d67621d31c7ab6c12) should we make some rules in ci or local check to avoid this @rakita ? if you want I can try to add such rules.

rakita:
Superseeded by https://github.com/bluealloy/revm/pull/3184

PR #3118, opened 2025-10-21
feat: impl `Sealable` for `Bytecode`

makcandrov:
Makes it possible to use [`Sealed`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Sealed.html) with [`Bytecode`](https://docs.rs/revm-bytecode/latest/revm_bytecode/bytecode/enum.Bytecode.html)


rakita (approved):
Doesn't hurt to have it

PR #3120, opened 2025-10-21
fix(jovian): fixes the DA footprint update storage slot. fix l1 fork associated with Jovian.

theochap:
## Description

This PR fixes:
- The DA footprint scalar's storage slot. Updated spec: https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/jovian/exec-engine.md#scalar-loading
- The l1 fork associated with Jovian. It should be `Prague` and not `Fusaka`


rakita:
Will switch target to release/v94 to make a patch of op-revm.

And will make an additional PR for main

edit: other way around :)

PR #3126, opened 2025-10-27
chore: change method calls to static context for tx size estimation

jackwener:
`&self` is useless


rakita:
Will close this as op-revm is getting its own repo, so would like to minimise unnecessary changes to it

PR #3130, opened 2025-10-28
docs: add descriptive comments to secp256r1 test vectors

Lil-Duckling-22:
Added grouping comments and format documentation to secp256r1 test vectors for better readability. 
Test cases are now clearly categorized by type (valid signatures, invalid message hashes, input length errors, and invalid components).


rakita:
There are already names in tests that describe if it is ok or not

PR #3139, opened 2025-11-03
fix: bal binary search cases

pochenai:
This PR fixes a bug in the BAL lookup method, where the binary search logic could return incorrect results in certain edge cases.
- When BalWrites length > 5:
  - case 1: The returned index should be decremented by 1 to align with BAL lookup semantics.
  - case 2: When bal_index is 0, the binary search function should return None.
- Handles the case where the searched `bal_index` does not exist in the write list correctly.


rakita (approved):
lgtm, good bug found.

Have slighlty simplified the code

PR #3142, opened 2025-11-05
feat: add `set_nonce` on `JournalState`

zerosnacks:
Foundry requires this in https://github.com/foundry-rs/foundry/blob/806792b21fd5ae9b9d2480b37a8a616ad09b544e/crates/evm/core/src/backend/mod.rs#L1457

### DIFF
--- a/crates/context/interface/src/journaled_state/account.rs
+++ b/crates/context/interface/src/journaled_state/account.rs
@@ -137,6 +137,17 @@ impl<'a, ENTRY: JournalEntryTr> JournaledAccount<'a, ENTRY> {
         true
     }
 
+    /// Sets the nonce of the account.
+    ///
+    /// Touches the account in all cases.
+    #[inline]
+    pub fn set_nonce(&mut self, nonce: u64) {
+        self.touch();
+        self.account.info.set_nonce(nonce);
+        self.journal_entries
+            .push(ENTRY::nonce_changed(self.address));
+    }
+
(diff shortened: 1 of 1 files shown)


crates/context/interface/src/journaled_state/account.rs
```diff
@@ -137,6 +137,17 @@ impl<'a, ENTRY: JournalEntryTr> JournaledAccount<'a, ENTRY> {
...
 
+    /// Sets the nonce of the account.
+    ///
+    /// Touches the account in all cases.
+    #[inline]
+    pub fn set_nonce(&mut self, nonce: u64) {
+        self.touch();
+        self.account.info.set_nonce(nonce);
+        self.journal_entries
+            .push(ENTRY::nonce_changed(self.address));
```
rakita, line 148: We need new entry for this, nonce bump by one is only one that happen in EVM.

I am fine to change curenct nonce_changed entry to contain previous nonce value, to facilitate this.

rakita: I can take over this

zerosnacks: Thanks :+1: 

For https://github.com/foundry-rs/foundry/pull/12461 we've found a temporary workaround but this method would still be preferred

Normal prio

rakita:
superseded by https://github.com/bluealloy/revm/pull/3163

PR #3143, opened 2025-11-05
fix(op-revm): return error when enveloped_tx is missing

Keemosty12:
Fixes a missing case from #3055 where `tx_cost_with_tx` in `l1block.rs:272` still uses `.expect()` with the same `enveloped_tx()` call.

### DIFF
--- a/crates/op-revm/src/l1block.rs
+++ b/crates/op-revm/src/l1block.rs
@@ -266,9 +266,9 @@ impl L1BlockInfo {
     /// Internally calls [`L1BlockInfo::tx_cost`].
     pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> Option<U256> {
         // account for additional cost of l1 fee and operator fee
-        let enveloped_tx = tx.enveloped_tx().cloned()?;
+        let enveloped_tx = tx.enveloped_tx()?;
         let gas_limit = U256::from(tx.gas_limit());
-        Some(self.tx_cost(&enveloped_tx, gas_limit, spec))
+        Some(self.tx_cost(enveloped_tx, gas_limit, spec))
     }
 
     /// Calculate additional transaction cost.


crates/op-revm/src/l1block.rs
```diff
@@ -264,15 +264,13 @@ impl L1BlockInfo {
     /// Calculate additional transaction cost with OpTxTr.
     ///
     /// Internally calls [`L1BlockInfo::tx_cost`].
-    #[track_caller]
-    pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> U256 {
+    pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> Result<U256, String> {
```
rakita: Better to return Option<U256> here, and leave for caller to create the error message.

Keemosty12: will do, thanks!

Keemosty12: Hi @rakita, updated, pls check, let me know if there's anything else needs tweaking. Appreciate your time.

crates/op-revm/src/l1block.rs
```diff
@@ -261,15 +264,11 @@ impl L1BlockInfo {
...
     /// Internally calls [`L1BlockInfo::tx_cost`].
-    #[track_caller]
-    pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> U256 {
+    pub fn tx_cost_with_tx(&mut self, tx: impl OpTxTr, spec: OpSpecId) -> Option<U256> {
         // account for additional cost of l1 fee and operator fee
-        let enveloped_tx = tx
-            .enveloped_tx()
-            .expect("all not deposit tx have enveloped tx")
-            .clone();
+        let enveloped_tx = tx.enveloped_tx().cloned()?;
```
rakita:
```suggestion
        let enveloped_tx = tx.enveloped_tx()?;
```

probably dont need clone here, not sure why was cloned in first place

Keemosty12: You're right, removed, thanks!

PR #3144, opened 2025-11-05
feat(precompiles): add performant PrecompileError::OtherCowStr variant

rakita:
Using &'static str does not require creating a string so it is more performant.

### DIFF
--- a/crates/precompile/src/interface.rs
+++ b/crates/precompile/src/interface.rs
@@ -277,8 +277,8 @@ pub enum PrecompileError {
 
 impl PrecompileError {
     /// Returns another error with the given message.
-    pub fn other(err: String) -> Self {
-        Self::Other(Cow::Owned(err))
+    pub fn other(err: impl Into<String>) -> Self {
+        Self::Other(Cow::Owned(err.into()))
     }
 
     /// Returns another error with the given static string.


crates/precompile/src/interface.rs
```diff
@@ -272,7 +272,11 @@ pub enum PrecompileError {
     /// Fatal error with a custom error message
     Fatal(String),
     /// Catch-all variant for other errors
+    ///
+    /// Prefere using `OtherStr` instead of `Other` if possible as it is more performant.
```
rakita:
```suggestion
    /// Prefer using `OtherStr` instead of `Other` as it is more performant.
```

crates/precompile/src/interface.rs
```diff
@@ -271,14 +271,19 @@ pub enum PrecompileError {
...
+    Other(Cow<'static, str>),
 }
 
 impl PrecompileError {
     /// Returns another error with the given message.
-    pub fn other(err: impl Into<String>) -> Self {
-        Self::Other(err.into())
+    pub fn other(err: String) -> Self {
+        Self::Other(Cow::Owned(err))
+    }
```
klkvr: let's avoid breaking this? given that we don't extend capabilities of the method either

rakita: Didn't see that it was different, reverting

rakita: done

PR #3145, opened 2025-11-06
feat: refactor ssload/sstore to JournaledAccount

rakita:
This will help with perf as condition for account to be loaded will be not needed.

sload/sstore are moved to JournaledAccount for better accesss to it.

This does not solve foundry API problem where they need to directly inject storage into account. Maybe doing `unsafe_sstore` would be a solution, but will leave this for later.

### DIFF
--- a/crates/context/interface/src/journaled_state/account.rs
+++ b/crates/context/interface/src/journaled_state/account.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 
 use super::entry::JournalEntryTr;
+use auto_impl::auto_impl;
 use core::ops::Deref;
 use database_interface::Database;
 use primitives::{
@@ -18,11 +19,10 @@ use state::{Account, Bytecode, EvmStorageSlot};
 use std::vec::Vec;
 
 /// Trait that contains database and journal of all changes that were made to the account.
+#[auto_impl(&mut, Box)]
 pub trait JournaledAccountTr: Deref<Target = Account> {
     /// Database error type.
     type DatabaseError;
-    /// Entry type.
-    type Entry: JournalEntryTr;
 
     /// Creates a new journaled account.
     fn sload(
@@ -140,7 +140,6 @@ impl<'a, ENTRY: JournalEntryTr, DB: Database> JournaledAccountTr
     for JournaledAccount<'a, ENTRY, DB>
 {
     type DatabaseError = <DB as Database>::Error;
-    type Entry = ENTRY;
 
     fn sload(
         &mut self,


crates/context/interface/src/journaled_state/account.rs
```diff
@@ -3,26 +3,143 @@
...
+    /// Transaction ID.
+    transaction_id: usize,
+    /// Database used to load storage.
+    db: &'b mut DB,
+}
+
+impl<'a, 'b, ENTRY: JournalEntryTr, DB: Database> JournaledAccount<'a, 'b, ENTRY, DB> {
+    /// Creates a new journaled account.
+    #[inline(never)]
+    pub fn sload(
```
mattsse: from precompile pov this doesn't provide any additional benefits because in there we can not contrain any types, so we'd need to write additioal dyn compat wrappers for this I believe @klkvr

crates/context/interface/src/journaled_state.rs
```diff
@@ -16,6 +16,19 @@ use primitives::{
...
 use std::{borrow::Cow, vec::Vec};
 
+/// Result type that contains database error.
+pub type JournaledAccountLoadResult<'a, 'b, JOURNAL> = Result<
+    StateLoad<
+        JournaledAccount<
+            'a,
+            'b,
+            <JOURNAL as JournalTr>::JournalEntry,
+            <JOURNAL as JournalTr>::Database,
```
mattsse: if these types are both coming from the JournalTr itself, shouldnt the JournaledAccount then just be generic over that?

crates/context/interface/src/journaled_state/account.rs
rakita: wdym by can not contain any types? 

Are you talking about this trait? https://github.com/alloy-rs/evm/blob/15975d06bea380a025e59d757367d80d16cacc5a/crates/evm/src/traits.rs#L47

crates/context/interface/src/journaled_state.rs
rakita: It would be problematic as JournalInner is used in a foundry that does not contain JournalTr

rakita: I can probably drop one lifetime from structure, and restrict it directly in where it is used

crates/context/interface/src/journaled_state/account.rs
klkvr: yeah for `EvmInternals` we don't propagate any context about `DB` generic so we'd need to somehow erase it

rakita: Have added `JournaledAccountTr` that hides generic except for the Error. I see that Error type is fixed so it should be okay.

rakita:
Superseeded by https://github.com/bluealloy/revm/pull/3201

PR #3146, opened 2025-11-06
chore(interpreter): deprecate public otry! macro

sashass1315:
Deprecated revm_interpreter::otry! The macro is unused internally across the workspace but is part of the public API and documented on docs.rs. Guides users toward the idiomatic let Some(x) = expr else { return; }; pattern for early return in instruction functions.


rakita:
makes sense

PR #3148, opened 2025-11-06
feat: process precompile logs to inspector

rakita:
Having ability to consume logs created from precompile.

Changes `Inspector::log` and removed interpreter from it, and adds `Inspector::log_full` that contains it. 

Logs from precompile are going to call `log`

### DIFF
--- a/crates/inspector/src/either.rs
+++ b/crates/inspector/src/either.rs
@@ -37,7 +37,7 @@ where
     #[inline]
     fn log(&mut self, context: &mut CTX, log: Log) {
         match self {
-            Either::Left(inspector) => inspector.log(context, log.clone()),
+            Either::Left(inspector) => inspector.log(context, log),
             Either::Right(inspector) => inspector.log(context, log),
         }
     }
@@ -45,7 +45,7 @@ where
     #[inline]
     fn log_full(&mut self, interp: &mut Interpreter<INTR>, context: &mut CTX, log: Log) {
         match self {
-            Either::Left(inspector) => inspector.log_full(interp, context, log.clone()),
+            Either::Left(inspector) => inspector.log_full(interp, context, log),
             Either::Right(inspector) => inspector.log_full(interp, context, log),
         }
     }


crates/handler/src/frame.rs
```diff
@@ -190,14 +191,20 @@ impl EthFrame<EthInterpreter> {
         let gas_limit = inputs.gas_limit;
 
         if let Some(result) = precompiles.run(ctx, &inputs).map_err(ERROR::from_string)? {
+            let mut logs = Vec::new();
             if result.result.is_ok() {
                 ctx.journal_mut().checkpoint_commit();
             } else {
+                // clone logs that precompile created, only possible with custom precompiles.
+                // checkpoint.log_i will be always correct.
+                logs = ctx.journal_mut().logs()[checkpoint.log_i..].to_vec();
```
klkvr, line 200: can we avoid cloning here somehow? given that those logs will be removed anyway? maybe `checkpoint_revert` could return the reverted logs?

another option is to just not invoke the `.log` with logs that were removed, don't think there are many usecases for that

rakita: > can we avoid cloning here somehow? given that those logs will be removed anyway? maybe checkpoint_revert could return the reverted logs?

You need to clone them, as revert sets the len of log vec, it does not discard old log. 

It looks like we will need to clone logs in every case, as accessing them as ref is not possible as we are sending context and log toggether to Inspector

> another option is to just not invoke the .log with logs that were removed, don't think there are many usecases for that

Idk if revert is handled in inspectors, so imo it is better to have same behaviour.

klkvr: > It looks like we will need to clone logs in every case, as accessing them as ref is not possible as we are sending context and log toggether to Inspector

ah I see, then let's revert `&Log` change to not force caller to clone log second time?

klkvr: my concern here is that this will now clone precompile logs even when there's no inspector at all

rakita: > my concern here is that this will now clone precompile logs even when there's no inspector at all

Maybe we can extract checkpoint so commit/revert mechanism to be outside and be part of handler so inspector can overrride this. I dont mind it rn, and would merge it, but will need to take a look for solution in the future.

crates/inspector/src/either.rs
```diff
@@ -35,10 +35,18 @@ where
     }
 
     #[inline]
-    fn log(&mut self, interp: &mut Interpreter<INTR>, context: &mut CTX, log: Log) {
+    fn log(&mut self, context: &mut CTX, log: Log) {
         match self {
-            Either::Left(inspector) => inspector.log(interp, context, log),
-            Either::Right(inspector) => inspector.log(interp, context, log),
+            Either::Left(inspector) => inspector.log(context, log.clone()),
```
klkvr:
```suggestion
            Either::Left(inspector) => inspector.log(context, log),
```

crates/inspector/src/either.rs
```diff
@@ -35,10 +35,18 @@ where
...
+            Either::Right(inspector) => inspector.log(context, log),
+        }
+    }
+
+    #[inline]
+    fn log_full(&mut self, interp: &mut Interpreter<INTR>, context: &mut CTX, log: Log) {
+        match self {
+            Either::Left(inspector) => inspector.log_full(interp, context, log.clone()),
```
klkvr:
```suggestion
            Either::Left(inspector) => inspector.log_full(interp, context, log),
```

crates/inspector/src/traits.rs
```diff
@@ -104,8 +106,23 @@ pub trait InspectorEvmTr:
         }
 
         let frame_input = frame_init.frame_input.clone();
+        let logs_i = ctx.journal().logs().len();
         if let ItemOrResult::Result(mut output) = self.frame_init(frame_init)? {
             let (ctx, inspector) = self.ctx_inspector();
+            // for precompiles send logs to inspector.
+            if let FrameResult::Call(CallOutcome {
+                was_precompile_called,
+                precompile_call_logs,
```
klkvr, line 115: do we actually need the `was_precompile_called`? `precompile_call_logs` will just be empty in this case

rakita: `was_precompile_called` precompile_call_logs, will be empty if it is precompile but it is not reverted. ~Can potentially merge those two. And make a Option~ it does not look nice

PR #3150, opened 2025-11-07
fix(op): Ensure L1Block account is always loaded

rakita:


klkvr:
should we fix this in `State<DB>` maybe?

crates/database/src/states/state.rs
```diff
@@ -258,33 +258,36 @@ impl<DB: Database> Database for State<DB> {
...
-                .unwrap_or_default())
+        // If account is not found in cache, it will be loaded from database.
+        let account = if let Some(account) = self.cache.accounts.get_mut(&address) {
+            account
         } else {
-            unreachable!("For accessing any storage account is guaranteed to be loaded beforehand")
-        }
+            self.load_cache_account(address)?;
+            // safe to unwrap as account is loaded a line above.
+            self.cache.accounts.get_mut(&address).unwrap()
```
mattsse, line 267: this is fine

PR #3151, opened 2025-11-07
Refactor: Remove code duplication in `first_frame_input`

prestoalvarez:
Eliminates duplicated bytecode extraction logic in `Handler::first_frame_input` by introducing a `target_account` variable to determine whether to use the original account or the delegated account for EIP-7702 transactions.


rakita (commented):
Does not exactly work bcs of mutability

rakita:
Will close it, if you find a solution please open PR again

PR #3153, opened 2025-11-10
chore: deprecate InterpreterAction::new_stop

Snezhkko:
Deprecate InterpreterAction::new_stop because it constructs a STOP result with zero gas, which contradicts EVM semantics where STOP preserves the remaining gas. Consumers should use new_halt(InstructionResult::Stop, gas) or interpreter.halt(InstructionResult::Stop) to preserve gas.


rakita:
Will leave it as is, it is good as a placeholder if there is need to quickly create with stop

PR #3156, opened 2025-11-10
feat: generic Context::new

klkvr:
Generalizes `Context::new` over spec


mattsse (approved):
this into call is appropriate imo

PR #3166, opened 2025-11-12
test: validate CLZ opcode is not activated before Osaka hardfork

Wodann:
This PR adds a test that validates that the EVM throws an exceptional halt if the CLZ opcode is used in hardforks before Osaka.

This is merely added to ensure test coverage.


rakita:
All opcodes checks are covered by eest tests, so there is no need to duplicate

Wodann:
For future reference, where can I find the eest tests?

PR #3167, opened 2025-11-12
chore: add ModExp precompile unit test

anaPerezGhiglia:
- Add unit test that validates that the right ModExp precompile is used for each spec_id


rakita:
This is already covered by eest tests, so there is not need to duplicate

PR #3171, opened 2025-11-13
fix call stack depth limit check

viktorking7:
Fixes an off-by-one error in the call stack depth validation that allowed frames to be created at depth 1024, exceeding the EVM consensus limit of 1024 active frames.


rakita:
Test does not agree with this. This is not a bug.

PR #3172, opened 2025-11-13
Remove artificial balance inflation in disabled balance check mode

anim001k:
When `is_balance_check_disabled` is enabled, the code was forcing the caller's balance to be at least `tx.value()`, effectively creating ETH out of thin air. This violates balance invariants and can cause incorrect behavior in test scenarios.
Removed the conditional logic that artificially inflated the balance. The function now correctly calculates the new balance as `balance.saturating_sub(gas_balance_spending)` regardless of the balance check setting.


rakita:
It is expected to do this

PR #3176, opened 2025-11-14
refactor: reuse bound tx and remove redundant context.tx() calls

Bashmunta:
his refactor binds the transaction once in validate_tx_env() and derives tx_type from the bound reference, eliminating redundant calls to context.tx(). It also reuses the bound tx for the EIP-3860 initcode size check instead of calling context.tx() again. The change improves clarity and marginally reduces method calls without changing semantics or borrow behavior. Repeated immutable borrows from ContextTr::tx() are cheap, but centralizing on a single borrow makes intent clearer and avoids unnecessary lookups via all().


rakita:
Will be merged later

PR #3179, opened 2025-11-16
fix: return G2NotInSubgroup for G2 subgroup check failure

Fibonacci747:
The arkworks backend returned Bls12381G1NotInSubgroup from read_g2() when the G2 subgroup check failed. This was inconsistent with both the logical context (G2 subgroup validation) and the blst backend, which correctly returns Bls12381G2NotInSubgroup. Aligning the error variant ensures consistent behavior across backends, corrects diagnostics for callers that rely on precise error reporting, and matches EIP-2537 expectations where MSM and pairing must enforce subgroup checks for their respective groups.


rakita:
Will be merged later

PR #3181, opened 2025-11-18
perf: parallelize 7702 tx sender recovery

pochenai:
revm currently processes the auth_list of EIP-7702 transactions sequentially. Some [7702 transactions](https://etherscan.io/tx/0x9c206b4692f7ed4e09a4e46332e22fdc4376921b5e57592944c8d373a4c6c428) contain up to 200 authorization entries, and with EIP-7928 this becomes a significant performance bottleneck.

This PR introduces parallel sender recovery for 7702 transactions. The speedup scales almost linearly with both the number of authorization entries and the available CPU cores.


pochenai:
Closed, as Dragan pointed out Reth can do this via Either.

PR #3182, opened 2025-11-20
feat: sort accounts by address in blockchaintest output

RyanBGG:
Closes #3178. Sort accounts by address in error printout for better readability.

PR #3186, opened 2025-11-21
chore: sort storage keys and test files in blockchaintest output

RyanBGG:
Realized there are more section need to be optimize like https://github.com/bluealloy/revm/pull/3182

Any though?

### DIFF
--- a/bins/revme/src/cmd/blockchaintest.rs
+++ b/bins/revme/src/cmd/blockchaintest.rs
@@ -87,14 +87,12 @@ pub fn find_all_json_tests(path: &Path) -> Vec<PathBuf> {
     if path.is_file() {
         vec![path.to_path_buf()]
     } else {
-        let mut files: Vec<_> = WalkDir::new(path)
+        WalkDir::new(path)
             .into_iter()
             .filter_map(Result::ok)
             .filter(|e| e.path().extension() == Some("json".as_ref()))
             .map(DirEntry::into_path)
-            .collect();
-        files.sort();
-        files
+            .collect()
     }
 }
 


bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -87,12 +87,14 @@ pub fn find_all_json_tests(path: &Path) -> Vec<PathBuf> {
...
     } else {
-        WalkDir::new(path)
+        let mut files: Vec<_> = WalkDir::new(path)
             .into_iter()
             .filter_map(Result::ok)
             .filter(|e| e.path().extension() == Some("json".as_ref()))
             .map(DirEntry::into_path)
-            .collect()
+            .collect();
+        files.sort();
```
rakita: It is better for file names to be in order of reading

bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -574,7 +576,9 @@ fn print_error_with_state(
         );
         if !storage.is_empty() {
             eprintln!("    Storage ({} slots):", storage.len());
-            for (key, value) in storage.iter().take(5) {
+            let mut sorted_storage: Vec<_> = storage.iter().collect();
+            sorted_storage.sort_by_key(|(key, _)| *key);
+            for (key, value) in sorted_storage.iter().take(5) {
```
rakita: This is nice, i would even remove that 5 slot limit and print all of them

bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -599,7 +603,9 @@ fn print_error_with_state(
...
```
rakita: Makes sense, would remove 5 limit, for tests we want to read them all

bins/revme/src/cmd/blockchaintest.rs
RyanBGG: Yep, added `files.sort()` to make test execution order consistent across different filesystems.

bins/revme/src/cmd/blockchaintest.rs
RyanBGG: Removed

RyanBGG:
> Makes sense, left few nit comments

Sorry for the delay fix. Just recovered from devConnect

bins/revme/src/cmd/blockchaintest.rs
rakita: I meant, lets revert this, sorting files would give arbitrary order if files are in multiple folders.

RyanBGG: > I meant, lets revert this, sorting files would give arbitrary order if files are in multiple folders.

Ohhh shoot, just saw this comment! sorry for the long time silence!!!

Got it, reverted the file sorting part. Only keeping the storage key sorting now.

PR #3188, opened 2025-11-22
fix: missing ForkSpec variants in blockchaintest fork mapping

maradini77:
Added explicit mappings for `MergeEOF`, `MergeMeterInitCode`, `MergePush0`, and `BPO1ToBPO2AtTime15k` in `fork_to_spec_id` to prevent them from incorrectly defaulting to `SpecId::OSAKA`. This ensures blockchain tests run with the correct hardfork configuration.


rakita:
It is fine if they are not supported

PR #3191, opened 2025-11-26
refactor: optimize vector initialization with size hints

jackwener:
This pull request optimizes memory allocation for several functions by pre-allocating vectors with an estimated capacity based on iterator size hints. This change can improve performance by reducing the number of reallocations needed when populating vectors.


rakita (approved):
This is good, waiting for CI

PR #3195, opened 2025-11-27
feat: DatabaseCommitExt + increment_balances

### LINKED ISSUE #3194: [feature] Create `DatabaseCommitExt` trait with reusable utility functions
It would be nice for some utility functions like `State::increment_balances` to be implemented for all types that are `DatabaseCommit`.

POC
```rust
impl<T: Database + DatabaseCommit> DatabaseCommitExt for T {
    // default implementation
}

/// EVM database commit interface.
pub trait DatabaseCommitExt: Database + DatabaseCommit {
    /// Iterates over received balances and increment all account balances.
    ///
    /// Update will create transitions for all accounts that are updated.
    fn increment_balances(
        &mut self,
        balances: impl IntoIterator<Item = (Address, u128)>,
    ) -> Result<(), Self::Error> {
        // Make transition and update cache state
        let balances = balances.into_iter();
        let mut transitions = HashMap::with_capacity(balances.size_hint().0);
        for (address, balance) in balances {
            if balance == 0 {
                continue;
            }
            let mut original_account = match self.basic(address)? {
                Some(acc_info) => Account::from(acc_info),
                None => Account::new_not_existing(0),
            };
            original_account.info.balance = original_account
                .info
                .balance
                .saturating_add(U256::from(balance));
            original_account.mark_touch();
            transitions.insert(address, original_account);
        }
        self.commit(transitions);
        Ok(())
    }
}

```

0xForerunner:
closes #3194 

It would be nice for some utility functions like State::increment_balances to be implemented for all types that are DatabaseCommit.

Happy to add some more `State` methods into this trait as well once I've gotten some initial feedback.

### DIFF
--- a/crates/database/interface/src/lib.rs
+++ b/crates/database/interface/src/lib.rs
@@ -191,9 +191,6 @@ pub trait DatabaseCommitExt: Database + DatabaseCommit {
         let mut transitions: HashMap<Address, Account> = HashMap::default();
         transitions.reserve(balances.size_hint().0);
         for (address, balance) in balances {
-            if balance == 0 {
-                continue;
-            }
             let mut original_account = match self.basic(address)? {
                 Some(acc_info) => Account::from(acc_info),
                 None => Account::new_not_existing(0),


crates/database/interface/src/lib.rs
```diff
@@ -172,3 +172,39 @@ impl<T: DatabaseRef> DatabaseRef for WrapDatabaseRef<T> {
...
+        // Make transition and update cache state
+        let balances = balances.into_iter();
+        let mut transitions = HashMap::with_capacity(balances.size_hint().0);
+        for (address, balance) in balances {
+            if balance == 0 {
+                continue;
+            }
+            let mut original_account = match self.basic(address)? {
+                Some(acc_info) => Account::from(acc_info),
+                None => Account::new_not_existing(0),
```
0xForerunner, line 196: I'm not sure this is the correct Status that we want to be committing. Would love if someone who knows more could comment. I'm also not confident that setting `transaction_id` to 0 is okay here.

crates/database/interface/src/lib.rs
```diff
@@ -172,3 +172,39 @@ impl<T: DatabaseRef> DatabaseRef for WrapDatabaseRef<T> {
...
+            };
+            original_account.info.balance = original_account
+                .info
+                .balance
+                .saturating_add(U256::from(balance));
+            original_account.mark_touch();
```
0xForerunner, line 202: I assume marking as touched is correct here.

mattsse:
ref https://github.com/alloy-rs/evm/pull/234

crates/database/interface/src/lib.rs
```diff
@@ -172,3 +172,40 @@ impl<T: DatabaseRef> DatabaseRef for WrapDatabaseRef<T> {
...
+    fn increment_balances(
+        &mut self,
+        balances: impl IntoIterator<Item = (Address, u128)>,
+    ) -> Result<(), Self::Error> {
+        // Make transition and update cache state
+        let balances = balances.into_iter();
+        let mut transitions: HashMap<Address, Account> = HashMap::default();
+        transitions.reserve(balances.size_hint().0);
+        for (address, balance) in balances {
+            if balance == 0 {
```
rakita: It would be good to skip this condition, there is assumption (that maybe can not relevant in this case, but it is true to internal evm) that even if balance is zero the account will get touched and read (in case of bal added to reads).

This is just to align that logic.

crates/database/interface/src/lib.rs
rakita: `transition_id` being 0 is fine, it is used internally inside evm, not in db.

crates/database/interface/src/lib.rs
rakita: This is okay

crates/database/interface/src/lib.rs
0xForerunner: Gotcha, so just to be clear, I should remove `if balance == 0`?

rakita: yes, lets remove that condition

0xForerunner: removed!

0xForerunner:
Should be good to go!

PR #3197, opened 2025-11-28
feat: DatabaseCommit::commit_iter

### LINKED ISSUE #3196: [Feature] Implement DatabaseCommit::commit_iter for more versatility.
```rust
fn commit(&mut self, changes: HashMap<Address, Account>)
```
is overly restrictive and forces additional computation by constructing a `HashMap`. I suggest adding an additional overridable trait method for use cases that don't require a `HashMap`.

```rust
/// EVM database commit interface.
#[auto_impl(&mut, Box)]
pub trait DatabaseCommit {
    /// Commit changes to the database.
    fn commit(&mut self, changes: HashMap<Address, Account>);

    /// Commit changes to the database with an iterator.
    ///
    /// Implementors of [`DatabaseCommit`] should override this method when possible for efficiency.
    fn commit_iter(&mut self, changes: impl IntoIterator<Item = (Address, Account)>) {
        let changes: HashMap<Address, Account> = changes.into_iter().collect();
        self.commit(changes);
    }
}
```

0xForerunner:
closes #3196 

```rust
fn commit(&mut self, changes: HashMap<Address, Account>)
```
is overly restrictive and forces additional computation by constructing a `HashMap`. I suggest adding an additional overridable trait method for use cases that don't require a `HashMap`.

This PR provides a way around reconstructing a HashMap. Additionally I've overridden the implementation for State which should give a bit of added efficiency.

I have included a breaking change by returning an `Iterator` in `apply_evm_state`. This allows passing through an iterator rather than unnecessarily collecting into a Vec which should be more efficient. If we don't want a breaking change then I'm happy to revert just that part.

### DIFF
--- a/crates/database/src/states/cache.rs
+++ b/crates/database/src/states/cache.rs
@@ -89,6 +89,7 @@ impl CacheState {
     }
 
     /// Applies output of revm execution and create account transitions that are used to build BundleState.
+    #[inline]
     pub fn apply_evm_state(
         &mut self,
         evm_state: impl IntoIterator<Item = (Address, Account)>,


crates/database/interface/src/lib.rs
```diff
@@ -72,6 +72,16 @@ pub trait Database {
 pub trait DatabaseCommit {
     /// Commit changes to the database.
     fn commit(&mut self, changes: HashMap<Address, Account>);
+
+    /// Commit changes to the database with an iterator.
+    ///
+    /// Implementors of [`DatabaseCommit`] should override this method when possible for efficiency.
+    ///
+    /// Callers should prefer using [`DatabaseCommit::commit`] when they already have a [`HashMap`].
+    fn commit_iter(&mut self, changes: impl IntoIterator<Item = (Address, Account)>) {
```
rakita, line 81: This makes a lot of sense, even for it to become the main required fn in future.

crates/database/src/states/cache.rs
```diff
@@ -89,14 +89,14 @@ impl CacheState {
...
-        for (address, account) in evm_state {
-            if let Some(transition) = self.apply_account_state(address, account) {
-                transitions.push((address, transition));
-            }
-        }
-        transitions
+    pub fn apply_evm_state<'a>(
+        &'a mut self,
+        evm_state: impl IntoIterator<Item = (Address, Account)> + 'a,
+    ) -> impl Iterator<Item = (Address, TransitionAccount)> + 'a {
```
rakita: Hm is this maybe a problem. If we don't collect/iterate over returned Iterator, would the changes still be applied?

rakita: lol, it is: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=30b3aaa62884a42cf12eaf5f99262a5c

Lets return vec here, it will get optimized by compiler, and returning a iterator is a footgun.
 
Would additionally add `#[inline]` as a hint.

crates/database/src/states/state.rs
```diff
@@ -315,9 +318,18 @@ impl<DB: Database> Database for State<DB> {
...
+        if let Some(s) = self.transition_state.as_mut() {
+            s.add_transitions(transitions)
+        }
+    }
+
+    fn commit_iter(&mut self, changes: impl IntoIterator<Item = (Address, Account)>) {
+        let transitions = self.cache.apply_evm_state(changes);
+        if let Some(s) = self.transition_state.as_mut() {
+            s.add_transitions(transitions)
+        }
```
borngraced, line 332: Since HashMap implements [IntoIterator](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#impl-IntoIterator-for-HashMap%3CK,+V,+S%3E), commit can just delegate to commit_iter..I believe

```rust
fn commit(&mut self, changes: HashMap<Address, Account>) {
        self.commit_iter(changes)
    }
```

this removes redundancy and avoids future updates in multiple locations if if the transition logic ever changes ofc

0xForerunner: Well either commit depends on commit_iter or vice versa haha. We can't have both I chose my way because it doesn't have breaking changes.

crates/database/interface/src/lib.rs
0xForerunner: Yeah I made it this way to avoid breaking changes, but we could absolutely invert the dependency between commit and commit_iter.

crates/database/src/states/cache.rs
0xForerunner: I don't think the collection into Vec will be optimized by the compiler, but this also isn't really a huge deal.

> If we don't collect/iterate over returned Iterator, would the changes still be applied?

Ah you're right. This could be a foot gun, although I do think the caller would get a clippy lint for the unused iterator. I'll revert this though, thanks for the feedback!

crates/database/src/states/state.rs
borngraced: makes sense then

crates/database/src/states/cache.rs
0xForerunner: Reverted!

crates/database/interface/src/lib.rs
rakita: `Database/DatabaseCommit` are one of oldest traits in repo, it is good that we are not breaking it

crates/database/src/states/cache.rs
```diff
@@ -89,14 +89,17 @@ impl CacheState {
...
     /// Applies output of revm execution and create account transitions that are used to build BundleState.
-    pub fn apply_evm_state(&mut self, evm_state: EvmState) -> Vec<(Address, TransitionAccount)> {
-        let mut transitions = Vec::with_capacity(evm_state.len());
-        for (address, account) in evm_state {
-            if let Some(transition) = self.apply_account_state(address, account) {
-                transitions.push((address, transition));
-            }
-        }
-        transitions
+    pub fn apply_evm_state(
```
rakita, line 93:
```suggestion
    #[inline]
    pub fn apply_evm_state(
```

PR #3198, opened 2025-12-01
feat(statetest): spec-compliant expected exception handling

bshastry:
This PR makes the revme statetest runner spec-compliant for handling expected exceptions, matching the behavior of geth and other clients.

## Problem

Previously, revme silently skipped tests when:
- `tx_type()` returned `None` for invalid tx combinations (e.g., blob+create)
- AND `expect_exception` was set

This meant tests with expected exceptions were not actually validated, they just returned "All tests passed" without executing anything.

### 1. Remove early transaction type rejection
- `tx_type()` now always returns a `TransactionType` (not `Option`)
- Invalid combinations (blob+create, 7702+create) are validated during EVM execution, not during parsing
- This matches geth's `toMessage()` behavior

### 2. Add missing validations in handler
- Added `BlobCreateTransaction` validation for EIP-4844
- Added `Eip7702CreateTransaction` validation for EIP-7702
- Both check `tx.kind().is_create()` and return appropriate errors

### 3. Proper exception matching
- New `exception_map` module maps EEST exception strings to REVM errors
- Supports pipe-separated multiple exceptions (`exc1|exc2`)
- Comprehensive mapping for 30+ exception types

### 4. Always execute and validate
- Removed silent `continue` in runner.rs
- Tests with expected exceptions now execute and validate the error matches
- State roots are computed and output (matching geth behavior)

## Results
- 2297/2303 corpus tests passing (99.7%)
- State roots match geth exactly for expected exception tests
- Remaining 6 failures are pre-existing skip list issues

### DIFF
--- a/bins/revme/src/cmd/statetest.rs
+++ b/bins/revme/src/cmd/statetest.rs
@@ -1,3 +1,4 @@
+pub mod exception_map;
 pub mod merkle_trie;
 mod runner;
 pub mod utils;
--- a/bins/revme/src/cmd/statetest/exception_map.rs
+++ b/bins/revme/src/cmd/statetest/exception_map.rs
@@ -74,7 +74,10 @@ fn error_matches_single_exception(error: &InvalidTransaction, exception: &str) -
 
         // === Gas Limit Errors ===
         "TransactionException.INTRINSIC_GAS_TOO_LOW" => {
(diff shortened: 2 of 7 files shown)


rakita (commented):
Supportive, need to flush the mapping.

rakita:
We are fine without it, hive is run from reth side and mapping exist there, and i didn't find problems on revm side around this

PR #3199, opened 2025-12-01
feat: Restrict Database::Error. JournaledAccountTr

rakita:
Few changes in here, Database::Error now is restricted to `Sync+Send+core::error::Error`, it is a breaking change, but usage of this type should not be very disruptive. 

Additionally, `JournaledAccountTr` is added that wraps JournaledAccount and removes `JournalEntryTr` from `JournalTr`

Adds `ErasedError` to JournaledAccount functions, which allows erasing Database::Error from it and simplifying the usage in alloy/evm. Have tried propagating ErasedError even more from Journal, but this change will be a lot bigger, so i skipped it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,7 +123,7 @@ indicatif = "0.18"
 plain_hasher = "0.2"
 rstest = "0.26.0"
 serde_derive = "1.0"
-thiserror = "2.0"
+thiserror = { version = "2.0", default-features = false }
 triehash = "0.8"
 walkdir = "2.5"
 
--- a/crates/database/interface/Cargo.toml
+++ b/crates/database/interface/Cargo.toml
@@ -37,6 +37,12 @@ tokio = { workspace = true, optional = true }
 
 [features]
 default = ["std"]
-std = ["serde?/std", "primitives/std", "state/std", "either/std"]
+std = [
+    "serde?/std",
+    "primitives/std",
+    "state/std",
+    "either/std",
+    "thiserror/std",
+]
 serde = ["dep:serde", "primitives/serde", "state/serde", "either/serde"]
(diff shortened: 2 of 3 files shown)


crates/op-revm/src/handler.rs
```diff
@@ -404,6 +408,8 @@ where
             acc.bump_nonce();
             acc.incr_balance(U256::from(mint.unwrap_or_default()));
 
+            drop(acc); // Drop acc to avoid borrow checker issues.
```
klkvr, line 411: hmm which change triggered this to cause issues with lifetimes?

klkvr (approved):
lgtm, though would be cool to avoid manual drops

crates/op-revm/src/handler.rs
rakita: It is with the introduction of GAT in JournalTr::JournaledAccount, the problem here is that for a generic JournaledAccount compiler does not know if manual drop is implemented, so it can't free the acc before the end of the function  (NLL Non-Lexical Lifetimes) as objects are freed/dropped in the end.

It sucks a little, but this is an idiomatic solution for this problem.

crates/context/interface/src/journaled_state/account.rs
```diff
@@ -3,17 +3,100 @@
...
+
+    /// Increments the balance of the account.
+    ///
+    /// Touches the account in all cases.
+    fn incr_balance(&mut self, balance: U256) -> bool;
+
+    /// Decrements the balance of the account.
+    ///
+    /// Touches the account in all cases.
+    fn decr_balance(&mut self, balance: U256) -> bool;
```
mattsse, line 58: what's the bool that this returns?

PR #3201, opened 2025-12-02
feat: JournaledAccount sload/sstore

rakita:
Superseeds https://github.com/bluealloy/revm/pull/3145

### DIFF
--- a/crates/context/interface/src/journaled_state/account.rs
+++ b/crates/context/interface/src/journaled_state/account.rs
@@ -157,7 +157,9 @@ impl<'a, DB: Database, ENTRY: JournalEntryTr> JournaledAccount<'a, DB, ENTRY> {
         }
     }
 
-    /// Creates a new journaled account.
+    /// Loads the storage slot.
+    ///
+    /// If storage is cold and skip_cold_load is true, it will return [`JournalLoadError::ColdLoadSkipped`] error.
     ///
     /// Does not erase the db error.
     #[inline(never)]
@@ -212,7 +214,7 @@ impl<'a, DB: Database, ENTRY: JournalEntryTr> JournaledAccount<'a, DB, ENTRY> {
 
     /// Stores the storage slot.
     ///
-    /// If account is not present it will load from database.
+    /// If storage is cold and skip_cold_load is true, it will return [`JournalLoadError::ColdLoadSkipped`] error.
     ///
     /// Does not erase the db error.
     #[inline]
--- a/crates/context/src/journal/inner.rs
+++ b/crates/context/src/journal/inner.rs
@@ -696,6 +696,8 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
     /// Use this function only if you know what you are doing. It will not mark the account as warm or cold.
     /// It will not bump transition_id or return if it is cold or warm loaded. This function is useful
     /// when we know account is warm, touched and already loaded.
+    ///
+    /// It is useful when we want to access storage from account that is currently being executed.
     #[inline]
     pub fn get_account_mut<'a, 'db, DB: Database>(
         &'a mut self,
@@ -715,6 +717,7 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
             self.transaction_id,
         ))
     }
+
     /// Loads account. If account is already loaded it will be marked as warm.
     #[inline(never)]
     pub fn load_account_mut_optional<'a, 'db, DB: Database>(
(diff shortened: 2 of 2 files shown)


crates/context/interface/src/journaled_state/account.rs
```diff
@@ -17,6 +24,24 @@ pub trait JournaledAccountTr {
     /// Returns the account.
     fn account(&self) -> &Account;
 
+    /// Creates a new journaled account.
+    fn sload(
+        &mut self,
+        key: StorageKey,
+        skip_cold_load: bool,
+    ) -> Result<StateLoad<&mut EvmStorageSlot>, JournalLoadErasedError>;
```
klkvr: doc is wrong

crates/context/interface/src/journaled_state/account.rs
```diff
@@ -96,38 +121,164 @@ pub trait JournaledAccountTr {
...
+            transaction_id,
+            db,
+        }
+    }
+
+    /// Creates a new journaled account.
+    ///
+    /// Does not erase the db error.
+    #[inline(never)]
+    pub fn sload_inner(
```
klkvr: this is a weird name for `pub` fn, should this just be `sload`?

crates/context/src/journal/inner.rs
```diff
@@ -748,46 +797,45 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
...
+            .map(|s| s.map(|s| s.present_value))
+    }
 
-                (value, is_cold)
-            }
+    /// Loads storage slot.
+    ///
+    /// If account is not present it will return [`JournalLoadError::ColdLoadSkipped`] error.
+    #[inline]
+    pub fn sload_unsafe<DB: Database>(
```
klkvr: is this a hack to avoid perf overhead?

rakita: Not exact the hack, but it directly gets the account, so it is smaller code footprint, and it is not heavy on sload.

rakita: Althouth reading it again, should probably rename it

crates/context/interface/src/journaled_state/account.rs
rakita: Regarding name.

I have renames it to `sload_preserve_error`, does it sound better?

crates/context/src/journal/inner.rs
rakita: I have renamed it to `sload_assume_account_present` does it sound better @klkvr ?

crates/context/interface/src/journaled_state/account.rs
```diff
@@ -96,38 +121,164 @@ pub trait JournaledAccountTr {
...
     /// Mutable account.
     account: &'a mut Account,
     /// Journal entries.
     journal_entries: &'a mut Vec<ENTRY>,
+    /// Access list.
+    access_list: &'a HashMap<Address, HashSet<StorageKey>>,
+    /// Transaction ID.
+    transaction_id: usize,
+    /// Database used to load storage.
+    db: &'a mut DB,
```
mattsse, line 136: does this even need a &mut if DB traits are autompl for mut?

crates/context/interface/src/journaled_state/account.rs
```diff
@@ -96,38 +121,164 @@ pub trait JournaledAccountTr {
...
+            transaction_id,
+            db,
+        }
+    }
+
+    /// Creates a new journaled account.
+    ///
+    /// Does not erase the db error.
+    #[inline(never)]
+    pub fn sload_preserve_error(
```
mattsse: this naming is a bit weird, why not just sload?

EDIT: oh I see because the trait uses that already that returns erased errors

but maybe we can come up with something more appropriate than preserve_error

crates/context/src/journal/inner.rs
```diff
@@ -690,56 +749,46 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
...
+                    self.journal.push(ENTRY::account_warmed(address));
                 }
-                StateLoad {
-                    data: account,
-                    is_cold,
-                }
+                (account, is_cold)
             }
             Entry::Vacant(vac) => {
-                // Precompiles among some other account(coinbase included) are warm loaded so we need to take that into account
```
mattsse, line 700: is this comment no longer relevant?

crates/context/interface/src/journaled_state/account.rs
rakita: Yeah it can't be sload, open for naming suggestions

crates/context/interface/src/journaled_state/account.rs
rakita: JournalInner uses `&mut DB`, so I went with that. There may be some lifetime benefits, but I would like to check DB in follow-up PR. Having only DB would be cleaner.

crates/context/src/journal/inner.rs
rakita: Will add the comment back

crates/context/interface/src/journaled_state/account.rs
rakita: Some names suggested by claude:

Image

wdyt @mattsse @klkvr

PR #3202, opened 2025-12-03
refactor(revme): optimize error vector allocation in test runner

Keemosty12:
Changed `thread_errors` vector initialization to pre-allocate memory using `num_threads` in `statetest/runner.rs` to prevent unnecessary memory reallocations when collecting errors from parallel test execution.


rakita:
No need for this

PR #3206, opened 2025-12-04
chore(context): remove TODO in local context tests

Keemosty12:
Inline `b.consume()` call in `frame_stack` test to remove the TODO comment, simplifying the test code logic.


rakita:
Will remove TODO in other PR

PR #3211, opened 2025-12-07
fix: FastLZ match length calculation to avoid overflow and bad lengths

anim001k:
**Description:**
- Fix `cmp` to avoid unsigned underflow when `q > r` and stop on first mismatch.
- Guard `flz_match` against `l == 0` to preserve minimal encoded length.
- Ensures FastLZ length estimates stay in sync with reference and no panics on crafted input.


rakita:
Will close this as op revm is getting its own repo

PR #3212, opened 2025-12-07
fix(database): verify handle belongs to current runtime before block_in_place

forkfury:
## Fix

Prevents potential deadlock when `WrapDatabaseAsync` is created with a handle from a different runtime via `with_handle()`. The code now verifies that the stored handle belongs to the current runtime before using `block_in_place`.

## Changes

- Add pointer equality check to ensure handle belongs to current runtime before calling `block_in_place`
- Update comments to clarify the runtime ownership requirement

### DIFF
--- a/crates/database/interface/src/async_db.rs
+++ b/crates/database/interface/src/async_db.rs
@@ -200,7 +200,7 @@ impl HandleOrRuntime {
                         !matches!(
                             current.runtime_flavor(),
                             tokio::runtime::RuntimeFlavor::CurrentThread
-                        ) && std::ptr::eq(handle as *const _, &current as *const _)
+                        ) && core::ptr::eq(handle as *const _, &current as *const _)
                     }
                     Err(_) => false,
                 };


crates/database/interface/src/async_db.rs
```diff
@@ -191,13 +191,17 @@ impl HandleOrRuntime {
...
-                        current.runtime_flavor(),
-                        tokio::runtime::RuntimeFlavor::CurrentThread
-                    ),
+                    Ok(current) => {
+                        // Only use block_in_place if we're in a multi-threaded runtime
+                        // and the handle belongs to the current runtime (same handle instance).
+                        !matches!(
+                            current.runtime_flavor(),
+                            tokio::runtime::RuntimeFlavor::CurrentThread
+                        ) && std::ptr::eq(handle as *const _, &current as *const _)
```
rakita:
```suggestion
                        ) && core::ptr::eq(handle as *const _, &current as *const _)
```

PR #3214, opened 2025-12-08
fix: always validate fees

klkvr:
this logic might be a source of footguns for EVMs with custom transactions as it's not immediately obvious that some of the checks are skipped based on thet type byte


rakita:
The check should be put at the place of custom transaction

rakita:
Having configurable validation should help with this https://github.com/bluealloy/revm/issues/3215 will prio this after GasParams is delivered

PR #3216, opened 2025-12-09
feat(transaction): Introduced configurable transaction validation

### LINKED ISSUE #3215: Configurable tx validation
Custom transaction should have the ability to choose what kind of validation they expect from revm, this could be done as a bitflags that has few options (Check base fee, check auth list, check blob fee, etc). This would allow custom transactions to opt into the validations they need but still allow Ethereum transactions to do their default checks.

It should be an enum
```
pub enum ValidationTypes {
  None,
  ByTxType,
  Custom(Validations),
}
``` 

where `ValidationChecks` is the list of bitflags.

developeruche:
Closes #3215

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3585,6 +3585,7 @@ dependencies = [
  "alloy-eip2930",
  "alloy-eip7702",
  "auto_impl",
+ "bitflags",
  "either",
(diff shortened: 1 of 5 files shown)


rakita:
Current approach is fine for now

PR #3218, opened 2025-12-10
feat(create): Implement Cache for CreateInputs::created_address

### LINKED ISSUE #3217: Cache CreateInputs::created_address
`CreateInputs::created_address` will eventually be called by revm. When there are inspectors however one might call this function before it's passed to revm, and revm will call it again thus re-computing 1/2 keccaks (1 for CREATE; 1 bytecode hash + 1 for CREATE2).
We could try making fields private in `CreateInputs` and having an internal cache for the result of this function.

Here are few uses (foundry, revm-inspectors compute this in the inspector fn create):

Image

developeruche:
Closes #3217

### DIFF
--- a/crates/handler/src/execution.rs
+++ b/crates/handler/src/execution.rs
@@ -31,6 +31,12 @@ pub fn create_init_frame(
                 return_memory_offset: 0..0,
             }))
         }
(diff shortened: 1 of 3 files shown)


rakita (approved):
cool
lgtm

DaniPopes:
the fields shouldnt be pub cuz it invalidates the cache but its not reset

developeruche:
> the fields shouldnt be pub cuz it invalidates the cache, but it's not reset

Got you... I would expose a getter method for the fields and make them private.

PR #3220, opened 2025-12-10
Remove redundant tx fetch in Optimism handler gas accounting

anim001k:
- drop duplicate `ctx.tx()` binding in `last_frame_result`, avoiding shadowing without changing behavior
- keep gas handling unchanged while reducing noise


rakita:
clippy pending

anim001k:
> clippy pending

@rakita clippy fixed

PR #3222, opened 2025-12-10
fix(create): Fix CreateInputs::created_address Cache invalidation

### LINKED ISSUE #3221: Fix CreateInputs::created_address Cache invalidation
See #3218 

Comment: https://github.com/bluealloy/revm/pull/3218#issuecomment-3637377780

developeruche:
Closes #3221


developeruche:
> We would need setter functions for these so they can be changed in the Inspector::call.

Setter functions have been introduced, the cache is going to reset whenever any set is done save, "set_gas_limit" and "set_value" as they both don't affect the resulting create address.

PR #3223, opened 2025-12-11
feat: Implement EIP-8024 for Amsterdam

frangio:
CFI'd as of last week.

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -30,14 +30,13 @@ pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
         } else {
             let push_offset = opcode.wrapping_sub(opcode::PUSH1);
             let dupn_offset = opcode.wrapping_sub(opcode::DUPN);
-            let skip =
-                if push_offset < 32 {
-                    2 + push_offset as usize
-                } else if dupn_offset < 3 {
-                    2
-                } else {
-                    1
-                };
+            let skip = if push_offset < 32 {
+                2 + push_offset as usize
+            } else if dupn_offset < 3 {
+                2
+            } else {
+                1
+            };
             // SAFETY: Iterator access range is checked in the while loop
             iterator = unsafe { iterator.add(skip) };
         }


crates/bytecode/src/opcode.rs
```diff
@@ -623,9 +623,9 @@ opcodes! {
     // 0xE3
     // 0xE4
     // 0xE5
-    // 0xE6
-    // 0xE7
-    // 0xE8
+    0xE6 => DUPN     => stack_io(0, 1), immediate_size(1);
+    0xE7 => SWAPN    => stack_io(0, 0), immediate_size(1);
+    0xE8 => EXCHANGE => stack_io(0, 0), immediate_size(1);
```
frangio, line 628: The `stack_io` values are not accurate here, they depend on the immediate, unless we want to put the worst case values here. I don't know what this is used for.

crates/interpreter/src/instructions/stack.rs
```diff
@@ -60,3 +59,173 @@ pub fn swap<const N: usize, WIRE: InterpreterTypes, H: ?Sized>(
...
+    let r = k % 16;
+    if q < r {
+        Some((q + 1, r + 1))
+    } else {
+        Some((r + 1, 29 - q))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
frangio, line 145: The test cases are the same listed in the EIP.

crates/interpreter/src/instruction_result.rs
```diff
@@ -348,6 +351,9 @@ impl<HaltReasonTr: From<HaltReason>> From<InstructionResult> for SuccessOrHalt<H
             InstructionResult::InvalidExtDelegateCallTarget => {
                 Self::Internal(InternalResult::InvalidExtDelegateCallTarget)
             }
+            InstructionResult::InvalidImmediateEncoding => {
+                Self::Halt(HaltReason::OpcodeNotFound.into())
```
frangio, line 355: I didn't want to add a new `HaltReason` for this because IIUC it would be a breaking change for implementers of `From<HaltReason>`.

frangio:
I can't reproduce the nightly test failures.

rakita:
> I can't reproduce the nightly test failures.

Interesting error, jumpdest analysis found that the last opcode is 0x00, but it is jumped over, so it concluded it does not need to append zero bytes. Will see how to do this in an efficient way in `analyze_legacy` function.

rakita:
Imo `DUPN PUSH1 0x01` variant is better

gakonst:
Unbelievably bullish change

rakita:
It will take a few months before this becomes active so will move this PR to draft

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -29,13 +29,16 @@ pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
...
+            } else if dupn_offset < 3 {
+                2
             } else {
-                // SAFETY: Iterator access range is checked in the while loop
-                iterator = unsafe { iterator.add(1) };
-            }
+                1
+            };
+            // SAFETY: Iterator access range is checked in the while loop
+            iterator = unsafe { iterator.add(skip) };
```
rakita, line 41: Ah, this can generate a different jumptable for previous bytecode, as we are now jumping over one more immediate

rakita: Okay, will need to rework this so jump table is same as previous bytecodes.

PR #3226, opened 2025-12-15
feat: move gas refund tracking to journal level

### LINKED ISSUE #2145: Refactor: migrate from call local refund to global journal refund

Himess:
Closes #2145


rakita:
The current approach is fine, not the finest but fine for now.

PR #3229, opened 2025-12-16
feat: move GasParams to Cfg

rakita:
Moves GasParams to the `Cfg` so that it can be accessed with Context easily from instructions. GasParams is removed from Frame/Interpreter.

Gas calculation stuff is moved from Interpreter to cfg.

CfgEnv functions now require setting gas_params and `CfgEnv::with_spec` got deprecated and it is pointing to `with_spec_and_mainnet_gas_params` as alternative. 

Performance boost is because the first time initialisation of Precompiles was inside codspeed scope, with this PR this is moved outside, and simple check if spec is same is done there. Precompiles is still initialised only once, but now outside of codspeed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ ee-tests = { path = "crates/ee-tests", package = "revm-ee-tests", version = "0.1
 # alloy
 alloy-eip2930 = { version = "0.2.1", default-features = false }
 alloy-eip7702 = { version = "0.6.1", default-features = false }
-alloy-eip7928 = { version = "0.2.0", default-features = false, git = "https://github.com/alloy-rs/eips.git", rev = "734beaf" }
+alloy-eip7928 = { version = "0.3.0", default-features = false }
 alloy-primitives = { version = "1.5.0", default-features = false }
 
 # alloy in examples, revme or feature flagged.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,8 +137,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-eip7928"
-version = "0.2.0"
-source = "git+https://github.com/alloy-rs/eips.git?rev=734beaf#734beafea409bdd87b3a4e33ac72ac68654bb8dd"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6adac476434bf024279164dcdca299309f0c7d1e3557024eb7a83f8d9d01c6b5"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021-2025 draganrakita
+Copyright (c) 2021-2026 draganrakita
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/bins/revme/LICENSE
+++ b/bins/revme/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021-2025 draganrakita
+Copyright (c) 2021-2026 draganrakita
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/bins/revme/src/cmd/blockchaintest.rs
+++ b/bins/revme/src/cmd/blockchaintest.rs
@@ -406,7 +406,7 @@ fn validate_post_state(
                         *address,
                         "code".to_string(),
                         format!("0x{}", hex::encode(&expected_account.code)),
-                        format!("0x{}", hex::encode(actual_code.bytecode())),
+                        format!("0x{}", hex::encode(actual_code.original_byte_slice())),
                     );
                 }
             } else {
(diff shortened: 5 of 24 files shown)


crates/context/src/cfg.rs
```diff
@@ -1,12 +1,21 @@
...
+
 /// EVM configuration
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug)]
+#[derive_where(Eq, PartialEq; SPEC)]
 #[non_exhaustive]
 pub struct CfgEnv<SPEC = SpecId> {
+    /// Specification for EVM represent the hardfork
+    spec: SPEC,
```
klkvr: why is this not `pub` anymore?

crates/context/src/cfg.rs
```diff
@@ -55,6 +62,14 @@ pub struct CfgEnv<SPEC = SpecId> {
...
     /// with initials cap of 30M.
     pub tx_gas_limit_cap: Option<u64>,
+    /// Gas params for the EVM.
+    pub gas_params: GasParams,
+    /// Gas params override function.
+    ///
+    /// If this is set, it will override the gas params with this function.
+    #[cfg_attr(feature = "serde", serde(skip))]
+    #[derive_where(skip)]
+    pub gas_params_override: Option<fn(SPEC) -> GasParams>,
```
klkvr: hmm not sure I understand what this is for

crates/primitives/src/hardfork.rs
```diff
@@ -210,3 +211,10 @@ impl core::fmt::Display for SpecId {
         write!(f, "{}", <&'static str>::from(*self))
     }
 }
+
+/// Utility interface that sets the spec for structs that implements it.
+#[auto_impl(&mut, Box)]
+pub trait SetSpecTr<SPEC: Into<SpecId> + Clone> {
+    /// Sets the spec for the struct.
+    fn set_spec(&mut self, spec: SPEC);
+}
```
klkvr: Unclear to me why this trait is needed

if we really need to we can define this setter directly on types

crates/context/src/evm.rs
```diff
@@ -95,3 +96,19 @@ impl<CTX, INSP, I, P, F> DerefMut for Evm<CTX, INSP, I, P, F> {
...
+    P: SetSpecTr<<<CTX as ContextTr>::Cfg as Cfg>::Spec>,
+{
+    /// Sets the spec for the EVM.
+    #[inline]
+    fn set_spec(&mut self, spec: <<CTX as ContextTr>::Cfg as Cfg>::Spec) {
+        self.ctx.set_spec(spec.clone());
+        self.precompiles.set_spec(spec.clone());
+        self.instruction.set_spec(spec);
+    }
+}
```
klkvr: why would I want to call this? I've already configured EVM with proper spec and CTX at construction time

crates/handler/src/precompile_provider.rs
```diff
@@ -55,11 +69,26 @@ impl EthPrecompiles {
...
+        self.spec = spec;
+    }
+}
+
 impl Clone for EthPrecompiles {
     fn clone(&self) -> Self {
         Self {
             precompiles: self.precompiles,
             spec: self.spec,
+            spec_override_fn: self.spec_override_fn,
```
klkvr: why do we need to change precompiles in scope of this PR? not obvious what this field is for

crates/handler/src/instructions.rs
```diff
@@ -57,23 +63,19 @@ where
...
-            instruction_table_gas_changes_spec(spec),
-            GasParams::new_spec(spec),
-            spec,
-        )
+        Self::new(instruction_table_gas_changes_spec(spec), spec)
+    }
+
+    /// Returns `EthInstructions` with mainnet spec.
+    pub fn new_mainnet_with_spec(spec: SpecId) -> Self {
+        Self::new(instruction_table_gas_changes_spec(spec), spec)
```
klkvr, line 55: can this take `GasParams` as input? so that all instructions have direct access to it without expecting `Host` to provide it

crates/context/src/cfg.rs
rakita: It is a small helper, and if not set, it will do nothing.

The idea is to allow changes to GasPaarms that is spec aware.

crates/context/src/evm.rs
rakita: I wanted to add new option so spec can be set if needed. I think you are right here, and it should be simplified, which will remove all set_spec that are not needed. All of them can be set at construction time.

crates/handler/src/instructions.rs
rakita: It does not work like that, instructions have access to context and interpreter

Cargo.toml
```diff
@@ -60,7 +60,7 @@ ee-tests = { path = "crates/ee-tests", package = "revm-ee-tests", version = "0.1
 # alloy
 alloy-eip2930 = { version = "0.2.1", default-features = false }
 alloy-eip7702 = { version = "0.6.1", default-features = false }
-alloy-eip7928 = { version = "0.2.0", default-features = false, git = "https://github.com/rakita/alloy-eips.git", rev = "734beaf" }
+alloy-eip7928 = { version = "0.2.0", default-features = false, git = "https://github.com/alloy-rs/eips.git", rev = "734beaf" }
```
mattsse: I think we can use 0.3 here now?

mattsse (approved):
I believe this makes sense

but would like @klkvr to revisit this once more before release

crates/handler/src/precompile_provider.rs
rakita: Removed this

crates/primitives/src/hardfork.rs
rakita: Remove it, it was a small helper not essential

crates/context/src/evm.rs
rakita: Removed it

crates/context/src/cfg.rs
rakita: Made it pub

PR #3231, opened 2025-12-17
perf(database): use fixed hashmaps in cache db

shekhirin:
Alloy provides specialized hashmap types for fixed byte keys, such as `Address` and `B256`. We should use them for `CacheDB`.


rakita (approved):
awesome

rakita:
Would be good to explore if we can use these more. @DaniPopes, how do they specialise, use a different hasher?

PR #3236, opened 2025-12-18
Align example precompile logs with actual execution outcomes

maradini77:
**Description:**
- Guard success/reward messages in `example-custom-precompile-journal` so they reflect real execution results.
- Summaries now report read/write/verify status based on actual outcomes, avoiding misleading logs.
- Ensured demo output mirrors verified storage value and reward transfer status.


rakita:
It is example, no need for expand it

PR #3237, opened 2025-12-18
fix: Remove redundant `LocalContext::new` constructor

maradini77:
**Description:**
- Drop the unused `LocalContext::new` helper that only wrapped `Default`.
- Reduce API noise and avoid maintaining duplicate initialization paths.


rakita:
Having new is fine, even with default

PR #3238, opened 2025-12-18
perf: use fixed bytes hashmaps from alloy-core

shekhirin:
Follow-up to https://github.com/bluealloy/revm/pull/3231 that makes use of alloy-core fixed bytes hashmaps throughout the code for `Address`, `U256`, and `B256` keys.


rakita:
Not sure why we are seeing 5% perf degradation

rakita:
Will read it in detail, but PR makes sense

gakonst:
@shekhirin @DaniPopes @rakita do we want this?

DaniPopes:
https://github.com/bluealloy/revm/pull/3358

PR #3239, opened 2025-12-18
deps: enable foldhash feature of alloy-primitives

shekhirin:
Enables `map-foldhash` feature for `alloy-primitives`


rakita:
Closing this, as it is exposed in the revm crate, so it can be enabled if needed. If it is in default it can mess things up for dependencies

PR #3241, opened 2025-12-19
tempory modification: expose EXT

benli5510:


rakita:
Will close this PR for now, as it does not look a good way forward. Be free to reopen it if you find a better solution

PR #3242, opened 2025-12-19
perf: avoid collect

klkvr:
we can avoid a collect here

### DIFF
--- a/crates/database/src/states/state.rs
+++ b/crates/database/src/states/state.rs
@@ -386,6 +386,9 @@ impl<DB: Database> DatabaseCommit for State<DB> {
         let transitions = self.cache.apply_evm_state_iter(changes, |_, _| {});
         if let Some(s) = self.transition_state.as_mut() {
             s.add_transitions(transitions)
+        } else {
+            // Advance the iter to apply all state updates.
+            transitions.for_each(|_| {});
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/database/src/states/cache.rs
```diff
@@ -90,22 +90,20 @@ impl CacheState {
 
     /// Applies output of revm execution and create account transitions that are used to build BundleState.
     #[inline]
-    pub fn apply_evm_state<F>(
-        &mut self,
-        evm_state: impl IntoIterator<Item = (Address, Account)>,
+    pub fn apply_evm_state<'a, F, T>(
```
mattsse: not sure if this is necessary, but perhaps we should mark this as must_use

klkvr:
moved this to a separate `apply_evm_state_iter` method that's `pub(crate)` and is only used by `State<DB>`

crates/database/src/states/state.rs
```diff
@@ -383,16 +383,18 @@ impl<DB: Database> Database for State<DB> {
...
-            self.bal_state.commit_one(*address, account);
-        });
+        let transitions = self
+            .cache
+            .apply_evm_state_iter(changes, |address, account| {
+                self.bal_state.commit_one(*address, account);
+            });
         if let Some(s) = self.transition_state.as_mut() {
             s.add_transitions(transitions)
         }
```
rakita, line 406: You need here `else { transitions.collect::<Vec<_>>()}`

klkvr: oh good catch 6107473

rakita: This is why it is a footgun, very hard to catch :)

PR #3243, opened 2025-12-20
refactor(handler): eliminate code duplication in transaction validation

letmehateu:
Created a helper function `validate_priority_fee_for_tx` to reduce code duplication. The function is used for Eip1559, Eip4844, and Eip7702 transaction types, eliminating three identical calls with the same parameters.

### DIFF
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -63,7 +63,7 @@ pub fn validate_priority_fee_tx(
 
 /// Validate priority fee for transactions that support EIP-1559 (Eip1559, Eip4844, Eip7702).
 #[inline]
-fn validate_priority_fee_for_tx<TX: ContextTr>(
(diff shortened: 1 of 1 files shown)


crates/handler/src/validation.rs
```diff
@@ -61,6 +61,22 @@ pub fn validate_priority_fee_tx(
     Ok(())
 }
 
+/// Validate priority fee for transactions that support EIP-1559 (Eip1559, Eip4844, Eip7702).
+#[inline]
+fn validate_priority_fee_for_tx<CTX: ContextTr>(
+    context: &CTX,
```
rakita: You could do it with TX in place of CTX

letmehateu: > You could do it with TX in place of CTX

Changed, thank you

PR #3247, opened 2025-12-22
refactor: re-export statetest-types from revm crate behind test-types feature

rakita:
## Summary
- Update `statetest-types` to depend on individual revm crates (`primitives`, `bytecode`, `state`, `context`, `context-interface`, `database`) instead of `revm` to avoid circular dependency
- Add `statetest-types` as an optional dependency to the `revm` crate
- Add `test-types` feature flag to `revm` that enables `statetest-types` re-export
- Update `revme` to import `statetest_types` from `revm` instead of directly from `revm-statetest-types`

## Test plan
- [x] `cargo fmt --all` passes
- [x] `cargo clippy --workspace --all-targets --all-features` passes

### DIFF
--- a/crates/revm/src/lib.rs
+++ b/crates/revm/src/lib.rs
@@ -26,6 +26,10 @@ pub use primitives;
 #[doc(inline)]
 pub use state;
 
+#[cfg(feature = "test-types")]
+#[doc(inline)]
+pub use statetest_types;
+
 // Export items.
 
 pub use context::{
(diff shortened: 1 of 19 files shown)


crates/revm/src/lib.rs
```diff
@@ -26,6 +26,10 @@ pub use primitives;
 #[doc(inline)]
 pub use state;
 
+#[cfg(feature = "test-types")]
+#[doc(inline)]
+pub use statetest_types;
```
rakita, line 31: This is only part that touched revm lib, other changes are revme related and not consensus sensitive.

PR #3248, opened 2025-12-22
feat(revme): add map-foldhash feature

letmehateu:
Adds `map-foldhash` feature to revme that enables foldhash-based hash maps from alloy-primitives. This allows users to opt-in to faster hash implementation when using revme.


rakita:
revme is binary, this PR does not have effect

PR #3249, opened 2025-12-23
perf:write directly to mem rather than use copy_non_overlapping

malik672:
https://godbolt.org/z/badnnzEaT

write directly to mem rather than use copy_non_overlapping

we know the layout 

took this from swap_non_overlapping


rakita:
If you in place of `ptr.sub(n)` use `self.data.as_mut_ptr().add(len-n)` you are getting the same assembly as your PR

https://godbolt.org/z/683aPhvPc

Will close this

PR #3250, opened 2025-12-23
feat: dynamically link gmp

gakonst:
GMP gives us higher performance modexp which is necessary for long-tail Ethereum blocks e.g. involving Aztec or ZKSync.

Rug doesn't let us dynamically link GMP so as to make use of it w/o having to relicense Revm/Reth.

Instead, we bind to it over FFI dynamically using the `use-system-libs` feature of https://gitlab.com/tspiteri/gmp-mpfr-sys/-/blob/master/build.rs#L1010.

This enables using GMP in Reth

### DIFF
--- a/crates/precompile/Cargo.toml
+++ b/crates/precompile/Cargo.toml
@@ -74,7 +74,8 @@ ark-std = { workspace = true }
 rstest.workspace = true
 
 [features]
-default = ["std", "gmp", "secp256k1", "blst", "c-kzg", "portable"]
+default = ["std", "secp256k1", "blst", "c-kzg", "portable"]
+
 std = [
 	"primitives/std",
 	"k256/std",


crates/precompile/Cargo.toml
rakita:
```suggestion
default = ["std", "gmp", "secp256k1", "blst", "c-kzg", "portable"]
```

DaniPopes (approved):

crates/precompile/Cargo.toml
```diff
@@ -74,7 +74,7 @@ ark-std = { workspace = true }
 rstest.workspace = true
 
 [features]
-default = ["std", "secp256k1", "blst", "c-kzg", "portable"]
+default = ["std", "gmp", "secp256k1", "blst", "c-kzg", "portable"]
```
rakita:
```suggestion
default = ["std", "secp256k1", "blst", "c-kzg", "portable"]

```

rakita:
ci error not related and it was already passing before

PR #3261, opened 2025-12-29
perf: avoid loading bytecode in extcodehash

MoNyAvA:
`EXTCODEHASH` only needs the account’s emptiness status and code_hash, not the full bytecode. Previously we were calling 
load_account_info_skip_cold_load and `berlin_load_account!` with load_code = true, which forced bytecode loading from the database even though extcodehash only reads is_empty() and code_hash. This change switches both branches to use load_code = false, so we no longer pull code blobs into memory. The cold/warm account accounting and gas behaviour are unchanged, and the implementation remains consistent with EIP-1052/EIP-161 semantics while reducing unnecessary IO and allocations.


rakita (approved):
good find! Merging it after CI

PR #3262, opened 2025-12-29
refactor(revme): consolidate find_all_json_tests in dir_utils

letmehateu:
Eliminates code duplication by consolidating three implementations of `find_all_json_tests` into a single, complete implementation in the shared `dir_utils` module.


letmehateu:
sorry, maybe i shouldn't close my PR, because i'm not sure my changes could degrade performance.

rakita (commented):
Can you merge with main, and revert other changed.

Main change of using only one `find_all_json_tests` lgtm.

Perf is not related to this PR, hashmap is flaky

PR #3263, opened 2025-12-30
perf: add optimization for push and pop

malik672:
make push 5.4x faster
Pop not much changes but we can add likely and unlikely instrincts

https://godbolt.org/z/8a5GKj4WT

### DIFF
--- a/crates/interpreter/src/interpreter/stack.rs
+++ b/crates/interpreter/src/interpreter/stack.rs
@@ -223,7 +223,7 @@ impl Stack {
     #[cfg_attr(debug_assertions, track_caller)]
     pub fn push(&mut self, value: U256) -> bool {
         // In debug builds, verify we have sufficient capacity provisioned.
-        assume!(self.data.capacity() >= STACK_LIMIT);
+        debug_assert!(self.data.capacity() >= STACK_LIMIT);
         let len = self.data.len();
         if len == STACK_LIMIT {
             return false;


crates/interpreter/src/interpreter/stack.rs
```diff
@@ -153,7 +153,16 @@ impl Stack {
...
+        //Todo: can you likely instrincts to show the else branch is more likely to be hit
+        let len = self.data.len();
+        if len == 0 {
+            Err(InstructionResult::StackUnderflow)
+        } else {
+            unsafe {
+                self.data.set_len(len - 1);
+                Ok(core::ptr::read(self.data.as_ptr().add(len - 1)))
+            }
+        }
```
rakita, line 165: pop is very similar to this, so would leave it as is

https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#2709

malik672: True but: 
//TODO: you can likely instrincts to show the else branch is more likely to be hit

stdlib `pop` can't do this because the writers don't know what you will do with `pop`

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -165,7 +174,9 @@ impl Stack {
     #[cfg_attr(debug_assertions, track_caller)]
     pub unsafe fn pop_unsafe(&mut self) -> U256 {
         assume!(!self.data.is_empty());
-        self.data.pop().unwrap_unchecked()
+        let new_len = self.data.len() - 1;
+        self.data.set_len(new_len);
+        core::ptr::read(self.data.as_ptr().add(new_len))
```
rakita: same with pop_unsafe, underlying code is probably same

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -215,10 +226,15 @@ impl Stack {
...
+        let len = self.data.len();
+        if len == STACK_LIMIT {
             return false;
         }
-        self.data.push(value);
+        unsafe {
+            let end = self.data.as_mut_ptr().add(len);
+            core::ptr::write(end, value);
+            self.data.set_len(len + 1);
+        }
```
rakita, line 235: This is nice! we didn't optimize this function as it is not used that much, PUSH opcodes use optimized push_slice

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -215,10 +226,15 @@ impl Stack {
     pub fn push(&mut self, value: U256) -> bool {
         // In debug builds, verify we have sufficient capacity provisioned.
         debug_assert!(self.data.capacity() >= STACK_LIMIT);
```
rakita, line 226: It looks there was a bug here

```suggestion
        debug_assert!(self.data.capacity() > STACK_LIMIT);
```

Capacity equal to STACK_LIMIT is checked in the line below.

crates/interpreter/src/interpreter/stack.rs
rakita: Branch prediction would do its thing here, but let's try it, there is ` if primitives::hints_util::unlikely(len == 0) {` that can be used here

crates/interpreter/src/interpreter/stack.rs
malik672: Mistake, it's meant to be `assume`

crates/interpreter/src/interpreter/stack.rs
malik672: Nahhh in the old `pop_unsafe` does `if else` if len == 0 `pop` panicks: safety
However in the new `pop_unsafe` if `len == 0` we get a `UB`

malik672: should I leave it or change it ?

crates/interpreter/src/interpreter/stack.rs
malik672: done

crates/interpreter/src/interpreter/stack.rs
rakita: `unwrap_unchecked` would be `UB` as it does not panic, would leave it as it is

malik672: done

crates/interpreter/src/interpreter/stack.rs
rakita: Will revert this line, it was correct previously. `assume` would not do anything here as we are handling pointers.

crates/interpreter/src/interpreter/stack.rs
```diff
@@ -214,11 +223,16 @@ impl Stack {
     #[cfg_attr(debug_assertions, track_caller)]
     pub fn push(&mut self, value: U256) -> bool {
         // In debug builds, verify we have sufficient capacity provisioned.
-        debug_assert!(self.data.capacity() >= STACK_LIMIT);
-        if self.data.len() == STACK_LIMIT {
+        assume!(self.data.capacity() >= STACK_LIMIT);
```
rakita:
```suggestion
        debug_assert!(self.data.capacity() >= STACK_LIMIT);
```

PR #3264, opened 2025-12-30
fix(database): make DatabaseCommit dyn-compatible

haythemsellami:
# Summary

This PR fixes dyn-compatibility of DatabaseCommit for downstream users like Foundry that use dyn DatabaseExt (which extends DatabaseCommit).

# Problem

While integrating recent revm codebase with Foundry, I discovered that DatabaseCommit is not dyn-compatible due to how #[auto_impl] generates implementations.

Foundry defines DatabaseExt which extends DatabaseCommit:

```rust
pub trait DatabaseExt: Database<Error = DatabaseError> + DatabaseCommit + Debug { ... }
```

And uses it as a trait object throughout the codebase:

```rust
  pub struct FoundryEvm<'db, I: InspectorExt> {
      db: &'db mut dyn DatabaseExt,
      // ...
  }
```

The issue is that `#[auto_impl(&mut, Box)]` generates implementations like:

```rust
  impl<T: DatabaseCommit + ?Sized> DatabaseCommit for &mut T {
      fn commit_iter(&mut self, changes: impl IntoIterator<Item = (Address, Account)>) {
          T::commit_iter(*self, changes) 
      }
  }
```

This fails because `commit_iter` uses `impl` Trait in argument position (which requires Self: Sized). When T is ?Sized (like dyn DatabaseCommit), the generated code tries to call a method that requires Sized on a ?Sized type.

# Solution

1. commit_iter now uses `&mut dyn Iterator`: this keeps the method in the vtable and callable on trait objects
2. impl dyn DatabaseCommit provides ergonomic commit_from_iter(impl IntoIterator) wrapper
3. Add tests for dyn-compatibility including a compile-time check mirroring Foundry's approach

### DIFF
--- a/crates/database/interface/src/lib.rs
+++ b/crates/database/interface/src/lib.rs
@@ -10,7 +10,7 @@ use core::convert::Infallible;
 use auto_impl::auto_impl;
 use primitives::{address, Address, HashMap, StorageKey, StorageValue, B256, U256};
 use state::{Account, AccountInfo, Bytecode};
-use std::{boxed::Box, vec::Vec};
+use std::vec::Vec;
 
 /// Address with all `0xff..ff` in it. Used for testing.
 pub const FFADDRESS: Address = address!("0xffffffffffffffffffffffffffffffffffffffff");
@@ -87,12 +87,9 @@ pub trait Database {
 ///
 /// # Dyn Compatibility
 ///
-/// This trait is dyn-compatible. The `commit_iter` method has a `Self: Sized` bound
-/// which excludes it from the trait's vtable, allowing `dyn DatabaseCommit` to work.
-///
-/// Note: We manually implement this trait for `&mut T` and `Box<T>` instead of using
-/// `#[auto_impl]` because auto_impl generates impls for `T: ?Sized` that try to delegate
-/// `commit_iter` to the inner type, which fails when `T` is `?Sized` (like `dyn DatabaseCommit`).
+/// This trait is dyn-compatible. The `commit_iter` method uses `&mut dyn Iterator`
+/// which allows it to be called on trait objects while remaining in the vtable.
+#[auto_impl(&mut, Box)]
 pub trait DatabaseCommit {
     /// Commit changes to the database.
     fn commit(&mut self, changes: HashMap<Address, Account>);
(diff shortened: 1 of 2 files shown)


crates/database/interface/src/lib.rs
```diff
@@ -244,7 +307,9 @@ pub trait DatabaseCommitExt: Database + DatabaseCommit {
             // Unfortunately must collect here to short circuit on error
             .collect::<Result<Vec<_>, _>>()?;
 
-        self.commit_iter(transitions);
+        // Use commit() instead of commit_iter() since we already have a collected Vec.
+        // This keeps these methods dyn-compatible without requiring Self: Sized.
+        self.commit(transitions.into_iter().collect());
```
rakita: This is going to be a perf penalty as it now recreates the HashMap. Should we enforce Sized here too?

rakita: Have dug litttle bit deeper on this, adding Sized basically removed this function from `Box<dyn Trait>`. Adding `&mut dyn IntoIterator` would allowed for trait to be boxable but still to have IntoIter function. Assumption here is that commit_iter would handle a lot of accesses to hashmaps, so the impact of dynamic dispatch of dyn would not be noticeable.

There is an additional nice small trick that we can do, this should make api litle nicer, but didn't try it to to be sure.

```
impl dyn DatabaseCommit {
    pub fn temp(&mut self, iter: impl IntoIterator<..>) {}
}

Comment (haythemsellami) [reply crates/database/interface/src/lib.rs]: I added `impl dyn DatabaseCommit` with a `commit_from_iter` fn, updated both `increment_balances()` and `drain_balances()` to use HashMap directly to avoid double collection.

File: crates/database/interface/src/lib.rs [hunk 2]
```diff
@@ -94,12 +102,84 @@ pub trait DatabaseCommit {
...
+    /// which is not dyn-compatible. Without this bound, the entire `DatabaseCommit` trait
+    /// would become not dyn-compatible, breaking code that uses `dyn DatabaseCommit` or
+    /// traits that extend it (e.g., Foundry's `DatabaseExt`).
+    ///
+    /// The `Sized` bound excludes this method from the trait's vtable while keeping the
+    /// trait itself dyn-compatible. Code using trait objects should use [`commit`](Self::commit)
+    /// instead, collecting the iterator into a `HashMap` first if needed.
+    fn commit_iter(&mut self, changes: impl IntoIterator<Item = (Address, Account)>)
+    where
+        Self: Sized,
```
rakita: This should be a `&mut dyn IntoIterator` so we can remove `Self: Sized` and allow usage of this function in parent traits

crates/database/interface/src/lib.rs
```diff
@@ -241,10 +322,9 @@ pub trait DatabaseCommitExt: Database + DatabaseCommit {
                 original_account.mark_touch();
                 Ok((address, original_account))
             })
-            // Unfortunately must collect here to short circuit on error
-            .collect::<Result<Vec<_>, _>>()?;
+            .collect::<Result<_, _>>()?;
```
rakita: We should revert this to vec, collecting to HashMap is a lot more expensive than collecting Vec.

I would try using `commit_from_iter` or figure out what needs to be changed here

crates/database/interface/src/lib.rs
haythemsellami: Do u mean here `&mut dyn Iterator` ?

haythemsellami:
@rakita Lmk if the below is a good path, will update the PR description later:
  1. commit_iter now uses `&mut dyn Iterator` instead of `impl IntoIterator with Self: Sized`
    - This keeps the method in the vtable and callable on trait objects
    - No more Self: Sized bound needed
  2. Re-enabled `#[auto_impl(&mut, Box)]` on `DatabaseCommit`: manual impls no longer needed
  3. `increment_balances` / `drain_balances` reverted to collect to Vec, then call commit_iter
  4. impl dyn DatabaseCommit provides ergonomic commit_from_iter(impl IntoIterator) wrapper

crates/database/interface/src/lib.rs
```diff
@@ -244,7 +270,7 @@ pub trait DatabaseCommitExt: Database + DatabaseCommit {
             // Unfortunately must collect here to short circuit on error
             .collect::<Result<Vec<_>, _>>()?;
 
-        self.commit_iter(transitions);
+        self.commit_iter(&mut transitions.into_iter());
```
rakita, line 273: Can this work with `commit_from_iter`?

haythemsellami: Tried that, but `commit_from_iter` is a method on dyn DatabaseCommit, not a trait method. (increment_balances/drain_balances uses Self: DatabaseCommit)

This line doing the same thing as if we will be using `commit_from_iter` anyway...

PR #3265, opened 2025-12-30
perf: add Display for HaltReason and OutOfGasError

RyanBGG:
Adds Display implementations for HaltReason and OutOfGasError to improve error messages. 

Previously these types only had Debug, so errors showed up as `Halt { reason: OutOfGas(Basic) }` instead of readable messages like "out of gas".


RyanBGG:
BTW This was inconsistent with InvalidTransaction and InvalidHeader which already have Display impls.

PR #3271, opened 2026-01-05
chore(interpreter): remove unused StackTr::is_empty

Olexandr88:
Remove unused StackTr::is_empty default method
No behavior changes, cleanup only


rakita:
Clippy lint, and it is good function

PR #3273, opened 2026-01-05
docs: clarify timestamp default in BlockEnv

RyanBGG:
Was reading through the block env code and noticed timestamp is set to ONE while everything else is ZERO. Took me some time to figure out if there's a technical reason for this.

Added a quick comment to save future readers the confusion. If there's actually a specific reason it needs to be 1 (like some validation I missed or edge case), let me know and I can update the comment to be more specific.

### DIFF
--- a/crates/context/src/block.rs
+++ b/crates/context/src/block.rs
@@ -104,6 +104,7 @@ impl Default for BlockEnv {
         Self {
             number: U256::ZERO,
             beneficiary: Address::ZERO,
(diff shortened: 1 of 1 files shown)


rakita:
Every block field need to be set to something to make sense, one is slightly better than zero as used in instruction, but there is no real reason why it is like that

RyanBGG:
> Every block field need to be set to something to make sense, one is slightly better than zero as used in instruction, but there is no real reason why it is like that

Make sense! thanks for explanation!

PR #3274, opened 2026-01-05
chore: remove unused InterpreterAction::new_stop helper

Snezhkko:
InterpreterAction::new_stop was never used and created an InterpreterResult::Stop with Gas::new(0), which is inconsistent with all other interpreter exit paths that preserve the actual gas state. This helper is not part of the documented public API and having it around is a footgun for external users


rakita:
WIll leave it as is for now

Issue #3277, opened 2026-01-06
Make a bytecode test utility that would make bytecode that insert input directly to memory


rakita:
If we want to insert 0x322..322 into memory from bytecode we need to do `PUSH32 __, MSTORE, PUSH20 __, MSTORE` etc. Make a utility fn that would take arbitrary input and offset and would make return arrey of opcodes that would push mstore this input into memory.

PR #3280, opened 2026-01-06
fix: add a test utility function to generate memory input opcodes

silathdiir:
close https://github.com/bluealloy/revm/issues/3277

### DIFF
--- a/crates/bytecode/Cargo.toml
+++ b/crates/bytecode/Cargo.toml
@@ -46,7 +46,13 @@ std = [
   "phf?/std",
 ]
 hashbrown = ["primitives/hashbrown"]
-serde = ["dep:serde", "primitives/serde", "bitvec/serde", "phf?/serde"]
+serde = [
+	"dep:serde",
+	"primitives/serde",
+	"bitvec/serde",
+	"phf?/serde",
+	"rand/serde"
+]
 parse = ["phf", "paste"]
 
 # Deprecated, please use `serde` feature instead.


crates/bytecode/src/utils.rs
```diff
@@ -19,3 +19,60 @@ pub unsafe fn read_i16(ptr: *const u8) -> i16 {
...
+/// Bytecode test utilities
+#[cfg(test)]
+pub mod test {
+    use crate::opcode;
+    use alloy_primitives::{hex, Bytes};
+    use anyhow::Result;
+    use std::str::FromStr;
+
+    /// Constructs bytecode for inserting input into memory
+    pub fn build_memory_input_opcodes(hex_input: &str) -> Result<Vec<u8>> {
```
rakita: No need for this to be string,  &[u8] is fine. Additionally it should have a `offset`

crates/bytecode/src/utils.rs
```diff
@@ -19,3 +19,60 @@ pub unsafe fn read_i16(ptr: *const u8) -> i16 {
...
+        let input_bytes = Bytes::from_str(hex_input)?;
+
+        // Iterate for each 32 bytes to prepend PUSH* and append MSTORE opcodes
+        for chunk in input_bytes.chunks(32) {
+            let push_opcode = opcode::PUSH0 + chunk.len() as u8;
+            opcodes.push(push_opcode);
+
+            opcodes.extend_from_slice(chunk);
+
+            opcodes.push(opcode::MSTORE);
```
rakita, line 46: MSTORE uses two items, where one is offset https://www.evm.codes/

crates/bytecode/src/utils.rs
```diff
@@ -19,3 +19,60 @@ pub unsafe fn read_i16(ptr: *const u8) -> i16 {
 pub unsafe fn read_u16(ptr: *const u8) -> u16 {
     u16::from_be_bytes(unsafe { ptr.cast::<[u8; 2]>().read() })
 }
+
+/// Bytecode test utilities
+#[cfg(test)]
+pub mod test {
+    use crate::opcode;
+    use alloy_primitives::{hex, Bytes};
```
rakita: use `primitives::Bytes` no need to introduce alloy_primitives or anyhow

crates/bytecode/src/utils.rs
silathdiir: Fixed in commit [35ea2a2](https://github.com/bluealloy/revm/pull/3280/commits/35ea2a2c3c6fb48e3aada54782f6f8ad857e27cb)

crates/bytecode/src/utils.rs
silathdiir: Fixed in commit [35ea2a2](https://github.com/bluealloy/revm/pull/3280/commits/35ea2a2c3c6fb48e3aada54782f6f8ad857e27cb)

crates/bytecode/src/utils.rs
silathdiir: Removed `alloy-primitives` dev-dependency in commit [35ea2a2](https://github.com/bluealloy/revm/pull/3280/commits/35ea2a2c3c6fb48e3aada54782f6f8ad857e27cb), since fixed the argument to bytes.

silathdiir:
Hi @rakita , I updated this PR according to your code review, please help review again when you have time.

PR #3281, opened 2026-01-06
feat: improve error reporting

RyanBGG:
Saw reth recently updated their error output with colors and it looks pretty nice. Figured we could use the same approach here.

This swaps out the basic eprintln! for color-eyre, which gives us colored error messages and better formatted backtraces. makes debugging much easier.

ref: https://github.com/paradigmxyz/reth/blob/main/Cargo.toml#L551


rakita:
I will be agains this, output of the revme is used by some tooling to parse json so I am not sure if adding colour would disrupt this i would close this PR

PR #3284, opened 2026-01-07
fix(database): return error instead of panic when block not found in AlloyDB

### LINKED ISSUE #3278: AlloyDB: block_hash_async_ref may panic on pruned nodes or missing blocks
## Description

In `crates/database/src/alloydb.rs`, the `block_hash_async_ref` function uses 
`unwrap()` which may panic if the block doesn't exist:

```rust
async fn block_hash_async_ref(&self, number: u64) -> Result<B256, Self::Error> {
    let block = self
        .provider
        .get_block_by_number(number.into())
        .await?;
    // SAFETY: If the number is given, the block is supposed to be finalized...
    Ok(B256::new(*block.unwrap().header().hash()))  // <- May panic
}## Problem
```

While the SAFETY comment assumes blocks will always exist, this may not hold true for:
- **Pruned nodes**: Old block data may be deleted
- **Light clients**: May not have all block data
- **Invalid block numbers**: Requesting future or non-existent blocks

## Suggested Solutions

1. **Improve documentation**: Make the assumptions and limitations clearer
2. **Add error handling**: Return an error instead of panicking

doocho:
- Change DBTransportError struct to AlloyDBError enum
- Add BlockNotFound variant for missing block errors
- Handle None case in block_hash_async_ref gracefully

Closes #3278

### DIFF
--- a/crates/database/src/alloydb.rs
+++ b/crates/database/src/alloydb.rs
@@ -20,7 +20,6 @@ pub enum AlloyDBError {
     /// Block not found for the given block number.
     ///
     /// This can occur when:
-    /// - The block number is in the future
     /// - The node has pruned the block data
     /// - Using a light client that doesn't have the block
     BlockNotFound(u64),


crates/database/src/alloydb.rs
```diff
@@ -12,23 +12,43 @@ use primitives::{Address, StorageKey, StorageValue, B256};
...
+/// Error type for AlloyDB database operations.
 #[derive(Debug)]
-pub struct DBTransportError(pub TransportError);
+pub enum AlloyDBError {
+    /// Transport error from the underlying provider.
+    Transport(TransportError),
+    /// Block not found for the given block number.
+    ///
+    /// This can occur when:
+    /// - The block number is in the future
```
rakita:
```suggestion
```

This option is not true as evm will request only last 256 block hashes from the current block.

PR #3285, opened 2026-01-07
perf: optimize Gas::record_cost with unlikely hint

joeybright53:
Replaces checked_sub with direct comparison in Gas::record_cost, adds unlikely hint for the out-of-gas branch to improve branch prediction.


rakita:
codspeed does not show any improvement, and as this is in a very hot path, the branch prediction will do its thing very well.

Will close this PR and leave the code as is.

PR #3286, opened 2026-01-07
fix(bal): ensure sorted writes in Alloy type conversions

doocho:
## Summary
`From`/`TryFrom` implementations for Alloy EIP-7928 types now use `Self::new()` 
to ensure writes are sorted by `block_access_index`.

## Problem
`BalWrites::get()` relies on sorted data for correct binary/linear search.
However, Alloy types (`Vec<BalanceChange>`, etc.) do not guarantee ordering.
Using direct struct initialization (`Self { writes: ... }`) bypassed sorting,
potentially causing incorrect lookups.

## Solution
Replace `Self { writes: ... }` with `Self::new(...)` which sorts the data.

## Changes
- `alloy.rs`: 4 conversions updated (`BalanceChange`, `NonceChange`, `StorageChange`, `CodeChange`)


rakita:
Will close this for now, as sorting introduces a penalty.

PR #3287, opened 2026-01-07
perf(state): remove redundant clones in account_info.rs

andrewshab3:
Removes three redundant `.clone()` calls


rakita:
Those are tests

PR #3291, opened 2026-01-08
fix: correct debug log key/value for blob gas fields in blockchaintest

maximevtush:
The debug output in blockchain test error messages incorrectly labeled the log key as `"Blob excess gas"` while printing the entire `BlobExcessGasAndPrice` struct, which contains both `excess_blob_gas` and `blob_gasprice` fields.

Split the log output into two separate lines with correct labels:
- `"Blob excess gas"` now correctly shows only `excess_blob_gas`
- `"Blob gas price"` now correctly shows only `blob_gasprice`


maximevtush:
@rakita why was added the ai label?

rakita:
I aggregate those small typos into one big PR, idea is to remove typos PR and ai bots that farm them.

PR #3292, opened 2026-01-08
feat(inspector): add test bootstrap for EVM variants

### LINKED ISSUE #3275: Add Inspector testing bootstrap for evm variants that extend revm
Make TestInspector public and add default test that we can run on top of it.

https://github.com/bluealloy/revm/blob/f383efd9fe8f5bcdf7b6d1ec7715c91e411f7895/crates/inspector/src/inspector_tests.rs#L47-L53

Default test need to depend on execution API

RyanBGG:
Closes #3275

Added default_tests module with run_tests and run_extended_tests functions that EVM variants can use to verify their inspector integration. 

Test suite covers CALL, CREATE, LOG, JUMP operations and precompile calls as mentioned in the review PR. Also added verify_callbacks helper to check if all inspector hooks are being called properly.

Usage example in inspector_tests.rs shows how to run the tests with a closure that sets up your EVM and returns the inspector.

### DIFF
--- a/crates/inspector/src/inspector_tests.rs
+++ b/crates/inspector/src/inspector_tests.rs
@@ -628,4 +628,60 @@ mod tests {
 
         assert!(evm.inspector.get_step_count() > 0);
     }
+
+    #[test]
+    fn test_default_tests_bootstrap() {
+        use crate::test_inspector::default_tests;
+
+        // Example: run basic tests
+        let result = default_tests::run_tests(|bytecode| {
+            let bytecode = Bytecode::new_raw(bytecode);
+            let ctx = Context::mainnet().with_db(BenchmarkDB::new_bytecode(bytecode));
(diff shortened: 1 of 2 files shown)


Suryakant-Soni:
Hey @RyanBGG , thanks for picking this up, I was checking #3275  and noticed rakita's [review comment](https://github.com/bluealloy/revm/pull/3282#pullrequestreview-3639613386) on #3282 , a few things are not yet added I think like 

- `verify_callbacks` only checks `step` / `step_end` ( not the other 8 hooks ) , `initialize_interp` isn't recorded, the runners are not calling `verify_callbacks`, and no bytecode triggers `SELFDESTRUCT`.

Since this PR is open for quite a while, I'd like to open a follow-up PR build on top of your commit so your authorship is preserved. Let me know in the next few days if you are actively working on it, otherwise I would go ahead, thanks.

PR #3295, opened 2026-01-09
test: reusing state / state cache across blocks

datnguyencse:
Hi, thanks a lot for `revm` - great work 🙏

I have two short, related questions about **reusing `State` / state cache across blocks**.

#### 1. Storage fallback when reusing cache

In my use case, I want to **reuse the state cache across blocks** for performance.

I noticed that when a storage slot is not found in cache, but the cached account has status
`InMemoryChange` or `DestroyedChanged`, storage reads may return incorrect values if the cache is reused for a new block.

I made a small PR so that in this case, `revm` **falls back to the underlying DB** when the value is missing from cache.

Does this behavior align with the intended cache semantics, or does it break any assumptions in `revm`?

Refer Test: https://github.com/bluealloy/revm/pull/2196/files

#### 2. Safety of reusing State across blocks

More generally, is it **safe to reuse `State` across blocks** if the bundle and transitions are reset?

If not, is it at least safe to reuse **only the state cache** across blocks?

My assumption is that as long as missing data can correctly fall back to the DB, correctness should be preserved - but I’d appreciate your confirmation.

Thanks a lot for your guidance!

### DIFF
--- a/crates/database/src/states/account_status.rs
+++ b/crates/database/src/states/account_status.rs
@@ -58,9 +58,7 @@ impl AccountStatus {
     pub fn is_storage_known(&self) -> bool {
         matches!(
             self,
-            Self::LoadedNotExisting
-                | Self::Destroyed
-                | Self::DestroyedAgain
+            Self::LoadedNotExisting | Self::Destroyed | Self::DestroyedAgain
         )
     }
 
--- a/crates/database/src/states/state.rs
+++ b/crates/database/src/states/state.rs
@@ -1029,7 +1029,8 @@ mod tests {
 
         // Block 1: Deploy the contract with an initial slot.
         let first_block_slot_key = U256::from(1);
-        let first_block_slot_value = state::EvmStorageSlot::new_changed(U256::ZERO, U256::from(1), 0);
+        let first_block_slot_value =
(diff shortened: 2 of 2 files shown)


rakita:
Was closed accidentally as it was tagged in a different PR that got merged. Sorry about that

rakita:
Change does not make sense, all statuses represent known storage so it does not make sense to fetch it from database.

1. How do you see that it is a wrong value. Assumption is that all storage values are inside cache (as account got created or destroyed inside the block), and that cache will not evict values from it.

2. Should be safe, and iirc it was used inside reth for initial sync to run one instance accross blocks. But would test this to be sure

datnguyencse:
Thanks for the explanation.

In the case of reth, they reuse state to sync a historical branch of blocks, then merge transitions and extract the bundle only at the last block. In that workflow, reusing state does not cause issues.

However, in our case (live sync), bundles and transitions are merged after each block. Under this model, reusing the state (or the state cache) across blocks can lead to incorrect values and errors.

Hopefully, you’ll be able to take this difference into account when testing and verifying the behavior. Reusing state between blocks is extremely important for correctness in live sync scenarios.

Thank you very much for your time and for looking into this.

PR #3302, opened 2026-01-14
revme/blockchaintest: clarify summary counters

leopardracer:
The previous summary mixed units (tests vs files), which could mislead both humans and scripts consuming `--json`.

- Fix `revme blockchaintest` summary counters by reporting **passed tests** separately from **failed/skipped files**
- Update JSON summary keys to match the actual units (`passed_tests`, `failed_files`, `skipped_files`)
- Rename the final error from `TestsFailed` to `FilesFailed` to reflect the real meaning


rakita:
Will be added to next batch. Thanks

PR #3307, opened 2026-01-14
perf(interpreter): tail-call dispatch optimization

gakonst:
## Summary

Replaces interpreter loop with trampoline-based dispatch to reduce function call overhead and improve branch prediction.

**Expected gain: 3-5%**

## Implementation

Uses stable Rust trampoline pattern:
- `StepResult` enum: `Continue` or `Finished`
- Main loop acts as trampoline, repeatedly "tail-calling" next step
- Unrolled variant (4x) for additional gains

## Key Components

- `step_with_result()`: executes one instruction, returns StepResult
- `run_with_tail_call()`: main trampoline loop
- `run_with_tail_call_unrolled()`: 4x unrolled variant
- `TailCallInterpreter` trait: extension trait for Interpreter

## Benefits

- Reduced loop condition checking overhead
- Better branch prediction (linear vs 256-way switch)
- Reduced return address stack pressure
- Compiler optimization opportunities

## Note

True tail-call elimination requires Rust's experimental `become` keyword (nightly-only). This uses the stable trampoline approach.

## Inspiration

Based on analysis of [guillotine EVM](https://github.com/evmts/guillotine)'s tail-call dispatch architecture.

### DIFF
--- a/crates/interpreter/src/lib.rs
+++ b/crates/interpreter/src/lib.rs
@@ -25,6 +25,10 @@ pub mod interpreter_action;
 /// Type traits and definitions for interpreter customization.
 pub mod interpreter_types;
 
+/// Tail-call dispatch optimization - reduces interpreter loop overhead.
+/// Expected gain: 3-5%.
+pub mod tail_call;
+
 // Reexport primary types.
 pub use context_interface::{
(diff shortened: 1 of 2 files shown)


DaniPopes:
This is not even correct

PR #3312, opened 2026-01-14
refactor(handler): extract duplicate ContextError handling

andrewshab3:
Introduces `take_error` helper to deduplicate `ContextError` unwrapping logic


andrewshab3:
@rakita sorry, can you check please?

PR #3299, opened 2026-01-10
perf: Use O(1) ring buffer cache for block hashes instead of BTreeMap

malik672:
Replaces the `BTreeMap<u64, B256>` used for block hash caching in `State` with a new `BlockHashCache` struct that implements a fixed-size ring buffer(Technically anyway)

from benchmark: 
```
Benchmark | Ring Buffer | BTreeMap | Speedup
-- | -- | -- | --
Insert 256 blocks | 382 ns | 4.35 µs | 11.4×
Insert 1000 continuous | 1.17 µs | 30.2 µs | 25.8×
Sequential get (256) | 93.8 ns | 1.58 µs | 16.8×
Get last 10 blocks | 6.37 ns | 82.1 ns | 12.9×
Get miss (256) | 94.3 ns | 2.60 µs | 27.6×
Single lookup | 1.11 ns | 10.2 ns | 9.2×
Mixed workload | 1.46 ns | 31.4 ns | 21.5×

```

PR #3319, opened 2026-01-16
fix(database): BlockHashCache incorrectly returns zero for block 0

bshastry:
Fixes regression caused by #3299 

The ring buffer cache introduced in #3299 was initialized with `(0, B256::ZERO)` for all entries. This caused `get(0)` to incorrectly match the default entry and return `B256::ZERO` instead of `None`, preventing the database fallback from being invoked.

Fix: Use `Option<B256>` instead of `B256` to explicitly distinguish between "not cached" (`None`) and "cached with value" (`Some(hash)`).

### DIFF
--- a/crates/database/src/states/block_hash_cache.rs
+++ b/crates/database/src/states/block_hash_cache.rs
@@ -38,14 +38,13 @@ impl BlockHashCache {
 
     /// Retrieves the block hash for the given block number, if it exists in the cache.
     #[inline]
-    pub const fn get(&self, block_number: u64) -> Option<B256> {
+    pub fn get(&self, block_number: u64) -> Option<B256> {
         let index = (block_number % BLOCK_HASH_HISTORY) as usize;
         let (stored_block_number, stored_hash) = self.hashes[index];
-        if let Some(stored) = stored_block_number {
-            if stored == block_number {
-                return Some(stored_hash);
-            }
+        if Some(block_number) == stored_block_number {
+            Some(stored_hash)
+        } else {
+            None
         }
-        None
     }
 }


crates/database/src/states/block_hash_cache.rs
```diff
@@ -2,13 +2,16 @@ use primitives::{alloy_primitives::B256, BLOCK_HASH_HISTORY};
...
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BlockHashCache {
     /// A fixed-size array holding the block hashes.
     /// Since we only store the most recent 256 block hashes, this array has a length of 256.
-    /// The reason we store block number alongside its hash is to handle the case where it wraps around, so we can verify the block number.
-    hashes: Box<[(u64, B256); BLOCK_HASH_HISTORY_USIZE]>,
+    /// The reason we store block number alongside its hash is to handle the case where it wraps around,
+    /// so we can verify the block number. Uses `Option<B256>` to distinguish between "not cached"
+    /// and "cached with value" - this is important because block 0 is a valid block number.
+    hashes: Box<[(u64, Option<B256>); BLOCK_HASH_HISTORY_USIZE]>,
```
rakita: A slightly better approach would be to have Option<u64> as block number is used to deduce if block hash is up to date

crates/database/src/states/block_hash_cache.rs
```diff
@@ -22,24 +25,24 @@ impl BlockHashCache {
...
         Self {
-            hashes: Box::new([(0, B256::ZERO); BLOCK_HASH_HISTORY_USIZE]),
+            hashes: Box::new([(0, None); BLOCK_HASH_HISTORY_USIZE]),
         }
     }
 
     /// Inserts a block hash for the given block number.
     #[inline]
-    pub const fn insert(&mut self, block_number: u64, block_hash: B256) {
+    pub fn insert(&mut self, block_number: u64, block_hash: B256) {
```
rakita: No need to remove consts, right?

crates/database/src/states/block_hash_cache.rs
```diff
@@ -22,26 +25,27 @@ impl BlockHashCache {
...
     /// Retrieves the block hash for the given block number, if it exists in the cache.
     #[inline]
     pub const fn get(&self, block_number: u64) -> Option<B256> {
         let index = (block_number % BLOCK_HASH_HISTORY) as usize;
         let (stored_block_number, stored_hash) = self.hashes[index];
-        if stored_block_number == block_number {
-            Some(stored_hash)
-        } else {
-            None
+        if let Some(stored) = stored_block_number {
```
rakita: Can be simplified with if `Some(block_number) == stored_block_number {`

PR #3325, opened 2026-01-18
deps: update lockfile

Olexandr88:
Updates Cargo.lock with latest compatible dependency versions, cargo-audit reports one transitive advisory without patched version


rakita:
I need to do this on my own, can't be external PR

PR #3335, opened 2026-01-20
fix(bytecode): improve analyze_legacy and stack tests

rakita:
## Summary
- Improve `analyze_legacy` padding logic for SWAPN/DUPN/EXCHANGE edge cases using `prev_opcode` tracking
- Simplify the main while loop to only handle PUSH opcodes
- Update stack.rs tests for new Interpreter API

## Test plan
- Existing tests pass including new edge case tests for SWAPN/DUPN/EXCHANGE with STOP

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -47,11 +47,11 @@ pub fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
     let overflow = (iterator as usize) - (end as usize);
     let mut padding = overflow + (opcode != opcode::STOP) as usize;
 
-    // Additional padding if previous opcode is SWAPN/DUPN/EXCHANGE and bytecode
-    // doesn't end with STOP. This handles edge cases like [SWAPN, STOP] where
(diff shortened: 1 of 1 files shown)


rakita:
This fixed the inconsistency in padding that immediates would do

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -201,6 +201,13 @@ mod tests {
             assert_eq!(padded_bytecode[0], op);
             assert_eq!(padded_bytecode[1], opcode::STOP);
             assert_eq!(padded_bytecode[2], opcode::STOP);
+
+            // test when there is only opcode with immediate data missing.
+            let bytecode = vec![op];
+            let (_, padded_bytecode) = analyze_legacy(bytecode.into());
+            assert_eq!(padded_bytecode.len(), 2);
+            assert_eq!(padded_bytecode[0], op);
+            assert_eq!(padded_bytecode[1], opcode::STOP);
```
frangio, line 210: Is this right? This corresponds to a single instruction. My understanding was you always want a final STOP instruction. In that case, you'd expect 2 bytes of padding where the first is an immediate and the second is an instruction.

PR #3338, opened 2026-01-20
fix(bytecode): add padding for final opcode missing immediates

frangio:
Follow up to https://github.com/bluealloy/revm/pull/3335#discussion_r2709408565

The test incorrectly accepted `[OPCODE] -> [OPCODE, STOP]` when opcode is DUPN/SWAPN/EXCHANGE.

Renamed variables from `opcode`/`prev_opcode` to `last/prev_byte` because the byte is not guaranteed to be an instruction opcode (it could be an immediate).


rakita (approved):
lgtm

PR #3348, opened 2026-01-25
fix: remove redundant balance check in self-transfer

zeroprooff:
When `from == to` in `transfer_loaded()`, the balance check was pointless. Self-transfer doesn't change anything - subtract and add cancel out. Removed the unnecessary `get_mut` call and balance comparison.


rakita:
This is by consensus

PR #3357, opened 2026-01-27
fix(bal): fix populated account if pre-state was none

jsign:
When basic was `None`, the function populated account info from BAL but returned early without writing it back, causing the data to be lost.

Realized about this while doing some experimentation coding a subblock-prover for Reth in the context of zkVMs, and trying to run some BAL tests.

cc @mattsse

### DIFF
--- a/crates/database/interface/src/bal.rs
+++ b/crates/database/interface/src/bal.rs
@@ -130,9 +130,16 @@ impl BalState {
     #[inline]
     pub fn basic_by_account_id(&self, account_id: usize, basic: &mut Option<AccountInfo>) -> bool {
         if let Some(bal) = &self.bal {
+            let is_none = basic.is_none();
             let mut bal_basic = core::mem::take(basic).unwrap_or_default();
-            bal.populate_account_info(account_id, self.bal_index, &mut bal_basic)
+            let changed = bal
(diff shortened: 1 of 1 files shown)


crates/database/interface/src/bal.rs
```diff
@@ -130,16 +130,10 @@ impl BalState {
     #[inline]
     pub fn basic_by_account_id(&self, account_id: usize, basic: &mut Option<AccountInfo>) -> bool {
         if let Some(bal) = &self.bal {
-            let is_none = basic.is_none();
             let mut bal_basic = core::mem::take(basic).unwrap_or_default();
             bal.populate_account_info(account_id, self.bal_index, &mut bal_basic)
                 .expect("Invalid account id");
 
-            // if it is not changed, check if it is none and return it.
-            if is_none {
```
rakita, line 139: There was a problem around empty and non existing account for eip7702 this is why this condition exist here.

But condition is not good, it should take bool from `let changed = bal.populate_account_info` and only `if !changed && is_none { return true;}` it true (aka it got changed) put basic to Some(bal_basic)

jsign: Makes sense! Pushed a commit to consider the `changed` criteria.

PR #3360, opened 2026-01-27
refactor(precompile): simplify array parsing in blake2 precompile

andrewshab3:
Simplifies parsing of state vector `h` and message block `m` in the Blake2F precompile by replacing verbose iterator chains with `core::array::from_fn`.


rakita:
Previous is more readable

PR #3364, opened 2026-01-28
perf: Replace HashSet/BitVec with u64 bitmask/Hashset for precompile lookups

malik672:
ReplacesHashSet<Address> + BitVec precompile storage with a simple u64 bitmap.
If Mainnet is the most popular use for REVM it's possible to seperate mainnet execution and only use bitmap to handle precompiles


malik672:
stupid git issues

PR #3368, opened 2026-01-28
perf: Replace HashSet/BitVec with u64 bitmask/Hashset for precompile lookups

malik672:
ReplacesHashSet

+ BitVec precompile storage with a simple u64 bitmap.
If Mainnet is the most popular use for REVM it's possible to seperate mainnet execution and only use bitmap to handle precompiles


rakita:
Current approach looks okay

malik672:
> Current approach looks okay

😂😂
I genuinely think there's something here, the profiler shows I reduced time but the vec trunked the perf

PR #3371, opened 2026-01-29
supposed perf: use bitmask for ethereum and optimism

malik672:
this should be a perf but checking with codspeed


malik672:
done with this  @rakita

malik672:
if op-revm is going to be removed at #3405 then possibly we can make this better by 1.2 percent probably by removing support for fjord

PR #3372, opened 2026-01-30
perf!: drop the Box from call/create inputs

DaniPopes:


rakita:
i dont think we should include this change.

The problem with Call/Create Input is that it is created inside the Interpreter and then moved outside of it. We could potentially slim it down and leave it fields inside the parent Interpreter that we will read from child, but i feel this can get messy.

Issue #2853, opened 2025-08-06
Proposal: Make jump_table Optional in LegacyAnalyzedBytecode


mlyknown:
I would like to make the jumptable constructible from static data, which can significantly reduce memory usage—especially when running in a zkVM environment.

To support this, I propose that the jump_table field in LegacyAnalyzedBytecode be made optional (Option<...>). This would allow users to provide their own jump check implementation if needed. Alternatively, if you have any suggestions for a better solution, I am open to feedback.

If this change is acceptable, I am happy to submit a pull request implementing it.

rakita:
`which can significantly reduce memory usage—especially when running in a zkVM environment.` can you give me some data on this?

Jumptable is essential for consensus, without it, EVM will not function to the spec.

mlyknown:
Reducing memory usage—especially RAM.Supporting static data for the jump table is likely a better approach for this, as it allows the table to be stored in ROM rather than taking up valuable RAM.
Currently, the jump table uses bitvec, which does not support construction from static data—so it always occupies RAM. By comparison, the bytecode uses the bytes type, which supports from_static and can be constructed from static data, allowing it to reside in ROM instead of RAM.
In zkVM environments—such as Jolt—ROM is generally more resource-friendly than RAM.

rakita:
I asked for some data and measurements as I don't have a feel for zkVM performance, so I don't want to guess. 

`BitVec` does support generating from a static slice. `BitVec::from_slice` does this.

mlyknown:
I don't have concrete data yet—this is just my initial understanding.

As far as I know, [BitVec::from_slice](https://github.com/ferrilab/bitvec/blob/5fb855073acc2ed045094ed89d8daf8c765f0135/src/vec.rs#L121) will reallocate memory. Would you consider supporting static slices in the future?

If you don’t plan to make this change, feel free to close the issue for now. I’ll reopen it if I gather more data or specific measurements later.

rakita:
@mlyknown i see you are right, BitVec cloned the memory. I would definitely consider it, but I don't have a replacement for it rn,

mlyknown:
```
use bitvec::{order::Lsb0, slice::BitSlice, vec::BitVec};
use once_cell::race::OnceBox;
use primitives::{bytes, hex};
use std::fmt::Debug;

/// A table of valid `jump` destinations. Cheap to clone and memory efficient, one bit per opcode.
[...]

PR #3375, opened 2026-01-30
refactor!: flatten Bytecode

DaniPopes:
Bytecode is now a single struct, with a single enum BytecodeKind for legacy and eip7702
- arced internally
- 16 -> 8 bytes
- removes the Arc in JumpTable since it's already present in Bytecode
- makes analyze fn private

### DIFF
--- a/crates/bytecode/src/legacy/jump_map.rs
+++ b/crates/bytecode/src/legacy/jump_map.rs
@@ -1,7 +1,7 @@
 use bitvec::vec::BitVec;
-use core::fmt;
+use core::{cmp::Ordering, fmt, hash};
 use primitives::hex;
-use std::{borrow::Cow, cmp::Ordering};
+use std::{borrow::Cow, vec::Vec};
 
 /// A table of valid `jump` destinations.
 ///
@@ -55,9 +55,9 @@ impl Ord for JumpTable {
     }
 }
 
-impl std::hash::Hash for JumpTable {
+impl hash::Hash for JumpTable {
     #[inline]
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
         self.as_slice().hash(state);
     }
 }


DaniPopes:
what is it used in?

mattsse:
@DaniPopes stored on disk

https://github.com/paradigmxyz/reth/blob/29072639d6cf38063fea9ff8505c44c5ac4f3b95/crates/primitives-traits/src/account.rs#L149-L152

DaniPopes:
shouldn't be affected

crates/bytecode/src/bytecode/mod.rs
```diff
@@ -0,0 +1,437 @@
...
+    }
+}
+
+impl Eq for Bytecode {}
+
+impl core::hash::Hash for Bytecode {
+    #[inline]
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.kind().hash(state);
+        self.original_byte_slice().hash(state);
```
rakita:
```suggestion
        self.original_byte_slice().hash(state);
```
nit, only bytecode is enough.

crates/bytecode/src/legacy/analysis.rs
```diff
@@ -111,13 +105,6 @@ mod tests {
         assert_eq!(padded_bytecode.len(), bytecode.len() + 2);
     }
 
-    #[test]
-    fn test_empty_bytecode_requires_stop() {
-        let bytecode = vec![];
-        let (_, padded_bytecode) = analyze_legacy(bytecode.into());
-        assert_eq!(padded_bytecode.len(), 1); // Just STOP
-    }
```
rakita, line 119: nit, why was this removed?

crates/bytecode/src/legacy/jump_map.rs
```diff
@@ -1,65 +1,33 @@
...
+/// It is immutable and memory efficient, with one bit per byte in the bytecode.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct JumpTable {
-    /// Cached pointer to table data to avoid Arc overhead on lookup
-    table_ptr: *const u8,
-    /// Number of bits in the table.
-    len: usize,
-    /// Actual bit vec
-    table: Arc<Bytes>,
+    table: Box<[u8]>,
```
rakita: Would use `Arc<[u8]>` here, there was a report that Arcing (static allocation) is better for zkVM. Should be similar in performance, right

crates/bytecode/src/legacy/analysis.rs
DaniPopes: tested through Bytecode itself, analysis fn doesn't handle this case anymore

crates/bytecode/src/legacy/jump_map.rs
DaniPopes: I see https://github.com/bluealloy/revm/issues/2853

DaniPopes: implemented with cow

rakita:
@mattsse this looks good, no new asserts were added and jumpdest size is same as the previous. iirc I have added few tests when the problem happened previously. I think this is okay to merge, but leave it to your discretion

PR #3384, opened 2026-01-30
refactor(interpreter): Instruction impl bounds

zeroprooff:
Remove `Host` bound from Copy/Clone impls - not needed since the struct only holds a fn pointer and u64.


rakita:
Let it stay for consistency with the others requirements

PR #3386, opened 2026-01-31
rm unncesscary assert

malik672:
`step_by` and `chunks_exact` do some `assert x = 0` underneath

example: 
`  #[stable(feature = "chunks_exact", since = "1.31.0")]
    #[rustc_const_unstable(feature = "const_slice_make_iter", issue = "137737")]
    #[inline]
    #[track_caller]
    pub const fn chunks_exact(&self, chunk_size: usize) -> ChunksExact<'_, T> {
        assert!(chunk_size != 0, "chunk size must be non-zero");
        ChunksExact::new(self, chunk_size)
    }`

PR #3393, opened 2026-02-02
perf: optimize push_slice

malik672:
same logic as #3386


malik672:
1 cycle diff not enough to matter

PR #3398, opened 2026-02-04
fix: remove dead cargo-asm helper functions from interpreter

prestoalvarez:
This PR removes a commented-out `asm_step` / `asm_run` helper block from `interpreter.rs` that was previously marked as “used for cargo asm”.


rakita:
It is useful to have these there

PR #3399, opened 2026-02-04
perf: wire shared stack arena into frame creation  

malik672:
Threads the pre-allocated arena buffer `(Arc<Vec<U256>>)` from Evm through the frame creation chain. Nested `CALL/CREATE` frames now use `Stack::new_with_arena` instead of allocating a fresh Vec per frame.


malik672:
`validate.env` is the new bottleneck

malik672:
clearly the benchmark might not show it but a significant number of txs on mainnet are type 2

DaniPopes:
I don't think this matters that much, it's pretty unsafe unlike sharedmemory impl, and very out of date

PR #3407, opened 2026-02-07
refactor: deduplicate bytecode extraction in first_frame_input

zeroprooff:
Remove duplicated code block that extracts (code, hash) from account. The same 4 lines were repeated in both if/else branches.


rakita:
Thanks will be added in one of followups.

PR #3408, opened 2026-02-07
perf(interpreter): direct dispatch for hot opcodes in run_plain

gakonst:
## Summary

Replace indirect function-pointer dispatch with a direct `match` for ~90 common opcodes in the `run_plain` interpreter loop. This allows LLVM to inline small instruction bodies (PUSH, DUP, SWAP, arithmetic, bitwise, memory, control flow) directly into the hot loop, eliminating indirect call overhead.

Inspired by [NethermindEth/nethermind#10120](https://github.com/NethermindEth/nethermind/pull/10120) which applies similar hot/cold path splitting to the .NET EVM interpreter.

## Changes

- `crates/interpreter/src/interpreter.rs`: `run_plain` now uses a `match` on the opcode for the most common instructions (STOP, JUMPDEST, PUSH0-32, POP, DUP1-16, SWAP1-16, ADD/MUL/SUB/DIV/MOD/ADDMOD/MULMOD, LT/GT/EQ/ISZERO/AND/OR/XOR/NOT/SHL/SHR, MLOAD/MSTORE/MSTORE8, JUMP/JUMPI, CALLDATALOAD/CALLDATASIZE/CALLER/CALLVALUE/ADDRESS/GAS), falling back to the instruction table for everything else.
- Gas accounting and `InstructionTable` lookup remain unchanged — static gas is still read from the table entry, preserving spec-dependent gas changes.
- No trait or API changes. Custom instruction tables still work via the fallback path.

## Benchmark Results (reth-bench)

Benchmarked via [reth CI](https://github.com/paradigmxyz/reth/tree/auto-opt/20260207-fast-dispatch) on mainnet blocks:

**Normal blocks (2000 blocks):**
| Metric | Baseline | Feature | Change |
|--------|----------|---------|--------|
| Mean | 31.95ms | 31.9ms | -0.16% ≈ |
| P50 | 29.82ms | 29.93ms | +0.36% ≈ |
| Mgas/s | 951.42 | 952.94 | +0.16% ≈ |

**Big blocks (55 blocks, ~1.4 Ggas):**
| Metric | Baseline | Feature | Change |
|--------|----------|---------|--------|
| Mean | 1140.78ms | 1189.97ms | +4.31% ≈ |
| P50 | 797.02ms | 801.01ms | +0.5% ≈ |
| Mgas/s | 874.35 | 838.21 | -4.13% ≈ |

Results are within noise on both workloads. In reth's real workload, state access (SLOAD/SSTORE) and state root computation dominate — interpreter dispatch overhead is a small fraction. However, this change may benefit EVM-execution-heavy workloads (e.g., fuzzing, testing, foundry) where state I/O is not the bottleneck.

## Testing

- All 30 interpreter tests pass
- Full workspace `cargo check` passes
- reth compiles and benchmarks successfully with the patched revm

### DIFF
--- a/crates/interpreter/src/interpreter.rs
+++ b/crates/interpreter/src/interpreter.rs
@@ -317,14 +317,150 @@ impl<IW: InterpreterTypes> Interpreter<IW> {
     }
 
     /// Executes the interpreter until it returns or stops.
+    ///
+    /// Uses direct dispatch for the most common opcodes to enable inlining
+    /// and avoid indirect function pointer call overhead. Less common opcodes
+    /// fall back to the instruction table.
     #[inline]
     pub fn run_plain<H: Host + ?Sized>(
         &mut self,
         instruction_table: &InstructionTable<IW, H>,
         host: &mut H,
     ) -> InterpreterAction {
+        use bytecode::opcode::*;
+        use crate::instructions::{arithmetic, bitwise, control, memory, stack, system};
+
         while self.bytecode.is_not_end() {
-            self.step(instruction_table, host);
+            let opcode = self.bytecode.opcode();
+            self.bytecode.relative_jump(1);
+
+            let instruction = unsafe { instruction_table.get_unchecked(opcode as usize) };
+
+            if self.gas.record_cost_unsafe(instruction.static_gas()) {
+                self.halt_oog();
+                continue;
+            }
+
+            match opcode {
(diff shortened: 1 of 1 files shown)


gakonst:
Closing — benchmark results were neutral. Interpreter dispatch overhead is not a bottleneck in reth's workload.

PR #3412, opened 2026-02-08
perf(database): use retain to filter storage slots in-place

### LINKED ISSUE #3374: perf: collect in apply_account_state is wasteful
The following collect is wasteful, we should be able to reuse the map since all we're doing is `retain`ing changed slots and shrinking the value type (EvmStorageSlot -> StorageSlot).

cc @rakita for ideas on how to do this.

Ideally fixing this wouldn't be too invasive.

https://github.com/bluealloy/revm/blob/fc750ca23d50a44419d736a2b52dc084477dd75d/crates/database/src/states/cache.rs#L213-L218

0xMars42:
Closes #3374

`apply_account_state` builds `changed_storage` by chaining
`into_iter().filter().map().collect()`. Because `filter()` yields
`size_hint (0, N)`, `collect()` cannot pre-allocate the right capacity
and may re-hash as the map grows.

Replace the filter step with `retain()` on the original map, then
`into_iter().map().collect()`. This gives `collect()` an exact
`size_hint (K, K)` where K is the number of changed slots, so only a
single allocation of the correct size is needed.

A zero-alloc path would require changing `StorageWithOriginalValues`
across TransitionAccount/BundleAccount/reverts, which is too invasive
for this fix.

### DIFF
--- a/crates/database/src/states/cache.rs
+++ b/crates/database/src/states/cache.rs
@@ -215,11 +215,12 @@ impl CacheState {
         let is_created = account.is_created();
         let is_empty = account.is_empty();
 
-        // Transform evm storage to storage with previous value.
-        let changed_storage = account
-            .storage
+        // Filter unchanged storage slots in-place, then convert to StorageSlot.
+        // retain() avoids allocating a new map and gives collect() an exact size hint.
+        let mut storage = account.storage;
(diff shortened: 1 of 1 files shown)


rakita:
This does nothing

PR #3415, opened 2026-02-10
fix: correct doc comment for `--time` flag in bench command

bluebluesaku:
The doc comment for the `time` field incorrectly started with "Samples represents..." which made `--help` output confusing since it read the same as the `--samples` description. Fixed it to say "Time represents...".


rakita:
Thanks, small PRs as comments get bumped in one saparate PR

PR #3416, opened 2026-02-10
feat(revme): validate block gas used in blockchain tests

rakita:
## Summary
- Track cumulative gas used across all transactions in each block during blockchain test execution
- Validate cumulative gas against expected `gas_used` from block headers
- Use EIP-7778 logic for Amsterdam+ (block gas = max(spent, floor_gas), excluding refunds) and standard `used()` for earlier specs

## Test plan
- [ ] Run blockchain tests to verify gas validation passes for existing test vectors
- [ ] Confirm new `BlockGasUsedMismatch` error surfaces correctly on mismatches

### DIFF
--- a/crates/handler/src/post_execution.rs
+++ b/crates/handler/src/post_execution.rs
@@ -115,6 +115,7 @@ pub fn output<CTX: ContextTr<Journal: JournalTr>, HALTREASON: HaltReasonTr>(
         },
         SuccessOrHalt::Revert => ExecutionResult::Revert {
             gas: result_gas,
+            logs,
             output: output.into_data(),
         },
         SuccessOrHalt::Halt(reason) => {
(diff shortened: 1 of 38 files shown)


bins/revme/src/cmd/blockchaintest.rs
```diff
@@ -881,8 +887,18 @@ fn execute_blockchain_test(
...
                         }
+                        block_completed = false;
                         break; // Skip to next block
                     }
+                    // EIP-7778: Block gas accounting without refunds.
+                    // For Amsterdam+, block gas = max(spent, floor_gas).
+                    // For pre-Amsterdam, block gas = used() = max(spent - refunded, floor_gas).
+                    let gas = result.result.gas();
+                    cumulative_gas_used += if spec_id.is_enabled_in(SpecId::AMSTERDAM) {
+                        gas.spent().max(gas.floor_gas())
```
rakita, line 898: EIP-7778 diff

crates/handler/src/post_execution.rs
mattsse, line 20: isn't this the same as like `Gas::finalize|freeze` unsure if this needs to be a standalone function?

mattsse (commented):
I dont have a better naming suggestion for ResultGas but maybe @klkvr has

all of this makes sense, smol bikeshed re the conversion fn

PR #3419, opened 2026-02-11
feat(revme): add --omit-progress flag to statetest command

bluebluesaku:
The blockchaintest command already has --omit-progress to suppress the progress bar, but statetest was missing it. This is annoying when profiling with samply/perf since the progress bar keeps flooding the terminal. Added the same flag so both commands behave consistently.


rakita:
Waiting for CI to finish

bluebluesaku:
> Waiting for CI to finish

Looks goooood!

PR #3420, opened 2026-02-11
perf: reduce per-call overhead in CALL/CREATE hot paths

gakonst:
## Summary
Reduces per-call overhead in revm's CALL/CREATE instruction paths.

## Changes
- **`#[inline]` on `load_acc_and_calc_gas`** — was `#[inline(never)]`, preventing LLVM from optimizing the hot CALL dispatch path. This function runs on every CALL/STATICCALL/DELEGATECALL/CALLCODE.
- **Cache `host.gas_params()`** in `load_account_delegated` — avoids two repeated calls before the first mutable borrow of `host`.
- **Cache `return_memory_offset`** in `make_call_frame` — avoids 4 redundant `Range::clone()` calls across the early-return, precompile, and normal paths.

## Testing
- `cargo check --workspace` — clean
- `cargo clippy` — clean  
- `cargo test` — all pass
- `cargo fmt --all --check` — clean

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -209,7 +209,7 @@ impl EthFrame<EthInterpreter> {
         }
 
         // Get bytecode and hash - either from known_bytecode or load from account
-        let (bytecode, bytecode_hash) = if let Some((hash, code)) = inputs.known_bytecode.take() {
+        let (bytecode, bytecode_hash) = if let Some((hash, code)) = inputs.known_bytecode.clone() {
             // Use provided bytecode and hash
             (code, hash)
         } else {


crates/interpreter/src/instructions/contract/call_helpers.rs
```diff
@@ -51,7 +51,7 @@ pub fn resize_memory(
 }
 
 /// Calculates gas cost and limit for call instructions.
-#[inline(never)]
+#[inline]
```
yongkangc, line 54: why did we remove never here?

rakita:
Closing this, as it is going to be overwriten by https://github.com/bluealloy/revm/pull/3426

PR #3421, opened 2026-02-11
fix(database): allow EIP161 state clear for empty Loaded and Changed accounts

msheth-circle:
## Summary                                                                                                                                                            
                                                                                                                                                                        
 - Fix `on_touched_empty_post_eip161()` panic when called on a `Loaded` `AccountStatus`                                                                                
 - Allow `Loaded` and `Changed` → `Destroyed` transition unconditionally

On EVM chains where the native token is an ERC20, an EOA can hold the native token (balance > 0, nonce=0, no code) and spend it all as an ERC20 via a meta-transaction (relayer pays gas, no nonce change on the EOA), making the account empty. In this scenario, the account was loaded from DB with `Loaded` status, and since the balance change is tracked in the Journal — not through `CacheAccount::change()` — the `CacheAccount` status remains `Loaded` when `touch_empty_eip161` is called.

The fix allows `Loaded` → `Destroyed` unconditionally. This is safe because `touch_empty_eip161` is only reachable when `account.is_empty()` is true on the Journal output, which requires no code (`code_hash == KECCAK_EMPTY`). A contract in `Loaded` status cannot reach this path — SELFDESTRUCT is the only way to remove code, and it's handled by the `is_selfdestructed()` early return before this point. This is consistent with [geth](https://github.com/ethereum/go-ethereum/blob/995fa79bf505542f3bdf0dbf8528b3dc8ad05bbc/core/state/statedb.go#L764), which has no status distinction and simply deletes any dirty empty account in `Finalise()`.

## Test plan                                                                                                                                                          

 - [x] Added unit test for `Loaded` → `Destroyed` transition in `on_touched_empty_post_eip161`                                                                         
 - [x] Added `#[should_panic]` test confirming `Changed` still panics                                                                                                  
 - [x] `cargo test -p revm-database` — 31 passed                                                                                                                       
 - [x] `cargo nextest run --workspace` — 369 passed, 0 failed                                                                                                          
 - [x] `cargo clippy --workspace --all-targets --all-features` — 0 warnings

### DIFF
--- a/crates/database/src/states/account_status.rs
+++ b/crates/database/src/states/account_status.rs
@@ -96,8 +96,6 @@ impl AccountStatus {
     }
 
     /// Returns the next account status on touched empty account post state clear EIP (EIP-161).
-/// Returns the next account status on touched empty account post state clear EIP (EIP-161).
-    /// If current status is [AccountStatus::Changed].
     pub fn on_touched_empty_post_eip161(&self) -> Self {
         match self {
             // Account can be touched but not existing. The status should remain the same.


crates/database/src/states/account_status.rs
```diff
@@ -108,9 +108,11 @@ impl AccountStatus {
...
             // Transition to destroy the account.
             Self::DestroyedAgain | Self::DestroyedChanged => Self::DestroyedAgain,
-            // Account statuses considered unreachable.
-            Self::Loaded | Self::Changed => {
-                unreachable!("Wrong state transition, touch empty is not possible from {self:?}");
+            // Account can be loaded and become empty.
+            Self::Loaded => Self::Destroyed,
+            // Changed account cannot be empty.
+            Self::Changed => {
+                panic!("Wrong state transition, touch empty is not possible from Changed");
```
msheth-circle:
```suggestion
                unreachable!("Wrong state transition, touch empty is not possible from {self:?}");
```

crates/database/src/states/account_status.rs
```diff
@@ -108,8 +108,10 @@ impl AccountStatus {
             Self::InMemoryChange | Self::Destroyed | Self::LoadedEmptyEIP161 => Self::Destroyed,
             // Transition to destroy the account.
             Self::DestroyedAgain | Self::DestroyedChanged => Self::DestroyedAgain,
-            // Account statuses considered unreachable.
-            Self::Loaded | Self::Changed => {
+            // Account can be loaded and become empty.
+            Self::Loaded => Self::Destroyed,
+            // Changed account cannot be empty.
+            Self::Changed => {
```
rakita: let make both Loaded and Changed become Destroyed. This case is not something that can happen in eth protocol, but I'm okay relaxing constraints on it.

crates/database/src/states/account_status.rs
```diff
@@ -108,8 +108,10 @@ impl AccountStatus {
...
                 unreachable!("Wrong state transition, touch empty is not possible from {self:?}");
             }
```
msheth-circle:
```suggestion
            // Account can become empty.
            Self::Changed | Self::Loaded => Self::Destroyed,
```

crates/database/src/states/account_status.rs
```diff
@@ -99,7 +99,7 @@ impl AccountStatus {
     ///
     /// # Panics
     ///
```
msheth-circle:
```suggestion
/// Returns the next account status on touched empty account post state clear EIP (EIP-161).
```

crates/database/src/states/account_status.rs
```diff
@@ -319,6 +321,16 @@ mod test {
             AccountStatus::DestroyedChanged.on_touched_empty_post_eip161(),
             AccountStatus::DestroyedAgain
         );
+        assert_eq!(
+            AccountStatus::Loaded.on_touched_empty_post_eip161(),
+            AccountStatus::Destroyed
+        );
```
msheth-circle, line 319:
```suggestion
        assert_eq!(
            AccountStatus::Loaded.on_touched_empty_post_eip161(),
            AccountStatus::Destroyed
        );
        assert_eq!(
            AccountStatus::Changed.on_touched_empty_post_eip161(),
            AccountStatus::Destroyed
        );
```

crates/database/src/states/account_status.rs
```diff
@@ -319,6 +321,16 @@ mod test {
...
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_on_touched_empty_post_eip161_changed() {
+        AccountStatus::Changed.on_touched_empty_post_eip161();
     }
```
msheth-circle:
```suggestion
```

crates/database/src/states/account_status.rs
```diff
@@ -98,8 +98,8 @@ impl AccountStatus {
     /// Returns the next account status on touched empty account post state clear EIP (EIP-161).
     ///
     /// # Panics
```
msheth-circle:
```suggestion
```

crates/database/src/states/account_status.rs
```diff
@@ -98,8 +98,8 @@ impl AccountStatus {
...
-    ///
-    /// If current status is [AccountStatus::Loaded] or [AccountStatus::Changed].
+/// Returns the next account status on touched empty account post state clear EIP (EIP-161).
+    /// If current status is [AccountStatus::Changed].
```
msheth-circle:
```suggestion
```

crates/database/src/states/account_status.rs
msheth-circle: @rakita sounds good, updated

msheth-circle: Also, do you mind creating a patch release for it?

msheth-circle:
@rakita Mind taking another look?

crates/database/src/states/account_status.rs
rakita: Patch release would be a stretch with just this

rakita:
Merging after CI

msheth-circle:
Thanks @rakita! CI is looking good

rakita:
Will make release soon, in next few days

PR #3425, opened 2026-02-13
fix: correct misleading copy-paste comment in osaka_run

Fallengirl:
The osaka_run docstring said "Gas cost of berlin is modified from byzantium" which was copy-pasted from berlin_run and is wrong — osaka modifies berlin, not byzantium. Added missing EIP-7883 reference and fixed the description.


rakita:
Thanks you will be added in one of followup commits

PR #3427, opened 2026-02-16
bench: add subcall benchmarks (1000-call variants)

rakita:
## Summary
- Add three subcall benchmark variants to `revme bench` for measuring CALL opcode performance
- **subcall_1000_transfer_1wei**: 1000 subcalls each transferring 1 wei (measures value transfer overhead)
- **subcall_1000_same_account**: 1000 subcalls with no value to same account (baseline subcall cost)
- **subcall_1000_nested**: 1000 subcalls where each target does another subcall (measures nested call frame overhead)

## Test plan
- [x] `cargo build -p revme` compiles successfully
- [x] `cargo run -p revme --release -- bench subcall` runs all 3 variants (~157-299µs per iteration)


rakita:
Just a bench in revme, it does not touch crates at all

PR #3429, opened 2026-02-17
fix(context): use EIP-7907 code size limits for Prague+

bluebluesaku:
max_code_size() and max_initcode_size() were hardcoded to EIP-170/EIP-3860 limits regardless of the active spec. The eip7907 constants were already defined in primitives but never actually used. Now these methods check the spec and return the correct limits for Prague+ (0xC000 for code, 0x12000 for initcode), matching the pattern already used by tx_gas_limit_cap(). Manual overrides via limit_contract_code_size/limit_contract_initcode_size still take priority.

Quick question @rakita — is there a timeline for the next revm release? This fix could affect downstream consumers like reth.

### DIFF
--- a/crates/context/src/cfg.rs
+++ b/crates/context/src/cfg.rs
@@ -2,7 +2,7 @@
 pub use context_interface::Cfg;
 
 use context_interface::cfg::GasParams;
-use primitives::{eip170, eip3860, eip7825, hardfork::SpecId};
+use primitives::{eip170, eip3860, eip7825, eip7907, hardfork::SpecId};
 
 /// EVM configuration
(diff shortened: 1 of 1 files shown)


rakita:
This is invalid

bluebluesaku:
> This is invalid

could reopen this?
Good catch, my bad — I mixed up the activation hardfork. EIP-7907 is OSAKA, not Prague. The existing tests in validation.rs already test against SpecId::OSAKA which should've been a clear signal. Fixed it now, all handler tests pass.

@rakita

rakita:
This is not protocol change, it is not included in eth

bluebluesaku:
> This is not protocol change, it is not included in eth

Thanks for correcting

PR #3432, opened 2026-02-20
fix(precompile): use big-endian resize for modexp output padding

Wollac:
## Summary

- Replace `left_pad_vec` with a new `resize_be` utility for modexp output normalization.
- `left_pad_vec` incorrectly truncates trailing (least-significant) bytes when the output exceeds `mod_len`. This is incorrect for big-endian values.
- The new `resize_be` utility correctly handles both padding (left-pad with zeroes) and truncation (drop leading bytes).
- Avoid an extra allocation by resizing the output vector in place instead of creating a new one.

## Motivation

`crypto().modexp()` is a trait method that can be overridden, so the output may exceed `mod_len` bytes and contain leading zeroes. The previous code silently corrupted results in that case by truncating from the wrong side.`resize_be` only assumes the mathematical invariant that the output value is `< modulus`, not any specific byte length.

### DIFF
--- a/crates/precompile/src/modexp.rs
+++ b/crates/precompile/src/modexp.rs
@@ -2,7 +2,7 @@
 //! and reprices in berlin hardfork with [`EIP-2565`](https://eips.ethereum.org/EIPS/eip-2565).
 use crate::{
     crypto,
-    utilities::{left_pad, resize_be, right_pad_vec, right_pad_with_offset},
+    utilities::{left_pad, left_pad_vec_be, right_pad_vec, right_pad_with_offset},
     Precompile, PrecompileError, PrecompileId, PrecompileOutput, PrecompileResult,
 };
 use core::cmp::{max, min};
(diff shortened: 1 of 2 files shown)


rakita (commented):
I see, this only can happen if `crypto().modexp` returns some badly formated data.

for style lets create `left_pad_vec_be` that would do a similar thing to left_pad_vec

rakita:
@Wollac can you check if changes are okay. And will merge afterwards

Wollac:
Yes, this fixes the issue I observed. 
One tradeoff worth noting: Due to the `Cow`s it'll most likely lead to an additional `Vec` allocation even for the most common case where `output` already has the perfect length. The original `resize_be` approach tried to avoid this by modifying the vec in place.

rakita:
@Wollac i see, will check this in saparate PR. Will merge this as we will be releasing new version soon(tm)

PR #3435, opened 2026-02-23
fix(inspector): output actual returnData in EIP-3155 tracer

bluebluesaku:
The returnData field in the EIP-3155 tracer output was hardcoded to "0x" for every step. The spec requires it to contain the actual data returned by the last CALL/STATICCALL/etc. Now we read it from interp.return_data.buffer() during step() and output the real hex-encoded value. This was a leftover TODO rather than an intentional choice — the field type was &'static str which made it impossible to hold dynamic data.

### DIFF
--- a/crates/inspector/src/eip3155.rs
+++ b/crates/inspector/src/eip3155.rs
@@ -1,7 +1,7 @@
 use crate::{inspectors::GasInspector, Inspector};
 use context::{Cfg, ContextTr, JournalTr, Transaction};
 use interpreter::{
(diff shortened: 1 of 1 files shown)


rakita:
It will become very noisy with to put return data on every opcode. And it is a penalty to convert bytes to str every time

bluebluesaku:
> It will become very noisy with to put return data on every opcode. And it is a penalty to convert bytes to str every time

Thanks. Not realized this point before.

PR #3436, opened 2026-02-24
feat: add revert inspector example

0xstepit:
Add a simple example demonstrating how to handle and inspect a revert from a contract call.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1876,6 +1876,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


rakita:
This probably can be done with sol macro

rakita:
Will close this PR as imo example does not contribute a lot here, it is more about how solidity parses creates the error.

stefanopitton:
Makes sense, the example was more on the Solidity side than revm.

PR #3437, opened 2026-02-24
refactor(gas): remove unreachable branch in log2floor

### LINKED ISSUE #2910: Unreachable statement in log2floor function
There's a cleanup that can be done in the [`log2floor`](https://github.com/bluealloy/revm/blob/ae1655da953356fa6707e529bec09f80dd88abde/crates/interpreter/src/gas/calc.rs#L75) function because the
```rust
if l == 0 {
```
statement is unreachable.

Proof:
This branch could only be reached if `value.as_limbs()[i].leading_zeros() == 64` when the loop has been iterated 4 times. But this condition is only true if `value.as_limbs()[i] == 0u64`, which is covered above. Meaning that the two branches are mutually exclusive.
Thus, we can just remove the statement.

```diff
diff --git a/crates/interpreter/src/gas/calc.rs b/crates/interpreter/src/gas/calc.rs
index 08ec9c0f..17330d84 100644
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -71,12 +71,7 @@ const fn log2floor(value: U256) -> u64 {
         if value.as_limbs()[i] == 0u64 {
             l -= 64;
         } else {
-            l -= value.as_limbs()[i].leading_zeros() as u64;
-            if l == 0 {
-                return l;
-            } else {
-                return l - 1;
-            }
+            return l - value.as_limbs()[i].leading_zeros() as u64 - 1;
         }
         if i == 0 {
             break;
```

Tried to optimize further the hotpath by checking for zeroeness of `value` as a first condition which would exclude the last limb to be `0`, reducing the loop iteration by one but it made the performance a little bit worse for `value != 0`, which makes sense:

```diff
diff --git a/crates/interpreter/src/gas/calc.rs b/crates/interpreter/src/gas/calc.rs
index 08ec9c0f..e843f6bb 100644
--- a/crates/interpreter/src/gas/calc.rs
+++ b/crates/interpreter/src/gas/calc.rs
@@ -65,25 +65,22 @@ pub const fn create2_cost(len: usize) -> Option<u64> {
 
 #[inline]
 const fn log2floor(value: U256) -> u64 {
+    if value.leading_zeros() == 256 {
+        return 0;
+    }
     let mut l: u64 = 256;
     let mut i = 3;
     loop {
         if value.as_limbs()[i] == 0u64 {
             l -= 64;
-        } else {
-            l -= value.as_limbs()[i].leading_zeros() as u64;
-            if l == 0 {
+            if i == 2 {
                 return l;
-            } else {
-                return l - 1;
             }
+            i -= 1;
+        } else {
+            return l - value.as_limbs()[i].leading_zeros() as u64 - 1;
         }
-        if i == 0 {
-            break;
-        }
-        i -= 1;
     }
-    l
 }
 
 /// `EXP` opcode cost calculation.
```

saylor-mik87786:
Fixes #2910

Removes the unreachable `if l == 0` check in `log2floor`, adds unit tests to  verify correctness.

### DIFF
--- a/crates/context/interface/src/cfg/gas_params.rs
+++ b/crates/context/interface/src/cfg/gas_params.rs
@@ -806,6 +806,14 @@ pub(crate) fn log2floor(value: U256) -> u64 {
         }
     }
     0
+pub(crate) fn log2floor(value: U256) -> u64 {
+    for i in (0..4).rev() {
+        let limb = value.as_limbs()[i];
+        if limb != 0 {
+            return i as u64 * 64 + 63 - limb.leading_zeros() as u64;
+        }
+    }
+    0
 }
 
 /// Gas identifier that maps onto index in gas table.


crates/context/interface/src/cfg/gas_params.rs
rakita:
```suggestion
pub(crate) fn log2floor(value: U256) -> u64 {
    for i in (0..4).rev() {
        let limb = value.as_limbs()[i];
        if limb != 0 {
            return i as u64 * 64 + 63 - limb.leading_zeros() as u64;
        }
    }
    0
```

More simplifications on this

crates/context/interface/src/cfg/gas_params.rs
```diff
@@ -791,19 +791,21 @@ pub(crate) const fn log2floor(value: U256) -> u64 {
...
-    l
+pub(crate) fn log2floor(value: U256) -> u64 {
+    for i in (0..4).rev() {
+        let limb = value.as_limbs()[i];
+        if limb != 0 {
+            return i as u64 * 64 + 63 - limb.leading_zeros() as u64;
+        }
+    }
+    0
 }
```
rakita, line 817:
```suggestion
pub(crate) fn log2floor(value: U256) -> u64 {
    for i in (0..4).rev() {
        let limb = value.as_limbs()[i];
        if limb != 0 {
            return i as u64 * 64 + 63 - limb.leading_zeros() as u64;
        }
    }
    0
}
```

rakita: Ok this didn't remove all code

rakita:
Superseeded by https://github.com/bluealloy/revm/pull/3440

PR #3443, opened 2026-02-25
fix: Refactor hints_util likely/unlikely to remove duplicated cold path logic`

anim001k:
Refactor `revm-primitives` `hints_util::likely` and `hints_util::unlikely` to share the common cold-path logic while preserving their public API and behavior.


rakita:
I like previous better, it is simpler and it has less functions

PR #3444, opened 2026-02-25
refactor: move EIP-161 state clear into journal finalize

decofe:
## Summary

Move `has_state_clear` / `set_state_clear_flag` out of the database layer (`CacheState` / `State<DB>`) and into `JournalInner::finalize()`, where the `SpecId` is already available.

## Motivation

`set_state_clear_flag` couples the database layer to chainspec knowledge. Block executors need extra chainspec-aware setup before each block just to handle a deprecated pre-Spurious Dragon behavior. This prevents executors from working with arbitrary `Database` implementations without that setup.

## Changes

- `JournalInner::finalize()`: normalize pre-EIP-161 empty touched accounts before returning `EvmState`
  - `LoadedAsNotExisting` → `mark_created()` (materialize)
  - Already existing → `unmark_touch()` (preserve)
- `CacheState`: remove `has_state_clear` field, `set_state_clear_flag()`, and pre-EIP-161 branch in `apply_account_state()` — always uses post-EIP-161 semantics
- `State<DB>`: remove `set_state_clear_flag()`
- `StateBuilder`: remove `with_state_clear` field and `without_state_clear()`
- `revme` statetest runner: remove `set_state_clear_flag` calls

Zero perf impact for post-Spurious Dragon blocks (the `finalize()` iteration is gated behind `!spec.is_enabled_in(SPURIOUS_DRAGON)`).

Prompted by: arsenii

### DIFF
--- a/crates/database/src/states/account_status.rs
+++ b/crates/database/src/states/account_status.rs
@@ -109,30 +109,6 @@ impl AccountStatus {
         }
     }
 
-    /// Returns the next account status on touched or created account pre state clear EIP (EIP-161).
-    /// Returns `None` if the account status didn't change.
-    ///
-    /// # Panics
-    ///
-    /// If current status is [AccountStatus::Loaded] or [AccountStatus::Changed].
-    pub fn on_touched_created_pre_eip161(&self, had_no_info: bool) -> Option<Self> {
-        match self {
-            Self::LoadedEmptyEIP161 => None,
-            Self::DestroyedChanged => {
-                if had_no_info {
-                    None
-                } else {
-                    Some(Self::DestroyedChanged)
-                }
-            }
-            Self::Destroyed | Self::DestroyedAgain => Some(Self::DestroyedChanged),
-            Self::InMemoryChange | Self::LoadedNotExisting => Some(Self::InMemoryChange),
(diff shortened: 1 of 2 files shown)


crates/context/src/journal/inner.rs
```diff
@@ -214,13 +214,34 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
...
         selfdestructed_addresses.clear();
 
-        let state = mem::take(state);
+        let mut state = mem::take(state);
+
+        // Pre-EIP-161 normalization: adjust empty touched accounts so the database
+        // layer can always apply post-EIP-161 commit semantics (destroy empty touched
+        // accounts). For pre-Spurious Dragon blocks, we prevent destruction by either
+        // marking the account as created (materialized) or clearing the touched flag.
+        if !cfg.spec.is_enabled_in(SPURIOUS_DRAGON) {
```
rakita, line 227: This is perf penalty. The reason why we have it outside is so we dont iterate over all accounts

klkvr: this is only pre-SPURIOUS_DRAGON so it's fine?

PR #3446, opened 2026-02-26
chore: preserve account_id in BAL storage fallback

saylor-mik87786:
BalDatabase now falls back to `db.storage_by_account_id(address, account_id, storage_key)` when BAL storage misses, preserving `account_id` for downstream account-indexed storage lookups.


rakita:
Previous is correct

PR #3452, opened 2026-02-27
feat(precompile): add herumi/mcl as alternative backend for bn254

rakita:
## Summary

- Adds `bn254-mcl` feature flag to use [herumi/mcl](https://github.com/herumi/mcl) as the fastest backend for BN254 curve operations (G1 add, G1 scalar mul, pairing check)
- Enabled by default; when active, mcl takes priority over substrate-bn and arkworks backends
- Uses Rust FFI bindings from https://github.com/herumi/mcl/pull/237

## Test plan

- [x] All bn254 precompile tests pass with `bn254-mcl` feature (`test_bn254_add`, `test_bn254_mul`, `test_bn254_pair`)
- [x] Builds cleanly with no warnings
- [x] Default features include `bn254-mcl`


rakita:
We are fine without it for now

PR #3459, opened 2026-02-28
fix(context): unify EIP-4844 missing target error path

saylor-mik87786:
TxEnvBuilder::build now returns `TxEnvBuildError::DeriveErr(DeriveTxTypeError::MissingTargetForEip4844)` for EIP-4844 transactions without a call target, and updates the EIP-4844 not-call unit test to assert the unified derive error.


rakita:
Will leave it as is

PR #3460, opened 2026-02-28
fix: forward account_id in BalDatabase::storage_by_account_id fallback

bluebluesaku:
Fix BalDatabase::storage_by_account_id fallback to call self.db.storage_by_account_id() instead of self.db.storage()

Without this fix, when the BAL layer misses, the account_id is silently discarded and the inner database falls back to address-based lookup, defeating any index-based fast path the inner database may implement.


rakita:
This is going to introduce a bug

PR #3461, opened 2026-02-28
Remove unused `call_depth` field from `TestInspector`

gap-editor:
**Description:** `call_depth` is incremented in `call()` and decremented in `call_end()` but never read anywhere — not in assertions, not in event matching, not in any public API. It's dead state that adds noise to the struct. Call nesting is already recoverable from the `events` vector via the `Call { outcome: None }` matching pattern.


rakita:
It is fine for it to stay

PR #3482, opened 2026-03-06
fix(revme): return error instead of process::exit in statetest runner

bluebluesaku:
When tests fail without thread panics (`n_errors > 0 && n_thread_errors == 0`),
`runner::run()` calls `process::exit(1)` instead of returning an `Err`. This
skips all `Drop` impls and bypasses the error reporting in `main()`:

```
main()                          // Err => eprintln, never reached
  └─ MainCmd::run()             // ?, never reached
      └─ statetest::Cmd::run()  // ?, never reached
          └─ runner::run()      // process::exit(1) here
```

Add a `TestsFailed(usize)` variant to `TestErrorKind` and return it properly:

```diff
     if n_thread_errors == 0 {
-        std::process::exit(1);
+        return Err(TestError {
+            name: String::new(),
+            path: String::new(),
+            kind: TestErrorKind::TestsFailed(n_errors),
+        });
     }
```

### DIFF
--- a/bins/revme/src/cmd/statetest/runner.rs
+++ b/bins/revme/src/cmd/statetest/runner.rs
@@ -42,6 +42,8 @@ pub enum TestErrorKind {
     StateRootMismatch { got: B256, expected: B256 },
     #[error("unknown private key: {0:?}")]
     UnknownPrivateKey(B256),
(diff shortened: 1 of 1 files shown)


rakita:
It is fine

PR #3494, opened 2026-03-11
feat: add reverting address to ExecutionResult::Revert

decofe:
Adds `address: Option<Address>` to `ExecutionResult::Revert` so consumers can identify which contract or precompile produced the revert.

**Motivation:** Chains with custom precompiles (e.g. Tempo) have multiple precompiles that share ABI error selectors (e.g. `Unauthorized` = `0x82b42900` exists in both ValidatorConfig and TIP-20). When a precompile reverts via `PrecompileOutput::new_reverted()`, the revert data flows through `InstructionResult::Revert` → `ExecutionResult::Revert`, but only the raw bytes are preserved—no address. The RPC layer's `FromRevert::from_revert()` receives just `Bytes` and cannot disambiguate which precompile the selector came from.

**Changes:**
- `ExecutionResult::Revert` gains `address: Option<Address>`
- `post_execution::output()` accepts and passes the address
- `Handler::execution_result()` extracts `TxKind::Call(addr)` → `Some(addr)`, `TxKind::Create` → `None`
- Same for op-revm's `execution_result`

**Testing:**
```
cargo test --workspace  # all pass
cargo clippy --workspace --all-targets  # clean
cargo fmt --all --check  # clean
```

Co-Authored-By: Georgios Konstantopoulos <17802178+gakonst@users.noreply.github.com>

Prompted by: georgios

### DIFF
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -14,7 +14,7 @@ use context_interface::{
     Cfg, ContextTr, Database, JournalTr, Transaction,
 };
 use interpreter::{interpreter_action::FrameInit, Gas, InitialAndFloorGas, SharedMemory};
-use primitives::U256;
+use primitives::{TxKind, U256};
 use state::Bytecode;
 
 /// Trait for errors that can occur during EVM execution.
@@ -480,7 +480,13 @@ pub trait Handler {
     ) -> Result<ExecutionResult<Self::HaltReason>, Self::Error> {
         take_error::<Self::Error, _>(evm.ctx().error())?;
 
-        let exec_result = post_execution::output(evm.ctx(), result, result_gas);
+        // Extract target address for revert context.
+        let address = match evm.ctx().tx().kind() {
+            TxKind::Call(addr) => Some(addr),
+            TxKind::Create => None,
+        };
+
+        let exec_result = post_execution::output(evm.ctx(), result, result_gas, address);
 
         // commit transaction
         evm.ctx().journal_mut().commit_tx();
--- a/crates/handler/src/post_execution.rs
+++ b/crates/handler/src/post_execution.rs
@@ -6,7 +6,7 @@ use context_interface::{
     Block, Cfg, ContextTr, Database, LocalContextTr, Transaction,
 };
 use interpreter::{Gas, InitialAndFloorGas, SuccessOrHalt};
-use primitives::{hardfork::SpecId, U256};
+use primitives::{hardfork::SpecId, Address, U256};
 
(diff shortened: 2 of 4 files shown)


crates/handler/src/post_execution.rs
```diff
@@ -99,6 +99,7 @@ pub fn output<CTX: ContextTr<Journal: JournalTr>, HALTREASON: HaltReasonTr>(
     // FrameResult should be a generic that returns gas and interpreter result.
     result: FrameResult,
     result_gas: ResultGas,
+    address: Option<Address>,
```
borngraced, line 102: Since `post_execution::output` already takes `context: &mut CTX`, you can use that to derive the address lazily inside the `SuccessOrHalt::Revert` match arm, instead of adding a new param and paying the derivation cost on every call regardless of outcome...like so:

```rust
        SuccessOrHalt::Revert => {
            let address = match context.tx().kind() {
                TxKind::Call(addr) => Some(addr),
                TxKind::Create => None,
            };
            ExecutionResult::Revert {
                gas: result_gas,
                logs,
                output: output.into_data(),
                address,
            }
        }
```

crates/op-revm/src/handler.rs
```diff
@@ -349,7 +349,13 @@ where
...
         take_error::<Self::Error, _>(evm.ctx().error())?;
 
-        let exec_result = post_execution::output(evm.ctx(), frame_result, result_gas)
+        // Extract target address for revert context.
+        let address = match evm.ctx().tx().kind() {
+            revm::primitives::TxKind::Call(addr) => Some(addr),
+            revm::primitives::TxKind::Create => None,
+        };
+
+        let exec_result = post_execution::output(evm.ctx(), frame_result, result_gas, address)
```
borngraced, line 358: can be reverted after my suggestion above

crates/handler/src/handler.rs
```diff
@@ -480,7 +480,13 @@ pub trait Handler {
...
         take_error::<Self::Error, _>(evm.ctx().error())?;
 
-        let exec_result = post_execution::output(evm.ctx(), result, result_gas);
+        // Extract target address for revert context.
+        let address = match evm.ctx().tx().kind() {
+            TxKind::Call(addr) => Some(addr),
+            TxKind::Create => None,
+        };
+
+        let exec_result = post_execution::output(evm.ctx(), result, result_gas, address);
```
borngraced, line 489: same as above :D

borngraced (changes requested):

rakita:
Obsolete

PR #3503, opened 2026-03-14
fix: make DummyHost return defaults instead of errors for storage ops

decofe:
Changes `selfdestruct`, `load_account_info_skip_cold_load`, `sstore_skip_cold_load`, and `sload_skip_cold_load` on `DummyHost` to return `Ok(Default::default())` instead of `Err(LoadError::...)`, so callers can use `DummyHost` without hitting errors on storage operations.

Co-Authored-By: DaniPopes <57450786+DaniPopes@users.noreply.github.com>

Prompted by: DaniPopes


DaniPopes (approved):
This used to have a hashmap backing it i believe, so it was used in revmc benchmarks, but after upgrading it started erroring.

PR #3512, opened 2026-03-19
fix(handler): use cfg gas_params in validate_initial_tx_gas

bluebluesaku:
`validate_initial_tx_gas` was constructing a fresh `GasParams::new_spec(spec)` instead of reading from `cfg.gas_params()`, ignoring custom gas parameter overrides. This causes validation to reject valid transactions on custom chains with modified gas schedules.


rakita:
Current is fine, new function was added with gas params

PR #3516, opened 2026-03-23
feat(examples): add multi-transaction block simulation example

danieljrc888:
## Summary

- Adds a new `multi_tx_block_sim` example demonstrating how to simulate an entire block of transactions using REVM
- Shows two execution strategies: per-transaction commit (`transact_commit`) and batch commit (`transact_many_commit`)
- Covers contract deployment, multi-account interaction, value transfers, gas tracking, and balance reporting

## What the example demonstrates

1. Configuring a realistic block environment (number, timestamp, coinbase, basefee)
2. Deploying a counter contract and having multiple accounts interact with it
3. Tracking cumulative gas usage across all transactions in a block
4. Comparing per-transaction commit vs batch commit strategies

## Test plan

- [x] Example compiles (`cargo build -p example-multi-tx-block-sim`)
- [x] Example runs successfully (`cargo run -p example-multi-tx-block-sim`)
- [ ] CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1902,6 +1902,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "example-multi-tx-block-sim"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "revm",
+]
+
+[[package]]
 name = "example-my-evm"
 version = "0.0.0"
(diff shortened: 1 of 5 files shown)


rakita:
Will close this as there is alread example of block exec

PR #3519, opened 2026-03-23
Amsterdam devnet3: EIP-8037 state gas (reservoir model)

rakita:
## Summary

Implements **EIP-8037 state gas** using the reservoir model for Amsterdam devnet3. This is a major change to how gas accounting works in revm.

### Core changes

- **GasTracker struct** (`context/interface/src/cfg/gas.rs`): New centralized gas tracker with `remaining`, `reservoir`, `state_gas_spent`, and `refunded` fields. The reservoir holds execution gas exceeding `TX_MAX_GAS_LIMIT`, creating a separate budget for state gas operations.
- **Gas struct updated** (`interpreter/src/gas.rs`): Delegates to `GasTracker` internally. Adds `record_state_cost()` (draws from reservoir first, then spills to remaining) alongside existing `record_cost()` for regular gas. The `GAS` opcode returns only `remaining` (excludes reservoir per EIP-8037).
- **GasParams / GasId** (`context/interface/src/cfg/gas_params.rs`): New parameterized gas cost system with `GasId` constants (1-45) and `GasParams::get(GasId)` lookup, replacing some hardcoded gas values.
- **ResultGas** (`context/interface/src/result.rs`): Extended to carry `state_gas_spent` and `reservoir` through execution results, enabling proper gas accounting in receipts and post-execution.

### Handler changes

- **Frame gas propagation** (`handler/src/frame.rs`): Child frames receive reservoir from parent. On return, parent absorbs child's reservoir and accumulates `state_gas_spent`.
- **Pre-execution** (`handler/src/pre_execution.rs`): EIP-7702 auth list state gas split into separate accounting via `split_eip7702_refund()`.
- **Post-execution** (`handler/src/post_execution.rs`): Floor gas (EIP-7623) and refund cap calculations updated to work with reservoir model. State gas spent excluded from regular refund caps.
- **Validation** (`handler/src/validation.rs`): Transaction validation updated for state gas budget.
- **System calls** (`handler/src/system_call.rs`): Gas handling updated for reservoir model.

### Instruction changes

- **SSTORE** (`interpreter/src/instructions/host.rs`): State gas charging for storage writes.
- **CREATE/CREATE2** (`interpreter/src/instructions/contract.rs`): State gas for code deposit and contract creation. Static call check ordering aligned with execution-specs `generic_create()`.
- **CALL variants** (`interpreter/src/instructions/contract/call_helpers.rs`): State gas for account creation (value transfer to new accounts).
[...]

PR #3521, opened 2026-03-24
feat: rakita/state-gas branch without ee-tests 

rakita:
Removal of tests as github can't handle big PRs


crates/context/interface/src/cfg/gas_params.rs
```diff
@@ -315,6 +315,42 @@ impl GasParams {
             table[GasId::tx_floor_cost_base_gas().as_usize()] = 21000;
         }
 
+        // EIP-8037: State creation gas cost increase
+        if spec.is_enabled_in(SpecId::AMSTERDAM) {
```
klkvr, line 319: @rakita how are we planning to trigger this on tempo side? we can't just enable `AMSTERDAM` i think because its going to be changing further?

crates/handler/src/precompile_provider.rs
```diff
@@ -94,42 +94,63 @@ impl<CTX: ContextTr> PrecompileProvider<CTX> for EthPrecompiles {
...
-            Err(e) => {
-                if e.is_fatal() {
-                    return Err(e.to_string());
+            Err(failure) => {
+                if failure.error.is_fatal() {
+                    return Err(failure.error.to_string());
+                }
+                // If the precompile tracked state gas, propagate it for reservoir refill.
+                if let Some(gas) = failure.gas {
+                    *result.gas.tracker_mut() = gas;
```
fgimenez: should we restore the reservoir after overwriting the tracker? same as we do in the Ok branch

crates/context/interface/src/cfg/gas_params.rs
rakita: Tempo has its own specific gas params [here](https://github.com/tempoxyz/tempo/blob/6974b23d59b3e1e9c0b3d98ff88d94f78d77d8ab/crates/revm/src/gas_params.rs#L23-L24)


EIP-8037 should use `cfg.enable_amsterdam_eip8037` in all other places

crates/handler/src/precompile_provider.rs
rakita: Yes, good catch.

This code is kinda flaky, GasTracker didn't exist in precompiles so with it it moved some things. Will see to fix it.

crates/op-revm/src/handler.rs
```diff
@@ -237,9 +240,15 @@ where
...
             //   - Regular transactions receive a refund on remaining gas as normal.
             if !is_deposit || is_regolith {
+                // Return unused regular gas.
                 gas.erase_cost(remaining);
             }
         }
+
+        // Restore state_gas_spent on all paths (lost by Gas::new_spent overwrite).
+        gas.set_state_gas_spent(state_gas_spent);
+        gas.set_reservoir(reservoir);
```
fgimenez, line 248: here we unconditionally do `gas.set_reservoir(reservoir)`, but in L1 EthHandler on revert/halt we do `gas.set_reservoir(reservoir + state_gas_spent)` i guess to refund rolled-back state gas, should we do the same here?

fgimenez: relevant code in L1 EthHandler https://github.com/bluealloy/revm/pull/3521/changes#diff-20fe56916ac1806f48ceed2ae9e0a8a90a2ccd77cd26d6809425969ed67d868bR484

rakita: I missed this one going back and forth, devnet 3 has a bug in this part of code. devnet3 [PR](https://github.com/bluealloy/revm/pull/3519) had this reverted and I didn't change it for op-revm. Imo fine to leave it as is and will bump it in later eth devnets.

crates/handler/src/precompile_provider.rs
rakita: I noticed that this logic is not used inside revm and only needed for alloy/evm. So i moved it there and reverted changes in revm

crates/handler/src/post_execution.rs
```diff
@@ -10,21 +10,36 @@ use primitives::{hardfork::SpecId, U256};
...
+        .state_gas_spent()
+        .saturating_add(init_and_floor_gas.initial_state_gas)
+        .saturating_sub(init_and_floor_gas.eip7702_reservoir_refund);
+
+    ResultGas::default()
+        .with_total_gas_spent(
+            gas.limit()
+                .saturating_sub(gas.remaining())
+                .saturating_sub(gas.reservoir()),
+        )
```
rakita, line 23: This is confusing, and I am not sure if it is correct.

stategas tests is passing but on hive `ResultGas::block_regular_gas_used` that does `total_gas_spent - state_gas` wound underflow (Still checking it)

Would love additional eyes on it. Gas is constructed in handler.rs in `fn last_frame_result` lines 440-480

crates/interpreter/src/gas.rs
```diff
@@ -3,27 +3,68 @@
...
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Gas {
     /// The initial gas limit. This is constant throughout execution.
     limit: u64,
-    /// The remaining gas.
-    remaining: u64,
-    /// Refunded gas. This is used only at the end of execution.
-    refunded: i64,
+    /// Tracker for gas during execution.
+    tracker: GasTracker,
```
fgimenez, line 23: iiuc `Gas::limit` and `GasTracker::gas_limit` store the same value, `Gas` never reads `tracker.limit()`, it always uses its own `self.limit` correct? 

if so maybe we could drop `gas_limit` from `GasTracker` or just inline the tracker fields directly into `Gas`, it's the only consumer, wdyt?

crates/context/interface/src/result.rs
fgimenez, line 53: this doc table is stale, `limit()`, `intrinsic_gas()`, and `remaining()` are removed from `ResultGas`, `spent()` deprecated

PR #3523, opened 2026-03-25
refactor(context-interface): deduplicate SSTORE refund sync in GasParams

anim001k:
- deduplicate the `sstore_set_refund` / `sstore_reset_refund` synchronization in `GasParams::new_spec_inner`
- move the shared logic into a small private helper so refund slots stay aligned with the corresponding SSTORE cost slots
- keep behavior unchanged while reducing the chance of future drift when SSTORE gas values are updated


rakita:
Current is fine

PR #3530, opened 2026-03-30
perf(database): avoid clones in BundleState::take_n_reverts

saylor-mik87786:
Replace split_at plus double to_vec in take_n_reverts with move-based split_off, reducing allocation and recomputation overhead when truncating reverts.


rakita:
Superseeded by https://github.com/bluealloy/revm/pull/3539/changes

PR #3533, opened 2026-03-31
Gate EIP-7702 delegation-follow behind Prague in call paths

ArshLabs:
## Summary
This PR adds explicit Prague gating for EIP-7702 delegation-follow behavior in call resolution paths.

Without explicit fork gating, delegation-shaped code could be followed in pre-Prague contexts in call paths where Prague rules should not apply.

## What changed
- Gate delegation extraction in `load_account_delegated` behind `SpecId::PRAGUE` in interpreter call helpers.
- Gate top-level call-frame delegation extraction in `create_init_frame` behind `SpecId::PRAGUE`.
- Align journal delegated-load extraction with the existing `is_eip7702_enabled` flag.
- Add regressions for both pre-Prague no-delegation and Prague delegation behavior across these paths.

## Tests
- `cargo fmt --check`
- `cargo build`
- `cargo test`
- `cargo clippy -- -D warnings`
- `cargo test -p revm-interpreter call_helpers::tests -- --nocapture`
- `cargo test -p revm-handler execution::tests -- --nocapture`
- `cargo test -p revm-context journal::inner::tests::load_account_delegated_ -- --nocapture`

### DIFF
--- a/crates/context/src/journal/inner.rs
+++ b/crates/context/src/journal/inner.rs
@@ -758,17 +758,19 @@ impl<ENTRY: JournalEntryTr> JournalInner<ENTRY> {
             account.is_cold,
         );
 
-        // load delegate code if account is EIP-7702
-        if let Some(address) = account
-            .info
-            .code
-            .as_ref()
-            .and_then(Bytecode::eip7702_address)
-        {
-            let delegate_account = self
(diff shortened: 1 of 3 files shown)


rakita:
This is not needed as deployment can't push for eip7702 type without enabling its tx.

PR #3535, opened 2026-04-01
fix(context): use storage_by_account_id fast path in sload

bluebluesaku:
`sload_concrete_error` was calling `db.storage(address, key)` directly on cold load, ignoring `account_id` even when set. This bypassed the `storage_by_account_id` fast path designed for IndexMap/BAL backends.


rakita (approved):
lgtm, will merge it after new release

DaniPopes (approved):

PR #3539, opened 2026-04-03
perf(database): avoid double cloning in take_n_reverts

saylor-mik87786:
Replace split_at plus double to_vec with move-based split_off in take_n_reverts.


rakita:
i see, lgtm

PR #3540, opened 2026-04-03
Add state gas reservoir refund for SSTORE 0→x→0

rakita:
## Summary
- When a storage slot goes 0→x→0 within a transaction, refund the `sstore_set_state_gas` back to the reservoir
- Adds `GasTracker::refund_state_cost()` and `GasParams::sstore_state_gas_refund()`
- Wired into the SSTORE instruction alongside the existing state gas charge

## Test plan
- [x] Updated `test_eip8037_sstore_set_then_clear_refund` — now expects zero net state gas for 0→1→0
- [x] All 37 EIP-8037 tests pass
- [ ] CI passes


crates/context/interface/src/cfg/gas.rs
```diff
@@ -145,6 +145,16 @@ impl GasTracker {
...
     }
 
+    /// Refunds state gas back to the reservoir.
+    ///
+    /// Used when a state change is undone (e.g. storage slot cleared back to zero),
+    /// returning the previously charged state gas to the reservoir.
+    #[inline]
+    pub fn refund_state_cost(&mut self, amount: u64) {
+        self.reservoir += amount;
+        self.state_gas_spent = self.state_gas_spent.saturating_sub(amount);
```
rakita, line 155: State gas is local to the call, so it can go to negative if we have a refund in one of the child calls. it needs to be local so we know how much gas is returned on revert/halt. Maybe the solution is to make it i64 so we can add it to parrent.

rakita:
Closing it as it was just a atempt

PR #3547, opened 2026-04-07
fix: update comments for initial tx gas related functions

AntonD3:
Updated comments for initial gas calculation related functions, comments didn't really correspond to the current functions behavior.


rakita:
Thanks, nit prs will be updated later in bulk

PR #3553, opened 2026-04-10
fix: reject transactions with nonce at u64::MAX

rakita:
## Summary
- Adds validation in `validate_tx_env` to reject transactions where the nonce equals `u64::MAX`, preventing nonce overflow when incrementing.

## Test plan
- [ ] Existing tests pass
- [ ] Verify `InvalidTransaction::NonceOverflowInTransaction` is returned for nonce = u64::MAX

### DIFF
--- a/crates/handler/src/validation.rs
+++ b/crates/handler/src/validation.rs
@@ -227,7 +227,7 @@ pub fn validate_tx_env<CTX: ContextTr>(
     }
 
     // Check that the transaction's nonce is not at the maximum value.
-    // Incrementing the nonce would overflow.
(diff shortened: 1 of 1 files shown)


crates/handler/src/validation.rs
```diff
@@ -226,6 +226,12 @@ pub fn validate_tx_env<CTX: ContextTr>(
         return Err(InvalidTransaction::CreateInitCodeSizeLimit);
     }
 
+    // Check that the transaction's nonce is not at the maximum value.
+    // Incrementing the nonce would overflow.
```
rakita:
```suggestion
    // Incrementing the nonce would overflow. Can't happen in the real world.
```

PR #3557, opened 2026-04-12
fix(bytecode): mark `Bytecode::new_analyzed` as unsafe

decofe:
`Bytecode::new_analyzed` accepts pre-analyzed bytecode but does not validate
that the padding is sufficient for safe execution of PUSH immediates. Calling
it with insufficient padding causes UB: `ExtBytecode::read_slice` performs an
out-of-bounds read via `core::slice::from_raw_parts`.

This PR:
- marks `new_analyzed` as `unsafe` with a safety doc referencing the `analyze_legacy` padding invariant
- fixes the serde `Deserialize` impl to re-analyze from original bytes instead of trusting serialized padding/jump table
- adds a `#[cfg(miri)]` reproducer that demonstrates the OOB read
- updates `MIGRATION_GUIDE.md`

Miri reproducer:
```
cargo +nightly miri test --test miri_read_slice_oob -p revm-interpreter
```

```
error: Undefined Behavior: pointer not dereferenceable: pointer must be
dereferenceable for 32 bytes, but got alloc16+0x1 which is only 1 byte
from the end of the allocation
 --> crates/interpreter/src/interpreter/ext_bytecode.rs:178:17
```

**Note for downstream (reth):** `reth-primitives-traits` calls `new_analyzed` in
its `Compact::from_compact` DB deserializer. That call site is safe (it restores
previously padded bytes), but will need `unsafe {}` + safety comment after
upgrading `revm-bytecode`.

Prompted by: pep

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -42,7 +42,7 @@
   * `bytecode.legacy_jump_table()` instead of matching `Bytecode::LegacyAnalyzed(...)`.
 * Constructors:
   * `Bytecode::new_legacy(raw_bytes)` — replaces `Bytecode::LegacyRaw(...)` (analysis happens on creation).
-  * `unsafe { Bytecode::new_analyzed(bytecode, original_len, jump_table) }` — replaces `Bytecode::LegacyAnalyzed(Arc::new(...))`. Marked `unsafe` because the caller must ensure bytecode padding is sufficient (see doc comment).
+  * `Bytecode::new_analyzed(bytecode, original_len, jump_table)` — replaces `Bytecode::LegacyAnalyzed(Arc::new(...))`.
   * `Bytecode::new_eip7702(address)` — replaces `Bytecode::Eip7702(Arc::new(...))`.
   * `Bytecode::new_raw(bytes)` — auto-detects EIP-7702 delegation.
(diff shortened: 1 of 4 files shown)


DaniPopes (approved):

MIGRATION_GUIDE.md
```diff
@@ -42,7 +42,7 @@
   * `bytecode.legacy_jump_table()` instead of matching `Bytecode::LegacyAnalyzed(...)`.
 * Constructors:
   * `Bytecode::new_legacy(raw_bytes)` — replaces `Bytecode::LegacyRaw(...)` (analysis happens on creation).
-  * `Bytecode::new_analyzed(bytecode, original_len, jump_table)` — replaces `Bytecode::LegacyAnalyzed(Arc::new(...))`.
+  * `unsafe { Bytecode::new_analyzed(bytecode, original_len, jump_table) }` — replaces `Bytecode::LegacyAnalyzed(Arc::new(...))`. Marked `unsafe` because the caller must ensure bytecode padding is sufficient (see doc comment).
```
rakita:
```suggestion
  * `Bytecode::new_analyzed(bytecode, original_len, jump_table)` — replaces `Bytecode::LegacyAnalyzed(Arc::new(...))`.
```

rakita (approved):

PR #3558, opened 2026-04-12
refactor!: return Result from instruction functions

DaniPopes:
Convert all instruction functions from returning () to returning
Result<(), InstructionResult> (InstructionExecResult). All macros now
use return Err(...) instead of interpreter.halt() + return.


DaniPopes:
perf changes are inlining / malloc / different machine noise.

crates/interpreter/src/instructions/arithmetic.rs
```diff
@@ -1,87 +1,97 @@
...
         *op2 = op1.wrapping_div(*op2);
     }
+    Ok(())
 }
 
 /// Implements the SDIV instruction.
 ///
 /// Performs signed division of two values from stack.
-pub fn sdiv<WIRE: InterpreterTypes, H: ?Sized>(context: InstructionContext<'_, H, WIRE>) {
+pub fn sdiv<WIRE: IT, H: ?Sized>(context: Icx<'_, H, WIRE>) -> Result {
```
rakita: I like abbreviations, but lets do `ITy` (for Interpreter types) and Ictx (for context, since cx was never used). 

Additionaly do search and replace of `WIRE: ITy` to maybe `IT: ITy` and `<'_, H, WIRE>` to `<'_, H, IT>`

rakita (approved):
one nit on naming, but otherwise lgtm

PR #3562, opened 2026-04-14
refactor(interpreter): dedup account loading across host instructions

DaniPopes:
Replace the duplicated `if berlin { berlin_load_account!() } else { load_account_info_skip_cold_load() }` pattern in balance, extcodesize, extcodehash, and extcodecopy with a single `load_account` function.

Pre-Berlin, `cold_account_additional_cost` is 0, so the cold load gas logic is a no-op and the function works correctly for all spec versions.

Removes the `berlin_load_account!` macro entirely and adds `Interpreter::halt_load_error` to map `LoadError` variants to the appropriate halt.


rakita (commented):
nice cleanup!

PR #3586, opened 2026-04-21
perf: avoid cloning precompiles on warmup

DaniPopes:
Removes the Box + collect in `.warm_precompiles(precompiles.warm_addresses().collect());`.
`Precompiles` already stores the addresses as a separate set so we can use that directly.


crates/handler/src/precompile_provider.rs
```diff
@@ -53,8 +52,8 @@ impl EthPrecompiles {
     }
 
     /// Returns addresses of the precompiles.
-    pub fn warm_addresses(&self) -> Box<impl Iterator<Item = Address>> {
-        Box::new(self.precompiles.addresses().cloned())
+    pub fn warm_addresses(&self) -> &AddressSet {
```
rakita, line 55: Acking that alloy/evm [uses](https://github.com/alloy-rs/evm/blob/6c6907452b9e20a1202c2b8a4b861d5f6526f667/crates/evm/src/precompiles.rs#L633) `AddressSet` too

PR #3588, opened 2026-04-21
perf(interpreter): optimize MLOAD/MSTORE with direct u256 memory ops

pepyakin:
Porting one of the optimizations from the batch. This one focuses on mstore and mload.

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -6,6 +6,7 @@ use core::{
     cmp::min,
     fmt,
     ops::Range,
+    ptr,
 };
 use primitives::{hex, B256, U256};
 use std::{rc::Rc, vec::Vec};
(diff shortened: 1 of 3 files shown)


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -89,6 +90,34 @@ impl MemoryTr for SharedMemory {
...
+    fn get_u256(&self, offset: usize) -> U256 {
+        let buffer = self.buffer_ref();
+        let start = self.my_checkpoint + offset;
+        assert!(start + 32 <= buffer.len(), "get_u256 out of bounds");
+        let ptr = unsafe { buffer.as_ptr().add(start) };
+        let limb3 = u64::from_be(unsafe { ptr::read_unaligned(ptr.cast::<u64>()) });
+        let limb2 = u64::from_be(unsafe { ptr::read_unaligned(ptr.add(8).cast::<u64>()) });
+        let limb1 = u64::from_be(unsafe { ptr::read_unaligned(ptr.add(16).cast::<u64>()) });
+        let limb0 = u64::from_be(unsafe { ptr::read_unaligned(ptr.add(24).cast::<u64>()) });
+        U256::from_limbs([limb0, limb1, limb2, limb3])
```
DaniPopes, line 103: U256::from_be_slice should be identical to this, if this manual impl is better then it's a bug in ruint

crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -89,6 +90,34 @@ impl MemoryTr for SharedMemory {
...
+        let mut buffer = self.buffer_ref_mut();
+        let start = self.my_checkpoint + offset;
+        assert!(start + 32 <= buffer.len(), "set_u256 out of bounds");
+        let ptr = unsafe { buffer.as_mut_ptr().add(start) };
+        unsafe {
+            ptr::write_unaligned(ptr.cast::<u64>(), limbs[3].to_be());
+            ptr::write_unaligned(ptr.add(8).cast::<u64>(), limbs[2].to_be());
+            ptr::write_unaligned(ptr.add(16).cast::<u64>(), limbs[1].to_be());
+            ptr::write_unaligned(ptr.add(24).cast::<u64>(), limbs[0].to_be());
+        }
```
DaniPopes, line 118: same here, to_be_bytes() + memcpy should be identical to the unrolled impl. i would first try to make it work in ruint

PR #3589, opened 2026-04-21
refactor(gas): rename InitialAndFloorGas.initial_total_gas to regular_gas

rakita:
## Summary

Refactors `InitialAndFloorGas` to store the regular and state portions of the intrinsic gas as separate fields and derive the total on demand, rather than keeping a pre-merged total plus a state subset. Also adds small accumulator helpers so call sites don't have to touch the fields directly.

### Field change

| Before                      | After               | Meaning                                     |
| --------------------------- | ------------------- | ------------------------------------------- |
| `initial_total_gas: u64`    | `regular_gas: u64`  | Regular (non-state) portion only            |
| `initial_state_gas: u64`    | `state_gas: u64`    | EIP-8037 state-gas portion                  |
| `floor_gas`                 | unchanged           | EIP-7623 floor                              |
| `eip7702_reservoir_refund`  | unchanged           | EIP-7702 state refund for existing auths    |

### New methods

- `total_gas(&self) -> u64` — saturating `regular_gas + state_gas`
- `add_regular_gas(&mut self, gas: u64)` — saturating `+=` on the regular portion
- `add_state_gas(&mut self, gas: u64)` — saturating `+=` on the state portion
- `initial_regular_gas()` retained as a transition alias for pre-rename call sites

### Migration

Downstream code that read `gas.initial_total_gas` should switch to `gas.total_gas()`; direct writes (`gas.initial_total_gas += x`) become `gas.add_regular_gas(x)` (if the increment was regular) or `gas.add_state_gas(x)` (if it was state). `gas.initial_state_gas` becomes `gas.state_gas`.

### Internal callers updated

- `GasParams::initial_tx_gas` — uses `add_regular_gas` / `add_state_gas`; dropped the trailing `initial_total_gas += initial_state_gas` merge.
- `validation.rs` — `total_gas()` for the intrinsic-vs-gas-limit check; `gas.regular_gas` for the EIP-8037 regular cap check.
- `post_execution.rs::build_result_gas` — reads `state_gas`.
- Doc comments in `pre_execution.rs` / `gas_params.rs` updated to the new field names.

## Test plan

- [x] `cargo check --workspace` — clean.
- [x] `cargo test --workspace --lib` — all green, including `test_initial_state_gas_for_create` (updated to use `state_gas` + `total_gas()`) and the handler validation suite.
- [x] `cargo clippy --workspace --all-targets` — no new warnings.

### DIFF
--- a/crates/context/interface/src/cfg/gas.rs
+++ b/crates/context/interface/src/cfg/gas.rs
@@ -263,63 +263,89 @@ pub const INITCODE_WORD_COST: u64 = 2;
 /// Gas stipend provided to the recipient of a CALL with value transfer.
 pub const CALL_STIPEND: u64 = 2300;
 
-/// Init and floor gas from transaction
+/// Init and floor gas from transaction.
+///
+/// Tracks the intrinsic gas split into regular and state-gas portions
+/// (EIP-8037). Use [`Self::total_gas`] for the combined amount,
+/// [`Self::add_regular_gas`] / [`Self::add_state_gas`] to accumulate into
+/// the respective portion.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct InitialAndFloorGas {
-    /// Initial gas for transaction.
-    pub initial_total_gas: u64,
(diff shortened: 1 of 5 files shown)


rakita:
Superseeded by `bal-devnet-6` pr

PR #3590, opened 2026-04-21
perf: no alloc for empty accounts

### LINKED ISSUE #3585: Make `Account::original_info` optional
Default account should start with `None`. Allows for cheaper `Account::default` since then we can also drop the static and the clone.

DaniPopes:
Avoids allocating a `Box<AccountInfo>` for `original_info` when the account is default/empty by using `Option<Box<AccountInfo>>` where `None` represents the default.

Key changes:

- `Account::original_info` is now `Option<Box<AccountInfo>>` — `None` means `Default::default()`
- `Bytecode::is_default()` works after serde roundtrip by returning the static default from deserialization when `original_len == 0`
- `AccountInfo::is_default()` optimization hint using `Bytecode::is_default()`
- Custom `PartialEq` for `Account` that normalizes `None` vs `Some(default)` via `original_info()` getter
- Backwards-compatible serde: missing field (old format) clones current info, `null` (new format) means `None`/default, present value is used directly

Closes https://github.com/bluealloy/revm/issues/3585.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1145,6 +1145,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63044e1ae8e69f3b5a92c736ca6269b8d12fa7efe39bf34ddb06d102cf0e2cab"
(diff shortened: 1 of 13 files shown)


rakita (approved):
Nice one!

PR #3593, opened 2026-04-22
refactor(primitives): add SpecId::NEXT, drop num_enum dependency

DaniPopes:
- Add `SpecId::NEXT` const that always points to the max variant, with `doc(alias = "MAX")` and a warning that it may be super experimental / not finalized.
- Remove `TryFromPrimitive` derive and `num_enum` dependency entirely.
- Implement `TryFrom<u8>` for `SpecId` and `From<SpecId>` for `u8` manually using transmute (bounded by `NEXT`).
- Make `try_from_u8` const.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3008,7 +3008,6 @@ version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff32365de1b6743cb203b710788263c44a03de03802daf96092f2da4fe6ba4d7"
 dependencies = [
- "proc-macro-crate",
  "proc-macro2",
  "quote",
  "syn 2.0.117",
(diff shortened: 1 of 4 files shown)


rakita:
Not sure what the problem with using `num_enum`. It is a small utility.

DaniPopes:
Not const fn, I can't make a [T; SpecId::COUNT] for each spec id without unsafe

rakita:
> Not const fn, I can't make a [T; SpecId::COUNT] for each spec id without unsafe

Is this used at all? If not, it doesn't matter it is not const?

DaniPopes:
It's a common pattern, as seen in precompiles and GasParams, to have T for each spec id. i use it in revmc for extra per-opcode data

PR #3599, opened 2026-04-22
perf: move memory_limit check into cold resize path

DaniPopes:
Move `limit_reached` from the hot `resize_memory` function into `resize_memory_cold`, keeping the fast path free of the feature-gated branch.

The `MemoryTr::limit_reached` signature is simplified to take `new_words` instead of `(offset, len)` since the caller has already computed it.

Also adds a test exercising the `memory_limit` path through `resize_memory_cold`.

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -129,7 +129,7 @@ impl MemoryTr for SharedMemory {
     #[cfg(feature = "memory_limit")]
     #[inline]
     fn limit_reached(&self, new_words: usize) -> bool {
-        self.my_checkpoint.saturating_add(new_words * 32) as u64 > self.memory_limit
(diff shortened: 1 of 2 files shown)


crates/interpreter/src/interpreter/shared_memory.rs
```diff
@@ -124,15 +124,12 @@ impl MemoryTr for SharedMemory {
...
     /// make the shared buffer length exceed the `memory_limit`.
     #[cfg(feature = "memory_limit")]
     #[inline]
-    fn limit_reached(&self, offset: usize, len: usize) -> bool {
-        self.my_checkpoint
-            .saturating_add(offset)
-            .saturating_add(len) as u64
-            > self.memory_limit
+    fn limit_reached(&self, new_words: usize) -> bool {
+        self.my_checkpoint.saturating_add(new_words * 32) as u64 > self.memory_limit
```
rakita:
```suggestion
        self.my_checkpoint.saturating_add(new_words.saturating_mul(32)) as u64 > self.memory_limit
```

prob need to saturate

PR #3601, opened 2026-04-23
refactor: remove SharedMemory in favor of per-interpreter owned Memory

DaniPopes:
Replace `SharedMemory` (`Rc<RefCell<Vec<u8>>>` with checkpoint-based child/parent context tracking) with a simple `Memory` struct that owns a `Vec<u8>` directly. Each interpreter frame now has its own memory with 4KiB initial capacity.

This removes significant complexity from the codebase:

- **Remove `CallInput` enum**: replaced the `SharedBuffer(Range<usize>)` / `Bytes` variants with just `Bytes` everywhere. Call input data is now copied from memory at call time.
- **Remove `FrameInit.memory` field**: no shared memory to pass between frames.
- **Simplify `LocalContextTr`**: removed `shared_memory_buffer()`, `shared_memory_buffer_slice()`, and the backing `shared_memory_buffer` field from `LocalContext`.
- **Simplify `MemoryTr` trait**: removed `set_data_from_global`, `global_slice`, `local_memory_offset` methods. Changed `slice()` return type from `Ref<'_, [u8]>` to plain `&[u8]`.
- **Remove shared memory lifecycle**: removed `new_child_context()` / `free_child_context()` calls from handler frame logic.
- **Simplify call helpers**: `get_memory_input_and_out_ranges` now returns `(Bytes, Range<usize>)` instead of `(Range<usize>, Range<usize>)`.

22 files changed, ~150 insertions, ~770 deletions.


rakita:
The current version is an optimisation on memory usage and access, there is no reason to remove it

DaniPopes:
Frames are pooled, the only thing this really optimizes is a single clone of calldata on CALL, which i dont think pulls its weight in 800 lines of code and making CALLDATALOAD slower too

PR #3602, opened 2026-04-23
refactor(precompile): use static OnceLock array for Precompiles init

DaniPopes:
Replaces per-spec `static OnceLock` instances with a single `static [OnceLock<Precompiles>; PrecompileSpecId::NEXT as usize + 1]` array indexed by `PrecompileSpecId`.

Adds `PrecompileSpecId::NEXT` and `is_enabled_in()`, and introduces a standalone `init_precompiles()` function that builds precompiles incrementally using `if spec.is_enabled_in(X)` guards, similar to `gas_table_spec`.

### DIFF
--- a/crates/precompile/src/lib.rs
+++ b/crates/precompile/src/lib.rs
@@ -436,8 +436,8 @@ impl PrecompileSpecId {
     }
 
     /// Returns the appropriate precompile Spec for the primitive [SpecId].
-    pub const fn from_spec_id(spec_id: primitives::hardfork::SpecId) -> Self {
-        use primitives::hardfork::SpecId::*;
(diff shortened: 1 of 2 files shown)


crates/precompile/src/lib.rs
```diff
@@ -87,143 +87,74 @@ pub struct Precompiles {
     /// Addresses of precompiles.
     addresses: AddressSet,
     /// Optimized addresses filter.
-    optimized_access: Vec<Option<Precompile>>,
-    /// `true` if all precompiles are short addresses.
-    all_short_addresses: bool,
```
DaniPopes, line 92: all_short_addresses was unused

PR #3605, opened 2026-04-23
feat(account): Optimized index type for account ID using `non-max`

### LINKED ISSUE #3583: Use optimized index type for account ID
Account ID is currently `Option<usize>` which is 16 bytes. This is very wasteful especially when it's almost always `None`. Use a custom type that wraps [`NonMaxU32`](https://docs.rs/nonmax/) so that `Option<AccountId>` becomes 4 bytes.

developeruche:
Closes #3583

### DIFF
--- a/crates/state/src/account_info.rs
+++ b/crates/state/src/account_info.rs
@@ -13,16 +13,19 @@ use nonmax::NonMaxU32;
 pub struct AccountId(NonMaxU32);
 
 impl AccountId {
-    /// Creates a new AccountId. Returns `None` if the value is `u32::MAX`.
+    /// Creates a new AccountId.
+    ///
+    /// Returns `None` if the value does not fit in the internal representation.
     #[inline]
-    pub fn new(id: u32) -> Option<Self> {
+    pub fn new(id: usize) -> Option<Self> {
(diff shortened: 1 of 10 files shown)


crates/state/src/bal.rs
```diff
@@ -146,7 +146,7 @@ impl Bal {
         let Some((_, bal_account)) = self.accounts.get_index(account_id) else {
```
DaniPopes: let's make Bal struct also use this ID, so that we avoid usize entirely.

DaniPopes (changes requested):

DaniPopes (approved):

crates/state/src/account_info.rs
```diff
@@ -5,6 +5,27 @@ use core::{
...
+impl AccountId {
+    /// Creates a new AccountId. Returns `None` if the value is `u32::MAX`.
+    #[inline]
+    pub fn new(id: u32) -> Option<Self> {
+        NonMaxU32::new(id).map(Self)
+    }
+
+    /// Gets the underlying u32 value.
+    #[inline]
+    pub const fn get(self) -> u32 {
```
DaniPopes: let's use usize in the public api

PR #3609, opened 2026-04-24
perf(precompile): vendor blake2b_simd for BLAKE2 compression

DaniPopes:
Vendor the portable and AVX2 BLAKE2b compression functions from [blake2b_simd](https://github.com/oconnor663/blake2_simd) (MIT), replacing the hand-rolled implementations.

The vendored code uses blake2b_simd's dispatch approach for runtime/compile-time AVX2 detection, and the implementations are adapted for EIP-152's variable round count.

## Changes to the vendored code

The original blake2b_simd code is designed for standard BLAKE2b (fixed 12 rounds, streaming/incremental hashing). The following adaptations were made for EIP-152:

**portable.rs** (from `blake2b/src/portable.rs`):
- Extracted only the `g`, `round`, and `compress_block` functions; removed the `compress1_loop` streaming wrapper and all `Finalize`/`LastNode`/`Stride`/`Count` machinery.
- Changed `compress_block` signature from `(block: &[u8; 128], words, count: u128, last_block: u64, last_node: u64)` to `(rounds: u32, h, m: &[u64; 16], t: &[u64; 2], f: bool)` matching EIP-152's F function.
- Replaced the 12 hardcoded `round(N, ...)` calls with a `for i in 0..rounds` loop.
- Removed byte-to-u64 deserialization (the precompile already parses m as `[u64; 16]`).
- Changed `Word` type alias to direct `u64`, `SIGMA` entries from `u8` to `usize`.

**avx2.rs** (from `blake2b/src/avx2.rs`):
- Extracted only the single-block `compress_block` function and its helpers; removed `compress1_loop`, `compress4_loop`, and all multi-hash (`Job`, transpose, count management) code.
- Changed signature to match EIP-152's F function (same as portable).
- Kept the 10 unrolled SIMD rounds with `remaining` counter and break checks for variable round support (same pattern as the previous revm implementation).
- Removed `compress4_loop` / `compress2_loop` / `transpose_*` / `Job` — only single-block compression is needed for EIP-152.

**mod.rs** (dispatch):
- Replaces the old single-file `blake2.rs` with a module containing `portable.rs` and `avx2.rs`.
- Dispatch function uses compile-time (`#[cfg(target_feature)]`) then runtime (`is_x86_feature_detected!`) AVX2 detection, falling back to portable. The old code only used compile-time `target_feature` gating.


crates/precompile/src/blake2/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15],
+    [14, 10, 4, 8, 9, 15, 13, 6, 1, 12, 0, 2, 11, 7, 5, 3],
+];
+
+/// BLAKE2b compression function F (EIP-152).
+///
+/// Dispatches to the best available implementation (AVX2 or portable).
+pub fn compress(rounds: u32, h: &mut [Word; 8], m: &[Word; 16], t: &[Word; 2], f: bool) {
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
rakita, line 51: @DaniPopes was this the only change, and the rest is refactoring?

DaniPopes: yes, i think impl is identical, maybe a bit newer revision from vendored crate

crates/precompile/src/blake2/portable.rs
```diff
@@ -0,0 +1,82 @@
...
+    v[a] = v[a].wrapping_add(v[b]).wrapping_add(y);
+    v[d] = (v[d] ^ v[a]).rotate_right(16);
+    v[c] = v[c].wrapping_add(v[d]);
+    v[b] = (v[b] ^ v[c]).rotate_right(63);
+}
+
+#[inline(always)]
+const fn round(r: usize, m: &[Word; 16], v: &mut [Word; 16]) {
+    // Select the message schedule based on the round.
+    let s = SIGMA[r % 12];
```
rakita, line 34: There is a bug here, it should be 10 rounds, not 12. This was a reason why this code is copied here as eth precompile has this difference.

Will make a fix PR

rakita: Code [here](https://github.com/bluealloy/revm/blob/d62933180e442319a93991fe4070dbeb466ef222/crates/precompile/src/blake2.rs#L77) and [here](https://github.com/bluealloy/revm/blob/d62933180e442319a93991fe4070dbeb466ef222/crates/precompile/src/blake2.rs#L178)

PR #3610, opened 2026-04-25
feat(state): Optimized index type for transaction ID using non-max

### LINKED ISSUE #3584: Use optimized index type for transaction ID
Same as #3583, but for `Account::transaction_id`.

yottaes:
Closes #3584.

Mirrors #3605 for `Account::transaction_id` and `EvmStorageSlot::transaction_id`.

Introduces a `TransactionId` newtype wrapping `nonmax::NonMaxU32`, so the field shrinks from 8 bytes to 4 bytes on 64-bit and any `Option<TransactionId>` benefits from niche optimization.

`JournalInner::transaction_id`, `JournaledAccount::transaction_id` and the `is_cold_transaction_id` / `mark_warm_with_transaction_id` / `with_warm` / `new_not_existing` / `EvmStorageSlot::new[_changed]` APIs now take `TransactionId` directly, so the id no longer flows around as `usize`.

API surface follows the same shape as `AccountId` in #3605: `new(u32) -> Option<Self>`, `get(self) -> u32`, plus a `ZERO` const and an `increment` helper for the journal hot-path.

## Breaking changes

* Public fields `Account::transaction_id`, `EvmStorageSlot::transaction_id` and `JournalInner::transaction_id` change type from `usize` to `TransactionId`.
* `Account::new_not_existing`, `Account::is_cold_transaction_id`, `Account::mark_warm_with_transaction_id`, `Account::with_warm`, `Account::with_warm_mark`, `EvmStorageSlot::new`, `EvmStorageSlot::new_changed`, `EvmStorageSlot::is_cold_transaction_id`, `EvmStorageSlot::mark_warm_with_transaction_id` now take `TransactionId` rather than `usize`.

Surface shape mirrors #3605, so downstream can migrate both at once.


borngraced (commented):
sorry for jumping into your PR uninvited 😅 I a question: 

the original `transaction_id` has it's type as `usize` however you chose to use `NonMaxU32` for the NewType instead of ``NonMaxUsize`. Any reason?

UPD: seen the issue https://github.com/bluealloy/revm/issues/3584. its all good

DaniPopes (approved):

PR #3616, opened 2026-04-26
fix(precompile): use SIGMA period 10 in blake2 portable path

rakita:
## Summary

The vendored portable BLAKE2b compression introduced in #3609 indexes `SIGMA` with `r % 12` against a 12-entry table where rows 10/11 duplicate rows 0/1. That layout is a shortcut from `blake2b_simd` (which only ever runs the fixed 12 rounds of BLAKE2b and indexes `SIGMA[r]` directly without modulo), but EIP-152 takes a `u32` round count, and `SIGMA` is defined with **period 10** (RFC 7693 §2.7). So for round `r >= 12` the portable path picked the wrong row:

| `r` | spec / AVX2 (`r % 10`) | portable (`r % 12`) |
|-----|------------------------|---------------------|
| 10  | SIGMA[0] | SIGMA[10] (= SIGMA[0]) ✓ |
| 11  | SIGMA[1] | SIGMA[11] (= SIGMA[1]) ✓ |
| 12  | SIGMA[2] | SIGMA[0] ✗ |
| 13  | SIGMA[3] | SIGMA[1] ✗ |
| ... | ... | ... |

The AVX2 path is unaffected — it has 10 unrolled rounds in a `loop { ... }` with an early break, so it already implements period-10 wraparound.

This trims `SIGMA` in `mod.rs` to the 10 spec rows and switches the portable index from `r % 12` to `r % 10`.

Caught running the EIP-152 state-test vectors against the `bal@v5.7.0` Amsterdam fixture set:

```
test_blake2b[fork_Amsterdam-state_test-valid-rounds-1024-call_opcode_CALL]      ✗ → ✓
test_blake2b[fork_Amsterdam-state_test-valid-rounds-1024-call_opcode_CALLCODE]  ✗ → ✓
test_blake2b_large_gas_limit[fork_Amsterdam-...case8-data9-call_opcode_CALL]    ✗ → ✓
test_blake2b_large_gas_limit[fork_Amsterdam-...case8-data9-call_opcode_CALLCODE]✗ → ✓
```

## Test plan

- [x] `cargo build -p revm-precompile`
- [x] `cargo run -p revme -- stest --keep-going ../fixtures_bal_v570/state_tests/for_amsterdam/` — 4 blake2 failures cleared, all tests pass


DaniPopes:
Why is this not caught in CI? We should run state tests on macos too?

DaniPopes (approved):

PR #3619, opened 2026-04-27
feat: show address or slot in BAL error

pepyakin:
When an item, either the address or a slot, is not found in the BAL it's useful to know what item is missing exactly.

This changest plumbs this info into the error.

We had to introduce a new error variant into the BalError enum because otherwise there is no way to report the address.


rakita (approved):
lgtm

DaniPopes (approved):

PR #3622, opened 2026-04-28
feat(eip8037): defer state-gas OOG to frame return

rakita:
## Summary

- Removes the inline EIP-8037 state-gas OOG check from SSTORE, SELFDESTRUCT, CREATE/CREATE2, CALL into empty accounts, and code deposit. The opcodes always succeed; `record_state_cost` saturates `reservoir`/`remaining` to zero and sets a new `state_gas_oog` flag on `GasTracker` when the budget cannot cover the charge.
- The check moves to `process_next_action`: if the flag is set on a frame that finished successfully, the result is converted to `OutOfGas` and the frame is reverted (state changes rolled back).
- `last_frame_result` zeroes the reservoir when `state_gas_oog` is set on a failed frame so the full transaction gas is consumed, matching standard OOG semantics (a previous `refill_reservoir` from a failed CREATE could otherwise re-inflate it).

This implements the rule: *"if the child frame does not have enough gas to cover all of its state creation, the call will revert."*

## Test plan

- [x] `cargo nextest run --workspace` — 327 passed
- [x] `cargo nextest run -p revm-ee-tests` — all EIP-8037 tests pass
- [x] Updated `test_eip8037_reservoir_refill_halt_state_gas_more` to assert full-gas consumption (was previously asserting reservoir refund). Regenerated the corresponding testdata.
- [x] Updated unit tests in `revm-interpreter::gas` for the new `record_state_cost` semantics (state_gas_spent always increments; OOG surfaces via the flag).
- [ ] CI green

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -95,12 +95,12 @@ pub trait Cfg {
     /// `cost_per_state_byte` and adds a hash cost for deployed bytecode.
     fn is_amsterdam_eip8037_enabled(&self) -> bool;
 
-    /// Returns the EIP-8037 `cost_per_state_byte` (CPSB) for the given block gas limit.
+    /// Returns the EIP-8037 `cost_per_state_byte` (CPSB).
     ///
     /// When [`Cfg::is_amsterdam_eip8037_enabled`] is `false` this returns `0`.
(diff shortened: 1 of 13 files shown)


rakita:
Bad way

PR #3626, opened 2026-04-28
refactor(interpreter): unify call handlers

DaniPopes:
This unifies the interpreter CALL, CALLCODE, DELEGATECALL, and STATICCALL handlers into a single const-generic handler keyed by opcode kind, matching the shared handling used by revmc. It also moves the CALL kind and CREATE2 boolean const parameters to the front of their generic parameter lists for clearer instruction table instantiation.


rakita (approved):
This is nice

rakita (approved):

PR #3627, opened 2026-04-28
chore: update alloy-eip7928 to newer version

pepyakin:
The main effect of that is moving from `type BalIndex = u64` to a newtype `BlockAccessIndex`.


mattsse (approved):

pepyakin:
Tried rerunning CodSpeed but it still red. I don't really expect any effect on perf by doing this though.

PR #3634, opened 2026-04-29
refactor(eip8037): defer state-gas to frame return via NewStateTracker

rakita:
## Summary

Switches EIP-8037 state-gas from immediate charging to a deferred model. State-creating opcodes bump a simple per-frame counter on the `Interpreter` (`new_state`); the totals are reconciled with the gas tracker at frame return inside `EthFrame::process_next_action`.

* **on ok**: `state_gas_refunded` added to the reservoir first, then `state_gas` charged (may OOG → result flipped to `OutOfGas` before journal commit)
* **on revert/halt**: counters dropped — state work didn't happen, no reservoir change

## Notes

- `NewStateTracker` (`crates/context/interface/src/cfg/new_state_tracker.rs`) is intentionally minimal: two `u64` counters and a small set of `add_*` / `remove_*` methods. Uses **hardcoded** Glamsterdam EIP-8037 values (`bytes_per_unit × CPSB_GLAMSTERDAM`) for now — does not read `GasParams` at runtime.
- Migrates the immediate-charge call sites (SSTORE in `host.rs`, SELFDESTRUCT in `host.rs`, CREATE/CREATE2 in `contract.rs`, CALL-with-value in `call_helpers.rs`) to bump `interpreter.new_state.add_*()`. Removes the `state_gas!` macro.
- `return_create` no longer commits/reverts the journal; it validates code, charges regular code-deposit + hash gas, bumps `add_code_deposit_bytes`, and lets `process_next_action` apply state-gas and finalize (`set_code` on success).
- `handle_reservoir_remaining_gas` is simplified — the deferred model removes the spill-on-revert distortion, so the old `logical_refund / baseline / excess` math is no longer needed.
- Parent's `return_result` Create-branch: on child failure → `interpreter.new_state.remove_create_account()` (replaces `refill_reservoir(create_state_gas)`).

## Test plan

- [x] `cargo build --workspace` clean
- [ ] 23 EIP-8037 tests in `revm-ee-tests` currently fail because the test suite overrides `sstore_set_state_gas` etc. via `GasParams` (e.g. `200_000`) but the new `NewStateTracker` ignores those — the numbers it produces are EIP-8037 Glamsterdam defaults (e.g. SSTORE = `32 × 1174 = 37_568`). Need to either update test expectations, switch `NewStateTracker` to read `GasParams`/`cpsb`, or both.
- [ ] Spot-check tracing/inspector output (state_gas no longer goes negative since charges are deferred)

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -2,8 +2,10 @@
 
 pub mod gas;
 pub mod gas_params;
+pub mod new_state_tracker;
 
 pub use gas_params::{GasId, GasParams};
+pub use new_state_tracker::NewStateTracker;
 
 use auto_impl::auto_impl;
 use core::{fmt::Debug, hash::Hash};
--- a/crates/context/interface/src/cfg/new_state_tracker.rs
+++ b/crates/context/interface/src/cfg/new_state_tracker.rs
@@ -1,0 +1,106 @@
+//! Per-frame state-gas accumulator (EIP-8037).
+//!
+//! State-creating opcodes (SSTORE 0→x, CREATE/CREATE2, CALL with value to an
+//! empty account, SELFDESTRUCT, code deposit) bump the counters here instead
+//! of charging the reservoir directly. At frame return the totals are
+//! reconciled with the gas tracker:
+//!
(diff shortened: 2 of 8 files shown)


rakita:
Obsolete by bal-devnet-6 changes

PR #3635, opened 2026-04-29
fix: code change log should record code and code hash

cuiweixie:


rakita:
Previous code will always be KECCAK_EMPTY, there is no override bytecode path in EVM

cuiweixie:
in eip-7702, there a account A delegated to B, code(a) is prefix+ B address, if redelegate A to account C, I think It will change from prefix+B address to prefix+C address.

cuiweixie:
refs to go-ethereum: https://github.com/ethereum/go-ethereum/blob/b2aa6987ded983b98a56ad14aff0708e9b003567/core/state/journal.go#L556-L558 
```
codeChange struct {
account  common.Address
prevCode []byte
}

func (ch codeChange) revert(s *StateDB) {
	s.getStateObject(ch.account).setCode(crypto.Keccak256Hash(ch.prevCode), ch.prevCode)
}
```

cuiweixie:
@rakita what do you think?

cuiweixie:
here is a fully reproduce unit test:
```
//! Regression test for bluealloy/revm#3635.
//!
//! `JournalEntry::CodeChange` revert must restore the previous code and
//! `code_hash`, not unconditionally clear them. The original buggy revert path
//! set the account's code to `None` and `code_hash` to `KECCAK_EMPTY`, which
//! corrupts an account that already carried code before `set_code` was called —
//! the realistic case being an EIP-7702 EOA that is re-delegated.
//!
//! This test drives the bug from outside: it submits a real EIP-7702 tx via the
//! public `transact_one` API. The tx re-delegates an already-delegated authority,
//! and the tx then fails during execution (a database error when loading the new
//! delegate's code). The handler's error path calls `JournalTr::discard_tx`,
//! which is where the buggy revert wipes out the authority's prior delegation.

use alloy_signer::{Either, SignerSync};
use alloy_signer_local::PrivateKeySigner;
use bytecode::Bytecode;
use context::{Context, TxEnv};
use context_interface::{journaled_state::JournalTr, transaction::Authorization, ContextTr};
use database::{CacheDB, EmptyDB};
use database_interface::{DBErrorMarker, Database};
use primitives::{
    address, hardfork::SpecId, Address, StorageKey, StorageValue, TxKind, B256, KECCAK_EMPTY, U256,
};
use revm_handler::{EvmTr, ExecuteEvm, MainBuilder, MainContext};
use state::AccountInfo;

/// DB wrapper that returns an error when `basic(fail_on)` is called.
/// Used to force the handler's error path (and thus `discard_tx`) to run after
/// `pre_execution` has already applied the EIP-7702 authorization.
#[derive(Debug)]
struct FailingLoadDb {
    inner: CacheDB<EmptyDB>,
    fail_on: Address,
}

#[derive(Debug, PartialEq, Eq)]
struct LoadError(Address);

impl core::fmt::Display for LoadError {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        write!(f, "forced db load failure on {}", self.0)
    }
}

impl core::error::Error for LoadError {}

impl DBErrorMarker for LoadError {}

impl Database for FailingLoadDb {
    type Error = LoadError;

    fn basic(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
        if address == self.fail_on {
            return Err(LoadError(address));
        }
        Ok(self.inner.basic(address).unwrap())
    }

    fn code_by_hash(&mut self, code_hash: B256) -> Result<Bytecode, Self::Error> {
        Ok(self.inner.code_by_hash(code_hash).unwrap())
    }

    fn storage(
        &mut self,
        address: Address,
        index: StorageKey,
    ) -> Result<StorageValue, Self::Error> {
        Ok(self.inner.storage(address, index).unwrap())
    }

    fn block_hash(&mut self, number: u64) -> Result<B256, Self::Error> {
        Ok(self.inner.block_hash(number).unwrap())
    }
}

#[test]
fn redelegating_an_already_delegated_eoa_then_failing_restores_prior_delegation() {
    let signer = PrivateKeySigner::random();
    let authority = signer.address();

    let prior_delegate: Address = address!("00000000000000000000000000000000000000aa");
    let new_delegate: Address = address!("00000000000000000000000000000000000000bb");
    let caller: Address = address!("1000000000000000000000000000000000000001");

    // Pre-existing state: `authority` is an EIP-7702 EOA already delegated to
    // `prior_delegate` from a previous (committed) transaction, so its nonce
    // has been bumped to 1.
    let prior_code = Bytecode::new_eip7702(prior_delegate);
    let prior_code_hash = prior_code.hash_slow();

    inner.insert_account_info(
        authority,
        AccountInfo {
            balance: U256::ZERO,
            nonce: 1,
            code: Some(prior_code.clone()),
            code_hash: prior_code_hash,
            ..Default::default()
        },
    );
    inner.insert_account_info(
        caller,
        AccountInfo {
            balance: U256::from(1_000_000_000_000_000_000u128),
            nonce: 0,
            ..Default::default()
        },
    );

    // The DB will error when execution tries to load `new_delegate`'s code,
    // which the handler does after pre_execution to follow the 7702 delegation.
    let db = FailingLoadDb {
        inner,
        fail_on: new_delegate,
    };

    // External tx: an EIP-7702 authorization that re-delegates `authority` to
    // `new_delegate` (chain_id 0 matches any chain).
    let auth = Authorization {
        chain_id: U256::ZERO,
        nonce: 1,
        address: new_delegate,
    };
    let signature = signer.sign_hash_sync(&auth.signature_hash()).unwrap();
    let signed_auth = auth.into_signed(signature);

    let ctx = Context::mainnet()
        .modify_cfg_chained(|cfg| cfg.set_spec_and_mainnet_gas_params(SpecId::PRAGUE))
        .with_db(db);
    let mut evm = ctx.build_mainnet();

    let tx = TxEnv::builder()
        .gas_limit(100_000)
        .authorization_list(vec![Either::Left(signed_auth)])
        .caller(caller)
        .kind(TxKind::Call(authority))
        .build()
        .unwrap();

    // Submit the tx via the public API. pre_execution applies the 7702 auth
    // (changing authority's code from delegation→Y to delegation→Z), then the
    // top-level call's bytecode load fails on Z, propagating the error. The
    // handler's `catch_error` invokes `discard_tx` to unwind the journal.
    let err = evm
        .transact_one(tx)
        .expect_err("tx must fail due to forced DB error");
    let msg = format!("{err:?}");
    assert!(
        msg.contains("LoadError"),
        "expected forced DB failure, got: {msg}",
    );

    // After the error path's discard_tx, authority's *prior* delegation to Y
    // must be restored. The buggy revert path instead leaves the code cleared
    // (code_hash = KECCAK_EMPTY, code = None).
    let ctx = evm.ctx();
    let acc = ctx
        .journal()
        .evm_state()
        .get(&authority)
        .expect("authority still tracked in state")
        .clone();

    assert_ne!(
        acc.info.code_hash, KECCAK_EMPTY,
        "regression bug: discard_tx cleared code_hash to KECCAK_EMPTY instead of restoring the prior delegation",
    );
    assert_eq!(
        acc.info.code_hash, prior_code_hash,
        "discard_tx must restore code_hash to the prior 7702 delegation hash",
    );
    assert_eq!(
        acc.info.code,
        Some(prior_code),
        "discard_tx must restore the prior 7702 delegation bytecode",
    );
}

```

PR #3637, opened 2026-04-30
fix(inspector): use build_result_gas in inspect_run_system_call

cuiweixie:
The inspect path constructed `ResultGas` inline from `gas.total_gas_spent()` (`limit - remaining`), while the non-inspect path used `build_result_gas`, which also subtracts the reservoir. When a system call left `reservoir > 0` (e.g. a child frame consuming state gas then reverting), the two paths reported different `total_gas_spent`, making observability depend on whether an inspector was attached.

Switch the inspect path to `build_result_gas` so both paths produce identical `ResultGas`. Add a regression test that triggers `reservoir > 0` via SSTORE + REVERT in a child under AMSTERDAM and asserts gas reporting matches between `system_call_one` and `inspect_one_system_call`.

### DIFF
--- a/crates/inspector/src/handler.rs
+++ b/crates/inspector/src/handler.rs
@@ -1,9 +1,8 @@
 use crate::{Inspector, InspectorEvmTr, JournalExt};
-use context::{
-    result::{ExecutionResult, ResultGas},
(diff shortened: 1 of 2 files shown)


rakita (approved):
good find

PR #3638, opened 2026-04-30
refactor: move global refund from call local to journal

### LINKED ISSUE #2145: Refactor: migrate from call local refund to global journal refund

alok-108:
Fixes #2145

Migrate gas refund tracking from call-local Gas struct to a global
refund field in Journal. This eliminates negative refunds leaking
from call frames and simplifies frame return values.

Key Changes:
- Integrated `refund` into `JournalInner` with full checkpointing support.
- Extended `JournalTr` and `Host` traits with unified refund accessors.
- Updated SSTORE, SELFDESTRUCT, precompiles, and the post-execution handler to use the global refund.
- Removed frame-local refund tracking from `GasTracker` and `Gas` structs.
- Verified with workspace-wide tests and updated EIP-3155/EIP-8037 logic.

### DIFF
--- a/crates/context/interface/src/cfg/gas.rs
+++ b/crates/context/interface/src/cfg/gas.rs
@@ -18,8 +18,6 @@ pub struct GasTracker {
     reservoir: u64,
     /// Total state gas spent so far.
     state_gas_spent: u64,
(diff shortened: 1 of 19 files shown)


rakita:
Closing this for now, it is low priority and probably will never be added

PR #3639, opened 2026-04-30
fix(precompile): halt with PrecompileOOG when precompile over-spends gas

cuiweixie:
When `record_regular_cost` returns false the over-spend was silently discarded, letting the call appear successful. Convert it into a PrecompileOOG halt instead, and add a regression test.


rakita:
Thanks

PR #3644, opened 2026-05-02
perf: opt all zeros check

cuiweixie:
Sanity benchmark (n=8):
- all_zero case: iter_all 232 ns → u64_safe 18.8 ns (12.4× faster)
- all_nonzero case: iter_all 5.4 ns → u64_safe 3.4 ns (1.6× faster)


DaniPopes:
iter().all() is already vectorized since the size of the array is known

PR #3654, opened 2026-05-05
test(ee-tests): switch to insta for snapshot assertions

rakita:
## Summary
- Replaces the hand-rolled `compare_or_save_testdata` / `TestdataConfig` helpers in `revm-ee-tests` with `insta::assert_json_snapshot!`. Snapshots now live alongside the tests under `crates/ee-tests/src/snapshots/` and can be reviewed/accepted with `cargo insta review`.
- Drops `serde_json`, `serde`, and `similar-asserts` dependencies from `revm-ee-tests`; adds `insta` (json feature) as a workspace dev-dep.
- Removes the old `tests/{testdata,revm_testdata,eip8037_testdata}/` JSON fixture trees in favor of the generated `*.snap` files.

## Test plan
- [x] `cargo test -p revm-ee-tests` → 48 passed
- [x] `cargo clippy -p revm-ee-tests --all-targets --all-features` clean
- [x] `cargo fmt --all --check` clean
- [x] Snapshots regenerated via `INSTA_UPDATE=always cargo test -p revm-ee-tests --lib` produced 40 `.snap` files matching the prior JSON fixtures

PR #3655, opened 2026-05-05
fix(ee-tests): sort snapshot map keys for hasher independence

rakita:
## Summary
Follow-up to #3654. Two CI jobs were broken on that PR's head and merged anyway:

- **`test stable/nightly/1.91 --all-features` (workspace)** — `test_selfdestruct_multi_tx` and `test_multi_tx_create` fail because `cargo nextest run --workspace --all-features` enables the `map-foldhash` feature on `primitives`, which changes `HashMap` iteration order and reorders keys in the serialized JSON state. The previous hand-rolled helper sidestepped this with `Value::sort_all_objects()` before comparison.
- **`docs`** — `cargo doc --workspace --all-features` errors with `unresolved link to insta` because `insta` is a dev-dep and the intra-doc link can't resolve.

This PR:
- Adds a `crate::assert_sorted_json_snapshot!` macro in `lib.rs` that wraps `insta::with_settings!({sort_maps => true}, ...)`, and routes all 40 call sites through it. Sorting maps gives stable JSON regardless of which hasher the workspace ends up with.
- Replaces the broken intra-doc link `` [`insta`] `` with plain `` `insta` crate ``.
- Regenerates the 6 snapshots whose key order was hasher-dependent.

## Test plan
- [x] `cargo nextest run --workspace --all-features` → 329 passed
- [x] `cargo test -p revm-ee-tests` → 48 passed (default features)
- [x] `cargo doc -p revm-ee-tests --all-features --no-deps --document-private-items` → no errors
- [x] `cargo fmt --all --check` and `cargo clippy -p revm-ee-tests --all-targets --all-features` → clean

### DIFF
--- a/crates/ee-tests/src/eip8037.rs
+++ b/crates/ee-tests/src/eip8037.rs
@@ -495,7 +495,7 @@ fn test_eip8037_sstore_new_slot() {
         result.gas().total_gas_spent() - baseline_result.gas().total_gas_spent(),
         STATE_GAS_SSTORE_SET
     );
-    insta::assert_json_snapshot!(&(baseline_result, result));
+    crate::assert_sorted_json_snapshot!(&(baseline_result, result));
 }
 
 /// 1.2 Two SSTOREs to same slot: only first charges state gas.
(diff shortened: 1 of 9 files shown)


rakita:
This fixes last merged commit

PR #3656, opened 2026-05-05
fix(inspector): notify on EIP-6780 no-op SELFDESTRUCT(self)

irnb:
Post-Cancun, when a contract is not created in the current transaction and
SELFDESTRUCT is called with `target == self`, the operation is a complete
state no-op: no balance transfer, no destruction. `JournalInner::selfdestruct`
correctly records nothing in this branch (see crates/context/src/journal/inner.rs,
trailing `else` arm).

However, the SELFDESTRUCT opcode still returns
`InstructionResult::SelfDestruct`, and `inspect_selfdestruct` only matched
`JournalEntry::AccountDestroyed | BalanceTransfer`. With neither variant
present, the inspector hook was silently skipped — leaving downstream tracers
(notably the Geth `callTracer` and Parity flat-trace builders in
`revm-inspectors`) to emit synthetic frames with `from = 0x000…0`,
omitted `to`, and omitted `value`. Consumers that strict-cast addresses
(e.g. Blockscout) would then reject these frames and crash their indexers.

Fix: pass the executing contract's address through to `inspect_selfdestruct`
and, when no journal entry matches, fire the hook with
`(contract, contract, 0)` — accurately reflecting that the opcode ran with
target == self and zero balance was transferred.

Adds a regression test `test_selfdestruct_self_target_eip6780_noop` that
exercises this path against a `BenchmarkDB`-pre-deployed contract calling
SELFDESTRUCT(self), and asserts the inspector receives one event with the
correct `(address, beneficiary, value)`

### DIFF
--- a/crates/inspector/src/handler.rs
+++ b/crates/inspector/src/handler.rs
@@ -6,11 +6,11 @@ use context::{
 use handler::{evm::FrameTr, EvmTr, FrameResult, Handler, ItemOrResult};
 use interpreter::{
     instructions::{GasTable, InstructionTable},
-    interpreter_types::{Jumps, LoopControl},
+    interpreter_types::{InputsTr, Jumps, LoopControl},
     FrameInput, Host, InitialAndFloorGas, InstructionResult, Interpreter, InterpreterAction,
     InterpreterTypes,
 };
-use primitives::hints_util::cold_path;
+use primitives::{hints_util::cold_path, Address, U256};
 use state::bytecode::opcode;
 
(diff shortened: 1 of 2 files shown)


rakita:
If there is a need to always do a hook on selfdestruct opcode than this should be done in Inspector `fn step`.

It looks correct to trigger Inspector::selfdestruct only if it happened.

rakita:
cc @mattsse should we use opcode for selfdestruct on opcode traces?

Hopefully, we can have a push in Amsterdam to remove selfdestruct.

PR #3657, opened 2026-05-05
fix: gracefully handle commits with non-cached accounts

klkvr:
Changes `State::commit` logic to use `original_info` to populate the cache instead of panicking


rakita:
Looks correct. We can potentially optimise it to fully remove self map insertion and update (as it will not be needed for BAL) but this is optimization for later

PR #3664, opened 2026-05-07
perf(interpreter): cache init_code_hash in CreateInputs for inspector reuse

cotter-circle:
## Summary

Adds a cached `init_code_hash()` method to `CreateInputs` and wires it into both `created_address()` and `make_create_frame`, so the keccak256 of init code is computed at most once.

## Why

Inspectors that call `CreateInputs::created_address()` (e.g. to check the target before frame init) trigger `keccak256(init_code)` via `create2_from_code`. When `make_create_frame` runs afterwards, it
computed the same hash independently since it bypassed the `OnceCell` cache. By routing both paths through a shared `init_code_hash` cache, inspectors that pre-derive the address no longer cause a
redundant hash over the full init code.

No change in behavior for standard revm without custom inspectors — this is purely a caching optimization for inspector use cases.

### DIFF
--- a/crates/handler/src/frame.rs
+++ b/crates/handler/src/frame.rs
@@ -15,14 +15,14 @@ use interpreter::{
     interpreter::{EthInterpreter, ExtBytecode},
     interpreter_action::FrameInit,
     interpreter_types::ReturnData,
-    CallInput, CallInputs, CallOutcome, CallValue, CreateInputs, CreateOutcome, CreateScheme,
-    FrameInput, Gas, InputsImpl, InstructionResult, Interpreter, InterpreterAction,
-    InterpreterResult, InterpreterTypes, SharedMemory,
+    CallInput, CallInputs, CallOutcome, CallValue, CreateInputs, CreateOutcome, FrameInput, Gas,
+    InputsImpl, InstructionResult, Interpreter, InterpreterAction, InterpreterResult,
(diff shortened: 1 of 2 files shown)


rakita:
The headline and description are really misleading. It would be a ddos bug if we do Keccak twice.

This is about Inspectors and caching init_code keccak.

Can you change the name/description and make it clear that this is Inspector perf related?

rakita (approved):
Pending PR name/description change

crates/handler/src/frame.rs
```diff
@@ -297,16 +297,11 @@ impl EthFrame<EthInterpreter> {
...
-            CreateScheme::Create2 { salt } => {
-                let init_code_hash = *init_code_hash.insert(keccak256(inputs.init_code()));
-                inputs.caller().create2(salt.to_be_bytes(), init_code_hash)
-            }
-            CreateScheme::Custom { address } => address,
-        };
+        // Create address — uses OnceCell cache so that if an inspector already called
+        // `created_address`, the expensive keccak256 is not recomputed.
+        let created_address = inputs.created_address(old_nonce);
+        let init_code_hash = matches!(inputs.scheme(), CreateScheme::Create2 { .. })
```
rakita, line 303: This is cleaner!

cotter-circle:
@rakita Apologies for the confusion. I updated title and description.

PR #3671, opened 2026-05-11
fix(database): AlloyDB returns None for non-existent accounts (#2454)

### LINKED ISSUE #2454: [Bug] AlloyDB should not always return a Some(AccountInfo)

satyakwok:
Closes #2454.

## The bug

`AlloyDB::basic_async_ref` always returns `Some(AccountInfo)`, even for accounts that don't exist on-chain. That breaks revm's EIP-161 state-clear check at `crates/state/src/lib.rs:43+`, which relies on the `Option<AccountInfo>` to distinguish existent-empty from non-existent.

Reporter [@nora-coder-dot](https://github.com/nora-coder-dot) hit this replaying tx `0x1ae5f96a...` (calls precompile `0x4`) via the `examples/block_traces` flow.

## Why the previous code couldn't fix it inline

Three parallel RPCs: `eth_getTransactionCount` + `eth_getBalance` + `eth_getCode`. All three return zeros for both **existent-but-empty** and **non-existent** accounts post Spurious Dragon. No signal disambiguates the two cases.

## The fix

Use `eth_getProof` instead. It returns `code_hash` as its own field — the only RPC signal that distinguishes existence vs emptiness:

| Account state | `code_hash` |
|---|---|
| non-existent | `B256::ZERO` |
| existent, empty | `KECCAK_EMPTY` |
| existent, non-empty | `keccak(code)` |

```rust
let proof = self.provider.get_proof(address, Vec::new()).block_id(self.block_number).await?;

if proof.code_hash == B256::ZERO {
    return Ok(None);  // EIP-161: account doesn't exist
}

let code = if proof.code_hash == KECCAK_EMPTY {
    Bytecode::new()  // skip the extra eth_getCode round-trip
} else {
    let code_bytes = self.provider.get_code_at(address).block_id(self.block_number).await?;
    Bytecode::new_raw(code_bytes.0.into())
};

Ok(Some(AccountInfo::new(proof.balance, proof.nonce, proof.code_hash, code)))
```

## References

[@rakita](https://github.com/rakita) flagged this approach in #2454 (Oct 2025) citing the discussion at ethereum/go-ethereum#28441 and the reference fix at a16z/helios#714 — both establish the `code_hash == ZERO` convention as the canonical non-existence signal.

## Differences from the previous attempt (#2468, closed)

| | #2468 | This PR |
|---|---|---|
| Public API | adds `spec_id: SpecId` to `AlloyDB::new` (breaking) | unchanged |
| Decision logic | check `nonce==0 && balance==0 && code.is_empty()` post Spurious Dragon | direct `code_hash` field from `eth_getProof` |
| Provider requirement | none new | `eth_getProof` (geth/erigon/reth/besu/nethermind all support it) |
| RPC count for empty / non-existent | still 3 in parallel | 1 |
| RPC count for accounts with code | 3 in parallel | 1 + 1 sequential |

## Trade-offs

- **No public API break.** `AlloyDB::new` signature unchanged.
- **RPC efficiency:** 1 call instead of 3 for empty / non-existent accounts (the common case in trace replays). 2 sequential calls when there's code to fetch — slightly worse than 3-in-parallel under high latency, but typical mixed workloads come out neutral to better.
- **Requires `eth_getProof`.** Every major Ethereum execution client supports it (it's been a standard JSON-RPC method since EIP-1186 / 2018).
- **Behaviour change for downstream callers:** anyone relying on always-Some now sees None for non-existent accounts. This matches the Ethereum spec, so any prior reliance was already a latent bug — but worth calling out in the changelog.

## Testing

The bug only manifests against a live RPC node, and the existing `can_get_basic` integration test is marked `#[ignore = "flaky RPC"]`. I'd be happy to add unit tests against a mocked `Provider` if you have a preferred mock pattern (didn't want to invent one). Manually reproducible by:

1. Pick any address that didn't exist at the test block (e.g. a freshly-generated EOA).
2. Call `basic_async_ref` via `WrapDatabaseAsync`.
3. On `main` → `Some(AccountInfo { ... all zeros ... })`. With this PR → `None`.

### DIFF
--- a/crates/database/src/alloydb.rs
+++ b/crates/database/src/alloydb.rs
@@ -8,7 +8,7 @@ use alloy_provider::{
 use alloy_transport::TransportError;
 use core::error::Error;
 use database_interface::{async_db::DatabaseAsyncRef, DBErrorMarker};
-use primitives::{Address, StorageKey, StorageValue, B256};
+use primitives::{Address, StorageKey, StorageValue, B256, KECCAK_EMPTY};
 use state::{AccountInfo, Bytecode};
 use std::fmt::Display;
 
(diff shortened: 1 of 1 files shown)


rakita:
This is clever! But `get_proof` is a lot more expensive than ordinary nonce/balance/bytecode fetch, so it is better not to use it.

I dont think there is solution for this one, will close it

satyakwok:
Thanks for the review.

That makes sense. I agree that using "eth_getProof" as the default path would be too expensive for AlloyDB, especially for tracing or replay workloads with many account lookups.

My main goal was to clarify the account-existence semantic issue: with only nonce, balance, and code, the RPC-backed path cannot reliably distinguish a non-existent account from an existing empty account.

Given the performance trade-off, treating this as a known limitation of RPC-backed state access sounds reasonable.

Appreciate the clarification.

PR #3672, opened 2026-05-12
fix(eip8037): align bal-devnet-7 system call gas limit

mark0-cn:
Update system calls to use the bal-devnet-7 gas limit formula instead of the legacy fixed 30M cap.

Changes:

- add SYSTEM_MAX_SSTORES_PER_CALL = 16

- derive SYSTEM_CALL_GAS_LIMIT as

  30_000_000 + SSTORE_SET_BYTES * CPSB_GLAMSTERDAM * SYSTEM_MAX_SSTORES_PER_CALL

- update the system_call test to assert the new limit

- sync stale EIP-8037 docs/comments with the current bal-devnet-7 behavior

This matches tests-bal@v7.0.0, where system calls receive a dedicated state-gas reservoir on top of the 30M regular-gas budget.

### DIFF
--- a/crates/ee-tests/src/eip8037.rs
+++ b/crates/ee-tests/src/eip8037.rs
@@ -1525,10 +1525,12 @@ fn test_eip8037_precompile_no_state_gas() {
 
 // ---- Category 7: Reservoir Refill ----
 //
-// The reservoir refill logic (handler_reservoir_refill) is invoked on HALT or REVERT:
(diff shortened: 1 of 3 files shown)


rakita (approved):
Mentioned [here](https://eips.ethereum.org/EIPS/eip-8037#system-contracts-and-system-transactions:~:text=The%20additional%20STATE_BYTES_PER_STORAGE_SET,or%20block_state_gas_used.)

PR #3678, opened 2026-05-13
fix(context): keep AccountLoad.is_empty on target account

gfischer7:
This change keeps `AccountLoad.is_empty` bound to the originally requested account in delegated loads, and aligns Host and Journal delegated-loading semantics to prevent inconsistent emptiness decisions.


rakita:
That funtion is not used anymore, so deprecating it makes sense. Would not merge this PR rn, but it does make sense that is_empty represents first account.

PR #3681, opened 2026-05-15
perf(database): preserve commit_iter semantics

NithinPJ998:
# Summary

This PR forwards `DatabaseCommit::commit_iter` through database wrappers instead of falling back to the default collect-then-commit path.

# Problem

`commit_iter` exists for callers that already have an iterator of state changes, but `CacheDB`, `BalDatabase`, `Either`, and `WrapDatabaseRef` currently lose that path. `CacheDB` also collapses repeated addresses when the default implementation collects into an `AddressMap`.

# Solution

- add a streaming `CacheDB::commit_iter` implementation shared with `commit`
- forward `commit_iter` through database wrappers
- add regression coverage for wrapper forwarding and repeated `CacheDB` account updates

### DIFF
--- a/crates/database/interface/src/bal.rs
+++ b/crates/database/interface/src/bal.rs
@@ -388,4 +388,13 @@ impl<DB: DatabaseCommit> DatabaseCommit for BalDatabase<DB> {
         self.bal_state.commit(&changes);
         self.db.commit(changes);
     }
+
+    fn commit_iter(&mut self, changes: &mut dyn Iterator<Item = (Address, Account)>) {
+        let bal_state = &mut self.bal_state;
+        let mut changes = changes.map(|(address, account)| {
+            bal_state.commit_one(address, &account);
+            (address, account)
+        });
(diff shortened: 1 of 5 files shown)


crates/database/interface/src/bal.rs
```diff
@@ -388,4 +388,13 @@ impl<DB: DatabaseCommit> DatabaseCommit for BalDatabase<DB> {
...
         self.db.commit(changes);
     }
+
+    fn commit_iter(&mut self, changes: &mut dyn Iterator<Item = (Address, Account)>) {
+        let bal_state = &mut self.bal_state;
+        let mut changes = changes.map(|(address, account)| {
+            bal_state.commit_one(address, &account);
+            (address, account)
+        });
+        self.db.commit_iter(&mut changes);
```
rakita, line 398: this could be a footgun if .db doesn't consume the iterator, but i guess this would be a bug in inner db impl

rakita (approved):
lgtm

PR #3682, opened 2026-05-15
feat(eip2780): reduce intrinsic transaction gas

rakita:
## Summary
- Implements [EIP-2780](https://github.com/misilva73/EIPs/blob/0a37332/EIPS/eip-2780.md) (Reduce intrinsic transaction gas) on top of EIP-8037 / EIP-8038.
- Replaces the legacy 21,000 stipend with a decomposed `TX_BASE_COST + to-based + value-based` model, plumbed through `GasParams::initial_tx_gas` via a new `Eip2780TxInfo`.
- Adds top-level execution charges in `EthFrame::make_call_frame` (depth == 0): extra `COLD_ACCOUNT_ACCESS` for EIP-7702-delegated recipients; `new_account_state_gas(cpsb)` for empty recipients with `tx.value > 0` (refunded on revert via the existing reservoir reconciliation in `last_frame_result`).
- Adds `enable_amsterdam_eip2780` config flag (auto-enabled at AMSTERDAM) and `Cfg::is_amsterdam_eip2780_enabled`. `TX_BASE_COST` and `TRANSFER_LOG_COST` are EIP-2780-specific constants in `primitives::eip2780`; `ACCOUNT_WRITE` / `CREATE_ACCESS` / `COLD_ACCOUNT_ACCESS` source from `primitives::eip8038` so a single change to the placeholder TBD values propagates everywhere.

## Notes
- Existing EIP-8037 ee-tests explicitly disable EIP-2780 so the state-gas-reservoir tests stay focused on that EIP. One EIP-7708 test's gas limit was bumped to fit the new ~207k intrinsic for value transfer to a new account.
- New `eip2780.rs` ee-tests cover self-transfer, EOA no-value, empty-with-value (state gas), create with/without value, precompile delta, and the legacy-fallback path.

## Test plan
- [x] `cargo nextest run --workspace` (347 passed, 1 skipped)
- [x] `cargo nextest run --workspace --no-default-features` (347 passed)
- [x] `cargo clippy --workspace --all-targets --all-features` clean
- [x] `cargo fmt --all --check` clean
- [x] `cargo doc --workspace --all-features --no-deps --document-private-items` clean
- [x] `typos` clean

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -95,6 +95,15 @@ pub trait Cfg {
     /// `cost_per_state_byte` and adds a hash cost for deployed bytecode.
     fn is_amsterdam_eip8037_enabled(&self) -> bool;
 
+    /// Returns whether EIP-2780 (Amsterdam) reduced intrinsic transaction gas
+    /// is enabled.
+    ///
+    /// When enabled, the legacy `21,000` intrinsic base is replaced with the
+    /// decomposed `TX_BASE_COST + to-based + value-based` model and top-level
+    /// execution charges are applied for empty recipients with value and
+    /// EIP-7702-delegated recipients.
+    fn is_amsterdam_eip2780_enabled(&self) -> bool;
+
     /// Returns the EIP-8037 `cost_per_state_byte` (CPSB).
     ///
     /// When [`Cfg::is_amsterdam_eip8037_enabled`] is `false` this returns `0`.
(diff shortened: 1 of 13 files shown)


mark0-cn:
Nice work on EIP-2780! I noticed two cases where the access-list warmth override specified in the EIP wasn't being      
  applied:                                                                                                                
                                                                                                                          
  1. Recipient warm override — eip2780_base_to_value_gas always charged COLD_ACCOUNT_ACCESS for tx.to even when it appeared in the transaction access list, violating the spec's rule: "Warmth discovered through an access list still   
 applies and overrides the cold cost."                                                                                   
  3. EIP-7702 delegation target — The depth-0 extra charge always used COLD_ACCOUNT_ACCESS regardless of whether the delegation target was pre-warmed. The underlying issue is that create_init_frame loads the delegation target before make_call_frame runs, so checking journal warmth at that point always returns warm — the access list must be checked directly instead.                                                                                                       
                                                            
  I've implemented both fixes with tests in #3683.

PR #3683, opened 2026-05-16
fix(eip2780): charge warm cost when tx.to or delegation target is in …

mark0-cn:
Per EIP-2780 [1]:

  "Warmth discovered through an access list still applies and overrides
   the cold cost."

  "The delegation target incurs an additional COLD_ACCOUNT_COST_CODE
   if cold, or WARM_STATE_READ if already warm (e.g., if it appeared
   in the access list)."

Two bugs fixed that violated the above rules:

**Bug 1 — Recipient warm override**
`eip2780_base_to_value_gas` always charged `COLD_ACCOUNT_ACCESS` for `tx.to` even when `tx.to` appeared in the transaction access list.

Fix: add `to_is_warm: bool` to `Eip2780TxInfo` and populate it by scanning `tx.access_list()` in `validate_initial_tx_gas`. When set, `eip2780_base_to_value_gas` charges `warm_storage_read_cost` instead of `COLD_ACCOUNT_ACCESS` for the recipient.

**Bug 2 — EIP-7702 delegation target warm/cold**
The depth-0 extra access charge for EIP-7702-delegated recipients always used `COLD_ACCOUNT_ACCESS` regardless of whether the delegation target had been pre-warmed by the access list.

The original implementation tried to check journal warmth at `make_call_frame` time, but `create_init_frame` (execution.rs) loads the delegation target unconditionally before `make_call_frame` runs, making the journal always report it as warm.

Fix: check the transaction access list directly (same approach as `to_is_warm`) instead of querying journal state. This correctly distinguishes addresses pre-warmed by the access list from those warmed incidentally during frame setup.

Tests added to `crates/ee-tests/src/eip2780.rs` covering both fixes.

[1] https://eip.tools/eip/2780

### DIFF
--- a/crates/context/interface/src/cfg/gas_params.rs
+++ b/crates/context/interface/src/cfg/gas_params.rs
@@ -1138,6 +1138,9 @@ impl GasParams {
             self.tx_create_access_cost()
         } else if to_is_self || info.to_is_precompile {
             0
+        } else if info.to_is_warm {
+            // EIP-2780: warmth from the access list overrides the cold cost.
+            self.warm_storage_read_cost()
         } else {
             eip8038::COLD_ACCOUNT_ACCESS
         };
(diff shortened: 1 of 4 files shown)


rakita:
Will not merge this as there are some inconsistencies and imo EIP should change

PR #3688, opened 2026-05-19
fix(database): return error for AlloyDB code-by-hash lookup

gfischer7:
AlloyDB now returns a structured error instead of panicking for code-by-hash lookups, preserving process stability when generic database callers hit this path.


rakita:
Panic can't be triggered

PR #3690, opened 2026-05-20
perf(handler): skip precompile bytecode loads

Dannyyy93:
## Summary

Avoid loading target bytecode when the initial transaction call is known to target a precompile. The frame input now only loads the account metadata needed for balance/value handling on that path, and leaves bytecode loading to the non-precompile path that actually executes account code.

## Why

Precompile execution does not consume the target account bytecode. This keeps databases from being asked for code by hash when the call will be handled by `PrecompileProvider::run`, matching the existing deferred-load behavior for call frame setup.

### DIFF
--- a/crates/handler/src/execution.rs
+++ b/crates/handler/src/execution.rs
@@ -19,21 +19,29 @@ pub fn create_init_frame<CTX: ContextTr>(
 
     match tx.kind() {
         TxKind::Call(target_address) => {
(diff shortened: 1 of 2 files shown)


rakita:
Precompiles should have KECCAK_EMPTY as code hash, in this case we shouldn't touch Database on loading. So we should be good

PR #3703, opened 2026-05-22
chore(bytecode): mark RETURN and REVERT as memory-modifying

NithinPJ998:
## Summary

- Mark RETURN and REVERT as memory-modifying opcodes because their shared return path expands memory for non-zero output ranges.
- Add coverage to `test_modifies_memory` for both terminating opcodes.


NithinPJ998:
Validation passes locally. The remaining CodSpeed failure appears unrelated to this bytecode-only diff and notes different runtime environments;

PR #3708, opened 2026-05-22
feat(eip7702): defer delegate code load via BytecodeLoad enum

rakita:
## Summary

- For an EIP-7702 delegated call target, the delegate account is no longer eagerly cold-loaded in the CALL helper. The warm/cold decision now drives gas charging only, and the actual code fetch is deferred to frame creation.
- Introduces a new `BytecodeLoad { LoadFrom(Address), Bytecode(Bytecode) }` enum (`state::BytecodeLoad`, re-exported from `bytecode`) that lets `load_account_delegated` return either an already-loaded bytecode or a pointer to "load from this address later".
- Adds `Host::is_account_warm(&self, address) -> bool` — a pure warm-set lookup that never reads from the DB and never promotes a cold account to warm. Mirrored on `JournalTr`, implemented in `JournalInner` (precompile/coinbase/access-list set + in-memory account's `is_cold_transaction_id`), and wired through `Journal`, `Context`, and the cheatcode_inspector example.
- `CallInputs.known_bytecode` becomes `(B256, BytecodeLoad)`. `make_call_frame` resolves a `LoadFrom` by calling `journal.load_account_with_code(addr)` **before** the child checkpoint is created, so the delegate's warmth survives if the child frame reverts (this caught a state-root regression on `test_pointer_reverts` during development).

## Test plan

- [x] `cargo check --workspace` (default features)
- [x] `cargo check --workspace --all-features`
- [x] `cargo check --workspace --no-default-features`
- [x] `cargo clippy --workspace --all-targets --all-features`
- [x] `cargo fmt --all --check`
- [x] `cargo nextest run --workspace` — 327 passed
- [x] `cargo run --release -p revme -- statetest test-fixtures/main/develop/state_tests` — all 2710 tests pass, including the EIP-7702 `test_pointer_reverts` cases that exercise the warm-on-revert path

### DIFF
--- a/crates/bytecode/src/bytecode/mod.rs
+++ b/crates/bytecode/src/bytecode/mod.rs
@@ -57,6 +57,48 @@ pub enum BytecodeKind {
     Eip7702,
 }
 
+/// Deferred bytecode reference produced when account loading is split from
+/// bytecode loading.
+///
+/// `Bytecode` carries an already-loaded [`Bytecode`]; `LoadFrom` carries an
+/// account [`Address`] whose code (and code hash) still needs to be fetched
+/// when the bytecode is actually required (e.g. on frame creation).
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum BytecodeLoad {
+    /// Code has not been loaded yet; load it from this address when needed.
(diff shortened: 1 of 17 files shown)


rakita:
Superseded by #3720, which sources the same branch (`eip7702-defer-delegate-load`) directly from the `bluealloy/revm` repo instead of a fork. Same changes and description.

PR #3710, opened 2026-05-22
feat: add `OnStateHook` for `State<DB>`

decofe:
## Summary
- add StateChangeSource and OnStateHook to revm database interface
- store an optional hook and current source on State
- invoke the hook from State::commit and State::commit_iter before applying changes

## Tests
- cargo check -p revm -p revm-database-interface -p revm-database

### DIFF
--- a/crates/database/src/states/state.rs
+++ b/crates/database/src/states/state.rs
@@ -7,7 +7,7 @@ use super::{
 use bytecode::Bytecode;
 use database_interface::{
     bal::{BalState, EvmDatabaseError},
-    Database, DatabaseCommit, DatabaseRef, EmptyDB,
+    Database, DatabaseCommit, DatabaseRef, EmptyDB, OnStateHook,
 };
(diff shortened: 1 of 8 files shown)


crates/database/src/states/state.rs
```diff
@@ -385,6 +405,23 @@ impl<DB: Database> DatabaseCommit for State<DB> {
...
+                .apply_evm_state_iter(changes, |address, account| {
+                    self.bal_state.commit_one(*address, account);
+                });
+            if let Some(s) = self.transition_state.as_mut() {
+                s.add_transitions(transitions)
+            } else {
+                // Advance the iter to apply all state updates.
+                transitions.for_each(|_| {});
+            }
+            return;
```
rakita: If we need to collect, let's just call `self.commit(changes)` to not duplicate code.

rakita (approved):
nit pending

mattsse (approved):

klkvr (approved):

PR #3714, opened 2026-05-24
feat: Bake EIP-8037 CPSB into gas params

decofe:
## Summary
- bake Glamsterdam CPSB into Amsterdam GasParams entries in new_spec_inner
- remove cpsb_override/cached local cpsb/Host::cpsb plumbing
- simplify state gas helper APIs so callers use pre-scaled gas values

## Tests
- cargo check -p revm-context-interface -p revm-context -p revm-interpreter -p revm-handler -p revm-inspector -p revm-ee-tests


rakita:
Have manually run fmt/clippy and state and block tests.

PR #3717, opened 2026-05-25
feat(context): implement EIP-8246 (remove SELFDESTRUCT burn)

rakita:
## Summary

Implements [EIP-8246: Remove SELFDESTRUCT Burn](https://eips.ethereum.org/EIPS/eip-8246), replacing the EIP-7708 burn-log mechanism.

Under EIP-8246, at transaction finalization a self-destructed account is no longer deleted/burned. Instead its **nonce is reset to 0, code and storage are cleared, and the balance is preserved**. If the resulting balance is `0` the account is empty and is removed by EIP-161 as before.

## Changes

- **Renamed `eip7708_emit_burn_remaining_balance_logs` → `eip8246_clear_selfdestructed_accounts`** (`JournalInner`). Instead of emitting `Burn` logs for self-destructed accounts that still hold a balance, it now clears their code/storage/nonce, keeps the balance, and removes the self-destruct flag so the account is preserved as a balance-only account. Zero-balance accounts keep the flag and are removed (EIP-161).
- **Self-destruct to self** (spec point 1): under Amsterdam the balance is no longer zeroed and no burn log is emitted; the account is cleared at finalization. Pre-Amsterdam behavior is unchanged.
- **Removed the now-dead `eip7708_burn_log` helper** and its `BURN_LOG_TOPIC` use.
- **Renamed the config flag** `eip7708_delayed_burn_disabled` → `eip8246_delayed_clear_disabled` (`CfgEnv` field `amsterdam_eip8246_delayed_clear_disabled`, `Cfg::is_eip8246_delayed_clear_disabled`, and the `JournalCfg` field), since it now gates delayed clearing rather than burn logging.

Because EIP-6780 is always active alongside EIP-8246, only same-tx-created accounts reach this path, so they never have database-resident storage and clearing the in-memory storage is sufficient.

### Gating note
The clear function gates on `AMSTERDAM && !eip8246_delayed_clear_disabled`. The old `eip7708_disabled` (transfer-log toggle) check was dropped from this path, since EIP-8246 is a state change independent of EIP-7708's logging.

## Testing

- Updated `test_eip7708_selfdestruct_to_self` → `test_eip8246_selfdestruct_to_self_preserves_balance`: asserts no burn log and that the created account keeps its balance with cleared code (`KECCAK_EMPTY`), nonce `0`, and no self-destruct flag.
- `cargo build --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --all --check`, and `cargo nextest run` (revm-context + revm-ee-tests, 93 tests) all pass.

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -73,14 +73,17 @@ pub trait Cfg {
     /// Returns whether EIP-7708 (ETH transfers emit logs) is disabled.
     fn is_eip7708_disabled(&self) -> bool;
 
-    /// Returns whether EIP-7708 delayed burn logging is disabled.
+    /// Returns whether the EIP-8246 delayed clearing of self-destructed accounts is disabled.
     ///
-    /// When enabled, revm tracks all self-destructed addresses and emits logs for
-    /// accounts that still have remaining balance at the end of the transaction.
-    /// This can be disabled for performance reasons as it requires storing and
-    /// iterating over all self-destructed accounts. When disabled, the logging
-    /// can be done outside of revm when applying accounts to database state.
-    fn is_eip7708_delayed_burn_disabled(&self) -> bool;
+    /// When enabled, revm tracks all self-destructed addresses and, at the end of the
+    /// transaction, clears the code, storage and nonce of any that still have a remaining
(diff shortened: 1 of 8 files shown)


rakita:
This look correct

PR #3718, opened 2026-05-26
chore: bump ark-* crates to 0.6 version

Pana:
To solve cargo audit issue: https://rustsec.org/advisories/RUSTSEC-2025-0055


rakita:
Will do it manually

PR #3726, opened 2026-05-27
feat(database): allow missing BAL storage reads

mediocregopher:
Adds an opt-in BAL mode where storage slots missing from the BAL fall through to the backing database instead of returning SlotNotFound. Account membership and declared storage writes remain strict, so the relaxed behavior only applies to absent read-only storage entries.


rakita:
This would mean that we dont have read in BAL. So prewarming for database is not happening for this storage.

What usecase are we targeting here, is it BAL without reads, but just writes?

rakita (approved):

mediocregopher:
> This would mean that we dont have read in BAL. So prewarming for database is not happening for this storage.
> 
> What usecase are we targeting here, is it BAL without reads, but just writes?

This is for [TIP-1043](https://github.com/tempoxyz/tempo/pull/3438) and any other chains that might want to do something different for their BAL. TIP-1043 will leave out storage_reads for specific txs, so some txs will have them (TIP-20 transfers) and others won't. At that point storage_reads will be hints, misses will come from cache/db.

mediocregopher:
Going to start a new branch with TIP-1043 changes for now and see what else is required

PR #3727, opened 2026-05-28
feat: `Account::is_changed`

klkvr:
Adds `is_changed` helper that allows to know if account's info has changed vs original one without cloning the original info

### DIFF
--- a/crates/state/src/lib.rs
+++ b/crates/state/src/lib.rs
@@ -203,9 +203,10 @@ impl Account {
     /// Returns true if account info was changed.
     #[inline]
     pub fn is_changed(&self) -> bool {
-        self.original_info
-            .as_deref()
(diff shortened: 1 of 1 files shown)


crates/state/src/lib.rs
```diff
@@ -200,6 +200,14 @@ impl Account {
         self.status.contains(AccountStatus::Touched)
     }
 
+    /// Returns true if account info was changed.
+    #[inline]
+    pub fn is_changed(&self) -> bool {
+        self.original_info
+            .as_deref()
+            .map_or(!self.info.is_default(), |original| self.info != *original)
```
mattsse: shouldn't these use map_or_else/unwrap_or_else because !self.info.is_default() is always computed here I believe

klkvr: yep fair fe8d5c6e

PR #3732, opened 2026-06-01
perf: avoid reallocations and cloning during commit

klkvr:
Right now when state hook is configured on `State<DB>`, it ends up cloning the entire state being commited before sending to it. This is done because the `apply_account_state` takes an owned account which is only useful to avoid a clone of `AccountInfo` (storage hashmap allocation is not reused)

This PR changes `apply_account_state` to take a `Cow<'_, Account>` and avoid cloning when state hook is configured

### DIFF
--- a/crates/database/src/states/transition_state.rs
+++ b/crates/database/src/states/transition_state.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use super::{StorageSlot, TransitionAccount};
 use primitives::{hash_map::Entry, Address, AddressMap, HashMap};
 use state::EvmStorage;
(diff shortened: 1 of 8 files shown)


crates/database/src/states/transition_state.rs
```diff
@@ -33,18 +36,46 @@ impl TransitionState {
...
+        address: Address,
+        account: TransitionAccount<Option<Cow<'_, EvmStorage>>>,
+    ) {
+        match self.transitions.entry(address) {
+            Entry::Occupied(entry) => entry.into_mut().update(account),
+            Entry::Vacant(entry) => {
+                _ = entry.insert(account.map_storage(|storage| {
+                    storage
+                        .map(|storage| {
+                            storage
```
DaniPopes, line 64: .retain maybe?

rakita (approved):
lgtm

crates/database/src/states/transition_state.rs
klkvr: we convert `EvmStorageSlot` -> `StorageSlot` here so we have to realloc

PR #3734, opened 2026-06-01
feat: add SSTORE gas-state policy hook

rakita:
## Summary

Adds a type-level SSTORE gas-state policy hook, allowing custom EVM variants to plug into SSTORE gas accounting without modifying the core instruction.

- Introduces a `GasStateTr` policy trait invoked after the storage write is journaled and before final gas/refund accounting, with a no-op default (`NoGasState`).
- Adds `instruction_table_with_gas_state` to build an instruction table with a custom SSTORE gas-state policy.

This is the base layer for follow-up work (TIP-1060 storage gas tokens, tracked in #3733 against the `tip1060` branch).

### DIFF
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -94,6 +94,15 @@ pub trait Cfg {
     /// via the reservoir model. EIP-8037 specifies concrete gas values based on
     /// `cost_per_state_byte` and adds a hash cost for deployed bytecode.
     fn is_amsterdam_eip8037_enabled(&self) -> bool;
+
+    /// Returns the TIP-1060 storage gas token account, if configured.
+    ///
+    /// When `Some`, the address points to the protocol account holding the
+    /// per-account storage-gas-token counters consumed/minted by the
+    /// `STATE_GAS_TOKEN` variant of the SSTORE instruction.
+    ///
+    /// Defaults to `None` (feature disabled).
+    fn storage_gas_token_contract(&self) -> Option<Address>;
 }
 
 /// What bytecode analysis to perform
--- a/crates/context/interface/src/cfg.rs
+++ b/crates/context/interface/src/cfg.rs
@@ -94,15 +94,6 @@ pub trait Cfg {
     /// via the reservoir model. EIP-8037 specifies concrete gas values based on
     /// `cost_per_state_byte` and adds a hash cost for deployed bytecode.
     fn is_amsterdam_eip8037_enabled(&self) -> bool;
-
-    /// Returns the TIP-1060 storage gas token account, if configured.
-    ///
-    /// When `Some`, the address points to the protocol account holding the
-    /// per-account storage-gas-token counters consumed/minted by the
-    /// `STATE_GAS_TOKEN` variant of the SSTORE instruction.
-    ///
-    /// Defaults to `None` (feature disabled).
-    fn storage_gas_token_contract(&self) -> Option<Address>;
 }
 
 /// What bytecode analysis to perform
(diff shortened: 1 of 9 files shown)


crates/context/interface/src/cfg.rs
```diff
@@ -94,6 +94,15 @@ pub trait Cfg {
...
     fn is_amsterdam_eip8037_enabled(&self) -> bool;
+
+    /// Returns the TIP-1060 storage gas token account, if configured.
+    ///
+    /// When `Some`, the address points to the protocol account holding the
+    /// per-account storage-gas-token counters consumed/minted by the
+    /// `STATE_GAS_TOKEN` variant of the SSTORE instruction.
+    ///
+    /// Defaults to `None` (feature disabled).
+    fn storage_gas_token_contract(&self) -> Option<Address>;
```
0xrusowsky: i think this is not needed cause it is a determinstic address, so tempo is aware of if without having to ask to the cfg

crates/context/interface/src/host.rs
```diff
@@ -19,6 +19,91 @@ pub enum LoadError {
...
+
+/// Derives the storage slot of an account's TIP-1060 gas token counter inside
+/// the storage gas token contract.
+///
+/// The counter is stored directly (no Solidity mapping hash): the slot is the
+/// account address left-padded into a [`StorageKey`].
+#[inline]
+pub fn storage_gas_token_slot(account: Address) -> StorageKey {
+    StorageKey::from_be_bytes(account.into_word().0)
+}
```
0xrusowsky: this can live on tempo

crates/context/interface/src/host.rs
0xrusowsky, line 1: actually all the types defined in `host.rs` can (imo should) live in tempo, as they are impl details

crates/context/interface/src/host.rs
```diff
@@ -19,6 +19,91 @@ pub enum LoadError {
...
+
+impl StorageCreationMode {
+    /// Decodes the 2-bit mode field (only the low two bits are considered).
+    #[inline]
+    pub const fn from_bits(bits: u8) -> Self {
+        match bits & 0b11 {
+            0 => Self::RefundTokens,
+            1 => Self::PreserveTokens,
+            2 => Self::DirectTokens,
+            _ => Self::PreserveTokens,
```
0xalpharush: Think we should reject 4 as reserved

crates/context/interface/src/host.rs
```diff
@@ -19,6 +19,91 @@ pub enum LoadError {
...
+/// Result of a TIP-1060 storage gas token counter update.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct GasTokenResult {
+    /// SStore result of the counter slot write, used to charge the extra
+    /// SSTORE-worth of gas. `None` when no write happened (a [`GasTokenOp::Consume`]
+    /// against an empty counter).
+    pub counter: Option<SStoreResult>,
+    /// Whether a token was actually consumed (a [`GasTokenOp::Consume`] that hit
+    /// a positive counter).
+    pub consumed: bool,
```
0xalpharush: Isn't this the redundant with `counter.is_some()` or `counter.is_none()`?

rakita: This struct was removed

crates/context/interface/src/cfg.rs
rakita: True, as all logic is in tempo

crates/context/interface/src/host.rs
rakita: This struct was removed

PR #3750, opened 2026-06-11
refactor: remove `GasStateTr`

klkvr:
Removes `GasStateTr` in favor of exposing `sstore_with_gas_accounting` allowing to customize gas accounting for the instruction directly


0xrusowsky (approved):
makes sense to me, pending @rakita

rakita (approved):

PR #3764, opened 2026-06-17
Use direct blst Pippenger for BLS MSM

decofe:
## Summary
- replace blst's Rust MultiPoint::mult wrapper with direct blst_p1s/blst_p2s_mult_pippenger calls for BLS12-381 MSM
- keep the existing single-point scalar multiplication fast path
- allocate scratch with blst's scratch sizing helpers and pass null-terminated pointer batches expected by the C API

## Tests
- cargo fmt -p revm-precompile
- cargo test -p revm-precompile bls12_381 --features blst
- cargo test -p revm-precompile --features blst
- cargo bench -p revm-precompile --features blst --bench bench --no-run

Prompted by: @Rjected


Rjected:
only marking ready to see if codspeed runs

4waan:
I have opened #3846  adding k = 4, 8, 16 and 31 to both MSM groups. If that lands first, this pr is gonna get a true signal on re running.

rakita:
bls precompiles are not even used that much, not wort optimizing

PR #3765, opened 2026-06-19
feat(context-interface): add tx-level skip base fee check to enable gasless tx

pochenai:
X Layer needs to support gasless transactions. Today, however, the base fee check is enforced as part of transaction validation, including the EIP-1559 fee checks.

For X Layer, simply tuning the EIP-1559 elasticity or denominator parameters to drive the base fee to zero is not a good fit, because the OP-Stack chain also has system transactions that still rely on the normal fee model. The existing `CfgEnv::disable_base_fee` switch is block-level, so using it for gasless transactions would require temporarily disabling the check for a single transaction and then restoring it afterward. That is fragile and makes per-transaction fee policy harder to express.

This PR adds a defaulted `skip_base_fee_check()` method to the `Transaction` interface. The method returns `false` by default, preserving existing behavior for all current transaction implementations. Custom transaction types can override it to opt out of the `gas_price >= basefee` validation on a per-transaction basis.

This allows chains with sponsored, gasless, or system-specific transaction flows to bypass the base fee check only for the transactions that require it, without changing block-wide validation configuration.

### DIFF
--- a/crates/context/interface/src/transaction.rs
+++ b/crates/context/interface/src/transaction.rs
@@ -86,6 +86,12 @@ pub trait Transaction {
     /// For Eip1559 it is max_fee_per_gas.
     fn gas_price(&self) -> u128;
 
+    /// Whether the `gas_price >= basefee` check should be skipped for this transaction
+    /// during validation. Defaults to `false`.
+    fn skip_base_fee_check(&self) -> bool {
+        false
+    }
+
     /// Access list for the transaction.
     ///
     /// Introduced in EIP-2930.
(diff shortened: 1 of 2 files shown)


pochenai:
reopened this PR, as we can't simply set `evm.block.basefee = 0`, because that would make the BASEFEE opcode return 0, which is incorrect. There might be opcodes or contracts that rely on BASEFEE to retrieve the actual base fee. We only want to exempt the transaction from paying the fee, without changing the semantics of other opcodes.

@rakita please take a look when you are avaliable

pochenai:
seems we can do it with the debug flag `disable_fee_check`, though it's not that elegant.

PR #3766, opened 2026-06-19
test(precompile): add benchmarkoor bls12-381 benches

Rjected:
Adds benchmarkoor-named Criterion coverage for the EIP-2537 BLS12-381 precompiles using the EF/Jochemnet compute cases from `NethermindEth/gas-benchmarks` release `amsterdam-repricings-v5.2.0`.

The BLS bench matrix now covers the cached `test_bls12_381[...]` cases, fixed MSM sizes `k=1,16,64,128`, pairing sizes `num_pairs=1,3,6,12,24`, and uncachable variants for add/MSM/map/pairing-style inputs. The MSM inputs keep the benchmarkoor scalar as `Spec.Q`; the uncachable MSM variants vary only the point input.

The MSM implementation now reduces scalars modulo the subgroup order before deciding whether to skip a term. This means scalar values equal to `r`/`Spec.Q` take the zero-scalar path and return encoded infinity when every term reduces to zero. The blst backend also skips MSM terms whose point is infinity after on-curve validation, avoiding the subgroup/scalar/MSM work for the later iterations of benchmarkoor-style uncachable MSM loops.

### DIFF
--- a/crates/precompile/bench/eip2537.rs
+++ b/crates/precompile/bench/eip2537.rs
@@ -260,7 +260,31 @@ fn bench_precompile(
     });
 }
 
-fn bench_uncachable_precompile(
+fn bench_feedback_precompile(
+    group: &mut BenchmarkGroup<'_, WallTime>,
+    name: &'static str,
+    precompile: &Precompile,
+    input: PrecompileInput,
+    ret_size: usize,
+) {
+    group.throughput(Throughput::Bytes(input.len() as u64));
+    group.bench_function(name, move |b| {
+        let mut input = input.clone();
+        b.iter(|| {
+            // Benchmarkoor's uncachable BLS loop uses args_offset=0, ret_offset=0, and
+            // args_size=CALLDATASIZE, so each call writes its output over the next input prefix.
+            let output = precompile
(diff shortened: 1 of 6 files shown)


mark0-cn:
I opened #3770 against this branch as a small follow-up.

While going through the MSM changes here I noticed the infinity-point skip (68ade1aa) only landed in the **blst** backend — the **arkworks** fallback still pushes points at infinity into the MSM (it only picked up the zero-scalar skip). #3770 brings arkworks to parity by mirroring the `read_g1_msm` / `read_g2_msm` helpers, and also:

- fixes the `read_scalar` doc comment in `blst.rs`, which now contradicts itself — it reduces modulo the subgroup order `r` via `blst_scalar_from_be_bytes`, but the doc still says it doesn't reduce;
- adds arkworks unit tests plus backend-agnostic integration tests asserting an infinity point paired with a non-zero scalar produces the same result as omitting it.

Output is unchanged in both backends (an infinity point contributes nothing to the MSM); it just avoids feeding infinity into the MSM and the subgroup check. The PR targets this branch, so the diff is only the follow-up — feel free to fold it in or use it as a reference. 🙏

PR #3767, opened 2026-06-19
perf(precompile): use malachite for std modexp

Rjected:
Switches the default std MODEXP backend from `aurora-engine-modexp` to `malachite::Natural` with `ModPow::mod_pow`. The `gmp` feature still takes precedence when explicitly enabled, and no-std builds continue to use the existing Aurora implementation.

The Malachite path imports base, exponent, and modulus as big-endian base-256 digits, handles zero modulus the same way as the previous backend, and reduces the base before calling `mod_pow` because Malachite requires the base to already be reduced modulo the modulus. The new dependency is wired only through `revm-precompile/std` with `default-features = false` and `naturals_and_integers` enabled.

Validated with:
- `cargo check -p revm-precompile`
- `cargo check -p revm-precompile --no-default-features`
- `cargo test -p revm-precompile modexp`
- `cargo test -p revm-precompile modexp --no-default-features`
- `cargo check -p revm-precompile --features gmp`
- `cargo test -p revm-precompile modexp --features gmp`
- `cargo check -p revm-precompile --all-features`
- `cargo bench -p revm-precompile --bench bench modexp --no-run`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1170,6 +1170,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7575182f7272186991736b70173b0ea045398f984bf5ebbb3804736ce1330c9d"
 
 [[package]]
+name = "bytemuck"
+version = "1.25.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


rakita:
Its licence is a problem it is LGPL

And it is probably going to be replaced in future https://eips.ethereum.org/EIPS/eip-8200

PR #3770, opened 2026-06-22
perf(precompile): skip bls msm infinity points in arkworks backend

mark0-cn:
Builds on #3766. That PR adds BLS12-381 MSM benches and teaches the **blst** backend to skip points at infinity and zero-reduced scalars in G1/G2 MSM. The **arkworks** fallback only got the zero-scalar skip — it still pushed infinity points into the MSM.

This PR brings the arkworks backend to parity:

- Add `read_g1_msm` / `read_g2_msm` in `arkworks.rs` mirroring the blst helpers: validate on-curve, return `None` for the point at infinity (skip it), then subgroup-check the non-infinity points. Wire them into `p1_msm_bytes` / `p2_msm_bytes`. Output is unchanged — an infinity point contributes nothing to the MSM — but it no longer enters the MSM (and skips the subgroup check, matching blst).
- Resolve the now-contradictory `read_scalar` doc comment in `blst.rs`: it reduces modulo the subgroup order `r` via `blst_scalar_from_be_bytes`, so the leftover "we do not reduce" note was wrong. Clarify the arkworks equivalent too. Reducing modulo `r` is sound because every MSM input point is subgroup-checked (order `r`), so `scalar · P == (scalar mod r) · P`.

### Tests
- arkworks unit tests mirroring blst's `read_g{1,2}_msm_skips_infinity`.
- Backend-agnostic integration tests `bls_g{1,2}msm_skips_infinity_points`, asserting an infinity point paired with a non-zero scalar yields the same result as omitting it.

Verified locally:
- `cargo test -p revm-precompile bls12_381` (blst, default) — 13 passed
- `cargo test -p revm-precompile --no-default-features --features std bls12_381` (arkworks) — 11 passed
- `cargo clippy -p revm-precompile` and `--no-default-features --features std` — clean

Targeting `dan/bls12-381-precompile-benches` so the diff is only these changes on top of #3766.

### DIFF
--- a/crates/precompile/src/bls12_381/arkworks.rs
+++ b/crates/precompile/src/bls12_381/arkworks.rs
@@ -127,6 +127,26 @@ fn read_g1(x: &[u8; FP_LENGTH], y: &[u8; FP_LENGTH]) -> Result<G1Affine, Precomp
     Ok(point)
 }
 
+/// Reads a G1 point for use in an MSM, returning `None` for the point at infinity.
+///
+/// Infinity is the identity, lies in every subgroup, and contributes nothing to the MSM
+/// result, so it can be skipped after the on-curve validation. Non-infinity points still
+/// require a subgroup check, exactly like [`read_g1`].
+#[inline]
+fn read_g1_msm(
+    x: &[u8; FP_LENGTH],
+    y: &[u8; FP_LENGTH],
+) -> Result<Option<G1Affine>, PrecompileHalt> {
+    let point = read_g1_no_subgroup_check(x, y)?;
+    if point.is_zero() {
+        return Ok(None);
+    }
+    if !point.is_in_correct_subgroup_assuming_on_curve() {
+        return Err(PrecompileHalt::Bls12381G1NotInSubgroup);
+    }
+    Ok(Some(point))
+}
+
 /// Reads a G1 point without performing a subgroup check.
 ///
 /// Note: Skipping subgroup checks can introduce security issues.
@@ -184,6 +204,28 @@ fn read_g2(
     Ok(point)
 }
 
+/// Reads a G2 point for use in an MSM, returning `None` for the point at infinity.
+///
+/// Infinity is the identity, lies in every subgroup, and contributes nothing to the MSM
+/// result, so it can be skipped after the on-curve validation. Non-infinity points still
+/// require a subgroup check, exactly like [`read_g2`].
+#[inline]
+fn read_g2_msm(
+    a_x_0: &[u8; FP_LENGTH],
+    a_x_1: &[u8; FP_LENGTH],
+    a_y_0: &[u8; FP_LENGTH],
+    a_y_1: &[u8; FP_LENGTH],
+) -> Result<Option<G2Affine>, PrecompileHalt> {
+    let point = read_g2_no_subgroup_check(a_x_0, a_x_1, a_y_0, a_y_1)?;
+    if point.is_zero() {
+        return Ok(None);
+    }
+    if !point.is_in_correct_subgroup_assuming_on_curve() {
+        return Err(PrecompileHalt::Bls12381G2NotInSubgroup);
+    }
+    Ok(Some(point))
+}
+
 /// Reads a G2 point without performing a subgroup check.
 ///
 /// Note: Skipping subgroup checks can introduce security issues.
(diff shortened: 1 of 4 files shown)


mark0-cn:
@Rjected

4waan:
The scalar reduction and the subgroup check are now coupled, but only implicitly. `read_scalar` unconditionally reduces mod r (`from_be_bytes_mod_order`), and that's result-preserving only because every point in the same loop was subgroup-checked by `read_g1_msm` (so ord(P) | r). Nothing in the types records that dependency.

If a later change swaps `read_g1_msm` for an unchecked reader (a "trusted input" fast path, say) while `read_scalar` keeps reducing, a cofactor point P (order h, not r) paired with scalar k = r silently collapses: the code computes [r mod r]P = [0]P = O, but the honest multiple is [r]P != O. That term just vanishes from the accumulation and thats a soundness break for any caller relying on the MSM being a faithful sum. Just a correctness guard.

So wrapping the checked point in a newtype (`SubgroupCheckedG1`) that's constructible only through the checking reader, and have `p1_msm` take that type. Then wiring in an unchecked reader fails to *compile* then the invariant holds all the time.

(apologies for the confusing previous unedited comment, the two fixes obviously don't go down the same way)

```
/// A G1 point proven to lie in the order-`r` subgroup.
///
/// `read_scalar` reduces every scalar mod `r`; that's result-preserving only
/// when `ord(P) | r`. This type carries that proof: it can only be built by a
/// constructor that runs the subgroup check, so a future unchecked reader can't
/// be pushed into the MSM accumulator without spelling out `assume_in_subgroup`.
#[repr(transparent)] // guarantees identical layout to G1Affine (see unwrap below)
#[derive(Clone, Copy)]
struct SubgroupCheckedG1(G1Affine);

impl SubgroupCheckedG1 {
    /// Reads a G1 MSM operand. `None` for infinity (identity, order `1 | r`);
    /// errors on any on-curve point outside the order-`r` subgroup.
    #[inline]
    fn read_msm(
        x: &[u8; FP_LENGTH],
        y: &[u8; FP_LENGTH],
    ) -> Result<Option<Self>, PrecompileHalt> {
        let point = read_g1_no_subgroup_check(x, y)?;
        if point.is_zero() {
            return Ok(None);
        }
        if !point.is_in_correct_subgroup_assuming_on_curve() {
            return Err(PrecompileHalt::Bls12381G1NotInSubgroup);
        }
        Ok(Some(Self(point)))
    }

    /// Only sanctioned way to build one without re-checking; greppable on purpose.
    #[cfg(test)]
    #[inline]
    fn assume_in_subgroup(point: G1Affine) -> Self {
        Self(point)
    }
}
```

Accumulator is `Vec<SubgroupCheckedG1>` unwrap once at end

```
pub(crate) fn p1_msm_bytes(
    point_scalar_pairs: impl Iterator<Item = Result<(G1Point, [u8; SCALAR_LENGTH]), PrecompileHalt>>,
) -> Result<[u8; G1_LENGTH], PrecompileHalt> {
    let (lower, _) = point_scalar_pairs.size_hint();
    let mut g1_points: Vec<SubgroupCheckedG1> = Vec::with_capacity(lower); // ← typed accumulator
    let mut scalars = Vec::with_capacity(lower);

    for pair_result in point_scalar_pairs {
        let ((x, y), scalar_bytes) = pair_result?;

        // Skip infinity; the type proves every surviving point is subgroup-checked.
        let Some(point) = SubgroupCheckedG1::read_msm(&x, &y)? else {
            continue;
        };

        let scalar = read_scalar(&scalar_bytes)?;
        g1_points.push(point);   // ← only a SubgroupCheckedG1 fits here
        scalars.push(scalar);
    }

    if g1_points.is_empty() {
        return Ok([0u8; G1_LENGTH]);
    }

    // Unwrap the proof at the single MSM boundary. `.map(|p| p.0)` over a
    // `Vec::into_iter()` reuses the buffer in place (identical layout), so this
    // is allocation-free — and it's the ONLY place a checked point becomes raw.
    let g1_points: Vec<G1Affine> = g1_points.into_iter().map(|p| p.0).collect();

    let result = p1_msm(g1_points, scalars); // ← p1_msm unchanged, still Vec<G1Affine>
    Ok(encode_g1_point(&result))
}
```

mark0-cn:
@4waan Thanks for pointing out the implicit coupling. Addressed in cc10da48.

- Added `SubgroupCheckedG1` and `SubgroupCheckedG2` transparent newtypes returned by the checked MSM readers.
- Changed `p1_msm` / `p2_msm` to accept only those checked operand types, so swapping in an unchecked reader now fails to compile.
- Unwrap the checked points only at the arkworks MSM boundary.
- Kept the infinity fast path and made the scalar-reduction invariant explicit in the docs.

Verified with:

- `cargo test -p revm-precompile --no-default-features --features std bls12_381` (11 passed)
- `cargo test -p revm-precompile bls12_381` (13 passed)
- `cargo clippy -p revm-precompile --no-default-features --features std --all-targets -- -D warnings`
- `cargo clippy -p revm-precompile --all-targets -- -D warnings`
- `cargo fmt --all --check`

PR #3780, opened 2026-06-28
fix(inspect): finalize journal on error to prevent EIP-2929 warm set leak

### LINKED ISSUE #3779: InspectEvm::inspect_tx skips finalize() on error, unlike ExecuteEvm::transact — journal (EIP-2929 warm set) leaks into the next tx
### Summary

`ExecuteEvm::transact` and `InspectEvm::inspect_tx` are intended to be the same "execute one tx **and finalize**" operation (one without an inspector, one with). But they place the `?` and `finalize()` in the **opposite order**, so they behave differently when the transaction returns `Err`:

- `transact` finalizes **before** propagating the error → the journal is cleared on error.
- `inspect_tx` propagates the error **before** finalizing → on error, `finalize()` is skipped and the journal is **not** cleared.

When the same `Evm` instance is reused for the next transaction (the normal pattern for block building / `transact_many`-style loops), the un-cleared journal from a failed tx leaks into the next one. In particular the **EIP-2929 warm/cold access set** survives, so the next transaction can be charged *warm* (100) instead of *cold* (2600/2100) for an account/slot that the failed tx happened to touch. This produces a different `gasUsed` → different receipts root / state root than a clean re-execution.

This is not purely theoretical. A real downstream example: in [[base/base](https://github.com/base/base)](https://github.com/base/base) v1.1.1, the block-building `EvmFactory::create_evm` (the non-inspector constructor) accidentally built the EVM with `inspect = true` (it called the `build_with_inspector_*` variant with a `NoOpInspector` instead of the `build_base_*` variant), so the production block-building path went through `inspect_tx`. A failed `nonce too low` tx then leaked its warm set into the next tx, the built block diverged from validators' re-execution, and the chain halted. The downstream code is at fault for enabling inspector mode on the building path, but the asymmetry below is what turned that mistake into a consensus-level gas divergence rather than a harmless no-op.

### Code

Latest `main` (`ae95b776`):

**`crates/handler/src/api.rs` — `ExecuteEvm::transact` (finalize before `?`):**
```rust
fn transact(&mut self, tx: Self::Tx) -> Result<ExecResultAndState<...>, Self::Error> {
    let output_or_error = self.transact_one(tx);
    // finalize will clear the journal
    let state = self.finalize();      // runs even on error
    let output = output_or_error?;    // error propagated AFTER finalize
    Ok(ExecResultAndState::new(output, state))
}
```

**`crates/inspector/src/inspect.rs` — `InspectEvm::inspect_tx` (`?` before finalize):**
```rust
/// Inspect the EVM and finalize the state.
fn inspect_tx(&mut self, tx: Self::Tx) -> Result<ExecResultAndState<...>, Self::Error> {
    let output = self.inspect_one_tx(tx)?;   // error propagated BEFORE finalize
    let state = self.finalize();             // skipped on error
    Ok(ExecResultAndState::new(output, state))
}
```

`InspectCommitEvm::inspect_tx_commit` (same file) has the same `let output = self.inspect_one_tx(tx)?;` ordering and is affected too.

### Impact

For any consumer that:
1. runs in inspector mode (`inspect_tx` / `inspect_tx_commit`), and
2. reuses the same `Evm` instance across transactions (so the journal is *not* dropped between txs),

a transaction that returns `Err` (e.g. `nonce too low`, validation failure) leaves its warmed accounts/slots in the journal. The next transaction then prices EIP-2929 accesses to those accounts as warm. Because the journal's `transaction_id` is also not incremented, the warm set is not invalidated for the next tx.

For a **block builder**, this is consensus-critical: the builder seals a block where a later tx used the (lower) warm gas, while a validator re-executing the canonical tx list — which does **not** include the failed tx — prices the same access cold. The two nodes compute different `gasUsed`/receipts root/state root for the same block, i.e. a **gas/state divergence** that the rest of the network cannot reproduce.

This only manifests when inspector mode is enabled on the execution path. Production block building normally runs without an inspector, but a consumer can reach this state by accident (e.g. constructing the building EVM with `inspect = true` even with a no-op inspector), at which point a single failed tx adjacent to a tx touching the same account can wedge the chain. We hit exactly this downstream.

### Minimal repro sketch

```text
db: account V with nonce = 1 (EOA), a probe contract P whose code does EXTCODESIZE(V),
    and a funded caller C.

reuse ONE evm with inspect = true:
  inspect_tx(tx{from: V, nonce: 0})   -> Err(NonceTooLow)   // V is warmed during validation, finalize skipped
  inspect_tx(tx{from: C, call: P})    -> Ok                  // EXTCODESIZE(V) priced WARM (100)

fresh evm with inspect = true:
  inspect_tx(tx{from: C, call: P})    -> Ok                  // EXTCODESIZE(V) priced COLD (2600)

=> the two probe txs report different gasUsed (delta = 2600 - 100 = 2500),
   solely because the failed tx's warm set leaked through the skipped finalize.
```
(With `transact` instead of `inspect_tx`, both report the cold price — `transact` finalizes on error.)

### Question

Is the `?`-before-`finalize` ordering in `inspect_tx` / `inspect_tx_commit` **intentional** (e.g. to preserve post-error journal state for debugging), or should it match `transact` / `transact_many` and finalize on error so the two paths are consistent?

If it's intended, it would help to document the divergent error/finalize contract on `inspect_tx` (and warn that the `Evm` must not be reused after an error in inspector mode). If it's an oversight, aligning `inspect_tx` with `transact` (finalize before propagating the error) would remove the asymmetry.

revm version checked: `main` @ `ae95b776` (also present in the v1.x line we depend on).

KyrinCode:
## Problem

Closes #3779

`InspectEvm::inspect_tx` (and its sibling inspect/system-call/commit variants) used
`self.X(tx)?` to short-circuit before calling `finalize()`. On error the handler calls
`discard_tx`, which reverts account values but keeps the entries in the state map and
increments `transaction_id`, so the EIP-2929 warm set and touched accounts leaked into
the next transaction.

The same latent bug also existed in `ExecuteEvm::transact_commit` and `ExecuteEvm::replay`.

## Solution

Fix all affected functions to finalize the journal on the error path, matching the existing
correct pattern in `ExecuteEvm::transact` and `transact_many`:

- **`crates/inspector/src/inspect.rs`** (8 functions): the `ExecResultAndState`-returning
  variants use "capture `Result` → `finalize()` unconditionally → `?`"; the `_commit`
  variants use `.inspect_err(|_| { let _ = self.finalize(); })`.
- **`crates/handler/src/api.rs`**: `transact_commit` (inspect_err pattern) and `replay`
  (`.inspect_err(...).map(...)` chain). `replay_commit` is fixed transitively via `replay`.

Each modified function documents the error-path behavior under a `# Outcome of Error`
section.

## Testing

Added two regression tests (`test_inspect_tx_finalizes_journal_on_error` and
`test_inspect_tx_commit_finalizes_journal_on_error`) that trigger a nonce-mismatch
`InvalidTransaction` (a state-dependent error raised *after* `load_accounts` warms the
caller/coinbase) and assert the journal state is empty afterwards.

Verified both tests **fail on the unfixed code** (`found 1 leftover accounts`) and **pass
after the fix** — confirming they genuinely reproduce the leak.

All checks green: `cargo fmt --all --check`, `cargo clippy --all-targets`,
`cargo test -p revm-inspector -p revm-handler` (19 passed), `typos`.

### DIFF
--- a/crates/inspector/src/inspect.rs
+++ b/crates/inspector/src/inspect.rs
@@ -25,8 +25,7 @@ pub trait InspectEvm: ExecuteEvm {
     /// # Outcome of Error
     ///
     /// If the transaction fails, the journal is finalized (cleared) so that the
-    /// next transaction starts from a clean state. This mirrors
-    /// [`ExecuteEvm::transact`][handler::ExecuteEvm::transact].
+    /// next transaction starts from a clean state. This mirrors [`ExecuteEvm::transact`].
     fn inspect_tx(
         &mut self,
         tx: Self::Tx,
--- a/crates/inspector/src/inspect.rs
+++ b/crates/inspector/src/inspect.rs
@@ -21,23 +21,37 @@ pub trait InspectEvm: ExecuteEvm {
     fn inspect_one_tx(&mut self, tx: Self::Tx) -> Result<Self::ExecutionResult, Self::Error>;
 
     /// Inspect the EVM and finalize the state.
+    ///
+    /// # Outcome of Error
+    ///
+    /// If the transaction fails, the journal is finalized (cleared) so that the
+    /// next transaction starts from a clean state. This mirrors
+    /// [`ExecuteEvm::transact`][handler::ExecuteEvm::transact].
     fn inspect_tx(
         &mut self,
         tx: Self::Tx,
     ) -> Result<ExecResultAndState<Self::ExecutionResult, Self::State>, Self::Error> {
-        let output = self.inspect_one_tx(tx)?;
+        // finalize the journal unconditionally, even on error, so the
+        // EIP-2929 warm set and state do not leak into the next transaction.
+        let output_or_error = self.inspect_one_tx(tx);
         let state = self.finalize();
+        let output = output_or_error?;
         Ok(ExecResultAndState::new(output, state))
     }
 
     /// Inspect the EVM with the given inspector and transaction, and finalize the state.
+    ///
+    /// # Outcome of Error
+    ///
+    /// Same as [`InspectEvm::inspect_tx`]: the journal is finalized on error.
     fn inspect(
         &mut self,
         tx: Self::Tx,
         inspector: Self::Inspector,
     ) -> Result<ExecResultAndState<Self::ExecutionResult, Self::State>, Self::Error> {
-        let output = self.inspect_one(tx, inspector)?;
+        let output_or_error = self.inspect_one(tx, inspector);
         let state = self.finalize();
+        let output = output_or_error?;
         Ok(ExecResultAndState::new(output, state))
     }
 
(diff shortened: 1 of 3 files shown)


rakita (approved):
Thanks for the fix

KyrinCode:
Received.Kyrin

crates/inspector/src/inspect.rs
```diff
@@ -21,23 +21,37 @@ pub trait InspectEvm: ExecuteEvm {
     fn inspect_one_tx(&mut self, tx: Self::Tx) -> Result<Self::ExecutionResult, Self::Error>;
 
     /// Inspect the EVM and finalize the state.
+    ///
+    /// # Outcome of Error
+    ///
+    /// If the transaction fails, the journal is finalized (cleared) so that the
+    /// next transaction starts from a clean state. This mirrors
+    /// [`ExecuteEvm::transact`][handler::ExecuteEvm::transact].
```
rakita:
```suggestion
    /// next transaction starts from a clean state. This mirrors [`ExecuteEvm::transact`].
```

PR #3788, opened 2026-07-03
add gas estimation example

plutohan:
Part of #1084.

Adds a `gas_estimation` example: estimates the gas limit of a call with a binary search over non-committing executions, the way `eth_estimateGas` works. Runs fully offline against a `CacheDB` with a small raw-opcode contract, and asserts the estimate covers the actual consumption.

Picked this one to not overlap with the `local_execution_tracer` example proposed in the issue thread.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1965,6 +1965,14 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


rakita:
No need for it

PR #3789, opened 2026-07-07
feat(amsterdam): EIP-2780 runtime gas phase (ethereum/EIPs#11844)

rakita:
Implements the EIP-2780 update from [ethereum/EIPs#11844](https://github.com/ethereum/EIPs/pull/11844) (head `9d2e4e17`): state-dependent transaction charges move from the intrinsic phase to a **runtime gas phase**, applied as the first frame is entered. Intrinsic gas stays the sole input to the validity check.

Everything is gated on the existing `CfgEnv::enable_amsterdam_eip2780`; older forks and 2780-disabled Amsterdam keep the previous code paths (the `eip8037.rs` ee-tests, which disable the flag, pass unchanged).

## Changes

- **Create transactions**: `create_state_gas` (183,600) is no longer intrinsic. It is charged in `make_create_frame` at depth 0, on the frame's gas tracker, only when the deployment target does not already exist — so a create to a pre-existing (balance-only) target pays no new-account state gas, and an initcode revert/halt unwinds the charge LIFO via `rollback_state_gas`. The old top-level refill in `last_frame_result` is gated to the non-2780 path.
- **EIP-7702 decomposition**: the pessimistic per-auth intrinsic charge (15,816 regular + 218,790 state) and its refunds are replaced by an intrinsic `REGULAR_PER_AUTH_BASE_COST = 7,816` (new `eip8038::EIP7702_PER_AUTH_BASE_REGULAR`) plus runtime charges accumulated while applying the auth list (`apply_auth_list_eip2780`): a non-existent authority pays `STATE_BYTES_PER_NEW_ACCOUNT × CPSB` state + `ACCOUNT_WRITE` regular; net-new delegation bytes pay `STATE_BYTES_PER_AUTH_BASE × CPSB` state, at most once per authority. No refunds.
- **Delegated recipient**: the delegation-target access follows the standard EIP-2929 warm/cold model (`WARM_ACCESS` if pre-warmed, e.g. by an access list) instead of a flat `COLD_ACCOUNT_ACCESS`, charged in the runtime phase (`apply_eip2780_runtime_gas`).
- **Runtime out-of-gas is a halt, not invalidity**: `apply_eip2780_runtime_gas` checks affordability of the whole runtime phase (new `InitialAndFloorGas::checked_initial_gas_and_reservoir`). On OOG it reverts a checkpoint spanning the runtime phase — undoing applied delegations — and sets `InitialAndFloorGas::runtime_oog`; `Handler::execution` / `inspect_execution` then include the transaction as an out-of-gas halt consuming all gas without entering the first frame.
- Affordable runtime charges are folded into `initial_regular_gas` / `initial_state_gas`, so they are deducted pre-frame and survive in-frame reverts (delegations persist on in-frame failure). The recipient new-account state gas stays on the frame tracker so a revert rolls it back with the transfer.

## Spec notes / ambiguities

- The PR body on ethereum/EIPs is stale (describes a warm-access-floor split); the head spec charges the recipient/authority at the cold rate unconditionally at intrinsic — implemented per the head.
- The `ACCOUNT_WRITE` bullet's "(i.e. the authority differs from `tx.to`)" parenthetical conflicts with Test Case 8 and the Interactions section ("`ACCOUNT_WRITE` for a non-existent authority"); implemented as `!existed`, matching the old model's net charges.
- Delegation set→clear→set on one authority within a tx charges the indicator bytes once, with no refund on clear ("at most once per authority").

## Testing

- 9 new ee-tests covering the reference cases: create to existing target (no state gas), included-as-halt runtime OOG for create/call/7702 (with delegation reverted), warm vs cold delegation target, new vs existing authority charges; plus a `gas_params` unit test for the intrinsic split.
- `cargo nextest run --workspace` (default / all / no-default features), clippy, fmt, typos: clean.
- Fixtures: pre-Amsterdam fork suites identical to main. `for_amsterdam`: 2663/2666 — the 3 failures (stInitCodeTest create-halt tests) encode the old devnet-6 refill-on-halt semantics; under the new spec the gas_left-drawn portion of the LIFO refill is consumed by the exceptional halt. Fixtures need regeneration for the updated spec.

Follow-up: op-revm needs alignment with the `runtime_oog` execution path.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -1,4 +1,16 @@
 
+# Unreleased
+
+### EIP-2780 runtime gas phase (ethereum/EIPs#11844)
+
+Gated on `CfgEnv::enable_amsterdam_eip2780`; older forks and 2780-disabled configs are unchanged.
+
+* Intrinsic gas no longer includes the create-transaction `create_state_gas` (charged at runtime, only when the deployment target does not exist) nor the pessimistic EIP-7702 per-auth charge — each authorization now costs `REGULAR_PER_AUTH_BASE_COST` (7,816, new `eip8038::EIP7702_PER_AUTH_BASE_REGULAR`) intrinsically, with the state-dependent remainder charged per authority at runtime and no refunds.
+* New struct field (breaks literal construction): `InitialAndFloorGas.runtime_oog` — set when the transaction passes the intrinsic check but cannot afford the runtime charges; the handler includes it as an out-of-gas halt consuming all gas, reverting applied delegations.
+* Additive: `InitialAndFloorGas::checked_initial_gas_and_reservoir`, `GasParams::tx_eip7702_state_gas_bytecode`, `pre_execution::apply_eip2780_runtime_gas`, `pre_execution::apply_auth_list_eip2780`.
+* `pre_execution::apply_eip7702_auth_list` runs the runtime gas phase (and returns no refund) when EIP-2780 is enabled.
+* The delegated-recipient delegation-target access is now warm/cold aware (`WARM_ACCESS` if pre-warmed) and charged in the runtime phase instead of a flat `COLD_ACCOUNT_ACCESS` at frame entry.
+
 # v113 tag (all crates v41.0.0)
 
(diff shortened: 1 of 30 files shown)


rakita:
Will move to glam-devnet-7 branch

PR #3792, opened 2026-07-08
fix(bytecode): avoid iterating legacy padding opcodes

weifanglab:
## Summary

`BytecodeIterator` now iterates over the original legacy bytecode slice instead of the padded execution buffer.

Legacy analysis may append padding bytes so the interpreter can safely decode truncated immediates. Those bytes are not part of the user-provided bytecode, so `iter_opcodes()` should not expose them as real opcodes.

I hava tested by the following cmd:

- `cargo test -p revm-bytecode`
- `cargo test -p revm-bytecode --all-features`
- `cargo clippy -p revm-bytecode --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `cargo check -p revm-bytecode --no-default-features`


rakita (approved):
lgtm

PR #3796, opened 2026-07-13
fix(inspector): forward EIP-7708 transfer logs

0xalpharush:
## Summary

Forward journal-created EIP-7708 ETH transfer logs through `Inspector::log` for frame initialization and SELFDESTRUCT transfers.

Also fix SELFDESTRUCT inspector callbacks so they use the relevant journal entry and report the correct value:

- ignore stale earlier SELFDESTRUCT journal entries when an account selfdestructs to itself
- report the preserved account balance for Amsterdam SELFDESTRUCT-to-self instead of the journal bookkeeping value

This branch is retargeted onto `glam-devnet-7` / #3795.

## Sequencing

This PR currently carries the stale-SELFDESTRUCT journal fix that is also isolated in #3797, because the Amsterdam preserved-value fix depends on inspecting the correct SELFDESTRUCT journal entry. If #3797 lands first, this branch should be rebased and the duplicate stale-journal commit dropped, leaving only the EIP-7708 log forwarding and Amsterdam preserved-value fixes. If this combined PR lands first, #3797 can be closed as a subset.

## Impact

Receipts already contained the EIP-7708 transfer logs, but revm inspector consumers could miss them because the inspector handler only forwarded LOG opcode and precompile logs. Execution and state were already correct; tracer output was incomplete.

SELFDESTRUCT inspector consumers could also observe misleading callback data in selfdestruct-to-self cases: stale journal entries could be reused, and Amsterdam preserved-balance cases could report `0` even when the account retained value.

## Root Cause

The inspector handler needed to forward journal-created EIP-7708 ETH transfer logs after frame initialization and SELFDESTRUCT transfers.

For SELFDESTRUCT, the inspector path also needed to inspect the most recent relevant `AccountDestroyed` journal entry for the target account. Amsterdam/EIP-8246 selfdestruct-to-self preserves the account balance and records `had_balance = 0` for revert bookkeeping, so the callback value must come from the preserved account balance instead of that journal field.

## Testing

- `cargo +nightly test -p revm-inspector eip7708` on the #3795 base
- `cargo +nightly test -p revm-inspector selfdestruct_to_self` on the #3795 base
- `structured_compare_amsterdam/crash-97865b0ccb6eca1d7c1c70cbe3229bbdb91d8ce6` passes in evm2-fuzzers when pinned to this branch at `09772073`

### DIFF
--- a/crates/inspector/src/handler.rs
+++ b/crates/inspector/src/handler.rs
@@ -12,7 +12,11 @@ use interpreter::{
     CallOutcome, CreateOutcome, FrameInput, Host, InitialAndFloorGas, InstructionResult,
     Interpreter, InterpreterAction, InterpreterTypes,
 };
-use primitives::{hints_util::cold_path, TxKind};
+use primitives::{
+    eip7708::{ETH_TRANSFER_LOG_ADDRESS, ETH_TRANSFER_LOG_TOPIC},
+    hints_util::cold_path,
+    Log, TxKind,
+};
 use state::bytecode::opcode;
 
 /// Trait that extends [`Handler`] with inspection functionality.
@@ -277,6 +281,7 @@ where
         }
 
         let opcode = interpreter.bytecode.opcode();
+        let logs_i = context.journal().logs().len();
         if let Err(e) = interpreter.step(instructions, gas_table, context) {
             cold_path();
             if interpreter.bytecode.action().is_none() {
@@ -286,6 +291,8 @@ where
 
         if (opcode::LOG0..=opcode::LOG4).contains(&opcode) {
             inspect_log(interpreter, context, &mut inspector);
+        } else {
+            inspect_eip7708_transfer_logs(context, &mut inspector, logs_i);
         }
 
         inspector.step_end(interpreter, context);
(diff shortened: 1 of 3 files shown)


crates/inspector/src/handler.rs
```diff
@@ -286,6 +293,8 @@ where
 
         if (opcode::LOG0..=opcode::LOG4).contains(&opcode) {
             inspect_log(interpreter, context, &mut inspector);
+        } else {
+            inspect_eip7708_transfer_logs(context, &mut inspector, logs_i);
```
rakita, line 297: We could remove the previous lines `LOG0..=` and just use `inspect_eip7708_transfer_logs` to report every log.

crates/inspector/src/handler.rs
```diff
@@ -301,13 +310,43 @@ where
...
+            if let Some(journal_i) = instruction_journal_i {
+                inspect_selfdestruct(context, &mut inspector, journal_i);
+            }
         }
     }
 
     next_action
 }
 
+#[inline]
```
rakita, line 322:
```suggestion
#[inline(never)]
```

crates/inspector/src/handler.rs
```diff
@@ -301,13 +310,43 @@ where
...
+        }
+    }
+}
+
+#[inline]
+fn is_eip7708_transfer_log(log: &Log) -> bool {
+    log.address == ETH_TRANSFER_LOG_ADDRESS
+        && log.data.topics().first() == Some(&ETH_TRANSFER_LOG_TOPIC)
+}
+
```
rakita, line 349: And remove this if we want to use it for every log


``` suggestion
```

rakita:
This is fixing the missing log for SELFDESTRUCT but i gave me good hint we are missing logs for transfers that are done outside of Interpreter in when `transfer_loaded` is called [here](https://github.com/bluealloy/revm/blob/30e04db0a6285a79f613dc404bfc04e4b757b0c9/crates/handler/src/frame.rs#L189) and on create [here](https://github.com/bluealloy/revm/blob/30e04db0a6285a79f613dc404bfc04e4b757b0c9/crates/handler/src/frame.rs#L323-L324)

This bug is related to glamsterdam. Good find!

rakita:
Superseed with https://github.com/bluealloy/revm/pull/3816

PR #3800, opened 2026-07-15
fix(bytecode): preserve iterator position after truncated push

pengqima:
## Summary

Make `BytecodeIterator::skip_immediate` consume at most the bytes that remain, keeping an exhausted iterator anchored to the original bytecode slice.

This is a follow-up to #3792.

## Problem

After #3792 switched `BytecodeIterator` to the original unpadded legacy bytecode, a truncated `PUSH` can make `get(immediate_size..)` return `None`.

The previous `unwrap_or_default()` fallback then creates an empty slice unrelated to the original bytecode allocation. A later call to `position()` uses `offset_from_unsigned` relative to the original start pointer, which requires both pointers to belong to the same allocation.

## Fix

Slice at `min(immediate_size, remaining.len())` instead. This preserves the existing iteration behavior—all available immediate bytes are consumed and iteration ends—while retaining the original slice provenance.

A regression test covers calling `position()` after a partially truncated `PUSH2`.

## Testing

- `cargo fmt --all --check`
- `cargo test -p revm-bytecode --all-features`
- `cargo clippy -p revm-bytecode --all-targets --all-features -- -D warnings`
- `cargo check -p revm-bytecode --no-default-features`

### DIFF
--- a/crates/bytecode/src/iter.rs
+++ b/crates/bytecode/src/iter.rs
@@ -59,12 +59,8 @@ impl<'a> BytecodeIterator<'a> {
 
         // Advance the iterator by the immediate size
         if immediate_size > 0 {
-            self.bytes = self
-                .bytes
-                .as_slice()
-                .get(immediate_size..)
-                .unwrap_or_default()
-                .iter();
+            let remaining = self.bytes.as_slice();
+            self.bytes = remaining[immediate_size.min(remaining.len())..].iter();
         }
     }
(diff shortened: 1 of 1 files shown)


pengqima:
@rakita Hi, Could you please review this PR at your convenience? Thank you very much.

rakita:
Failed test not related to this change. PR lgtm

PR #3801, opened 2026-07-15
fix(bytecode): skip DUPN/SWAPN/EXCHANGE immediates in analyze_legacy

cuiweixie:
## Summary
- Skip the 1-byte immediate of `DUPN`/`SWAPN`/`EXCHANGE` during legacy analysis (same as `PUSH`), so a `JUMPDEST` byte in that immediate is not marked as a valid jump target.
- Simplify end-of-bytecode padding: incomplete immediates are covered by iterator overflow; only append a final `STOP` when needed.
- Add regression tests for jump-table correctness and truncated opcode padding.

## Test plan
- [x] `cargo test -p revm-bytecode --lib dupn_swapn -- --nocapture`
- [x] `cargo test -p revm-bytecode --lib legacy::analysis::tests`
- [ ] Confirm the two new tests fail if the skip branch is removed

### DIFF
--- a/crates/bytecode/src/legacy/analysis.rs
+++ b/crates/bytecode/src/legacy/analysis.rs
@@ -13,11 +13,9 @@ pub(crate) fn analyze_legacy(bytecode: Bytes) -> (JumpTable, Bytes) {
     let start = range.start;
     let mut iterator = start;
     let end = range.end;
-    let mut prev_byte: u8 = 0;
     let mut last_byte: u8 = 0;
 
     while iterator < end {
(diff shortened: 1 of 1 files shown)


rakita (approved):
lgtm

rakita:
Have reread the EIP as I was suspicious why was not triggered before, previous code was corect and EIP expelicitly allows jumpdest to be found in immediate. Closing the PR.

From EIP https://eips.ethereum.org/EIPS/eip-8024#:~:text=JUMPDEST%20analysis%20is%20unchanged

PR #3802, opened 2026-07-15
fix(context): keep is_empty on EIP-7702 account in load_account_delegated

cuiweixie:
## Summary
- Stop overwriting `AccountLoad::is_empty` with the EIP-7702 delegate account's emptiness in `Host::load_account_delegated`.
- Align the Host default with `JournalInner::load_account_delegated`, so empty-account / creation-cost checks use the caller account.
- Add a regression test: non-empty EIP-7702 account delegating to an empty account must keep `is_empty == false`.

## Test plan
- [x] `cargo test -p revm-context-interface --lib load_account_delegated_keeps_caller_is_empty_not_delegate -- --nocapture`
- [ ] Confirm the test fails if `account_load.data.is_empty = delegate_account.is_empty` is restored

### DIFF
--- a/crates/context/interface/src/host.rs
+++ b/crates/context/interface/src/host.rs
@@ -169,7 +169,6 @@ pub trait Host {
                 .load_account_info_skip_cold_load(address, true, false)
                 .ok()?;
             account_load.data.is_delegate_account_cold = Some(delegate_account.is_cold);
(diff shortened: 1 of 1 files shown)


rakita:
This PR is more correct in CALL type of opcodes you use target is_empty to decide if account is going to be created or not. This function is not used in revm so this is not a consensus issue,

PR #3807, opened 2026-07-20
test(bytecode): document and cover the Bytecode serde implementation

4waan:
The `Bytecode` serde implementation had almost no coverage (only an empty-default round-trip) and no documentation of its design decisions. 
Opening the PR with two commits:

**docs**: module and type docs explaining that `BytecodeSerde` exists to fix the wire format independently of the in-memory representation (mirroring the pre-flatten enum shape from #3375), why the EIP-7702 variant stores only the delegated address, why the `original_len` check must precede `slice` (panic on untrusted input otherwise) and that the serialized `jump_table` is deliberately ignored on deserialization per #3557.

**tests**:
- `test_eip7702_roundtrip` / `test_legacy_roundtrip`: round-trip both variants, including designator reconstruction from the address alone and a non-trivial jump table (`PUSH1 0x5b` data byte vs. a real `JUMPDEST`).
- `test_legacy_rejects_oversized_original_len`: corrupt `original_len` in the serialized form must produce a serde error, not a panic in `Bytes::slice`.
- `test_legacy_ignores_serialized_jump_table`: implant a wrong jump table into the serialized form and assert deserialization re-analyzes instead of trusting it, pinning the security behavior from #3557.

Docs and tests only; no behavior affected.

### DIFF
--- a/crates/bytecode/src/bytecode/serde_impl.rs
+++ b/crates/bytecode/src/bytecode/serde_impl.rs
@@ -1,7 +1,19 @@
+//! Serde support for [`Bytecode`].
+//!
+//! Serialization goes through the private [`BytecodeSerde`] type.
+//! It fixes the wire format independently of the in-memory representation.
+//! Its shape mirrors the previous `Bytecode` enum, so data serialized before
+//! the internal flattening still deserializes.
+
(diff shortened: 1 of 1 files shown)


rakita:
This is no public enum

PR #3815, opened 2026-07-22
refactor(handler): validate state before tracking gas

klkvr:
Moves caller state validation and fee deduction into `Handler::validate` so chain-specific intrinsic gas adjustments finish before `GasTracker` is constructed.

This lets handler implementations customize state-based intrinsic gas without overriding `run_without_catch_error`, while preserving `pre_execution`’s `InitialAndFloorGas` hook for authorization-list metering.

Tests: `cargo fmt --all --check`; `cargo test -p revm-handler -p revm-inspector`

### DIFF
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -195,13 +195,14 @@ pub trait Handler {
 
     /// Validates the execution environment and transaction parameters.
     ///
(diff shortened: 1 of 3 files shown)


rakita:
CI error not related.

PR lgtm

Issue #3687, opened 2026-05-19
Potential soundness issue in `revm-interpreter` 35.0.1


varies-rust:
Hello, and thank you for maintaining this crate.

I found a sanitizer/Miri failure reachable through public crate APIs using safe Rust code. I may be missing crate-specific preconditions, but the behavior looks worth checking because safe callers should not be able to trigger undefined behavior.

## Summary

- Crate: `revm-interpreter`
- Version tested: `35.0.1`
- API paths with similar failures observed:
  - `ExtBytecode::read_u8/read_u16/read_slice`
  - `ExtBytecode::read_offset_u16`
  - `ExtBytecode::opcode/pc`
  - `ExtBytecode::relative_jump/absolute_jump`
  - `SharedMemory::set`
  - `SharedMemory::set_data`
  - `SharedMemory::resize`

## Observed diagnostics

```text
Miri undefined behavior: error: Undefined Behavior: memory access failed: attempting to access 2 bytes, but got alloc88868 which is only 1 byte from the end of the allocation
Unsafe precondition violation: unsafe precondition(s) violated: slice::get_unchecked_mut requires that the range is within the slice
```

## Reproduction

The snippets below are minimal readable reproducers. Each PoC also includes the source location most relevant to the reported failure.

### PoC 1: `test_revm_interpreter10::generated_test_10`

Relevant source location:

- `src/interpreter/ext_bytecode.rs:167-168`: `read_u16` performs an unchecked pointer read at the current instruction pointer.
- `src/interpreter/ext_bytecode.rs:182-188`: `read_offset_u16` applies an unchecked offset before reading two bytes.

Readable equivalent PoC:

```rust
use revm_interpreter::interpreter::ExtBytecode;
use revm_interpreter::interpreter_types::{Immediates, Jumps};

#[test]
fn poc() {
    let mut bytecode = ExtBytecode::default();

    bytecode.relative_jump(isize::MAX);
    let _ = bytecode.read_offset_u16(isize::MAX);
}
```

Observed diagnostic:

```text
error: Undefined Behavior: memory access failed: attempting to access 2 bytes, but got alloc88868 which is only 1 byte from the end of the allocation
```

### PoC 2: `test_revm_interpreter202::generated_test_202`

Relevant source location:

- `src/interpreter/shared_memory.rs:531`: internal `set_data` receives unchecked offsets and length.
- `src/interpreter/shared_memory.rs:543-545`: it uses unchecked slice ranges derived from those values.

[...]

PR #3817, opened 2026-07-23
fix: validate SharedMemory ranges in release builds

1sgtpepper:
## Summary

- Validate `SharedMemory` read and write ranges, checkpoint arithmetic, and child cleanup so invalid safe inputs panic in every profile instead of reaching unchecked operations.
- Preserve memory and checkpoint state when resize or child cleanup is rejected.
- Keep the checked copy fast path for valid `set_data` calls and cover the reported `usize::MAX`, alias invalidation, and failure-atomicity cases.

Addresses the `SharedMemory` cases in https://github.com/bluealloy/revm/issues/3687. The `ExtBytecode` cases remain out of scope to avoid conflicting with https://github.com/bluealloy/revm/pull/3545.

## Testing

- `cargo test -p revm-interpreter shared_memory::tests`
- `cargo test --release -p revm-interpreter shared_memory::tests`
- `cargo +nightly-2026-07-11 miri test -p revm-interpreter --lib shared_memory::tests`
- `cargo +nightly-2026-07-11 miri test --release -p revm-interpreter --lib shared_memory::tests`
- `cargo test -p revm-interpreter --all-features`
- `cargo check -p revm-interpreter --no-default-features`
- `cargo clippy -p revm-interpreter --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='--cfg docsrs -D warnings' cargo doc -p revm-interpreter --all-features --no-deps --document-private-items`
- `cargo fmt --all --check`
- GitHub CI: stable, nightly, no-default-features, docs, clippy, and 32-bit Ethereum tests

### DIFF
--- a/crates/interpreter/src/interpreter/shared_memory.rs
+++ b/crates/interpreter/src/interpreter/shared_memory.rs
@@ -283,11 +283,17 @@ impl SharedMemory {
     }
 
     /// Resizes the memory in-place so that `len` is equal to `new_len`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the checkpoint plus the new size overflows.
     #[inline]
     pub fn resize(&mut self, new_size: usize) {
-        self.buffer()
-            .dbg_borrow_mut()
-            .resize(self.my_checkpoint + new_size, 0);
+        let new_len = self
+            .my_checkpoint
+            .checked_add(new_size)
+            .expect("memory resize overflow");
+        self.buffer().dbg_borrow_mut().resize(new_len, 0);
     }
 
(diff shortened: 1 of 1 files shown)


1sgtpepper:
@rakita, I confirmed the −4.88% regression comes from validating `SharedMemory::invalid()` instead of allowing release UB. All functional CI and focused Miri tests pass. Is this tradeoff acceptable for this PR, or would you prefer an invalid-state redesign first?

rakita:
Will leave it as is and just close the PR, access to memory is done in a way that resize happens before

PR #3823, opened 2026-07-23
perf(precompile): use fused multi-miller loop for BLS12-381 pairing

4waan:
## What

Replace the per-pair miller loop and fp12 multiply accumulation in the blst `pairing_check` with a single `blst_miller_loop_n` call followed by one final exponentiation, sharing the per-bit fp12 squaring across all pairs.

This brings the blst backend to parity with the arkworks backend, which already fuses via `multi_pairing`.

## Result

Bit-identical output (the product of the individual miller loops equals the fused product) — only faster. Benchmark via the existing `eip2537` pairing bench, end-to-end through the precompile:

| pairs | change |
|------:|:------:|
| 1 | ~0% (nothing to fuse) |
| 2 | ~6% |
| 8 | ~21% |
| 16 | ~30% |

## Notes

- Uses the raw FFI `blst_miller_loop_n`, not blst's `miller_loop_n` Rust wrapper, which fans out to a thread pool — undesirable inside a precompile and unavailable on `no_std` targets. This keeps the computation single-threaded and deterministic.
- Pairing correctness is covered by the existing EIP-2537 execution-spec-tests.

### DIFF
--- a/crates/precompile/src/bls12_381/blst.rs
+++ b/crates/precompile/src/bls12_381/blst.rs
@@ -7,8 +7,8 @@ use crate::{
     PrecompileHalt,
 };
 use blst::{
-    blst_bendian_from_fp, blst_final_exp, blst_fp, blst_fp12, blst_fp12_is_one, blst_fp12_mul,
-    blst_fp2, blst_fp_from_bendian, blst_map_to_g1, blst_map_to_g2, blst_miller_loop, blst_p1,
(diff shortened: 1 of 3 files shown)


rakita:
Nice one!

Have pushed commit that would skip infinity points in kzg as [blst_miller_loop](https://github.com/supranational/blst/blob/14cd09aa002dce99e855ac7353b5ef848bbc1553/src/pairing.c#L408) had it. For bls precompile, we are already skipping those.

PR #3829, opened 2026-07-28
fix: check call depth in CREATE pre-access checks (EIP-8037)

rakita:
## Summary

Moves the call-depth limit check from frame creation into the CREATE opcode's pre-access checks. Under EIP-8037, a too-deep CREATE now pushes 0 without reading the destination account, keeping the derived address out of the EIP-7928 block access list and the warm set.

- Add `depth` to `InputsTr`/`InputsImpl` and thread it through frame creation so the opcode can see the current call depth
- `DummyHost` now reports EIP-8037 as enabled when constructed with an Amsterdam spec
- Add a test asserting that a CREATE at the depth limit pushes 0 without requesting a new frame

## Testing

- `cargo nextest run -p revm-interpreter create_too_deep`
- `cargo clippy --workspace --all-targets --all-features`
- `cargo fmt --all --check`

### DIFF
--- a/crates/context/interface/src/host.rs
+++ b/crates/context/interface/src/host.rs
@@ -210,6 +210,7 @@ pub trait Host {
 #[derive(Default, Debug)]
 pub struct DummyHost {
     gas_params: GasParams,
+    spec: SpecId,
 }
 
 impl DummyHost {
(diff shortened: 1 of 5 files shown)


rakita:
This is not reachable by 16M max gas limit, but either way it does not hurt to have it.

rakita:
Not possible to trigger with 16M gas limit, but the check should be there

PR #3833, opened 2026-07-31
fix(inspector): preserve selfdestruct trace payload

### LINKED ISSUE #3834: Inspector::selfdestruct is skipped for post-Cancun selfdestruct-to-self
## Problem

With EIP-6780 enabled, an existing contract that executes `SELFDESTRUCT` with itself as the beneficiary is not deleted and its balance is unchanged. `JournalInner::selfdestruct` therefore records no journal entry, which is correct for state journaling.

The inspected handler currently reconstructs `Inspector::selfdestruct` exclusively from `AccountDestroyed` and `BalanceTransfer` journal entries. Since neither exists in this case, the hook is skipped even though the opcode completed successfully.

## Impact

Downstream tracers can still observe `InstructionResult::SelfDestruct` but receive none of its payload. In reth's callTracer this produces a frame with a zero `from` address and missing `to` and `value`, while geth reports the executing account, beneficiary, and pre-opcode balance.

This also occurs through `DELEGATECALL`, where the account executing `SELFDESTRUCT` differs from the address whose code is running. The executing identity must therefore come from the interpreter input rather than the code address.

The malformed callTracer output was observed on Ethereum mainnet blocks `25644439` and `25644459`.

## Expected behavior

After a successful `SELFDESTRUCT`, `Inspector::selfdestruct` should be called with:

- the account executing the opcode;
- the beneficiary popped by the opcode;
- the executing account's balance immediately before the opcode.

The hook should fire even when the opcode has no state effect and creates no journal entry.

## Proposed fix

#3833 captures the tracing payload before the opcode consumes it and uses the journal-derived payload only as a compatibility fallback.

Related: #638 introduced the `value` parameter, and #1746 added support for the `BalanceTransfer` journal-entry case.

mo4islona:
Closes #3834

## Context

After Cancun, `SELFDESTRUCT` has no state effect when a contract that was not created in the current transaction names itself as the beneficiary. In that case, `JournalInner::selfdestruct` correctly records no journal entry.

`Inspector::selfdestruct` is currently reconstructed from those journal entries, so the hook never fires for this case. `revm-inspectors` still observes the successful `InstructionResult::SelfDestruct` and emits a callTracer frame from unset fields: `from` becomes the zero address, while `to` and `value` are absent. Geth instead invokes its tracer unconditionally with the executing account, beneficiary, and pre-opcode balance.

This is a follow-up to #1746, which added support for the `BalanceTransfer` journal entry but could not cover a selfdestruct that produces no entry at all. The issue is particularly easy to hit through `DELEGATECALL`, where the account executing the opcode differs from the address whose code is running.

## Fix

Capture the tracing payload immediately before executing `SELFDESTRUCT`, while all of it is still available:

- the executing account from `input.target_address()`, which preserves the correct identity under `DELEGATECALL`;
- the beneficiary from the top of the stack;
- the executing account's balance before the opcode.

After a successful `SELFDESTRUCT`, the inspector hook prefers this captured payload. The journal-derived path remains as a compatibility fallback for custom instruction tables. Failed selfdestructs still emit no event, and the non-inspected execution path is unchanged.

## Semantics

`Inspector::selfdestruct` now consistently exposes `value` as the executing account's balance immediately before the opcode, rather than the amount by which state changed. This restores the original inspector behavior and matches geth: `value` can be non-zero even when no funds move.

As a result:

- post-Cancun selfdestruct-to-self is reported instead of being silently dropped;
- delegated execution reports the delegator, not the code address;
- under Amsterdam/EIP-8246, created-in-transaction selfdestruct-to-self reports the pre-opcode balance instead of the journal entry's zero `had_balance`.

### DIFF
--- a/crates/inspector/src/handler.rs
+++ b/crates/inspector/src/handler.rs
@@ -6,12 +6,13 @@ use handler::{
     FrameResult, Handler, ItemOrResult,
 };
 use interpreter::{
-    instructions::{GasTable, InstructionTable},
-    interpreter_types::LoopControl,
+    instructions::{utility::IntoAddress, GasTable, InstructionTable},
+    interpreter_types::{InputsTr, Jumps, LoopControl, StackTr},
     FrameInput, GasTracker, Host, InitialAndFloorGas, InstructionResult, Interpreter,
     InterpreterAction, InterpreterTypes,
 };
-use primitives::hints_util::cold_path;
+use primitives::{hints_util::cold_path, Address, U256};
+use state::bytecode::opcode;
 
 /// Trait that extends [`Handler`] with inspection functionality.
 ///
@@ -259,11 +260,25 @@ where
     IT: InterpreterTypes,
 {
     let mut instruction_journal_i = None;
+    let mut selfdestruct;
     loop {
+        selfdestruct = None;
         inspector.step(interpreter, context);
         if interpreter.bytecode.is_end() {
             cold_path();
             break;
+        }
+
+        if interpreter.bytecode.opcode() == opcode::SELFDESTRUCT {
+            cold_path();
+            // SELFDESTRUCT pops its beneficiary and the journal only records state changes. In
+            // particular, EIP-6780 selfdestruct-to-self leaves no journal entry, so preserve the
+            // complete tracing payload before executing the opcode.
+            let contract = interpreter.input.target_address();
+            selfdestruct = interpreter.stack.top().and_then(|beneficiary| {
+                let balance = context.journal().evm_state().get(&contract)?.info.balance;
+                Some((contract, beneficiary.into_address(), balance))
+            });
         }
 
         instruction_journal_i = Some(context.journal().journal().len());
@@ -294,7 +309,7 @@ where
     if let InterpreterAction::Return(result) = &next_action {
         if result.result == InstructionResult::SelfDestruct {
             if let Some(journal_i) = instruction_journal_i {
-                inspect_selfdestruct(context, &mut inspector, journal_i);
+                inspect_selfdestruct(context, &mut inspector, journal_i, selfdestruct);
             }
         }
     }
@@ -343,6 +358,7 @@ fn inspect_selfdestruct<CTX, IT>(
     context: &mut CTX,
     inspector: &mut impl Inspector<CTX, IT>,
     journal_i: usize,
+    selfdestruct: Option<(Address, Address, U256)>,
 ) where
     CTX: ContextTr<Journal: JournalExt> + Host,
     IT: InterpreterTypes,
@@ -353,7 +369,9 @@ fn inspect_selfdestruct<CTX, IT>(
         .get(journal_i..)
         .and_then(|entries| entries.last());
 
-    if let Some(
+    if let Some((contract, beneficiary, balance)) = selfdestruct {
+        inspector.selfdestruct(contract, beneficiary, balance);
+    } else if let Some(
         JournalEntry::AccountDestroyed {
             address: contract,
             target: to,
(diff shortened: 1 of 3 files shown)


rakita:
No triggering hook if the destruction of the account has not happened looks good. Not sure what you want to fix

mo4islona:
**1. "No destruction, no hook" isn't the rule revm implements today.** In `JournalInner::selfdestruct`, the `address != target` arm pushes `BalanceTransfer` unconditionally, including for a zero balance:

```rust
} else if address != target {
    acc.info.balance = U256::ZERO;
    ...
    Some(ENTRY::balance_transfer(address, target, balance))
}
```

Post-Cancun, for an account not created in the transaction with zero balance, a selfdestruct to a *different* beneficiary destroys nothing and moves nothing — and still fires the hook. The same opcode naming itself fires nothing. The effective rule is `beneficiary != self`, which falls out of journaling rather than from a decision about what tracing should report.

**2. The frame is emitted regardless of the hook.** `revm-inspectors` decides a frame is a selfdestruct from `InstructionResult::SelfDestruct`, not from the hook having fired, so when the hook is skipped it builds the frame from unset fields and callTracer returns:

```json
{"from":"0x0000000000000000000000000000000000000000","gas":"0x0","gasUsed":"0x0","input":"0x","type":"SELFDESTRUCT"}
```

`parity_selfdestruct_action` is worse — it defaults all three fields, so `trace_` asserts a selfdestruct from `0x0` to `0x0`, indistinguishable from a real one. Seen on mainnet blocks 25644439 and 25644459, both through `DELEGATECALL`, identical across three RPC providers.

**3. Geth calls the tracer unconditionally** in `opSelfdestruct6780`, with the executing account, the beneficiary, and the balance immediately before the opcode — which is the semantics `Inspector::selfdestruct` had before it was reconstructed from journal entries.

An event with no state effect can't be recovered from a record of state effects, hence capturing before the opcode runs. It sits behind an opcode check with `cold_path()` in `inspect_instructions`, so the non-inspected path is untouched, and `contract` comes from `input.target_address()` so `DELEGATECALL` reports the delegator.

If you'd rather the hook stayed strictly state-derived, the alternative is to stop `revm-inspectors` emitting a frame without a payload — happy to go that way, though it can't fall back to `trace.address` (for `DELEGATECALL` that's the code address, not the executing identity). I can also invert the precedence here — journal entry first, captured payload only when there is none — to keep existing cases byte-identical.

mo4islona:
@rakita 

To make it concrete: the question isn't whether the hook should fire when nothing is destroyed. `revm-inspectors` decides a frame is a selfdestruct from `InstructionResult::SelfDestruct` alone, so the frame is emitted **whether or not the hook fired**. Skipping the hook doesn't suppress the event - it ships it with empty fields.

Same mainnet transaction (`0x22d29894e6cdb8cf4dbbf71c7d6ba8ec0d597c8f130c500c93b19762828b1d6d`, block 25735294), same `debug_traceTransaction` + callTracer request:

**reth v2.4.0**
```json
{"from":"0x0000000000000000000000000000000000000000","gas":"0x0","gasUsed":"0x0","input":"0x","type":"SELFDESTRUCT"}
```

**geth v1.17.1**
```json
{"from":"0xf92060aa5e079a0a90119bac558b8c93454d0467","gas":"0x0","gasUsed":"0x0","to":"0xf92060aa5e079a0a90119bac558b8c93454d0467","input":"0x","value":"0x0","type":"SELFDESTRUCT"}
```

Via `trace_` it's worse: all three action fields default, so it asserts a selfdestruct from `0x0` to `0x0` — indistinguishable from a genuine one.

"No hook when nothing is destroyed" would be a consistent position if the frame were dropped too. Today it's neither: the event is always reported, and in exactly this case with wrong data. The PR makes the reported payload match geth. This is the third occurrence on mainnet live traffic (blocks 25644439, 25644459, 25735294 — same protocol, recurring), not a synthetic edge case.

At SQD we index and archive debug traces for most EVM chains, and when every reth-based endpoint returns the same malformed frame, there is no way to even detect the corruption - a real pain for us.

PR #3836, opened 2026-08-02
perf(precompile): reformulate kzg verify to drop the G2 scalar mul

### LINKED ISSUE #3835: perf(precompile): kzg point-evaluation verify does an avoidable G2 scalar multiplication
The EIP-4844 point-evaluation precompile (`0x0a`) verify does an avoidable **G2 scalar multiplication** every call in the rust backends. It can be reformulated away. Both the `arkworks` and `blst` backends compute the check in the form:

```
e(commitment - [y]G1, -G2) · e(proof, [τ]G2 - [z]G2) == 1
```

Building `[τ]G2 - [z]G2` requires a **G2 scalar multiplication** `[z]G2` (Fp2 arithmetic — the expensive group), a G2 subtraction, and a normalization (Fp2 inversion), per call. By bilinearity, `e(proof, -[z]G2) = e([z]·proof, -G2)`, so `z` can be moved from G2 to G1. Both G2 pairing inputs then become the **compile-time constants** `-G2` and `[τ]G2`:

```
e(D, -G2) · e(proof, [tau]G2) == 1,   D = commitment - [y]G1 + [z]·proof
```

This replaces the G2 scalar mul (+ G2 sub + Fp2 inversion) with a **cheaper G1** scalar mul.

Additionally, because both G2 inputs are now constant:
- **arkworks**: they can be cached as `G2Prepared` (Miller-loop lines computed once). `multi_miller_loop` still fuses the per-bit squaring across pairs, so this is a strict win with fewer G2 ops, no loss of fusion.
- **blst**: less clear. blst's fused primitive (`blst_miller_loop_n`, see #3823) takes affine points, and its precomputed-lines primitive (`blst_miller_loop_lines`) is single-pair. So caching lines would forfeit the #3823 fusion (needs benchmark). The reformulation itself (dropping the G2 scalar mul) still applies over the fused `pairing_check`.

- Affects the `arkworks` and `blst` backends; `c-kzg` is external/unchanged.
- On `default` features the dispatcher prefers `c-kzg`, so this benefits the pure-Rust / `no_std` consumers (arkworks; blst-without-c-kzg) — e.g. zkVM-style builds.
-This is just a perf refactor of the pairing operands.

Reformulate as above in both pure-Rust backends.. cache the two constant G2 points where the backend allows it (without losing fusion).

4waan:
Move z from G2 to G1 in the EIP-4844 point-evaluation pairing check via bilinearity `e(proof, -[z]G2) = e([z]·proof, -G2)`, so both G2 pairing inputs collapse to the constants -G2 and [τ]G2:

  e(D, -G2) · e(proof, [τ]G2) == 1,  D = commitment - [y]G1 + [z]·proof

replacing the per-call G2 scalar multiplication (plus a G2 subtraction and its Fp2 inversion) with a single cheaper G1 scalar multiplication, in both the arkworks and blst backends(c-kzg is unchanged (external)).

- arkworks: cache the two now-constant G2 points as G2Prepared (Miller-loop lines computed once) and run multi_miller_loop + one final exponentiation, which still fuses the per-bit squaring across pairs. Fewer G2 ops and cached line prep with no loss of fusion.
- blst: keep the fused pairing_check (blst_miller_loop_n from #3823); only the reformulation applies. Precomputed lines are deliberately not used: blst has no fused-multi-loop-with-precomputed-lines primitive, so precomputing would forfeit the #3823 fusion. That trade is left for a follow-up benchmark.

Tests: add an all-finite tampered-y rejection (pins the pairing relation to the correct evaluation using the c-kzg vector as oracle) and a D=infintiy / finite-proof rejection (exercises the mirror infinity-filter path introduced by moving z to G1).

closes #3835

### DIFF
--- a/crates/precompile/src/kzg_point_evaluation.rs
+++ b/crates/precompile/src/kzg_point_evaluation.rs
@@ -176,4 +176,46 @@ mod tests {
         let t = verify_kzg_proof(&commitment, &z, &y, &proof);
         assert!(!t);
     }
+
+    /// A correct proof with the claimed evaluation `y` perturbed by one bit must
+    /// be rejected. All points are finite, so both pairing pairs are non-trivial:
+    /// this pins the reformulated pairing check (`e(D, -G2)·e(proof, [τ]G2) == 1`,
+    /// `D = commitment - [y]G1 + [z]·proof`) to the exact correct relation, using
+    /// the c-kzg `basic_test` vector as an independent oracle.
+    #[test]
+    fn test_correct_vector_tampered_y_invalid() {
+        let commitment = hex!("8f59a8d2a1a625a17f3fea0fe5eb8c896db3764f3185481bc22f91b4aaffcca25f26936857bc3a7c2539ea8ec3a952b7");
+        let z = hex!("73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000000");
+        // basic_test's correct y (1522…49e9) with its lowest bit flipped. The
+        // result (…49e8) is still far below the modulus, hence canonical, so it is
+        // rejected by the pairing check and not by the canonical-scalar guard.
+        let mut y =
+            hex!("1522a4a7f34e1ea350ae07c29c96c7e79655aa926122e95fe69fcbd932ca49e9").to_vec();
+        y[31] ^= 0x01;
+        let y: [u8; 32] = y.try_into().unwrap();
+        let proof = hex!("a62ad71d14c5719385c0686f1871430475bf3a00f0aa3f7b8dd99a9abc2160744faf0070725e00b60ad9a026a15b1a8c");
+
+        assert!(!verify_kzg_proof(&commitment, &z, &y, &proof));
+    }
+
+    /// Mirror of [`test_infinity_proof_invalid`] for the reformulated check: here
+    /// `D = commitment - [y]G1 + [z]·proof` is the point at infinity while `proof`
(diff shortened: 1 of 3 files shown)


4waan:
the reformulation cannot change how fast F compression runs in blake2
i ll extend the kzg bench to arkworks and blst backends directly though

4waan:
### Local benchmark (arkworks backend)

`kzg verify/arkworks`, criterion, same bench on `main` vs this branch
(`--no-default-features --features std,secp256k1`):

| base (main) | head (this PR) | change |
|:-----------:|:--------------:|:------:|
|  ~3.02 ms   |    ~2.22 ms    | **−26.5%** (CI [−28.6%, −23.3%], p<0.05) |

Wall-clock on macOS (not as clean as CodSpeed's instruction counts).. the
shift is well outside the noise. Eliminating the G2 scalar mul + caching the two
constant `G2Prepared` line sets accounts for it.

4waan:
also ran the head on for blst:
base:  [1.5521 ms  1.5761 ms  1.6156 ms]
           head:  [1.4309 ms  1.4476 ms  1.4687 ms]
           change:[ -9.4962%  -8.1132%  -6.9252%]  (p = 0.00)

rakita:
It looks like even c-kzg is not doing this optimisation for single check (for batch they do). Have made PR here: https://github.com/ethereum/c-kzg-4844/pull/662 and will wait for review there before merging it here

PR #3839, opened 2026-08-03
fix(database): preserve inner storage when flattening CacheDB

### LINKED ISSUE #3838: CacheDB::flatten() drops inner storage entries for accounts present in both caches
## Summary

`CacheDB::flatten()` can lose storage entries that exist only in the inner cache when the same account is also present in the outer cache.

This happens because `flatten()` merges accounts with `HashMap::extend`, which replaces the entire inner `DbAccount` with the outer one:

```rust
inner.cache.accounts.extend(accounts);
```

##  Reproduction

```rust
let account = Address::with_last_byte(42);
let slot_a = StorageKey::from(1);
let value_a = StorageValue::from(10);
let slot_b = StorageKey::from(2);
let value_b = StorageValue::from(20);

inner.insert_account_info(account, AccountInfo::default());
inner
    .insert_account_storage(account, slot_a, value_a)
    .unwrap();

let mut outer = inner.nest();
outer
    .insert_account_storage(account, slot_b, value_b)
    .unwrap();

let mut flattened = outer.flatten();

assert_eq!(flattened.storage(account, slot_a), Ok(value_a));
assert_eq!(flattened.storage(account, slot_b), Ok(value_b));
```

The first assertion currently fails because slot_a is no longer present after flattening:

```shell
left:  Ok(0)
right: Ok(10)
```

## Root cause

When the outer cache first loads the account, load_account() obtains only AccountInfo through basic_ref():

```rust
db.basic_ref(address)?
    .map(|info| DbAccount {
        info,
        ..Default::default()
    })
```

The inner account's cached storage is not copied. The outer account therefore contains only the storage slots accessed by the outer cache.

After flatten(), the outer DbAccount replaces the inner DbAccount. A lookup for slot_a then falls through to the underlying database. If the value existed only in the inner cache, it cannot be recovered and the underlying database may return zero.

## Expected behavior

Flattening a nested cache should preserve storage entries from the inner cache while allowing outer entries for the same slot to override them.

Account-state semantics also need to be preserved. In particular, StorageCleared and NotExisting must not be accidentally weakened by merging with an outer account.

## Proposed direction
Merge accounts with the same address instead of replacing the whole DbAccount:

- merge storage maps, with outer slots taking precedence;
- preserve the appropriate AccountState semantics;
- use the outer AccountInfo as the newer account information.

grandpig:
## Summary

Fixes https://github.com/bluealloy/revm/issues/3838

`CacheDB::flatten()` previously replaced an inner `DbAccount` with the outer `DbAccount` whenever both caches contained the same address. Since the outer account may contain only a subset of the account's storage, storage entries cached exclusively by the inner database could be lost.

## Changes

- Merge `DbAccount` values for addresses present in both caches.
- Preserve inner storage entries that are not present in the outer cache.
- Let outer storage entries override inner entries for the same slot.
- Preserve `StorageCleared` / `NotExisting` semantics.
- Add a regression test covering storage entries from both cache layers.

## Regression test

The test verifies that after flattening:

- an inner-only slot remains readable;
- an outer slot remains readable;
- outer values override inner values for the same slot.

## Validation

```text
cargo test -p revm-database
cargo fmt --all --check

### DIFF
--- a/crates/database/src/in_memory_db.rs
+++ b/crates/database/src/in_memory_db.rs
@@ -71,7 +71,7 @@ impl<ExtDb> CacheDB<CacheDB<ExtDb>> {
     /// Flattens a nested cache by applying the outer cache to the inner cache.
     ///
     /// The behavior is as follows:
-    /// - Accounts are overridden with outer accounts
+    /// - Accounts are merged, with outer account info and storage entries taking precedence
     /// - Contracts are overridden with outer contracts
     /// - Logs are appended
     /// - Block hashes are overridden with outer block hashes
@@ -88,7 +88,14 @@ impl<ExtDb> CacheDB<CacheDB<ExtDb>> {
             ..
         } = self;
 
-        inner.cache.accounts.extend(accounts);
+        for (address, account) in accounts {
+            match inner.cache.accounts.entry(address) {
+                Entry::Occupied(mut entry) => entry.get_mut().merge(account),
+                Entry::Vacant(entry) => {
+                    entry.insert(account);
+                }
+            }
+        }
         inner.cache.contracts.extend(contracts);
         inner.cache.logs.extend(logs);
         inner.cache.block_hashes.extend(block_hashes);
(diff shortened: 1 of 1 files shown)


grandpig:
@rakita Just a gentle follow-up on this one. The change is specifically about preserving inner cached storage when flattening nested CacheDBs while keeping StorageCleared / NotExisting semantics intact. I've added coverage for overlapping slots and the account-state edge cases. Would appreciate your thoughts when you have time. Thanks!

grandpig:
@rakita friendly ping~

PR #3840, opened 2026-08-03
perf(precompile): unroll blake2 portable rounds to const-fold the message schedule

### LINKED ISSUE #3837: perf(precompile): blake2 portable path loses upstream's const-folded message schedule
While benchmarking an unrelated EIP-4844 point-evaluation change (#3836), the CodSpeed bot reported a ~4.41% overall degradation. All five regressed benchmarks were `blake2/*` and thats code that PR does not touch at all. That report is a **false positive**: the KZG diff and the blake2 benches share a bench binary, so a change in one shifts code layout and codegen for the other, and CodSpeed's **Cachegrind-based** measurement is sensitive to exactly that (no causal link between that PR and blake2)

But "why would blake2 move at all?" was worth answering, so the blake2 path got audited and that audit surfaced the genuine, pre-existing inefficiency described below ( and quite literally nothing to do with #3836  and predates it)

Two things follow from that, both of which cut against reading too much into the CodSpeed numbers here:

- On x86 CI with `std`, dispatch selects the AVX2 implementation (`mod.rs:62-68`), so the portable path described in this issue is most likely **not** what CodSpeed measured in the first place.
- The `blake2/2_rounds`, `4_rounds`, `10_rounds` and `12_rounds` benchmarks do so little compression work that they are dominated by the 213-byte input parse and the `crypto()` -> `dyn Crypto::blake2_compress` vtable hop, which cannot be inlined into `run`. Movement in those benchmarks is not evidence about the round loop either way( only high round-count cases (100K/200K) isolate it)

So this issue rests on the direct profile and disassembly evidence below, not on the CI bot report that prompted the look.

`blake2::portable::compress` indexes the SIGMA message schedule with a **runtime** round index. Upstream `blake2b_simd` uses a compile-time constant index, which lets LLVM fold the entire schedule into immediates. As a result the portable path executes ~32 memory operations per round that upstream pays zero of, including a two-level dependent load chain on the critical path.

The AVX2 path in this repo does **not** have this problem.. it already unrolls 10 rounds with constant schedules. Only the portable path regressed. The two files even document the divergence in their own headers:

- `avx2.rs:9` "Replaced 12 hardcoded rounds with a loop over 10 unrolled rounds using a remaining counter."
- `portable.rs:9` "Replaced 12 hardcoded `round()` calls with a for loop."

## scope for performance in this issue:

`portable::compress` is not a rare fallback. Per the gating in `blake2/mod.rs:12-16` and
`mod.rs:54-71`, it is the only implementation used on:

- all non-x86 targets — aarch64 (Apple Silicon, AWS Graviton, ARM validators); there is no NEON path
- x86 without AVX2
- **every `no_std` build**, including the RISC-V zkVM targets revm supports (`mod avx2` requires
  `target_feature = "avx2"` or `feature = "std"`, so it isn't even compiled otherwise)

EIP-152 prices the precompile at 1 gas per round, so a caller can request very high round counts cheaply. Making the scalar compression function slower than it needs to be directly worsens worst-case block execution time on those nodes.

## root cause

`portable.rs:32-46`:

```rust
const fn round(r: usize, m: &[Word; 16], v: &mut [Word; 16]) {
    let s = SIGMA[r % 10];                                    // r is the runtime loop variable
    g(v, 0, 4, 8, 12, m[s[0] as usize], m[s[1] as usize]);
    ...
}
```

called from `portable.rs:70-72`:

```rust
for i in 0..rounds as usize {
    round(i, m, &mut v);
}
```

With a constant `r`, upstream gets `r % 10` const-folded, `SIGMA[const]` resolved to a literal row, and every `m[s[j]]` becomes a direct immediate-offset load. With a runtime `r`, none of that survives: the SIGMA row must be addressed at runtime, its 16 bytes loaded, and each message word then loaded at an address that *depends on* the byte just loaded.

Note the loop itself is not the problem.. fusing the calls is fine. The cost is that fusing turned the round index from a constant into a variable.

## Evidence

Measured on aarch64 (apple M2, macOS), release + debug symbols, `revm-precompile` bench target.

The CI report is a single aggregate number over a benchmark that mixes input parsing, precompile dispatch and the round loop, so it cannot attribute cost to any of them. So decided to bulge through via two ways:

1. **A sampling profile answers *where* time goes.** Before proposing to restructure the round loop it had to be established that the round loop is in fact the hotspot if compression were a small fraction of the benchmark, no amount of restructuring it would matter, and this issue would not be worth filing. The profile also makes the real call path visible, including the `crypto()` -> `dyn Crypto::blake2_compress` vtable hop that cannot be inlined into `run`.
2. **Disassembly answers *why*.** A profiler attributes samples to functions; it cannot show branches, load-dependency chains or const-folding. 

**1. Profile/ flamegraph** — `samply record` of `blake2/200K_rounds`, 15 s, 15,657 samples:

```
80.50%  revm_precompile::blake2::portable::compress   (self time)
 2.91%  revm_precompile::blake2::run
 remainder: criterion harness
```
Image

The round loop, not input parsing or precompile dispatch, dominates at high round counts.

**2. Disassembly** — `objdump -d --disassemble-symbols=..._portable8compress... <bench binary>`. The whole function is 252 instructions; the loop body spans `0x…c3d0`–`0x…c640` and covers **one round** (not unrolled), closed by a single `b.ne`. Instruction mix:

| instruction | count | what it is |
| --- | --- | --- |
| `ldrb` | 16 | every byte of the SIGMA row, at offsets `#0x0`–`#0xf`, reloaded each round |
| `ldr` | 16 | `m[s[j]]`, at addresses **dependent** on the `ldrb` results |
| `add` / `eor` / `ror` | 53 / 50 / 32 | the actual mixing work (8 × `g`) |
| `umulh` + `madd` + `lsr` | 1 each | `r % 10` by reciprocal multiply, folded into the SIGMA row address |

So each round performs ~32 memory ops against ~38 ALU ops of real mixing. Upstream's constant
schedule emits **zero** of the 16 `ldrb` and no modulo, and its message loads use immediate
offsets, so they issue independently instead of forming a serialized
`ldrb → use as index → ldr` chain two L1 hits deep before the first `add` can start.

**Note on bounds checks:** there are none. The only branches in the function are the loop
back-edge and the `rounds == 0` guard — LLVM range-analyzes the constant SIGMA table and proves
every index is in range. The cost here is purely the extra loads and the dependency chain, not
panic paths.

Will open a draft first that will: 
mirror what `avx2.rs` already does: unroll 10 physical rounds in the loop body, each with a constant round index so its SIGMA row const-folds, and keep EIP-152's variable round count with a `remaining` counter checked between rounds. SIGMA has spec period 10, so 10 unrolled rounds tile any round count exactly, and rounds 10/11 of standard BLAKE2b reuse rows 0/1 as required by RFC 7693 §2.7.

This changes no observable behaviour.. same output for every input, including the EIP-152 edge cases (`rounds = 0`, non-multiple-of-10 counts). Existing precompile tests cover it, and the disassembly gives an objective check: the 16 `ldrb` per round and the `umulh`/`madd` should disappear entirely.

4waan:
Closes #3837 

`portable::round` took the round index as a runtime `usize`, so `SIGMA[r % 10]` was resolved at runtime. It now takes the index as a `const` generic, and `compress` unrolls ten physical rounds with a `remaining` counter for EIP-152's variable round count.. the structure `avx2.rs` already uses. SIGMA has spec period 10 (RFC 7693 §2.7), so ten rounds tile any count, and re-entering the loop at round 0 is how standard BLAKE2b's rounds 10 and 11 reuse rows 0 and 1.

(No public API change.)

**In the emitted code** (aarch64, same function, same flags): the 16 `ldrb` schedule loads per round and the `umulh`/`madd` modulo are gone entirely, message loads drop from 16 per round to ~6 across all ten, and the loop goes from 157 to ~120 instructions per round. The dependent `ldrb → index → ldr` chain disappears because constant indices let the compiler keep most message words in registers. Cost is a ~5x larger function (252 → 1309 instructions), worth noting for icache pressure.

**In wall time**  native aarch64 runner (`ubuntu-24.04-arm`), existing `bench/blake2.rs` unchanged, baseline on `main` and comparison on this branch in the same job on the same VM:

| benchmark | main | this PR | change | 95% CI |
| --- | --- | --- | --- | --- |
| 2 rounds | 57.21 ns | 49.63 ns | −13.26% | [−13.28%, −13.24%] |
| 4 rounds | 84.35 ns | 72.50 ns | −14.05% | [−14.16%, −13.99%] |
| 10 rounds | 174.39 ns | 139.34 ns | −20.10% | [−20.15%, −20.07%] |
| 12 rounds | 204.00 ns | 160.39 ns | −21.38% | [−21.38%, −21.37%] |
| 64 rounds | 985.36 ns | 736.77 ns | −25.23% | [−25.28%, −25.18%] |
| 512 rounds | 7.7102 µs | 5.7014 µs | −26.05% | [−26.07%, −26.03%] |
| 1024 rounds | 15.3916 µs | 11.3850 µs | −26.03% | [−26.04%, −26.02%] |
| 100K rounds | 1.5000 ms | 1.1034 ms | −26.44% | [−26.48%, −26.39%] |
| 200K rounds | 2.9999 ms | 2.2071 ms | −26.43% | [−26.45%, −26.40%] |

All p < 0.05.

Two things stand out. The gain **scales with round count and plateaus near −26%**: at 2–12 rounds the fixed costs (213-byte input parse, plus the `crypto()` -> `dyn Crypto::blake2_compress` vtable hop that cannot be inlined into `run`) dilute it; from 512 rounds up the loop dominates and the measurement converges on the loop's own improvement. And that asymptote **agrees with the static instruction count** (−23.6% per round), which owes nothing to the machine that ran it.. the measurement landing slightly ahead is what you would expect when the removed instructions are serialized dependent loads rather than average ones.

## Tests

The compression function had no unit tests.. it was covered only indirectly through the execution-spec state tests. This PR adds:

- a **differential test** against a reference implementation using the previous runtime round
  index, over round counts `0..=40` plus 100/101/109/110/111/1000/4096, 64 pseudorandom
  `(h, m, t, f)` inputs each. Those counts cover an empty run, a partial tile, the wraparound at
  ten, and a full second tile. It calls `portable::compress` directly, so the portable path is
  exercised even on targets that would otherwise select AVX2.
- **EIP-152 vector 4** as an oracle independent of both implementations, checked against the
  portable path and the dispatcher.

Both were mutation-checked: mis-numbering one unrolled round (`round::<9>` → `round::<8>`) fails the differential test at exactly `rounds = 10` and fails the EIP-152 vector too.

### DIFF
--- a/crates/precompile/src/blake2/mod.rs
+++ b/crates/precompile/src/blake2/mod.rs
@@ -51,6 +51,10 @@ const SIGMA: [[u8; 16]; 10] = [
 /// BLAKE2b compression function F (EIP-152).
 ///
 /// Dispatches to the best available implementation (AVX2 or portable).
+// On targets with no SIMD path the cfgs below collapse to the single `portable::compress` call,
+// which is `const`, so clippy suggests making this `const` too. It cannot be: on x86 this performs
+// runtime AVX2 feature detection and calls an `unsafe` intrinsic implementation.
+#[allow(clippy::missing_const_for_fn)]
 pub fn compress(rounds: u32, h: &mut [Word; 8], m: &[Word; 16], t: &[Word; 2], f: bool) {
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     {
--- a/crates/precompile/src/blake2/mod.rs
+++ b/crates/precompile/src/blake2/mod.rs
@@ -83,7 +83,13 @@ pub fn compress(rounds: u32, h: &mut [Word; 8], m: &[Word; 16], t: &[Word; 2], f
 /// never timed.
 #[doc(hidden)]
 #[inline]
-pub fn compress_portable(rounds: u32, h: &mut [Word; 8], m: &[Word; 16], t: &[Word; 2], f: bool) {
+pub const fn compress_portable(
+    rounds: u32,
+    h: &mut [Word; 8],
+    m: &[Word; 16],
+    t: &[Word; 2],
+    f: bool,
+) {
     portable::compress(rounds, h, m, t, f);
 }
 
--- a/crates/precompile/src/blake2/mod.rs
+++ b/crates/precompile/src/blake2/mod.rs
@@ -136,3 +136,156 @@ pub fn run(input: &[u8], gas_limit: u64) -> EthPrecompileResult {
 
     Ok(EthPrecompileOutput::new(gas_used, out.into()))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Reference implementation: the compression function written with a runtime round index,
+    /// as it stood before the rounds were unrolled with a constant schedule. Deliberately kept
+    /// naive so that it is easy to check against RFC 7693 §3.2 by eye.
+    fn reference_compress(
+        rounds: u32,
+        words: &mut [Word; 8],
+        m: &[Word; 16],
+        t: &[Word; 2],
+        f: bool,
+    ) {
+        fn g(v: &mut [Word; 16], a: usize, b: usize, c: usize, d: usize, x: Word, y: Word) {
+            v[a] = v[a].wrapping_add(v[b]).wrapping_add(x);
+            v[d] = (v[d] ^ v[a]).rotate_right(32);
+            v[c] = v[c].wrapping_add(v[d]);
+            v[b] = (v[b] ^ v[c]).rotate_right(24);
+            v[a] = v[a].wrapping_add(v[b]).wrapping_add(y);
+            v[d] = (v[d] ^ v[a]).rotate_right(16);
+            v[c] = v[c].wrapping_add(v[d]);
+            v[b] = (v[b] ^ v[c]).rotate_right(63);
+        }
+
+        let mut v = [
+            words[0],
+            words[1],
+            words[2],
+            words[3],
+            words[4],
+            words[5],
+            words[6],
+            words[7],
+            IV[0],
+            IV[1],
+            IV[2],
+            IV[3],
+            IV[4] ^ t[0],
+            IV[5] ^ t[1],
+            IV[6] ^ if f { !0 } else { 0 },
+            IV[7],
+        ];
+
+        for r in 0..rounds as usize {
+            let s = SIGMA[r % 10];
+            g(&mut v, 0, 4, 8, 12, m[s[0] as usize], m[s[1] as usize]);
+            g(&mut v, 1, 5, 9, 13, m[s[2] as usize], m[s[3] as usize]);
+            g(&mut v, 2, 6, 10, 14, m[s[4] as usize], m[s[5] as usize]);
+            g(&mut v, 3, 7, 11, 15, m[s[6] as usize], m[s[7] as usize]);
+            g(&mut v, 0, 5, 10, 15, m[s[8] as usize], m[s[9] as usize]);
+            g(&mut v, 1, 6, 11, 12, m[s[10] as usize], m[s[11] as usize]);
+            g(&mut v, 2, 7, 8, 13, m[s[12] as usize], m[s[13] as usize]);
+            g(&mut v, 3, 4, 9, 14, m[s[14] as usize], m[s[15] as usize]);
+        }
+
+        for i in 0..8 {
+            words[i] ^= v[i] ^ v[i + 8];
+        }
+    }
+
+    /// splitmix64, so the inputs are varied but the failures are reproducible.
+    struct Rng(u64);
+
+    impl Rng {
+        fn next(&mut self) -> u64 {
+            self.0 = self.0.wrapping_add(0x9E37_79B9_7F4A_7C15);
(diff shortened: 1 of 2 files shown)


4waan:
**On CodSpeed:** 
`bench.yml` runs `ubuntu-latest` (x86_64), where `compress` dispatches to AVX2, so the code this PR changes is compiled but never executed.

**also a note on a clippy cascade:**

Unrolling makes `portable::compress` const-eligible, so it is now `const fn`. That in turn makes the `blake2::compress` dispatcher look const-eligible on targets where the cfgs collapse to the portable call alone. It cannot be on x86 it does runtime AVX2 detection and calls an `unsafe` implementation so `clippy::missing_const_for_fn` is allowed there with an explanation. Neither lint fires on main.

rakita:
Nice! On my run's I am getting around 10%, but this is good. Will refactor the code slightly but idea lgtm

PR #3848, opened 2026-08-07
bench(precompile): measure the blake2 portable compression path

4waan:
The blake2 benches call `run`, which reaches the compression function through `compress`. On x86 that detects AVX2 at runtime and returns before the portable implementation, so on the x86_64 CI runner the `blake2/*_rounds` benches only ever time the AVX2 code. The portable one is what `no_std`, zkVM and non-x86 targets actually execute, and it is currently unmeasured everywhere except on hosts with no SIMD path.

Add `blake2 portable/{2,10,12,64}_rounds`, calling the portable function directly so it is timed on every host (regardless of what the dispatcher would pick). The round counts cover how SIGMA's period-10 schedule tiles: 2 is a partial tile, 10 exactly one, 12 one plus a remainder (standard BLAKE2b), 64 several..

Reaching it needs `compress_portable`, a `#[doc(hidden)]` wrapper. The dispatcher is the only supported entry point, so the wrapper documents that it is not public API and exists for measurement.

Existing benchmark names are left untouched so their histories are carried.

### DIFF
--- a/crates/precompile/bench/blake2.rs
+++ b/crates/precompile/bench/blake2.rs
@@ -69,4 +69,47 @@ pub fn add_benches(group: &mut BenchmarkGroup<'_, criterion::measurement::WallTi
         let input = black_box(&inputs[8]); // 200000 rounds
         b.iter(|| blake2::run(input, u64::MAX).unwrap());
     });
+
+    add_portable_benches(group);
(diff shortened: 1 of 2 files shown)


rakita:
ack

PR #3860, opened 2026-08-15
fix(precompile): prevent stale short-address lookups after `get_mut`

### LINKED ISSUE #3859: `Precompiles::get_mut()` can leave optimized short-address lookups stale
## Summary

`Precompiles` maintains two representations of each precompile:

- `inner`, the authoritative address map;
- `optimized_access`, a cloned cache for addresses below `SHORT_ADDRESS_CAP`.

`get_mut()` currently returns a mutable reference to the entry in `inner` without invalidating or updating the corresponding cached clone. A caller can therefore replace a short-address precompile successfully, while subsequent `get()` calls continue returning the old precompile.

## Reproduction

```rust
use revm_precompile::{
    u64_to_address, Precompile, PrecompileId, PrecompileResult, Precompiles,
};

fn replacement_precompile(_: &[u8], _: u64, _: u64) -> PrecompileResult {
    unreachable!("the reproduction only checks which precompile was returned")
}

let address = u64_to_address(4);
let mut precompiles = Precompiles::istanbul().clone();

let replacement = Precompile::new(
    PrecompileId::custom("replacement"),
    address,
    replacement_precompile,
);

*precompiles.get_mut(&address).unwrap() = replacement;

assert_eq!(
    precompiles.get(&address).unwrap().id(),
    &PrecompileId::custom("replacement"),
);

aroralanuk:
> AI disclosure: This changes have been aided by Codex 5.6 Sol but I have manually reviewed them and I take responsibility for its correctness.

Fixes https://github.com/bluealloy/revm/issues/3859

## Summary

TLDR - currently `Precompiles` stores precompiles in both inner and optimized_access which is a cloned cache. Right now `get_mut()` modified only `inner`. Then, if you replace a short-address precompile you would leave `get()` returning and executing the old cached precompile.

### Approach

I've added a packed bitmap identifying short-address cache slots invalidated by `get_mut()`. The lookup states are:

| Cache | Dirty | Result |
|---|---:|---|
| `Some(precompile)` | false | Return the cached precompile |
| `None` | false | Return `None` without hashing |
| `None` | true | Consult the authoritative map |

A successful short-address `get_mut()` clears that cached clone and marks the slot dirty.

This changes also keeps the existing `Option<&mut Precompile>` API unchanged. The bitmap occupies 40 inline bytes on both 32-bit and 64-bit targets and requires no additional allocation.

## Performance

No normal cached-hit or clean-miss regression was observed. But, a lookup after `get_mut()` increases from approximately 1.19 ns to 2.44 ns because it intentionally consults the address map. This cost is isolated to the mutated slot and lasts until `extend()` rebuilds its cache entry.

## Testing

Added coverage for:

- replacing and executing a short-address precompile through `get_mut()`;
- cloned invalidated state;
- restoring the cache through `extend()`;
- unsuccessful `get_mut()` preserving a clean missing slot;
- bitmap word boundaries and address `SHORT_ADDRESS_CAP - 1`.

### DIFF
--- a/crates/precompile/bench/lookup.rs
+++ b/crates/precompile/bench/lookup.rs
@@ -1,0 +1,49 @@
+use criterion::Criterion;
+use revm_precompile::{
+    primitives::SHORT_ADDRESS_CAP, u64_to_address, Precompile, PrecompileId, PrecompileResult,
+    Precompiles,
+};
+use std::hint::black_box;
+
+fn unused_precompile(_: &[u8], _: u64, _: u64) -> PrecompileResult {
+    unreachable!("the lookup benchmark never executes the precompile")
+}
+
+pub fn benchmark_precompile_lookup(c: &mut Criterion) {
+    let short_hit = u64_to_address(4);
+    let short_miss = u64_to_address(SHORT_ADDRESS_CAP as u64 - 1);
+    let long_hit = u64_to_address(SHORT_ADDRESS_CAP as u64);
+    let long_miss = u64_to_address(SHORT_ADDRESS_CAP as u64 + 1);
+
+    let mut precompiles = Precompiles::latest().clone();
+    precompiles.extend([Precompile::new(
+        PrecompileId::custom("lookup-benchmark"),
+        long_hit,
+        unused_precompile,
+    )]);
+
+    let mut invalidated = precompiles.clone();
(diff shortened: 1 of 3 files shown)


rakita:
Hm lets deprecate `get_mut` it is not used in reth. And add fn insert function that would insert a new precompile that would overwrite old one

PR #3861, opened 2026-08-18
refactor(gas): rename cold_storage_cost to sstore_additional_cold_cost, document EIP-2929 warm double charge

rakita:
## What

- Renames `GasId::cold_storage_cost` (id 24) to `GasId::sstore_additional_cold_cost` — the value is the SSTORE-specific cold surcharge, not a generic cold storage cost, and the new name plus docs make the fork quirk explicit:
  - **Berlin..Osaka**: holds the full cold sload cost (2100). Since the warm base (`sstore_static`, 100) is always charged separately, the warm cost is **double charged** on a cold SSTORE (EIP-2929: cold no-op 2200, cold set 22100, cold reset 5000).
  - **Amsterdam**: EIP-8038 aligns this by folding the warm base into the cold cost, so the surcharge is only the premium above warm (2000), mirroring SLOAD accounting.
- The double-charge quirk and its Amsterdam fix are now documented on the `GasId`, the `GasParams` accessor, both table assignments, and `sstore_dynamic_gas`.
- `sstore_dynamic_gas` keeps its original signature (no new fork flag) — the fork difference stays encoded in the gas table.
- The SSTORE `skip_cold_load` sentry compares against `sstore_additional_cold_cost`: with the static warm base charged first, this is exactly EELS's `max(access_cost, CALL_STIPEND + 1)` BAL-recording rule on Amsterdam. Pre-Amsterdam the EIP-2200 stipend sentry makes the check unreachable.
- `GasParams::cold_storage_cost()` derived helper is removed along with the old name; total cold storage access cost is `cold_storage_additional_cost() + warm_storage_read_cost()` on every fork.

## Why

The old `cold_storage_cost` name hid an EIP-2929 quirk: on a cold SSTORE the table value silently re-charges the warm access from Berlin until Amsterdam, where EIP-8038 aligned it. Naming it as the SSTORE cold surcharge and documenting the double charge at the point of definition makes the gas table honest without changing any behavior.

Verified experimentally (previous iteration of this PR): charging only the cold premium on all forks (SLOAD-style) matches glamsterdam fixtures exactly but diverges by exactly −100 per cold SSTORE on every pre-Amsterdam fork (2039/2723 develop and 5921/10962 devnet fixtures fail), confirming the double charge is enshrined consensus, not a test bug. Full write-up (EELS spec excerpts, SSTORE pipeline / BAL-recording equivalence proof) in `report.md` in commit 8189048cc, removed from the final tree.

## Breaking change

`GasId::cold_storage_cost()` / `GasParams::cold_storage_cost()` are renamed to `sstore_additional_cold_cost` (same id 24, same table values — no behavior change). Downstream callers only need the rename.

## Testing

- `cargo nextest run --workspace --all-features`: all pass, ee-test golden snapshots unchanged
- `revme statetest test-fixtures/main/develop/state_tests`: all pass
- `revme statetest test-fixtures/devnet/state_tests` (glamsterdam v8.1.0 incl. Amsterdam): all pass


rakita:
Yeah this was an overcharge from Berlin, and this is why we need an additional flag that makes this more straightforward

PR #3868, opened 2026-08-19
fix(context): preserve error state in with_db

cuiweixie:
## Summary
- Preserve the existing `error` field when calling `Context::with_db`.
- Previously `with_db` reset `error` to `Ok(())`, which could silently drop a prior database error after swapping the database.

## Test plan
- [ ] `cargo nextest run -p revm-context`
- [ ] `cargo clippy --workspace --all-targets --all-features`


rakita:
Previous is fine; if db is swapped db error should be dropped

PR #3879, opened 2026-08-25
chore: add Bogota to SpecId

ArshLabs:
adds Bogota as the next `SpecId`


rakita:
There is no need for it

ArshLabs:
no worries!

PR #3885, opened 2026-08-27
feat(database): support state overrides with BAL reads

mattsse:
Adds a dedicated state-override layer to `State` so account, code, and nested storage overrides read before an attached BAL and the backing database. Diff and replacement storage modes accept nested iterators, and later committed execution state remains above the BAL.

This lets Reth position RPC execution with a BAL even when state overrides are present (paradigmxyz/reth#23734).


mattsse:
Closing to keep this work focused on the evm2 design.

PR #3892, opened 2026-09-03
fix(eip8037): place the system call state-gas margin in the reservoir

RealiCZ:
### Summary

EIP-8037 sizes the system call gas limit as `30_000_000 + STATE_BYTES_PER_STORAGE_SET × CPSB × SYSTEM_MAX_SSTORES_PER_CALL` and specifies that "the additional `STATE_BYTES_PER_STORAGE_SET × CPSB × SYSTEM_MAX_SSTORES_PER_CALL` is placed in `state_gas_reservoir` while the rest of the system call's gas is placed in `gas_left`" ([EIP-8037, "System contracts and system transactions"](https://eips.ethereum.org/EIPS/eip-8037#system-contracts-and-system-transactions)). `execution-specs` does the same: the Amsterdam system transaction is built with `execution_gas_grant=SYSTEM_TRANSACTION_GAS` (30M) and `state_gas_reservoir=StateGasCosts.STORAGE_SET * SYSTEM_MAX_SSTORES_PER_CALL` (`src/ethereum/forks/amsterdam/fork.py`, `process_system_transaction`).

#3672 adopted the new limit but kept building the system call gas through `tx_gas` with an all-zero `InitialAndFloorGas`. That path lifts the EIP-7825 cap to `u64::MAX` for system calls, so the whole 31,566,720 lands in `gas_left` and the reservoir is 0. The `SYSTEM_CALL_GAS_LIMIT` doc comment ("the base 30M regular-gas budget plus a state-gas reservoir") already describes the intended split.

Observable differences against the spec today:
- `GAS` inside a system contract reports 31,566,718 instead of 29,999,998.
- Every state-gas charge in a system call spills into regular gas (`state_gas_spilled` grows), so a revert inside the system call credits it back to `gas_left` instead of the reservoir.

The four mainnet system contracts (EIP-2935/4788/7002/7251) do not branch on `gasleft()`, so this does not change their results; it changes what a system contract can observe and how state gas is settled on its failure paths.

### Changes

- `SYSTEM_CALL_GAS_LIMIT` is now expressed as `SYSTEM_CALL_REGULAR_GAS_LIMIT` (30M) + `SYSTEM_CALL_STATE_GAS_RESERVOIR`; both constants are public.
- New `Handler::system_call_gas` (default method) builds the system call `GasTracker`: with EIP-8037 enabled, everything above the regular budget goes to the reservoir; without it the whole limit stays regular gas. `run_system_call` and `inspect_run_system_call` use it instead of `tx_gas`.
- Test `test_system_call_eip8037_state_gas_reservoir`: a system contract running `GAS PUSH0 SSTORE STOP` under `SpecId::AMSTERDAM` stores `29_999_998` and the fresh-slot SSTORE is accounted as `SSTORE_SET_BYTES × CPSB` state gas.

No change for configurations without EIP-8037; `total_gas_spent` of a system call is unchanged (it already subtracts the reservoir).

### Verification

- `cargo nextest run -p revm-handler -p revm-inspector`: 48 passed
- `cargo nextest run -p revm-ee-tests -E 'test(eip8037) | test(system_call)'`: 43 passed
- `cargo clippy -p revm-handler -p revm-inspector --all-targets --all-features`: clean; `cargo fmt --all --check`: clean; `RUSTDOCFLAGS=-D warnings cargo doc -p revm-handler --no-deps`: clean
- The new test fails on `main` (6014612c) with `left: 31566718, right: 29999998` and passes with this change.

### DIFF
--- a/crates/handler/src/handler.rs
+++ b/crates/handler/src/handler.rs
@@ -4,6 +4,7 @@ use crate::{
     frame::handle_reservoir_remaining_gas,
     post_execution::{self, build_result_gas},
     pre_execution::{self, apply_eip7702_auth_list, PreExecutionOutput},
+    system_call::SYSTEM_CALL_REGULAR_GAS_LIMIT,
     validation, EvmTr, FrameResult, ItemOrResult,
 };
 use context::{
@@ -129,7 +130,7 @@ pub trait Handler {
     ) -> Result<ExecutionResult<Self::HaltReason>, Self::Error> {
         // dummy values that are not used.
         let init_and_floor_gas = InitialAndFloorGas::new(0, 0);
-        let mut gas = self.tx_gas(evm, &init_and_floor_gas);
+        let mut gas = self.system_call_gas(evm);
         // System calls skip pre-execution, so the checkpoint that
         // [`Handler::execution`] settles is opened here.
         let checkpoint = evm.ctx().journal_mut().checkpoint();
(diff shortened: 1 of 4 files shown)


rakita (approved):
Cool, lgtm

PR #3894, opened 2026-09-03
feat(state): carry account extension data

decofe:
Adds opt-in `account-ext` support for opaque account payloads in AccountInfo, journaled state, and internal BAL replay. Payloads use an optional ThinArc: one pointer inline, no allocation for empty data, and one shared allocation for nonempty data. With the feature disabled, account layout and serialization remain extension-free.

Canonical EIP-7928 BAL conversion intentionally has no extension field; custom chains must retain the internal BAL representation.

Prompted by: @mediocregopher


crates/context/interface/src/journaled_state/entry.rs
```diff
@@ -117,6 +121,18 @@ pub enum SelfdestructionRevertStatus {
...
 pub enum JournalEntry {
+    /// Restore an account's previous chain-specific payload.
+    #[cfg(feature = "account-ext")]
+    ExtensionChange {
+        /// Account whose extension changed.
+        address: Address,
+        /// Payload before the change.
+        #[cfg_attr(
+            feature = "serde",
+            serde(default, skip_serializing_if = "state::AccountExtension::is_empty")
```
mediocregopher: Not necessary

PR #3905, opened 2026-09-09
fix: keep inner storage when flattening a nested CacheDB

### LINKED ISSUE #3838: CacheDB::flatten() drops inner storage entries for accounts present in both caches

kriss39:
Closes #3838.

`flatten()` merged accounts with `HashMap::extend`:

```rust
inner.cache.accounts.extend(accounts);
```

That replaces the whole `DbAccount` for any address held by both caches, so storage slots known only to the inner account are dropped — including slots the outer cache never touched.

## Why the merge is not just extending the storage maps

`storage()` answers a slot missing from the account map by falling through to `self.db`, unless the account is `StorageCleared` or `NotExisting`, in which case it returns zero without asking. In a nested `CacheDB`, `self.db` of the outer cache *is* the inner cache, so flattening has to reproduce the same answers:

- outer slot present → outer value, so outer wins per slot
- outer slot missing, outer account cleared its storage (or does not exist) → zero, and the inner slots were never reachable, so they must not survive the flattening
- outer slot missing otherwise → the inner account decides, so inner-only slots have to be kept

The last case has one more wrinkle. If the *inner* account is the one that cleared its storage, unknown slots used to resolve to zero there. Keeping the outer `Touched`/`None` state would send them to the external database instead, so the merged account keeps `StorageCleared`.

## Tests

Four tests, in `crates/database/src/in_memory_db.rs`:

- inner-only slot survives (the reproduction from the issue)
- outer value wins for a slot in both
- outer `replace_account_storage` still drops the inner slots
- inner `replace_account_storage` leaves the merged account `StorageCleared`

The first and last fail on `main` (`left: 0, right: 10` — the inner slot reads back as zero); the middle two pass before and after and are there to pin behaviour I did not want to change.

`cargo test --workspace`, `cargo clippy -p revm-database --all-targets --all-features` and `cargo fmt --all --check` are clean locally.

### DIFF
--- a/crates/database/src/in_memory_db.rs
+++ b/crates/database/src/in_memory_db.rs
@@ -71,7 +71,9 @@ impl<ExtDb> CacheDB<CacheDB<ExtDb>> {
     /// Flattens a nested cache by applying the outer cache to the inner cache.
     ///
     /// The behavior is as follows:
-    /// - Accounts are overridden with outer accounts
+    /// - Accounts are merged, with outer accounts taking precedence. Storage slots cached only
+    ///   by the inner account are kept, since the outer cache would have resolved them through
+    ///   the inner one. They are dropped when the outer account cleared its storage.
     /// - Contracts are overridden with outer contracts
     /// - Logs are appended
     /// - Block hashes are overridden with outer block hashes
@@ -88,7 +90,16 @@ impl<ExtDb> CacheDB<CacheDB<ExtDb>> {
             ..
         } = self;
 
-        inner.cache.accounts.extend(accounts);
+        for (address, outer_account) in accounts {
+            match inner.cache.accounts.entry(address) {
+                Entry::Vacant(entry) => {
+                    entry.insert(outer_account);
+                }
+                Entry::Occupied(entry) => {
+                    merge_account(entry.into_mut(), outer_account);
+                }
+            }
+        }
         inner.cache.contracts.extend(contracts);
         inner.cache.logs.extend(logs);
         inner.cache.block_hashes.extend(block_hashes);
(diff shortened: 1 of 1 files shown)


kriss39:
Closing in favour of #3839, which addresses the same flatten issue and predates this PR.

PR #3921, opened 2026-09-19
fix(inspector): forward frame_start/frame_end through Either<L, R>

kriss39:
## Motivation

`Inspector::frame_start`/`frame_end` were added in #3518 and forwarded in the `(L, R)` tuple impl, but the `Either<L, R>` impl in `crates/inspector/src/either.rs` was never updated. Wrapping an inspector in `Either` therefore silently replaces its `frame_start` override (including a `Some(FrameResult)` short-circuit) and its `frame_end` mutation with the trait's no-op defaults. Every other hook is forwarded, so nothing fails to compile and nothing signals the loss.

## Solution

Generalize the `Either<L, R>` impl over the `FI`/`FR` type parameters, matching the tuple impl, and delegate `frame_start`/`frame_end` to the active variant in the same `match` style as the other methods. `Either<NoOpInspector, TracerEip3155<W>>` and other default-parameter users still infer as before.

Adds a regression test with a counting inspector wrapped in `Either::Left`; it fails on `main` (`frame_start` is never observed) and passes with this change. `cargo nextest run --workspace`, `cargo clippy -p revm-inspector --all-targets --all-features`, `cargo check -p revm-inspector --no-default-features` and `cargo fmt --all --check` are clean.


rakita:
bot opened.

PR #3922, opened 2026-09-19
fix(handler): don't panic on overflowing gas cost when balance check is disabled

kriss39:
## Motivation

`calculate_caller_fee` only calls `ensure_enough_balance` when the balance check is enabled. That call (via `max_balance_spending`) is what guarantees `gas_limit * gas_price` fits in `u128`, but the `.expect()` right after it assumes the guarantee always holds:

```rust
let effective_balance_spending = tx
    .effective_balance_spending(basefee, blob_price)
    .expect("effective balance is always smaller than max balance so it can't overflow");
```

With `disable_balance_check = true` a transaction whose gas cost overflows `u128` therefore panics in `pre_execution.rs` instead of failing validation. `gas_limit` only needs to be within the EIP-7825 cap (so nothing earlier rejects it); `gas_price` alone is enough to overflow.

This is reachable for embedders that execute externally supplied transactions with the balance check disabled. reth's engine payload prewarm (`crates/engine/tree/src/tree/payload_processor/prewarm.rs`) runs every transaction of an incoming `engine_newPayload` block with `disable_balance_check = true`, and nothing in the payload validation path bounds `gas_price` before that.

## Solution

Propagate the `OverflowPaymentInTransaction` error that `effective_balance_spending` already returns for this case — the same error the enabled path produces through `ensure_enough_balance`. Returning the error rather than saturating also keeps the invariant `reward_beneficiary` relies on (`coinbase_gas_price * gas_used` is multiplied unchecked there).

Regression test under the `optional_balance_check` feature, next to the existing `test_disable_balance_check`; it panics on `main` and passes here. `cargo nextest run --workspace`, `cargo nextest run -p revm-ee-tests --features optional_balance_check`, `cargo clippy -p revm-handler --all-targets --all-features` and `cargo fmt --all --check` are clean.


rakita:
bot opened.

PR #3923, opened 2026-09-19
fix(context-interface): keep fake_exponential exact and non-panicking for large excess blob gas

kriss39:
## Motivation

`fake_exponential` evaluates the EIP-4844 Taylor series with `u128` intermediates. The per-iteration product `numerator_accum * numerator` overflows above roughly 192M excess blob gas with the Cancun update fraction (284M for Prague), long before the final blob price stops fitting in a `u128` (296M / 444M):

- in debug builds this panics ("attempt to multiply with overflow");
- revm's release profile has overflow checks off, so the accumulator wraps and the loop keeps running roughly `excess / denominator` iterations before returning a wrong price — seconds at `2^50`, far longer near `u64::MAX`.

The value comes from headers: `revme blockchaintest` feeds the fixture header's `excess_blob_gas` into `BlobExcessGasAndPrice::new_with_spec`, and embedders that build `BlockEnv` from RPC headers hit the same path through `set_blob_excess_gas_and_price`. `TestUnit::block_env` additionally converts `currentExcessBlobGas` with a bare `.to::<u64>()`, so an out-of-range fixture panics the statetest runner before even reaching the series (its sibling fields `current_gas_limit`, `current_base_fee` and `slot_number` already clamp with `try_into().unwrap_or(u64::MAX)`).

alloy-eips fixed the same function for the same reason (alloy-rs/alloy#2806, refined in #3961 to stay exact instead of saturating early).

## Solution

Keep the `u128` loop as the fast path with checked arithmetic, and fall back to 512-bit intermediates (`ruint` `U512`, already a dependency) when it would overflow. The result is exact for every value that fits in a `u128` and saturates to `u128::MAX` only when the mathematical result does not fit, which fails closed for any fee check that consumes it. The fallback is `#[cold]`; the fast path adds two checked ops, measured at ~2 ns on `BlockEnv::default()` (9.1 → 11.6 ns), so the `evm-build` benchmark is unaffected.

`TestUnit::block_env` clamps `currentExcessBlobGas` like its sibling fields.

Tests: big-integer reference values on both sides of the old overflow threshold and at the `u128` boundary for the Cancun and Prague fractions, saturation for `u64::MAX`, and a statetest fixture with `currentExcessBlobGas = U256::MAX`. Both fail on `main` (panic) and pass here. `cargo nextest run --workspace`, `cargo clippy -p revm-context-interface -p revm-statetest-types --all-targets --all-features`, `cargo check -p revm-context-interface --no-default-features` and `cargo fmt --all --check` are clean.


rakita:
bot opened.

kriss39:
> bot opened.
Ai just helped with running the tests and preparing the description. The rest of the infrastructure and solution are completely up to me.

PR #3924, opened 2026-09-19
fix(database): keep storage cleared after a selfdestructed account is touched again

Bruce039:
`CacheDB::commit_account` decides the new `AccountState` like this:

```rust
db_account.account_state = if is_newly_created {
    db_account.storage.clear();
    AccountState::StorageCleared
} else if db_account.account_state.is_storage_cleared() {
    AccountState::StorageCleared
} else {
    AccountState::Touched
};
```

After a selfdestruct the account is stored as `NotExisting` with its cached storage cleared. If a later transaction only touches that address again — a plain value transfer, which the journal reports as touched but not created — the `else` branch turns it into `Touched`. `storage()` and `storage_ref()` only treat `StorageCleared | NotExisting` as "storage known to be empty", so for a `Touched` account every slot that is not in the (now empty) cache map is looked up in the inner database, which still holds the pre-selfdestruct values. The old storage comes back.

This is visible with a two-level `CacheDB` (the shape a fork-backed database has): seed the inner one with a contract that has a non-zero slot, run a transaction that selfdestructs it, then send it 1 wei in a second transaction; reading the slot afterwards returns the old value. On Cancun and later the selfdestruct has to happen in the creating transaction (EIP-6780), so this mostly matters for pre-Cancun specs and for addresses that have storage without code; `State` is not affected because its status enum keeps a `DestroyedChanged` state that stays inside `is_storage_known()`.

The fix keeps the known-empty invariant: a `NotExisting` account that is touched again becomes `StorageCleared`, the same state a `StorageCleared` account keeps today. The added test does the selfdestruct-then-touch sequence on a nested `CacheDB` and checks `storage`, `storage_ref`, the committed account info and the resulting state; it fails on `main` with the resurrected value.

Checked with `cargo nextest run --workspace`, `cargo clippy -p revm-database --all-targets --all-features`, `cargo check -p revm-database --no-default-features` and `cargo fmt --all --check`.


rakita:
bot opened.

PR #3925, opened 2026-09-19
fix(state): keep same-index BAL write when a rebased commit repeats it

Bruce039:
EIP-7928 shares one `block_access_index` between the withdrawals and the post-execution system calls (and one between the pre-execution system calls), and asks implementations to compare every write "against the storage value as of immediately before the current `block_access_index`". `BalWrites::update_with_key` does that comparison against the `original` value the caller passes in. When the only entry recorded at the current index came from an earlier commit at that same index, the next commit's `original` is that entry's value rather than the pre-index value, so a locally unchanged commit looks like "restored to original" and the entry is popped.

`revme`'s `post_block_transition` runs into this on a valid block: it commits the withdrawals first and then the EIP-7002/7251 system calls at the same index. If a validator's withdrawal address is one of those system contracts, the balance write from the withdrawal is erased when the system call commits the (now unchanged) balance again, and the BAL built by the runner no longer matches the block's. The same applies to any BAL producer that commits more than once per index for the same account.

Fix: in the single-entry case, when the caller's `original` already equals the recorded write, refresh the entry instead of removing it; the caller was rebased on top of the index and cannot know the pre-index value. Everything else keeps the current behaviour, including dropping an entry when a commit with the pre-index baseline restores the pre-index value (covered by a test), and the multi-entry path is unchanged.

Tests: three unit tests on `BalWrites` (rebased no-op keeps the entry, rebased change refreshes it, pre-index restore still pops), and an end-to-end test in `revme` running `post_block_transition` on Prague with a withdrawal to `WITHDRAWAL_REQUEST_ADDRESS`, which fails on `main` with an empty balance write list. `cargo nextest run --workspace`, `cargo clippy -p revm-state --all-targets --all-features`, `cargo check -p revm-state --no-default-features` and `cargo fmt --all --check` pass.


rakita:
bot opened.

PR #3926, opened 2026-09-19
fix(database): keep original storage values when reverting a wiped account

Bruce039:
## Problem

`BundleState::revert` does not restore a storage slot's original value once that slot has passed through a wipe.

When an account is destroyed, the storage it had in the bundle is moved into the `AccountRevert` as `RevertToSlot::Some(present_value)` — in `AccountRevert::new_selfdestructed`, `new_selfdestructed_again`, the `DestroyedChanged` branch of `BundleAccount::update_and_create_revert`, and the wipe pre-drain in `BundleState::extend`. `previous_or_original_value` is dropped at all four sites. On revert the slot is no longer in `BundleAccount::storage` (the wipe removed it), so `BundleAccount::revert` re-creates it with `StorageSlot::new(value)`, which sets both `previous_or_original_value` and `present_value` to the restored value. The slot now reports `is_changed() == false`, and `to_plain_state(OriginalValuesKnown::Yes)` — the mode documented for a single continuous execution/revert session — omits it from the changeset.

Minimal sequence: block 1 changes slot `1` from `0` to `5`; block 2 selfdestructs the account; `revert(1)`. The bundle should now diff exactly like the bundle after block 1 alone, `storage: [(1, 5)]`, but it produces `storage: []`. The same happens when the two blocks are built as separate bundles joined with `BundleState::extend` (the `ExecutionOutcome` shape), and when block 3 recreates the account with different storage and both transitions are reverted one at a time. Anything that unwinds a bundle across a destroyed account and then reads `is_changed()` — plain-state diffs or a changeset-driven state root — sees a dropped write. Under EIP-6780 this needs the destroyed account to already have storage in the bundle, so it mostly affects pre-Cancun specs and storage-without-code addresses.

## Change

`AccountRevert` gets a `wiped_storage_originals: StorageKeyMap<StorageValue>` field, filled at the four sites above with the `previous_or_original_value` of each slot that is moved into the revert, and consumed in `BundleAccount::revert` when a `RevertToSlot::Some` entry is applied. The field is `#[serde(default)]`, so previously serialized reverts still deserialize; `RevertToSlot` itself is unchanged, so existing exhaustive matches on it keep compiling. Code that builds `AccountRevert` with a struct literal needs the new field (one such site exists in reth's invalid-block witness hook).

I picked an additive field over a new `RevertToSlot` variant because the latter would change a persisted enum that downstream code matches exhaustively. If a different representation is preferred I am happy to rework this; the regression tests stand on their own either way.

## Tests

Four tests in `states::state::tests`: revert of a selfdestruct on one long-lived `State`, the same across `BundleState::extend`, a destroy-then-recreate sequence reverted one transition at a time, and an `extend` case where the second bundle changes the slot again before destroying it (the first bundle's original must win). The first three fail on `main`. The existing `extend_on_destroyed_values` expectation was updated for the new field. `cargo nextest run --workspace`, `cargo clippy -p revm-database --all-targets --all-features`, `cargo check -p revm-database --no-default-features`, `--features serde` and `cargo fmt --all --check` are clean.


rakita:
bot opened.

PR #3927, opened 2026-09-19
fix(database): make State::storage_ref consistent with State::storage

0xShadowX:
`State` implements both `Database` and `DatabaseRef`, but `storage_ref` returns different values than `storage` for the same state in two situations.

**Destroyed or not-existing cached account.** `storage_ref` only looks at the cached storage and at `status.is_storage_known()` inside `if let Some(plain_account) = &account.account`. A cached account whose info is `None` — destroyed through `CacheAccount::selfdestruct`, or loaded as not existing — skips that block and falls through to `self.database.storage_ref`, returning the value the slot had before the selfdestruct. `storage` returns zero for the same account (`unwrap_or_default` on the missing info).

**Preloaded bundle, account not cached yet.** With `with_bundle_prestate`, `basic_ref` and `code_by_hash_ref` consult `bundle_state` before the database when the account is not in the cache. `storage_ref` does not, and reads the database directly; `storage` reads the bundle value, because `load_cache_account_with` seeds the cache from the bundle first. Nothing gets cached on the `DatabaseRef` path, so "load the account first" cannot make this consistent — `basic_ref` itself already bypasses that ordering.

This change makes `storage_ref` mirror the mutable path: an account without info yields zero, and an uncached account is looked up in the preloaded bundle (`BundleAccount::storage_slot` already returns zero for destroyed bundle accounts and `None` only for slots the bundle does not know, which then go to the database as before).

Two tests, one per case, each calling `storage` and `storage_ref` on the same `State`; both fail on `main`. `cargo nextest run --workspace`, `cargo clippy -p revm-database --all-targets --all-features`, `cargo check -p revm-database --no-default-features` and `cargo fmt --all --check` pass.


rakita:
bot opened.

PR #3928, opened 2026-09-19
fix(context): clear EIP-8037/EIP-2780 flags on spec downgrade

0xShadowX:
### What happens

`CfgEnv::set_spec_and_mainnet_gas_params` recomputes `gas_params` for the new spec but only ever sets `enable_amsterdam_eip8037` / `enable_amsterdam_eip2780` to `true`:

```rust
if spec.into().is_enabled_in(SpecId::AMSTERDAM) {
    self.enable_amsterdam_eip8037 = true;
    self.enable_amsterdam_eip2780 = true;
}
```

`with_spec_and_mainnet_gas_params` does the same through `self.enable_amsterdam_eip8037 || is_amsterdam`. So a `CfgEnv` that was once at AMSTERDAM and is re-spec'd to OSAKA (or any earlier fork) keeps state-gas and EIP-2780 accounting on while its gas params already describe the older fork. `validate_tx_env` skips the EIP-7825 gas-limit cap when EIP-8037 is enabled, so after such a downgrade a transaction with `gas_limit = 2^24 + 1` is accepted on OSAKA; `validate_initial_tx_gas` and the pre-execution state-gas handling are likewise still on the Amsterdam path.

Long-lived environments are re-spec'd with exactly this helper — it is the documented replacement for the deprecated `set_spec`, and tools that switch forks or replay older blocks call it on an env that may already have been at a later spec. With `SpecId::default()` still at OSAKA this is not hit by construction today, but it will be as soon as the default moves to AMSTERDAM.

### Change

Both setters derive the two flags from the target spec, the same way `max_code_size` and the gas params are derived, so upgrading enables them and downgrading disables them. `with_spec_and_gas_params` (the non-mainnet variant) is untouched and still preserves explicitly set flags for chains that opt in independently.

Tests: a unit test on both setters (downgrade clears, upgrade sets) and a handler test that re-specs AMSTERDAM → OSAKA on one `CfgEnv` and expects `TxGasLimitGreaterThanCap` for a `2^24 + 1` gas transaction. Both fail on `main`. `cargo nextest run --workspace`, `cargo clippy -p revm-context -p revm-handler --all-targets --all-features`, `cargo check -p revm-context --no-default-features` and `cargo fmt --all --check` are clean.


rakita:
bot opened.

